PR #1, opened 2015-11-23
[#108441598] Initial commit to create bastion host

combor:
# What
This creates a new-style bastion host, without NAT and using our own SSH keys rather than insecure-deployer.

It should be noted that it will automatically upload your SSH public key to amazon and assumes that your key is called ```~/.ssh/id_rsa``` and ```~/.ssh/id_rsa.pub```.

# Details
- Bastion host is created and hostname configured correctly.
- User data is used to pass hostname to cloud-init.
- Also we now use user ssh keys which are uploaded by terraform.
- NAT was removed.

# How to test

``` terraform apply -var env=piotr -state=piotr.tfstate```
```terraform destroy -var env=piotr -state=piotr.tfstate -var force_destroy=true```

# Who can test

Anyone but @combor and @Jonty

### DIFF
--- a/tf/aws-vpc.tf
+++ b/tf/aws-vpc.tf
@@ -1,0 +1,11 @@
+provider "aws" {
+  region      = "${var.region}"
+}
+
+resource "aws_vpc" "default" {
+  cidr_block = "${var.vpc_cidr}"
+  enable_dns_hostnames = true
+  tags {
+    Name = "${var.env}-paas"
+  }
+}
--- a/tf/bastion.tf
+++ b/tf/bastion.tf
@@ -1,0 +1,26 @@
+resource "aws_instance" "bastion" {
+  ami = "${lookup(var.ubuntu_amis, var.region)}"
+  instance_type = "t2.micro"
+  subnet_id = "${aws_subnet.infra.0.id}"
+  private_ip = "10.0.0.4"
+  associate_public_ip_address = true
+  vpc_security_group_ids = ["${aws_security_group.bastion.id}"]
+  key_name = "${aws_key_pair.cloudfoundry.key_name}"
+  source_dest_check = false
+  user_data = "#cloud-config\nhostname: ${var.env}-bastion"
+
+  root_block_device = {
+    volume_type = "gp2"
+    volume_size = 100
+  }
+
+  tags = {
+    Name = "${var.env}-bastion"
+  }
+
+  connection {
+    user = "ubuntu"
+    key_file = "~/.ssh/id_rsa"
+  }
+
+}
--- a/tf/dns-records.tf
+++ b/tf/dns-records.tf
@@ -1,0 +1,7 @@
+resource "aws_route53_record" "bastion" {
+  zone_id = "${var.dns_zone_id}"
+  name = "${var.env}-bastion.${var.dns_zone_name}."
+  type = "A"
+  ttl = "60"
+  records = ["${aws_instance.bastion.public_ip}"]
+}
--- a/tf/globals.tf
+++ b/tf/globals.tf
@@ -1,0 +1,13 @@
+variable "env" {
+  description = "Environment name"
+}
+
+variable "office_cidrs" {
+  description = "CSV of CIDR addresses for our office which will be trusted"
+  default     = "80.194.77.90/32,80.194.77.100/32"
+}
+
+variable "ssh_user" {
+  description = "Username used to ssh into VMs."
+  default     = "ubuntu"
+}
--- a/tf/infra-subnet.tf
+++ b/tf/infra-subnet.tf
@@ -1,0 +1,37 @@
+resource "aws_internet_gateway" "default" {
+  vpc_id = "${aws_vpc.default.id}"
+}
+
+resource "aws_route_table" "internet" {
+  vpc_id = "${aws_vpc.default.id}"
+  route {
+    cidr_block = "0.0.0.0/0"
+    gateway_id = "${aws_internet_gateway.default.id}"
+  }
+}
+
+resource "aws_route_table" "infra" {
+  vpc_id = "${aws_vpc.default.id}"
+  route {
+    cidr_block = "0.0.0.0/0"
+    gateway_id = "${aws_internet_gateway.default.id}"
+  }
+}
+
+resource "aws_subnet" "infra" {
+  count             = 2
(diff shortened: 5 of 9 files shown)

### CODE REVIEW

**Comment by keymon** [comment]
Hostname is created but it is the same in every deployment. 

can we change it to use a different name per deployment? for instance:
```
"${var.env}-bastion"
```
Actually that makes sense and it is consistent with the DNS name below. You can also add the field `fqdn`

**Comment by combor** [comment]
This is fixed now. FQDN is not included as it's not configured by cloud-init by just adding fqdn: key.

**Comment by alext** [comment]
I'm a little concerned that this is uploading the ssh secret key files to the bastion host.  Why is this needed? I'd expect only the public key to be needed to allow ssh access.

File: variables.tf [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+}
+
+variable "dns_zone_id" {
+  description = "Amazon Route53 DNS zone identifier"
+  default = "Z3SI0PSH6KKVH4"
+}
+
+variable "dns_zone_name" {
+  description = "Amazon Route53 DNS zone name"
+  default     = "cf.paas.alphagov.co.uk"
```
dcarley: Do we want to continue using `cf` in the domain name? Now is the time to decide. It'd be nice just to use `paas.alphagov.co.uk`. How would that affect the trial on `trial.cf.paas.alphagov.co.uk`?

globals.tf
```diff
@@ -0,0 +1,18 @@
...
+
+variable "office_cidrs" {
+  description = "CSV of CIDR addresses for our office which will be trusted"
+  default     = "80.194.77.90/32,80.194.77.100/32"
+}
+
+variable "web_access_cidrs" {
+  description = "CSV of CIDR addresses for which we allow web access"
+  default     = "80.194.77.90/32,80.194.77.100/32"
+}
```
alext: This variable doesn't appear to be being used. Can we leave it out if it's not.

outputs.tf
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+output "aws_secret_access_key" {
+	value = "${var.AWS_SECRET_ACCESS_KEY}"
+}
+
+output "aws_access_key_id" {
+       value = "${var.AWS_ACCESS_KEY_ID}"
+}
```
alext: Are any of these outputs used at present?  If not, can we again leave these out until we have a need for them.

alext:
We could do with thinking about the repository layout.  This is going to contain more than just the terraform scripts, so they could do with going in a subdirectory.

variables.tf
mtekel: AFAIK certificate wildcard covers only one level. So if you move trial.cf-paas to trial.cf.paas the cert won't cover it any more. Nevertheless, any reason to move trial to new domain or build pipeline?

combor: Ok, I have changed domain to paas.alphagov.co.uk

globals.tf
combor: ```env``` is used to provide prefix for environment. You specify it while running terraform like this: ```terraform apply --var env=alex``` . ```office_cidrs``` are used in security group to provide ssh access. I have removed web_access_cidrs.

outputs.tf
combor: It provides summary after deployment so IMHO it's good to have them.

combor:
It's hard to predict what will be layout requirement in the future. Right now we've got only bastion host in this repo so there is no point moving those files and have a lonely subdir in repo. I expect that it will be rearranged in next iterations/stories.

combor:
> I'm a little concerned that this is uploading the ssh secret key files to the bastion host. Why is this needed?

As I was told we need(ed) that for bosh init to be able to ssh into bosh vm. Not sure if this will be used anymore. I can remove it now as it's easy to add this later.

dcarley:
We should add as little as we need for each step. For example, if we don't need to SSH to the BOSH VM right now, then we shouldn't add anything related to it. That way it's clear why each piece of code exists and we won't unnecessarily carry forwards unused code from the alpha.

Jonty:
FWIW we could simply enable SSH agent forwarding rather than have private keys on the bastion, however that does come with it's own (lesser) set of security risks.

dcarley:
Or generate a new keypair on the bastion host that never leaves that environment. Let's figure that part out when we get there though.

variables.tf
```diff
@@ -0,0 +1,83 @@
...
+}
+
+variable "dns_zone_id" {
+  description = "Amazon Route53 DNS zone identifier"
+  default = "Z3SI0PSH6KKVH4"
+}
+
+variable "dns_zone_name" {
+  description = "Amazon Route53 DNS zone name"
+  default     = "paas.alphagov.co.uk"
```
jimconner: Be careful with this change - We don't currently have the zone paas.alphagov.co.uk delegated to Route53, so this will not be able to create the required DNS records.

mtekel: Speaking of dev/staging/prod, we might want to use different zones for these. As such, we could say that the current `cf` subdomain is for dev. Or go ahead and create AWS accounts, zones and setup delegation for dev/staging/prod already. I think we'd be specifying the zone by parameter then or have some logic to pick zone automatically based on env. name.

mtekel:
Regarding files/directories organization, I'd start with putting all terraform (related) files into e.g. `tf` or `terraform` directory...

variables.tf
dcarley: Very good point. Let's use `dev.paas.alphagov.co.uk` for beta development. However we will need to parameterise so that we can add other environments later:

- `ci.paas.alphagov.co.uk`
- `stage.paas.alphagov.co.uk`
- `prod.paas.alphagov.co.uk`

dcarley:
> Regarding files/directories organization, I'd start with putting all terraform (related) files into e.g. tf or terraform directory...

:+1:

combor:
We have decided to put dev as a domain prefix for now. We can change it later according to our needs.

Jonty:
There's quite a lot of commits on this branch now, make sure you squash them down before merging!

combor:
> There's quite a lot of commits on this branch now...

Done.

dcarley:
Discussed in person: we're closing this for the moment while we concentrate on the Concourse build option, which is preferable and doesn't require this. If that doesn't work out when we can come back to this branch. Thanks for your work on it @combor and @Jonty.

PR #3, opened 2015-12-04
[#109030954]  Concourse job to create minimal VPC

combor:
# What

Two concourse pipelines to set up initial VPC + S3 bucket and to destroy them on request.
S3 bucket has versioning enabled which is requirement for Concourse S3 resource. 
First pipeline is using Terraform to create VPC, networking, security groups and S3 bucket to save state and is triggered by a push to paas-cf repository. 
Second pipeline destroys all objects created by first pipeline and has to be executed manually.

Additionally this PR contains:
- Crown license
- ```.gitignore```
- build status in README file
- Travis configuration
- helper scripts to set up Concourse pipelines

You can find more details in commit messages.

# How to test
Use your Vagrant box to start concourse server.
Execute ```BRANCH=109030954-initial-vpc concourse/scripts/deploy.sh your_env_name```
Go to concourse to check whether pipelines are correctly created.
Push a commit to ```109030954-initial-vpc``` and check if cf-deploy was started and finished successfully. Please make sure you don't merge your test commits.
Execute ```BRANCH=109030954-initial-vpc concourse/scripts/destroy.sh your_env_name```
Watch ```fly``` output to check if terraform destroys VPC.

Please note that due to a bug destroy S3 bucket action will fail. We have created a Terraform [PR](https://github.com/hashicorp/terraform/pull/4168) to fix this.

# Who can test
Anyone but @combor, @mtekel , @actionjack , @jimconner


terraform/vpc/outputs.tf
```diff
@@ -0,0 +1,24 @@
...
+
+output "region" {
+  value = "${var.region}"
+}
+
+output "vpc" {
+  value = "${aws_vpc.myvpc.id}"
+}
+
+output "vpc-CIDR" {
```
keymon: I think it is better to keep this output name in lowercase, don't  you think so? This way we keep consistency.

terraform/vpc/subnet.tf
```diff
@@ -0,0 +1,38 @@
...
+    gateway_id = "${aws_internet_gateway.default.id}"
+  }
+}
+
+resource "aws_subnet" "infra" {
+  count             = "${var.infra_subnet_zone_count}"
+  vpc_id            = "${aws_vpc.myvpc.id}"
+  cidr_block        = "${lookup(var.infra_cidrs, concat("zone", count.index))}"
+  availability_zone = "${lookup(var.zones,       concat("zone", count.index))}"
+  map_public_ip_on_launch = true	# Really?
```
keymon: Good question: really? :)

keymon:
In the [tasks for the creation of the bucket](https://github.com/alphagov/paas-cf/pull/3/files#diff-b0145803d295296ad7a4284c3f402427R24) and for the VPC, it is written all the task and the script inline in the pipeline.

But [for the destroy tasks](https://github.com/alphagov/paas-cf/pull/3/files#diff-3069856755778afa318a7bb10bca0680R42) we setup a separated task file .yml and a external script. 

I think creating a separated `.yml` for the task is unnecessary, but for the sake of consistency, I would suggest to move this "apply" script that is in the pipeline as a inline script to a dedicated script.

What do you think?

concourse/scripts/deploy.sh
```diff
@@ -0,0 +1,18 @@
...
+
+[[ -z "${env}" ]] && echo "Must provide environment name" && exit 100
+bash "${SCRIPT_DIR}/deploy-pipeline.sh" "${env}" "${pipeline}" "${config}"
+
+
+export ATC_URL=${ATC_URL:-"http://192.168.100.4:8080"}
+export fly_target=${FLY_TARGET:-tutorial}
+
+fly unpause-pipeline --pipeline "${pipeline}"
+curl "${ATC_URL}/pipelines/${pipeline}/jobs/init-bucket/builds" -X POST
```
keymon: Do we really want to trigger the pipeline whenever we deploy it? It can be risky/undesirable.

Additionally:

 * the task can be triggered several times concurrently if you run the deploy script several time. In this case does not matter but it can be an issue. I am not sure how to prevent that. 

 * I think the intention of the next command `fly watch -j ${pipeline}/vpc` is wait for the current triggered pipeline. But actually it will watch the latest finished job, which does not need to be (and it never is) the one triggered in this execution of `deploy.sh`, but the previous one.

keymon: When you do the curl, you get the build number in the response in a `Location` header:

```
$ curl -I http://192.168.100.4:8080/pipelines/cf-deploy/jobs/init-bucket/builds -X POST
HTTP/1.1 302 Found
Location: /pipelines/cf-deploy/jobs/init-bucket/builds/15
Date: Mon, 07 Dec 2015 10:57:57 GMT
Content-Length: 0
Content-Type: text/plain; charset=utf-8
```

The fly watch command has a `-b` option to specify the build, but I am not sure how that works:

```
$ fly watch -h
error: Usage:
  fly [OPTIONS] watch [watch-OPTIONS]

Application Options:
  -t, --target=                 Concourse target name or URL (default: http://192.168.100.4:8080)

Help Options:
  -h, --help                    Show this help message

[watch command options]
      -j, --job=PIPELINE/JOB    Watches builds of the given job
      -b, --build=              Watches a specific build

```

terraform/vpc/subnet.tf
mtekel: I have actually removed this. Looks like it got back via the multiple rebases we did. I'll leave to decide this to whoever is using infra subnet first...

concourse/scripts/deploy.sh
keymon: the `-b` option is the build number, for example:

```
fly -t "tutorial" watch -j cf-deploy/init-bucket -b '16'
```

The issue is that we trigger the `cf-deploy/init-bucket` but watch the next task, `cf-deploy/vpc` which we don't know the build number (it might be different).

Additionally, we need to wait until the task is triggered by  `cf-deploy/init-bucket`  to execute watch against  `cf-deploy/vpc`, if not we will get  this error:

```
$ fly -t "tutorial" watch -j cf-deploy/vpc -b '13'
2015/12/07 11:10:29 build not found
```

Solutions for this could be:

 1. Write a script to keep polling the jobs until it finishes. 
 2. Extend the command `fly` to support watch all the pipeline, instead of a specific job.

keymon:
As commented in person with @mtekel [we might get in the situation that the jobs get triggered several times and run in parallel](https://github.com/alphagov/paas-cf/pull/3/files#r46808274). To avoid that, @mtekel mention that adding the option `serial: true` in the `jobs` section, will prevent that.

I have just tested it and it works. 

Can you add the `serial: true` option?

mtekel:
I explicitly did the 2 pipelines differently to show different approaches how it can be done. So that we can see and decide... I'd leave it be for now and see how it develops.

Regarding deploy. The idea of deploy.sh is to be equivalent to running make aws - essentially it's a deploy trigger/script you run when you want to deploy. But good point about multiple concurrent triggers. 

Regarding fly watching jobs - that only works for destroy, where you have just triggered the job. I have not looked at finding ways to wait for the new build and only watch that...

keymon:
Other feedback is add `set -e` in the scripts to force failure of the script if any command fails

keymon:
LGTM, merging

PR #7, opened 2015-12-11
[#109031802] Explain commits

saliceti:
# What
Some commits lack explanation. This readds them with more details.

# How to review
This doesn't change the code. `git diff master` should return nothing.

# Who can review
Anyone but @saliceti


dcarley:
Thanks for doing this :+1:

PR #6, opened 2015-12-11
[#109515840] Add readme

combor:
# What

README for paas-cf repository and Vagrantfile that simplifies Vagrant setup.

# Details

Basic information about how deployment chain operates and how to set-up your environment to be able to deploy your DEV Cloudfoundry instance.

# How to test

Proofreading is highly welcome. Also, please follow `USAGE` section to check whether instructions are correct.

# Who can test 
Anyone but @combor and @jimconner

PR #9, opened 2015-12-11
[#109515840] Fix typo in Vagrantfile vm.box

dcarley:
Looks like a vim whoopsie that repeated the text three times. This was fixed
in #6, but the commits were lost just prior to merging.


combor:
LGTM

PR #13, opened 2015-12-17
Pin concourse/lite version and remove comments

dcarley:
No story ID, because I'm on review duty.

#### Remove comments from Vagrantfile

These are like "getting started" instructions when you create a new
`Vagrantfile` with `vagrant init`. You can find the same information in the
Vagrant documentation so I don't think it's useful to clutter this file:

- https://docs.vagrantup.com/v2/

#### Pin concourse/lite version to 0.69.0

To ensure that we are all using the same version of Concourse. This is the
latest version box available from Atlas and I have tested it with the
`create-deployer` pipeline in `master`.

- https://atlas.hashicorp.com/concourse/boxes/lite

On important change, if you were using 0.68.0 previously, is that `fly hijack`
now sets the correct PWD and environment variables for a job. Which
makes debugging failed jobs much easier.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,6 +3,8 @@
 
 Vagrant.configure(2) do |config|
   config.vm.box = "concourse/lite"
+  config.vm.box_version = "0.69.0"
+
   config.vm.provider "virtualbox" do |vb|
      vb.memory = "2048"
   end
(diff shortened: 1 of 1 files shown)


actionjack:
:+1: Much cleaner!

PR #15, opened 2015-12-17
Remove references to terraform vars from README

alext:
The variables themselves were removed in 278e081, but this reference to
them was left behind.


dcarley:
Good catch

PR #21, opened 2016-01-05
[#110238314] s3init supports instance profile

mtekel:
## What
[Use IAM profiles/roles in new Concourse pipeline](https://www.pivotaltracker.com/projects/1275640/stories/110238314)
Enhance s3init script to support using instance profile provided credentials.
Remove explicit usage of AWS credentials in the create bosh init pipeline.

## Testing
Deploy concourse and bosh init pipeline. Run `bootstrap-s3-state` job. Watch the output and check the bucket for objects being created/not updated.

## Who
Anyone but @mtekel

### DIFF
--- a/concourse/scripts/s3init.sh
+++ b/concourse/scripts/s3init.sh
@@ -4,12 +4,29 @@ bucket=$1
 file=$2
 init_file=$3
 
+# Attempt to use instance profile if keys not configured
+if [[ -z "${AWS_SECRET_ACCESS_KEY}" ]] && [[ -z ${AWS_ACCESS_KEY_ID} ]] ; then
+  meta_url="http://169.254.169.254/latest/meta-data/iam/security-credentials/"
+  profile_name=$(curl -s ${meta_url})
+  instance_profile=$(curl -s ${meta_url}/${profile_name})
+
+      AWS_ACCESS_KEY_ID=$(echo "${instance_profile}" | awk -F\" '$2 == "AccessKeyId"     { print $4 }')
+     AWS_SECURITY_TOKEN=$(echo "${instance_profile}" | awk -F\" '$2 == "Token"           { print $4 }')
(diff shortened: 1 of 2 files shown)


concourse/scripts/s3init.sh
```diff
@@ -4,12 +4,29 @@ bucket=$1
...
 
+# Attempt to use instance profile if keys not configured
+if [[ -z "${AWS_SECRET_ACCESS_KEY}" ]] && [[ -z ${AWS_ACCESS_KEY_ID} ]] ; then
+  meta_url="http://169.254.169.254/latest/meta-data/iam/security-credentials/"
+  profile_name=$(curl -s ${meta_url})
+  instance_profile=$(curl -s ${meta_url}/${profile_name})
+
+      AWS_ACCESS_KEY_ID=$(echo "${instance_profile}" | grep AccessKeyId     | cut -d':' -f2 | sed 's/[^0-9A-Z]*//g')
+     AWS_SECURITY_TOKEN=$(echo "${instance_profile}" | grep Token           | cut -d':' -f2 | sed 's/[^0-9A-Za-z/+=]*//g')
+  AWS_SECRET_ACCESS_KEY=$(echo "${instance_profile}" | grep SecretAccessKey | cut -d':' -f2 | sed 's/[^0-9A-Za-z/+=]*//g')
```
dcarley: The following might be a little safer and simpler, because it's not using any regexes:
```
AWS_ACCESS_KEY_ID=$(echo "${instance_profile}" | awk -F\" '$2 == "AccessKeyId" { print $4 }'
AWS_SECURITY_TOKEN=$(echo "${instance_profile}" | awk -F\" '$2 == "Token" { print $4 }'
AWS_SECRET_ACCESS_KEY=$(echo "${instance_profile}" | awk -F\" '$2 == "SecretAccessKey" { print $4 }'
```
Although we might need to double check if any of these tools are still available if the container changes in alphagov/paas-docker-cloudfoundry-tools#7

mtekel: Thanks for suggestion, updated.

PR #22, opened 2016-01-06
[#109060102] Use specific containers in concourse tasks

keymon:
[#109060102 Create bootstrap pipeline to deploy BOSH](https://www.pivotaltracker.com/story/show/109060102)

What
----

Using specific containers with the minimum dependencies will help to
understand and track the dependencies added to different tools.

Requirements
------------------

#24 should be merged first to cover a bugfix in s3init.sh

Scope
-----

We use the specific containers `curl-ssl` to do curl and openssl logic and
ruby 2.2.3 instead of the multipurpose toolbox.

We use sh instead of bash in s3init script as the curl-ssl container does
not include bash.

How to test this?
-----------------

Test both pipelines create-deployer and create-microbosh as described in the
README.

It can be pipelines in an already deployed environment, and it is not needed
to wait for the bosh-init tasks to finish, as we only want to test the
tasks modified in this PR.

Who?
---

anyone but @keymon

### DIFF
--- a/concourse/scripts/s3init.sh
+++ b/concourse/scripts/s3init.sh
@@ -31,7 +31,7 @@ host=${bucket}.s3-${region}.amazonaws.com
 
 sign() {
   string=$1
-  echo -en "${string}" | openssl sha1 -hmac "${AWS_SECRET_ACCESS_KEY}" -binary | base64
+  /bin/echo -e -n "${string}" | openssl sha1 -hmac "${AWS_SECRET_ACCESS_KEY}" -binary | base64
 }
 
 put() {
--- a/concourse/scripts/s3init.sh
+++ b/concourse/scripts/s3init.sh
@@ -1,11 +1,11 @@
-#!/bin/bash
+#!/bin/sh
 set -x
 bucket=$1
 file=$2
 init_file=$3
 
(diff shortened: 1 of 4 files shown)


concourse/scripts/s3init.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
mtekel: This could break the new, instance-profile enabled s3init, as it is using [[

keymon: Then we either:

 * Create a new container including bash.
 * do not use `[[`

I think all of that can be done in the other story about refactor s3init.sh, as it was played later. If we block this PR longer, we block all the other cards.

dcarley: The S3 changes have already been merged. Looks like this branch needs a rebase anyway, possibly to pick that up. Could you rebase and test whether it works?

keymon:
@actionjack  #24 should be merged first to cover a bugfix in s3init.sh

Nothing really changes, but we need to rebase from master after #24 is merged.

keymon:
@actionjack #24 has been merged and I rebased this with master. Please do `git pull --rebase` to update your history.

PR #24, opened 2016-01-07
[#110238314] Fix bug in s3init when using static credentials

keymon:
https://www.pivotaltracker.com/story/show/110238314

What
----

After implementing IAM profile in s3init, it would fail using static credentials
because it tries to add the x-amz-security-token header.

To fix it, we must not add that header to the request when using static
credentials.

How to test it?
---------------

Run the create-deployer pipeline in a local vagrant, which will use s3init with
static credentials

Who?
----
Anyone but @keymon

### DIFF
--- a/concourse/scripts/s3init.sh
+++ b/concourse/scripts/s3init.sh
@@ -26,7 +26,6 @@ fi
 aws_path=/
 content_type='application/x-compressed-tar'
 acl="x-amz-acl:private"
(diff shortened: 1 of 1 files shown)


dcarley:
Not a blocker for this PR, but this reminds me that we have quite a lot of untested logic here. If we need to continue using it as a workaround in the future then it might benefit from being rewritten in a language that we can easily test.

PR #26, opened 2016-01-08
[#110334440] Fix race condition with Terraform state files.

alext:
This addresses race conditions in the concourse pipelines relating to the storage of terraform state files.

Amazon S3 only [provides eventual consistency](https://aws.amazon.com/s3/faqs/#What_data_consistency_model_does_Amazon_S3_employ) for reads on resources that have been updated. This meant that using `terraform remote` to store the state file resulted in the read in the next state in the pipline getting a stale copy of the state, which lead to the error.  Using concourse resources to manage all reading and writing of these state files avoids this problem because concourse caches the resources.

This PR also include the removal of some deprecated options for the S3 puts (see 3bb828c for details)

## How to review

Deploy an environment all the way up to microbosh following the instructions in the README (be sure to the the `BRANCH` option with this branch).  There should be no errors relating to missing tfvars etc. It should be possible to run the pipelines several times without error.

Destroy the environment completely.  Everything should be cleaned up correctly (including the S3 bucket)

## Who can review

Anyone buy @saliceti and myself.


bleach:
I'm mid-review with this. My create-deployer run failed to put `docker_image_resource.tgz` into the BOSH blobstore. I don't believe that's caused by these changes.

I also still need to test the destroy-deployer pipeline.

alext:
I've now rebased this after #27 has been merged.

bleach:
Now that this uses the EC2 vagrant box I have been able to create and destroy a deployer concourse using this  & I have no comments on the code, so merging.

PR #27, opened 2016-01-08
[#111132940] Use vagrant concourse-lite with AWS provider

keymon:
[#111132940 Switch to EC2 concourse-lite vagrant box](https://www.pivotaltracker.com/story/show/111132940)

# What

In order to improve the reliability of bootstrapping concourse (and therefore reduce the average and worst-case time taken) and be able of benefit of IAM instance profiles, we should switch to using the EC2 concourse-lite vagrant box for bootstrapping a concourse instance.

# Context

In this PR we want to:
 * Provide a functional Vagrantfile to start a AWS instance of the official concourse-lite image. 
 * concourse-lite must be setup with basic authentication.
 * concourse-lite must be accessible only from the office to the SSH port.
 * A ssh tunnel is created to connect to the concourse HTTP interface in port 8080.
 * Do the changes required to allow the AWS concourse-lite to bootstrap the final concourse.
 * Provide some deploy script to ease the bootstrap of the environment.
 * Document all the new process in the README.

Out of the scope:
 * Remove AWS credential references in the pipelines. Although the AWS concourse-lite has a IAM profile, this will be done in a different card.

# How to review.

To test this PR you must be able to deploy a final deployer concourse using the AWS vagrant instance. Test deployment of microbosh is not required.

As review the README.md is also required, follow the steps described there.

**Important:* Remember to setup a custom git branch to be used by the pipelines by setting:
```
```

# Who can review this?

Anyone but @keymon or @mtekel


dcarley:
I'm not sure if we can or should fix it here, but it's important to note that while access to Concourse is limited to the office + BasicAuth, the credentials are currently sent over plain HTTP. This is related to but not solved by the story to secure access to the deployer Concourse in [#109878560](https://www.pivotaltracker.com/story/show/109878560) /cc @bleach

alext:
shellcheck raises some issues with the new shell scripts mostly relating to quoting of variables.  Would be good to check these.

keymon:
@dcarley I understand. Not sure how we can easily fix it, maybe using ELB, but for that we have to create the ELB instance, or setting up a proxy like nginx a tunnel like stunnel using a provisioner in the vagrant file.

Other option: Do not open the port 8080 to the office, but only port 22, and for the user to connect using a SSH tunnel:

``` 
(cd vagrant && vagrant ssh-config) > vagrant/.vagrant/ssh-config
ssh -F vagrant/.vagrant/ssh-config default -L 8080:localhost:8080 -fN
```

Then we connect to http://localhost:8080

dcarley:
> Other option: Do not open the port 8080 to the office, but only port 22, and for the user to connect using a SSH tunnel:

I was wondering the same. That doesn't seem like it'd be too difficult to use - anyone else have any thoughts about it?

alext:
> anyone else have any thoughts about it?

That seems reasonable. If we did this there wouldn't be a need to add auth to this instance either.

combor:
> anyone else have any thoughts about it?

According to Vargant [documentation](https://docs.vagrantup.com/v2/cli/ssh.html) it should be possible to: 
```
vagrant ssh -- -f -L 8080:127.0.0.1:8080 -N
```

keymon:
I fixed the warnings from shellcheck

keymon:
Added support for the SSH tunnel and updated the README.

keymon:
@alext Fixed the hack to get the public IP doing this ed9f6c8cd95467077a95528a788109766982cad8

concourse/pipelines/create-deployer.yml
```diff
@@ -321,3 +327,84 @@ jobs:
...
+          TF_VAR_AWS_SECRET_ACCESS_KEY: {{aws_secret_access_key}}
+        run:
+          path: sh
+          args:
+          - -e
+          - -c
+          - -x
+          - |
+            cd paas-cf/terraform/vpc
+            terraform remote config -backend=s3 -backend-config={{tfstate_bucket}} -backend-config="key=vpc.tfstate"
```
alext: If #26 gets merged before this, these tasks should be updated to use a concourse resource to write the state files instead of using `terraform remote`.

mattbostock:
Should the `./vagrant/deploy.sh` script fail eariy if there is no `~/.ssh/insecure-deployer` file?

```
username: admin
password:
token saved
cat: /Users/matt/.ssh/insecure-deployer: No such file or directory
cat: /Users/matt/.ssh/insecure-deployer: No such file or directory
Concourse API target matt-bootstrap
```

vagrant/Vagrantfile
```diff
@@ -0,0 +1,41 @@
...
+    # us-east-1e in default VPC, 172.31.0.0/20 range
+    aws.subnet_id = 'subnet-b2e3a488'
+
+    # "Concourse Vagrant" security group
+    aws.security_groups = ['sg-ee21a597']
+
+    # Could be replaced by key per user in the future. But would require each user's key to be in AWS.
+    aws.keypair_name = 'insecure-deployer'
+    override.ssh.username = "ubuntu"
+    override.ssh.private_key_path = ENV['PRIVATE_KEY_PATH']
```
saliceti: There is a mismatch with the key specified in the vagrantfile named VAGRANT_PRIVATE_KEY_PATH
It probably works at the moment because insecure-deployer is loaded in ssh agent.

Instead of using that variable, we could actually rely on ssh agent, but this should be documented.

keymon: Fixed. I got rid of those vars and added a small comment in the README.

alext:
> Should the `./vagrant/deploy.sh` script fail eariy if there is no `~/.ssh/insecure-deployer` file?

Probably, but this isn't a problem that's been introduced by this PR, so should probably be addressed separately.

PR #28, opened 2016-01-08
[#110234556] Initiate CF deployment from Concourse

combor:
# What
Using the vanilla manifests and microbosh create a pipeline that deploys a CF instance. [Pivotal story link](https://www.pivotaltracker.com/story/show/110234556). 
# Details

This pipeline uses Terraform states created by VPC, Concourse and BOSH pipelines to create required AWS resources to deploy cloudfoundry. Terraform state is saved in S3 bucket. Next it generates CF manifest from vanilla sources and deploys CF using concourse bosh-deployment resource. Bosh secretes are taken from BOSH pipeline state. s3curl.pl script is provided to simplify S3 authentication.
Also, pipeline creation script is provided to simplify manifest upload to remote concourse instance.

During the test phase we have noticed that concourse instance doesn't have enough space to handle multiple builds especially if cf-deploy job downloads 2GB files during every run so we have increased available disk space to 50G.

Director UUID has been removed from vanilla CF manifests as it is provided by bosh-deplyment resource which is much more convenient than getting it manually. 

# Pre-requisites

- Concourse instance created by create-deployer pipeline
- Microbosh instance created by create-microbosh pipeline
- exported `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID`
- exported `FLY_TARGET` from previous runs

# How to test
Optionally you can provide `RELEASE_VERSION` and `STEMCELL_VERSION` environmental variables.

```
BRANCH=110234556_cf_deploy concourse/scripts/create-cf.sh <env_name>
```

Chelck if `deploy-cf` job finished successfully and CF has been deployed.

# Notes

- `deploy-cf` job downloads CF release during each run. It's more than 2GB so it takes some time to finish. Later we need to improve this and implement caching.
- there is no destroy pipeline so far so here are the steps how to destroy your deploymnet manually:

Hijack the cf-deployment container:

```
fly -t remote hijack -j "deploy-cloudfoundry/deploy-cf" sh
```

Now login to the microbosh instance

```
bosh target https://10.0.0.6:25555
```

username is admin, password is from `bosh-secrets.yml` inside the state bucket.

Check the existing deployments

```
bosh deployments
```

Then delete the cloudfoundry deployment

```
bosh delete deployment <DEPLOY_ENV>
```

exit the container.

Now hijack the cf-terraform container -

```
fly -t remote hijack -j "deploy-cloudfoundry/cf-terraform" sh
```

and find the build directory ( the only one in /tmp/build )

```
cd /tmp/build/<build directory>
```

source the terraform variables into the environment -

```
. terraform-variables/bosh.tfvars.sh
. terraform-variables/vpc.tfvars.sh
. terraform-variables/concourse.tfvars.sh

export AWS_DEFAULT_REGION=eu-west-1
export TF_VAR_AWS_ACCESS_KEY_ID=<access-key-id>
export TF_VAR_AWS_SECRET_ACCESS_KEY=<secret-access-key>
```

Then run terraform destroy -

```
cd paas-cf/terraform/cloudfoundry/
terraform destroy -force -var env=<DEPLOY_ENV> -state=../../../cf.tfstate
```
# Who can test 
anyone but @combor or @timmow


manifests/cf-manifest/outputs/cf-secrets.yml
```diff
@@ -0,0 +1,207 @@
+---
```
mtekel: Do we really want to have cf-secrets.yml in github? Maybe add to .gitignore? Also, IMO, this file should never be created within the repository at all...

keymon:
Do we really need to download the stemcell and the release? 

Bosh allows you to specify a remote url to download the release and the stemcell from, saving one from downloading it locally. See [here](https://bosh.io/docs/uploading-releases.html) and [here](https://bosh.io/docs/uploading-stemcells.html). 

I wonder if it was enough to just specify the url in the `put` of the bosh-deployment-resource 

``` 
      - put: cf-deployment
        params:
          manifest: cf-manifest/cf-manifest.yml
          stemcells: 
            -  https://s3.amazonaws.com/bosh-jenkins-artifacts/bosh-stemcell/aws/light-bosh-stemcell-${STEMCELL_VERSION}-aws-xen-hvm-ubuntu-trusty-go_agent.tgz
          releases:
            - https://bosh.io/d/github.com/cloudfoundry/cf-release?v=$VERSION
```

combor:
> I wonder if it was enough to just specify the url in the put of the bosh-deployment-resource

I'm not sure and haven't tested this. Bosh deployment resource doc says:

> stemcells: Required. An array of globs that should point to where the stemcells used in the deployment can be found.
> releases: Required. An array of globs that should point to where the releases used in the deployment can be found.

and example says:
> ```
- put: staging
  params:
    manifest: path/to/manifest.yml
    stemcells:
    - path/to/stemcells-*.tgz
    - other/path/to/stemcells-*.tgz
    releases:
    - path/to/releases-*.tgz
    - other/path/to/releases-*.tgz
```

I'm aware that downloading release every time sucks :) so I mentioned in in PR. We need to implement caching somehow.

File: manifests/cf-manifest/build_manifest.sh [hunk 2]
```diff
@@ -1,20 +1,18 @@
 #!/bin/sh
 
 set -e
-cd $(dirname $0)
 
-terraform_outputs=${TERRAFORM_OUTPUTS:-"outputs/terraform-outputs.yml"}
```
keymon: I think the commit 58b8032e30c2d2ff8a497410ec536c695 should be split in two, as these changes are not related to the director uuid change

combor: Fixed.

concourse/scripts/create-cf.sh
```diff
@@ -0,0 +1,39 @@
...
+set -e
+
+SCRIPT_DIR=$(cd "$(dirname $0)" && pwd)
+ATC_URL=${ATC_URL:-"http://192.168.100.4:8080"}
+FLY_TARGET=${FLY_TARGET:-$ATC_URL}
+
+env=${DEPLOY_ENV-$1}
+pipeline="deploy-cloudfoundry"
+config="${SCRIPT_DIR}/../pipelines/create-cf.yml"
+bosh_password=$("$SCRIPT_DIR"/s3curl.pl --id="$AWS_ACCESS_KEY_ID" --key="$AWS_SECRET_ACCESS_KEY" --  \
```
keymon: We should discuss if this is the best way to manage this secret. 

We are not sure about two things: 

First, the idea of generating the password, store it in S3 and then forcing us to query it from S3 to define the next pipeline.  I wonder if it is better to create the password in the bosh pipeline definition in a similar way we do in https://github.com/alphagov/paas-cf/blob/master/vagrant/environment.sh#L17

Second, having to maintain a script in perl, which also adds a new dependency.

This topic is touched by a story in the backlog https://www.pivotaltracker.com/projects/1275640/stories/110059806

README.md
```diff
@@ -159,6 +159,25 @@ This pipeline will:
    with an empty file.
  * Destroy the resources created by terraform.
 
+#### Deploy cloudfoundry with microbosh
+```
+# Optionally pass the current branch for the git resources
+export BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
keymon: There is a section below talking about setting the BRANCH variable. Shall we refer to it here?

concourse/scripts/create-cf.sh
dcarley: I second both of these concerns.

keymon:
Although it was not part of this PR (it was added in 4fdcafc1f43d873e13a81126b45448941990bc0e) can we change the name of the deployment from only the name of the environment (`$DEPLOY_ENV`) to something like `myenv-cloudfoundry`.

The idea would be change this line https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/manifests/cf-manifest/deployments/000-base-cf-deployment.yml#L12--1: 

```diff
diff --git a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
index 4eef983..8ac006a 100644
--- a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
@@ -9,7 +9,7 @@ meta:
   - name: cf
     version: latest
 
-name: (( grab meta.environment ))
+name: (( concat meta.environment "-cloudfoundry" ))
 
 releases: (( grab meta.releases ))
```

terraform/cloudfoundry/security-groups.tf
```diff
@@ -0,0 +1,34 @@
+resource "aws_security_group" "web" {
+  name = "${var.env}-web-cf"
```
keymon: The name is `${var.env}-web-cf` but the tag below is `${var.env}-cf-web`. 

Can you change both to `"${var.env}-cf-web"` as other resources are called `${var.env}-cf-*`

concourse/scripts/create-cf.sh
combor: Yes, that's a temporary solution until [Generate secrets automatically in an environment](https://www.pivotaltracker.com/n/projects/1275640/stories/110059806) is done. I'll make a note in Pivotal.

keymon:
About using URLs directly for the `bosh-deployment` resource, I tested it and it does not work:

The change was:

```
   - name: deploy-cf
     serial_groups: [ deploy ]
     serial: true
     plan:
-      - aggregate:
-        - task: cf-stemcell-tarball
-          config:
-            params:
-              STEMCELL_VERSION: {{stemcell-version}}
-            image: docker:///governmentpaas/curl-ssl
-            platform: linux
-            run:
-              path: sh
-              args: ["-c", "curl -L -J -o stemcell.tgz https://s3.amazonaws.com/bosh-jenkins-artifacts/bosh-stemcell/aws/light-bosh-stemcell-${STEMCELL_VERSION}-aws-xen-hvm-ubuntu-trusty-go_agent.tgz"]
-        - task: cf-release-tarball
-          config:
-            params:
-              RELEASE_VERSION: {{cf-version}}
-            platform: linux
-            image: docker:///governmentpaas/curl-ssl
-            run:
-              path: sh
-              args: ['-c', 'curl -f -L -o release.tgz https://bosh.io/d/github.com/cloudfoundry/cf-release?v=$RELEASE_VERSION']
-        - get: cf-manifest
-          passed: ['generate-manifest']
-          trigger: true
+      - get: cf-manifest
+        passed: ['generate-manifest']
+        trigger: true
       - put: cf-deployment
         params:
           manifest: cf-manifest/cf-manifest.yml
-          stemcells: ['cf-stemcell-tarball/stemcell.tgz']
+          stemcells:
+            - https://s3.amazonaws.com/bosh-jenkins-artifacts/bosh-stemcell/aws/light-bosh-stemcell-3104-aws-xen-hvm-ubuntu-trusty-go_agent.tgz
           releases:
-            - cf-release-tarball/release.tgz
+            - https://bosh.io/d/github.com/cloudfoundry/cf-release?v=225
diff --git a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
```

And it fails with:

```
glob 'https://s3.amazonaws.com/bosh-jenkins-artifacts/bosh-stemcell/aws/light-bosh-stemcell-3104-aws-xen-hvm-ubuntu-trusty-go_agent.tgz' matched no files
``` 

Commenting with @combor, we can just not use the bosh-deployment resource. IMO, the bosh operations should be a task, not a resource.

concourse/scripts/create-cf.sh
dcarley: Even if temporary, I think @alext's suggestion is right; we should rebase this against `master` and use [s3get.sh](https://github.com/alphagov/paas-cf/blob/master/concourse/scripts/s3get.sh) to avoid introducing additional dependencies.

keymon:
Just FYI, I hit this bug https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/issues/18 where a password in the registry with a special character `/` made the deployment crash.

Then, I am hitting often this problem https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/pull/17 which makes it fail the deployment, forcing it to be rerun, which has to destroy all the VMs again and start from scratch :(

terraform/globals.tf
```diff
@@ -60,3 +60,9 @@ variable "microbosh_static_private_ip" {
   description = "Microbosh internal IP"
   default     = "10.0.0.6"
 }
+
+variable "web_access_cidrs" {
```
saliceti: Can the default reference to variable office_cidrs instead?

terraform/cloudfoundry/variables.tf
```diff
@@ -0,0 +1,77 @@
+variable "region"     {
```
dcarley: It looks like some of this content duplicates and/or could be moved to `globals.tf`.

Does Terraform not complain about duplicate `variable` definitions?

keymon:
In this PR, [we add the variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in terraform](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/terraform/cloudfoundry/variables.tf#L38--1), to [write them in the outputs](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/terraform/cloudfoundry/outputs.tf#L17--1) so that is consumed [by the manifest](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/manifests/cf-manifest/deployments/aws/000-cf-infrastructure.yml#L8--1) to configure access to the cloudcontroller AWS buckets.

But we can use IAM instance profiles already now, following the changes described here: https://github.com/alphagov/cf-terraform/pull/102

Our [bosh is already deployed using IAM profiles](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/manifests/bosh-manifest/deployments/aws/090-bosh-aws-stub.yml#L2--1), so that should work.

But I am open to implement this in a different story. What do you think @dcarley?

terraform/cloudfoundry/variables.tf
```diff
@@ -0,0 +1,68 @@
...
+  description = "Amazon Route53 DNS zone name"
+  default     = "cf.paas.alphagov.co.uk"
+}
+
+variable "vpc_id" {
+  description = "id of VPC created in main 'vpc' terraform"
+}
+
+variable "AWS_ACCESS_KEY_ID" {
+  description = "AWS access key to be pass to the bosh CPI"
```
keymon: Despite [this comment](https://github.com/alphagov/paas-cf/pull/28#issuecomment-170898565) this description is wrong. The AWS credentials in this case are added to be passed to the cloudcontroller to be able to access the S3 buckets.

concourse/scripts/create-cf.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -e
+
+SCRIPT_DIR=$(cd "$(dirname $0)" && pwd)
+ATC_URL=${ATC_URL:-"http://192.168.100.4:8080"}
+FLY_TARGET=${FLY_TARGET:-$ATC_URL}
+
+env=${DEPLOY_ENV-$1}
+pipeline="deploy-cloudfoundry"
```
keymon: In the rest of pipelines, the name of the pipeline file definition and the script to upload it matches with the name of the pipeline (see [create-deployer](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/concourse/scripts/create-deployer.sh#L10--1) or [create-microbosh](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/concourse/scripts/create-microbosh.sh#L10--1).

Shouldn't we do the same for this pipeline?

In that case, I would suggest rename the file `create-cf.sh` to `deploy-cloudfoundry.sh` and `create-cf.yml` to `deploy-cloudfoundry.yml`.

terraform/cloudfoundry/variables.tf
```diff
@@ -0,0 +1,68 @@
+variable "zones" {
```
keymon: This is duplicated, [also in globals.tf](https://github.com/alphagov/paas-cf/blob/110234556_cf_deploy/terraform/vpc/globals.tf#L15--1)

terraform/cloudfoundry/dns-records.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_route53_record" "wildcard" {
+  zone_id = "${var.dns_zone_id}"
+  name = "*.${var.env}.${var.dns_zone_name}."
+  type = "CNAME"
+  ttl = "60"
+  records = ["dummy"]
```
keymon: Why we use the value "dummy" here, instead of using the public IP  of the web ELB: `${aws_elb.router.dns_name}`, as done [here](https://github.com/alphagov/cf-terraform/blob/master/aws/dns-records.tf#L6)?

timmow:
> In this PR, we add the variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY in terraform, to write them in the outputs so that is consumed by the manifest to configure access to the cloudcontroller AWS buckets.

> But we can use IAM instance profiles already now, following the changes described here: alphagov/cf-terraform#102

> Our bosh is already deployed using IAM profiles, so that should work.

> But I am open to implement this in a different story. What do you think @dcarley?

We did discuss this but realised the vanilla cloudfoundry manifests dont have the iam profile setup in them, and decided to defer this to another story

concourse/pipelines/create-cf.yml
```diff
@@ -0,0 +1,207 @@
...
+            - -e
+            - -c
+            - |
+              . terraform-variables/vpc.tfvars.sh
+              . terraform-variables/concourse.tfvars.sh
+              . terraform-variables/bosh.tfvars.sh
+              
+              STATE_BUCKET={{state_bucket}}
+              cd paas-cf/terraform/cloudfoundry
+              terraform remote config -pull=true -backend=s3 -backend-config=bucket=${STATE_BUCKET} -backend-config="key=cf.tfstate"
```
dcarley: We stopped using Terraform remote state in #26. Could you change this to match?

dcarley:
> We did discuss this but realised the vanilla cloudfoundry manifests dont have the iam profile setup in them, and decided to defer this to another story

That's a shame, but I think it makes sense now given how big the story has turned out to be. However the discussion should have been documented in Pivotal (and probably also the commit message that introduced them). Otherwise there's a good chance that we'd have overlooked it when the story was finished and it wouldn't be obvious that we're another step away from using IAM profiles. I'll create a new story now.

terraform/cloudfoundry/variables.tf
combor: Fixed.

terraform/cloudfoundry/dns-records.tf
combor: Because we don't have it available. It's not created. I have tried to keep this story as small as possible and do not recreate entire alpha environment as we may go in completely different way of setting up networking.

terraform/globals.tf
combor: Unfortunately we can't interpolate variables there.

concourse/pipelines/create-cf.yml
combor: Sure. I had to create one additional job to do s3init.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -0,0 +1,231 @@
...
+            args:
+            - -e
+            - -c
+            - |
+              paas-cf/concourse/scripts/s3init.sh {{state_bucket}} cf.tfstate paas-cf/concourse/init_files/terraform.tfstate.tpl
+          inputs:
+          - name: paas-cf
+      - put: cf-tfstate
+        params:
+          file: s3init/cf.tfstate
```
alext: This put is unnecessary. `s3init.sh` will write the file to S3 if there's nothing there already. Actually it doesn't write a local file anyway, so this won't work.

keymon: The script has been changed to write to local. The intention of @combor is have a pure pipeline with the dependencies well defined, where the tfstate is passed as a resource between all the jobs.

I think is good, because then we have a 3rd pattern to handle the s3 resource initialisation. We should discuss later which one adopt.

keymon:
Now, finally, I am merging :+1:

dcarley:
:clap:

PR #29, opened 2016-01-11
Prefer wget to curl for installing Fly

mattbostock:
`curl` is not installed by default on the Ubuntu image specified by the `Vagrantfile`, so use `wget` which is already installed.

Without this change, I had to run `sudo apt-get update && sudo apt-get install curl` before I could install Fly.


alext:
The `fly` command needs to be installed on the host, and not inside vagrant. Maybe the README needs a tweak to make this clearer.

mattbostock:
:+1: I've raised https://github.com/alphagov/paas-cf/pull/30 instead.

PR #31, opened 2016-01-11
[#109972384] Remove RDS

dcarley:
## What

Add a PostgreSQL job to the Deployer Concourse machine and configure ATC to
use that instead of Amazon RDS. I've taken this configuration from
concourse-lite, which does the same thing:

- https://github.com/concourse/concourse-lite/blob/0ca5964c0f4b0e25b03166f4dd0c403fd5c08426/templates/concourse_manifest.yml#L24-L40

We're doing this because:

- RDS is incredibly slow to provision and destroy. It's not uncommon for it
  to take up to 20 mins. Compiling `postgresql_9.3` can still take 3mins,
  but that's much faster.
- we are not *that* concerned about the data stored in PSQL at the moment.
  It contains build history and some other things that we will want to
  backup in the near future, but all of the important state is in S3 and
  our one worker is a single point of failure anyway.
- as an aside, using RDS is more expensive and increases our integration
  points, which will make things like automatically shutting down
  environments at night more difficult/less effective.

NB: This change will apply on top of an existing environment but you will:

- lose all pipeline configs and build history (which come from PostgreSQL).
  New builds will still work after deploying the pipelines though, because
  that state comes from S3.
- have a short outage of the Deployer Concourse while `bosh-init` replaces
  the instance and reattaches the disk.
- possibly need to unmount the disk manually, see:
  https://www.pivotaltracker.com/story/show/111149770

## How to review

1. Deploy and run the pipelines `create-deployer` on your "Bootstrap Concourse" and `create-microbosh` on your "Deployer Concourse" (if you haven't already)
2. Update the `create-deployer` pipeline on your "Bootstrap Concourse" to use this branch: `BRANCH=109972384-remove_rds ./vagrant/deploy.sh`
3. Run the `create-deployer` pipeline again and wait for it finish
4. Login to the "Deployer Concourse" again: https://github.com/alphagov/paas-cf#login-to-deployed-concourse
5. Deploy the pipelines to "Deployer Concourse" again: https://github.com/alphagov/paas-cf#pipelines-create-microbosh-and-destroy-microbosh
6. Run the `create-microbosh` pipeline again and confirm that it works (should have no changes)

## Who can review

Anyone but @dcarley


mtekel:
Tried to deploy, got an error in concourse-terraform step:
```
terraform-apply

fix these before continuing.

Errors:

  * 4 error(s) occurred:

* Required variable not set: vpc_id
* Required variable not set: subnet0_id
* Required variable not set: subnet1_id
* Required variable not set: subnet2_id
```

alext:
> Tried to deploy, got an error in concourse-terraform step:

That looks like the usual race condition that should be fixed by #26.

PR #34, opened 2016-01-11
[#109878560] Secure external access to concourse

saliceti:
# What
Allow external access to Concourse via HTTPS only.

We create an ELB with HTTPS in front of Concourse and close external access to Concourse via HTTP 8080.

The default certificate used is the wildcard self-signed certificate stored in `paas-pass` and uploaded manually to AWS.

# How to review
* Build a new Concourse and connect to `https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk/`.
* Check that insecure access via HTTP 8080 is closed.
* Note: the certificate is self-signed so fly won't accept it unless you login with the -k option

# Who can review
Anyone but @actionjack or @saliceti

### DIFF
--- a/terraform/concourse/outputs.tf
+++ b/terraform/concourse/outputs.tf
@@ -9,3 +9,7 @@ output "concourse_security_group" {
 output "concourse_security_group_id" {
   value = "${aws_security_group.concourse.id}"
 }
+
+output "concourse_elb_name" {
+  value = "${aws_elb.concourse.name}"
+}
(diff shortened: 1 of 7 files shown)


terraform/concourse/outputs.tf
```diff
@@ -9,3 +9,11 @@ output "concourse_security_group" {
 output "concourse_security_group_id" {
   value = "${aws_security_group.concourse.id}"
 }
+
+output "concourse_elb_name" {
+  value = "${aws_elb.concourse.name}"
+}
+
+output "concourse_dns_name" {
+  value = "${aws_route53_record.concourse.name}"
```
alext: Is this output actually used anywhere?  If not, can we remove it until it is needed.

alext:
This looks good to me.  The build is red apparently due to rubies.travis-ci.org being down.

PR #36, opened 2016-01-13
Quote string with variable interpolation in README

dcarley:
So that I paste the command into my shell (zsh) without it automatically
escaping the curly braces.


keymon:
LGTM

PR #37, opened 2016-01-13
[#111137450] Seed bosh-init installation_id to use cached CPI

alext:
This seeds the installation_id for bosh-init runs so that it picks up the cached CPI saving 3-5 mins for each bosh-init run.

See alphagov/paas-docker-cloudfoundry-tools#12 for full details.


manifests/bosh-manifest/deployments/aws/010-bosh-aws-cpi.yml
```diff
@@ -1,3 +1,6 @@
+# When updating the version of the CPI here, the version used in the
+# bosh-init container must also be updated so that the cached CPI compile
+# will be used.
```
dcarley: Could you move this comment just above or below `name: bosh-aws-cpi` (like `concourse-base.yml`) so that it stays relevant if we happen to add any more releases to this manifest.

mtekel:
The fact that it has to be done this way is just because how bosh-init works. If it had different/better behaviour, the PRs could be nicer. Nevertheless this fixes the problem in the most nicest and clean way. For me this is a +1 and should be merged.

manifests/bosh-manifest/deployments/aws/010-bosh-aws-cpi.yml
alext: done

dcarley:
I tried testing this by adding 8a1ff49d08dbb77072d8d151fdb601d851a6f013 to this branch, which uses the container that has been built from the other PR, and deploying to my Bootstrap Concourse with:
```
➜  paas-cf git:(wip_dcarley) BRANCH=wip_dcarley ./vagrant/deploy.sh dcarley
```
However it compiled the CPI again:
```
Started installing CPI
  Compiling package 'ruby_aws_cpi/a5b66d011ce1b31642ff148ea2c9097af65ff78c'... Finished (00:02:54)
  Compiling package 'bosh_aws_cpi/dfde48910067fbb2ef747bf0508f1bd9737214f1'... Finished (00:01:52)
  Installing packages... Finished (00:00:02)
  Rendering job templates... Finished (00:00:00)
  Installing job 'aws_cpi'... Finished (00:00:00)
Finished installing CPI (00:04:49)
```
Against a new installation ID:
```
root@aan5ahqomc3:/tmp/build/f7ebecd0-df55-47d1-6534-14f65d84a8b8# ls -al /root/.bosh_init/installations/
total 16
drwxr-xr-x 4 root root 4096 Jan 14 16:35 .
drwxr-xr-x 6 root root 4096 Jan 14 16:35 ..
drwxr-xr-x 5 root root 4096 Jan 14 10:38 44f01911-a47a-4a24-6ca3-a3109b33f058
drwxr-xr-x 5 root root 4096 Jan 14 16:38 9610dfb7-8925-49e9-648a-d4cd5189c8fc
```
Have I missed something?

alext:
> However it compiled the CPI again:
> Against a new installation ID:

That can happen if there's already a bosh-init state file in the S3 bucket. The seeding of the state with the correct installation_id only happens at the start of the first pipeline if no state file already exists.

dcarley:
Yep, that was it. Thanks @alext and @mtekel. I had to download the state file, modify the ID to match this PR, and upload it again:
```
➜  paas-cf git:(master) ✗ aws s3 cp s3://dcarley-state/concourse-state.json .
download: s3://dcarley-state/concourse-state.json to ./concourse-state.json
➜  paas-cf git:(master) ✗ vi concourse-state.json
➜  paas-cf git:(master) ✗ aws s3 cp concourse-state.json s3://dcarley-state/
upload: ./concourse-state.json to s3://dcarley-state/concourse-state.json
```
Worth nothing that other people won't be able to benefit from this until they do the same or create a new environment. Though that's fine, because It's a non-breaking change and our codebase is in an early state.

dcarley:
Waiting for alphagov/paas-docker-cloudfoundry-tools#12 to be built: https://hub.docker.com/r/governmentpaas/bosh-init/builds/

dcarley:
I had to manually trigger the hub.docker.com build. Noted in Pivotal.

PR #38, opened 2016-01-13
Don't trigger create-microbosh from paas-cf

alext:
We don't (yet) want this pipeline to be triggered on every commit to
git. This is different to all the other pipelines and therefore causes
surprise.


mtekel:
works

PR #41, opened 2016-01-13
[#110234556] Stop using the insecure deployer - concourse / deployer

timmow:
> Note: There is an alternative implementation in #43 both should be compared and discussed.

**What**

We want to stop using the insecure deployer key

This PR should reviewed in the following context:

* We are no longer using the insecure deployer key for deployer vm
* It is still possible to ssh to the deployer vm

**How to test this PR**
Check the branch out, run the deployment script.

```
./vagrant/deploy.sh
```

Then run the create-deployer pipeline

When the create-deployer pipeline has finished, check your concourse/0 vm does not use the insecure-deployer keypair

Verify you can ssh into the concourse vm - Find and download the deployer-key in the s3 state bucket, then ssh to the instance using this key. 

```
ssh -i <deployer-key-file-from-state-bucket> vcap@{DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk
```

**Who should review this PR**

* Anyone but me or @mtekel

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,17 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
    fly -t $FLY_TARGET login -k -c https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk
 ```
 
+#### ssh to deployed Concourse
+
+During the deploy process, a keypair is created in aws to create the instance
+with. This is uploaded to s3 and then discarded, to avoid the private key being
+left anywhere in the pipeline. To ssh to the instance, find its ip in the
+console and download the deployer-key file from the s3 state bucket, then run
+
+```
+ssh -i <deployer-key-file-from-state-bucket> vcap@${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk
+```
+
 ### Microbosh deployment from concourse bootstrap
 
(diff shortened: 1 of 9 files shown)


mtekel:
The idea here is to generate one-time keys for bosh-init and then throw them away. Have different keys for deployer, bosh-micro and VMs bosh deploys (this PR covers only deployer creds.). Because we still might want to ssh to the deployer (before we implemet other ways to access it, e.g. vault), we store the generated key in s3 bucket.

mtekel:
One more thing can be improved: we don't need to destroy the uploaded temporary key in the destroy pipeline, but could do that just after bosh-init completes. That would simplify this PR a bit

mtekel:
One inconvenience/open issue is that we always overwrite the private key even though bosh init doesn't re-create the instance. The old instance key is not lost, as bucket is versioned, but you'd have to find it manually. The fix is to check if sha1hash of the manifest has changed and skip whole [generate new key, upload new key, do bosh init] chain.

mtekel:
This PR covers only generating credentials for deployer. The new [PR](https://github.com/alphagov/paas-cf/pull/43) takes different approach: it generates key once and then uses it for all VMs of the environment: deployer, bosh init and also CF VMs.

PR #42, opened 2016-01-14
[#110234556] Fix bosh secret generation

mtekel:
s3init currently also downloads the file. That's why your init file
and main file name have to be different files. Otherwise you'll over-
write your init file in the checking process and instead initialize
with the file containing S3 API reply.


keymon:
LGTM, fixes the issue

PR #43, opened 2016-01-14
[#110234556] Stop using the insecure deployer

keymon:
https://www.pivotaltracker.com/story/show/109897246

**What**

We want to stop using the insecure deployer key

This PR should reviewed in the following context:

* We are no longer using the insecure deployer key for deployer vm
* It is still possible to ssh to the deployer vm

**How to test this PR**

 1. Check the branch out
 2. run the deployment as described in the README.

You can check the PR as you deploy, so:

Once the concourse deployed is created (`create-deployer` pipeline):

 * check your concourse/0 vm does not use the insecure-deployer keypair by checking in the AWS interface.
 * Verify you can ssh into the concourse vm: Find and download the `id_rsa` in the s3 state bucket, then ssh to the instance using this key. You can reuse the `s3get.sh` script:
    ```
./concourse/scripts/s3get.sh ${DEPLOY_ENV}-state id_rsa && chmod 400 id_rsa && ssh-add ./id_rsa
ssh vcap@<concourse_ip> -v -A  
```

   There should be a debug message indicating the SSH key used to login:
    ```
debug1: Offering RSA public key: /home/keymon/gds/paas-cf/id_rsa
debug1: Server accepts key: pkalg ssh-rsa blen 535
debug1: Authentication succeeded (publickey).
```

Once `create-microbosh` finishes, check the same for `bosh/0`. In this case, you must SSH from the concourse machine with agent forwarding (option `-A`)

```
ssh -v vcap@10.0.0.6
```

Finally, once Cloud Foundry is deployed you can check the same in the deployed VMs. Check the IP in the AWS console.

**Who should review this PR**

* Anyone but @keymon or @timmow  or @mtekel

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,6 @@ export AWS_ACCESS_KEY_ID=XXXXXXXXXX
 export AWS_SECRET_ACCESS_KEY=YYYYYYYYYY
 ```
 
-#### SSH credentials
-
-You must have insecure-deployer private key present in `~/.ssh/insecure-deployer` location, 
-as it is being used by pipeline creation scripts.
-
 #### Deployment of bootstrap concourse-lite
 
 Execute vagrant provision script `./vagrant/deploy.sh` to initialise a vagrant
@@ -109,6 +104,39 @@ export FLY_TARGET=$DEPLOY_ENV
 
 echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
    fly -t $FLY_TARGET login -k -c "https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk"
+```
+
+#### SSH to deployed Concourse and microbosh
+
+In the `create-deployer` pipeline when creating the initial VPC,
+a keypair is generated and uploaded to AWS to be used by deployed instances.
+`bosh-init` needs this key to be able to create a SSH tunnel to
+forward some ports to the agent of the new VM.
+
+Both public and private keys are also uploaded to S3 to be consumed by
+other jobs in the pipelines as resources and/or by us for troubleshooting.
+
+To manually ssh to the deployed concourse, learn its IP via AWS console and
+download the `id_rsa` file from the s3 state bucket.
+
+For example:
+
+```
+./concourse/scripts/s3get.sh "${DEPLOY_ENV}-state" id_rsa && \
+chmod 400 id_rsa && \
+ssh-add $(pwd)/id_rsa
+
+ssh vcap@<concourse_ip>
+```
+
+microbosh is deployed to use the same SSH key, although is not publically
+accessible. But you can connect via the concourse host and forward SSH
+credentials (option -A):
+
+```
+ssh vcap@<concourse_ip> -A
+
+ssh vcap@10.0.0.6
 ```
 
 ### Microbosh deployment from concourse bootstrap
--- a/concourse/pipelines/create-microbosh.yml
+++ b/concourse/pipelines/create-microbosh.yml
@@ -77,6 +77,15 @@ resources:
       secret_access_key: {{aws_secret_access_key}}
       key: {{pipeline_trigger_file}}
 
+  - name: ssh-private-key
+    type: s3
+    source:
+      bucket: {{state_bucket}}
+      access_key_id: {{aws_access_key_id}}
+      secret_access_key: {{aws_secret_access_key}}
+      versioned_file: id_rsa
+      region_name: {{aws_region}}
+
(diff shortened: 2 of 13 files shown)


keymon:
The idea in this repo is do changes that allows us: 
 * Generate a unique key per environment
 * But still be able to retrieve the SSH key for troubleshooting debugging.

This PR assumes that is OK to store the SSH key in concourse resources, because concourse has permissions to redeploy or do any change on the machine anyway using AWS API.

keymon:
One doubt I got is:
 * Shall we create a different container with `openssh` package to run `ssh-keygen`, and setup 2x tasks, one generate and other upload with s3init.
 * Or shall we install `openssh` package in the curl-ssl container and run this in a unique task https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/master/curl-ssl/Dockerfile#L3

timmow:
I agree this is simpler - just to add to the PR some discussion we had earlier today, the decision to implement the way we did in #41 came from the following in the pivotal story -

> - Ideally we can generate the key pair as close to the environment as possible so we don't have to pass it around. 

I'm not arguing for the approach in #41 but that was part of our thought process for doing it that way

README.md
```diff
@@ -111,6 +106,30 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
    fly -t $FLY_TARGET login -k -c "https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk"
 ```
 
+#### SSH to deployed Concourse and microbosh
+
+During the deploy process, a keypair is created in aws to create the instance
```
combor: Can you please rephrase it so it's clear what the process is?

concourse/pipelines/create-microbosh.yml
```diff
@@ -225,7 +240,8 @@ jobs:
 
             ./build_manifest.sh \
               ../../../bosh-secrets/bosh-secrets.yml \
-              ../../../terraform-outputs/terraform-outputs.yml \
+              ../../../terraform-outputs/bosh.terraform-outputs.yml \
```
combor: Can you refactor it in a way that is done [here](https://github.com/alphagov/paas-cf/pull/28/files#diff-35adbaaa48f824ce18c1838d3dc9342fR208)

You can specify `MANIFESTS_DIR` as a task parameter that acts a a prefix dir for manifests.

keymon: I was about to do it, but there are a lot of things to change. First the script is different in both cases (one uses args, other gets vars) and then I find that other things need refactoring (e.g. `inputs` and `outputs` should be before `run`, etc).

Also, the different pipelines use different "patterns" to solve the same problems.

I would rather suggest that we do a PR to clean up and homogenise the pipelines, and do not change it here.

README.md
keymon: Can you check now?

concourse/pipelines/create-microbosh.yml
combor: Yes, I wasn't aware that we've got two scripts with the same name to spruce manifests but in different directories and working in a different way. It would be good to have just one that we can parametrise. I'll put this as a suggestion in [Pipelines improvements](https://www.pivotaltracker.com/n/projects/1275640/stories/111700216).

README.md
combor: It's still quite not right: 
>`bosh-init` needs this key to be able  a know

[Line 113](https://github.com/alphagov/paas-cf/pull/43/files#diff-04c6e90faac2675aa89e2176d2eec7d8R113)

keymon: sorry, and now?

combor: LGTM now.

combor:
I guess that instructions to check removal of insecure_deployer for microbosh and CF are not relevant here?

combor:
OK, this works. Merging.

alext:
I'm getting errors in the destroy-deployer pipeline.  The `destroy-vpc` job fails with:

```
fix these before continuing.

Errors:

  * file: open /tmp/build/965b4427-bcc6-4015-57f4-4bf3980ef0e6/paas-cf/terraform/vpc/id_rsa.pub: no such file or directory in:

${file("${path.module}/id_rsa.pub")}
```

mtekel:
Yes, we forgot to transfer it from the other branch: https://github.com/alphagov/paas-cf/commit/8fc1a33c0f9822597a3eb70d8504e945e771430b#diff-68bd8f758ea6ed051aecfd861fcd67d8R169

PR #44, opened 2016-01-14
Use $DEPLOY_ENV instead of <deploy_env> in README

mtekel:
No need to type in deployment_environment name again, as it is set
already in $DEPLOY_ENV variable used one line above. Update README
to reflect this. This way we can also simply copy & paste the
command from README directly.


keymon:
:+1: handy, merging

PR #48, opened 2016-01-15
[#111303414] Use IAM role in bootstrap-concourse machine

alext:
## What

Remove most usage of access keys  from the bootstrap-concourse pipelines.

## Detail

This updates the `{create,destroy}-deployer` pipelines to use IAM profiles for Terraform and the s3init scripts

The only thing left using access keys after this is the concourse resources themselves. These will be switched over to use profiles in another story.

This PR is the companion to https://github.gds/government-paas/terraform-iam/pull/5 which defines the roles in use here.  I've manually created roles with matching permissions in the trial account so that this can be tested.

## How to review

Deploy a new concourse-lite instance (it'll need to be newly provisioned from this branch in order to be given the necessary role):

```
BRANCH=bootstrap_concourse_role ./vagrant/deploy.sh $DEPLOY_ENV
```

Run both the create-deployer and destroy-deployer pipelines and verify that they successfully create/destroy everything.

## Who can review

Anyone but me.

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -89,8 +89,6 @@ jobs:
         image: docker:///governmentpaas/curl-ssl
         params:
           AWS_DEFAULT_REGION: {{aws_region}}
-          AWS_ACCESS_KEY_ID: {{aws_access_key_id}}
-          AWS_SECRET_ACCESS_KEY: {{aws_secret_access_key}}
         inputs:
           - name: paas-cf
         outputs:
@@ -110,10 +108,6 @@ jobs:
         params:
           TF_VAR_env: {{deploy_env}}
           AWS_DEFAULT_REGION: {{aws_region}}
-          AWS_ACCESS_KEY_ID: {{aws_access_key_id}}
-          AWS_SECRET_ACCESS_KEY: {{aws_secret_access_key}}
-          TF_VAR_AWS_ACCESS_KEY_ID: {{aws_access_key_id}}
-          TF_VAR_AWS_SECRET_ACCESS_KEY: {{aws_secret_access_key}}
         inputs:
           - name: paas-cf
           - name: bucket-state
@@ -133,9 +127,6 @@ jobs:
     - task: s3init-concourse
       config:
         image: docker:///governmentpaas/curl-ssl
-        params:
-          AWS_ACCESS_KEY_ID: {{aws_access_key_id}}
-          AWS_SECRET_ACCESS_KEY: {{aws_secret_access_key}}
         run:
           path: sh
           args:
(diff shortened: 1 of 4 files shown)


mtekel:
This works, furthermore, I have removed AWS_DEFAULT_REGION from TF destroy pipeline:
```
@@ -130,7 +130,7 @@ jobs:
         - name: vpc-terraform-outputs
         - name: concourse-terraform-state
         params:
-          AWS_DEFAULT_REGION: {{aws_region}}
+          #AWS_DEFAULT_REGION: {{aws_region}}
           TF_VAR_env: {{deploy_env}}
         run:
           path: sh
@@ -162,7 +162,7 @@ jobs:
         image: docker:///governmentpaas/docker-terraform
         params:
             TF_VAR_env: {{deploy_env}}
-            AWS_DEFAULT_REGION: {{aws_region}}
+            #AWS_DEFAULT_REGION: {{aws_region}}
         inputs:
           - name: paas-cf
           - name: vpc-terraform-state
@@ -192,7 +192,7 @@ jobs:
           image: docker:///governmentpaas/docker-terraform
           params:
               TF_VAR_env: {{deploy_env}}
-              AWS_DEFAULT_REGION: {{aws_region}}
+              #AWS_DEFAULT_REGION: {{aws_region}}
           inputs:
             - name: paas-cf
             - name: bucket-terraform-state
```
I have tested destroy including this new commit. I'll leave it up to you if you want to squash it or merge.

dcarley:
I think we should leave all of the region changes to a [separate story](https://www.pivotaltracker.com/story/show/111508486). It deserves more attention to:
- consider whether we do need to be able to override the region it's deployed to (probably)
- pass it consistently and explicitly to Terraform with `-var region=`
- stop passing the `aws_region` variable with `fly` if it's no longer used

alext:
> I think we should leave all of the region changes to a separate story

I've rebased that back out again.  I agree it should be in a separate story.  There are some other things that need looking at relating to this, like the region being hardcoded in `s3init.sh` etc.

mtekel:
Current state is not good: AWS_DEFAULT_REGION is removed from creating terraform, but not from destroying part, indicating that you want to be able to override where you destroy, but not create. IMO we should clean up pipeline as much as possible. If TF currently specifies region elsewhere, I'd leave it so and then consider what is the best solution in that separate story, which might chose to re-introduce that env. variable (but remove duplicate region definitions from TF files) or do other solution.

dcarley:
We should retain `AWS_DEFAULT_REGION` in `create-deployer` too. @alext can you rebase 90d1b3207de3d94910c9353075a69bbf3d3c9893?

mtekel:
I would chose to not pile tech debt. DEFAULT region is currently not needed and can confuse people. Moreover, pipeline has already been tested with that variable removed...

alext:
> We should retain AWS_DEFAULT_REGION in create-deployer too.

Rebased back in.

saliceti:
Works fine, merging.

PR #51, opened 2016-01-18
[#110059806] Generate passwords automatically in an environment

actionjack:
## What
[Generate passwords automatically in an environment](https://www.pivotaltracker.com/story/show/110059806)

* Since the scope of this story is only secrets related to passwords and not certificates or private keys
  * I want to Move SSL keys and certs into a dedicated file so they can be handled separately later on.
* I want to modify the existing `deploy-cloudfoundry` pipeline to generate and cloud foundry secrets

## How to Review this PR

* Provision the environment using [vagrant aws](https://github.com/alphagov/paas-cf/blob/master/README.md) to bootstrap concourse in AWS
* Deploy cloudfoundry pipeline against the concourse deployed in AWS (not your concourse-lite) i.e:
 ```
$ fly --target remote login -k -c https://$DEPLOY_ENV-concourse.cf.paas.alphagov.co.uk

$ BRANCH=110059806-cf-secrets LOG_LEVEL=DEBUG  FLY_TARGET=remote \
CONCOURSE_DB_PASSWORD=your_atc_db_password CONCOURSE_ATC_PASSWORD=your_atc_password \
./concourse/scripts/deploy-cloudfoundry.sh smashmouth`
```
* Trigger s3init job in $DEPLOY_ENV-concourse.cf.paas.alphagov.co.uk deploy-cloudfoundry pipeline to test
 
## Who should review this PR

* This PR should be reviewed by any man, child or woman on the face of this earth with the exception of @saliceti or @actionjack who worked on this PR.

### CODE REVIEW

**Comment by actionjack** [comment]
@keymon: Good point I've updated the commit.

**Comment by dcarley** [comment]
@keymon Could you add comments inlines on the "files changed" tab rather than individual commits? That way the comments won't get lost when the branch is rebased and it's easier to review any comments that are still outstanding.

**Comment by keymon** [comment]
@dcarley Actually the trick is go to "files changed" and then click on the "unified" view. In unified view I think the commits are as you mention, but split view are in a specific commit.

File: concourse/scripts/generate-cf-secrets.sh [hunk 1]
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+set -x
+
+SCRIPTS_DIR=`dirname $0`
+OUTPUT_FILE=$1
+
+pwgen() {
+   openssl rand -base64 12 | sed 's/+/0/;s/\//0/'
```
keymon: Yes, I think this is good. It will avoid issues with invalid characters.

concourse/scripts/generate-cf-secrets.sh
```diff
@@ -0,0 +1,42 @@
...
+}
+cat > $OUTPUT_FILE <<EOF
+---
+  secrets:
+    vcap_password: $(pwgen)
```
keymon: This needs to be generated in a different way, it must be a sha-512 of the password. For instance  `pwgen | mkpasswd -m sha-512 -s` 

But it would be convenient if you store also the cleartext password in this file, to be used for troubleshooting...

concourse/scripts/generate-cf-secrets.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+set -x
```
keymon: can you remove this and make it less verbose?

keymon:
The pipeline looks funny because it looks like the generate passwords writes on paas-cf, but there it is not connected to the `cf-secrets` resource. But I guess this is the limitation of uploading the `cf-secrets.yml` using the task instead of the resource.

screenshot from 2016-01-18 18 10 43

keymon:
I wonder if you can also do the same change in https://github.com/alphagov/paas-cf/blob/110059806-cf-secrets/concourse/scripts/generate-bosh-secrets.sh#L4--1, by adding `sed 's/+/0/g;s/\//0/g'`

concourse/scripts/generate-cf-secrets.sh
keymon: Actually, this won't work in the case of repeated `+` or `/`. Can you please add the modifier `g` to sed:

`sed 's/+/0/g;s/\//0/g'`

alext: Instead of having to deal with 2 problems (now that we're using a regex), we could use `-hex` instead of `-base64` in the openssl command. That would ensure we don't have any problematic characters.

keymon:
LGTM, but waiting for CF to deploy to confirm.

dcarley:
> This branch has conflicts that must be resolved
Looks like this needs a rebase - it contains some commits that are already on `master`.

Depending on which PR gets merged first, we'll also need to update the additional s3 resource to use profiles per #57 (Hector knows/noted this).

keymon:
CF deploys OK, testing the part of bosh.

keymon:
Some unit testing I did:

```
 docker run -t -v $(pwd):/code governmentpaas/mksecrets /code/concourse/scripts/generate-bosh-secrets.sh  | tee /tmp/bosh-secrets.yml | ruby -ryaml -e "puts YAML::load(STDIN.read).to_yaml"
 docker run -t -v $(pwd):/code governmentpaas/mksecrets /code/concourse/scripts/generate-cf-secrets.sh  | tee /tmp/cf-secrets.yml | ruby -ryaml -e "puts YAML::load(STDIN.read).to_yaml"
```

keymon:
All LGTM, merging

PR #52, opened 2016-01-18
README: Replace ForwardAgent with ProxyCommand

dcarley:
Using `ProxyCommand` is preferable to `ForwardAgent` because it doesn't
expose your local agent to other users on the remote machine.

This is a reasonable blog post, for further reading on about the issue:

- https://heipei.github.io/2015/02/26/SSH-Agent-Forwarding-considered-harmful/

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,14 +129,11 @@ ssh-add $(pwd)/id_rsa
 ssh vcap@<concourse_ip>
 ```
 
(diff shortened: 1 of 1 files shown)


saliceti:
`ProxyCommand` works fine. It's a bit more complex to use, I think we need a Makefile, for this and other commands.

PR #53, opened 2016-01-18
[#109878560 #109897246] README: De-dupe incorrect SSH instructions

dcarley:
These instructions were no longer working because:

- 48db141 changed the hostname from an EIP (which you could SSH to) to an
  ELB (which only forwards HTTP[S])
- 3a5e3f1 (and preceding commits) changed the SSH key that's used


jimconner:
LGTM - merging.

PR #54, opened 2016-01-19
Ensure bosh-init state is pushed on failure.

alext:
If the bosh-init deployment job fails for some reason, we still want to
push the state file so that future runs know about the VM if it's been
created, and can retry without conflicting with this VM.


mtekel:
Works as expected (like in other pipelines).

PR #55, opened 2016-01-19
[#111778374] Remove use of access keys in deployer concourse pipelines

alext:
This removes the remaining usage of access keys in the deployer-concourse pipelines

This does not cover removing them from the concourse resources themselves (covered in a separate story), or from the cloudfoundry manifest (also covered in a separate story).

This relies on some additional permissions that are defined in https://github.gds/government-paas/terraform-iam/pull/6, and have been manually applied to the trial account.

## How to review

Run the create-microbosh and deploy-cloudfoundry pipelines from this branch and verify that they run successfully.

Run the corresponding destroy pipelines and verify that they also work as expected.

## Who can review

Anyone but me.


mtekel:
works...

PR #56, opened 2016-01-19
[#111590274] implement nat

mtekel:
### What
Implement new AWS NAT gateways and use them for CF subnets. [story](https://www.pivotaltracker.com/projects/1275640/stories/111590274).

### Testing
Deploy new environment or apply against existing. Verify that:
* NAT gateway was created in each zone and placed into infrastructure subnet
* CF subnet routing table has `0.0.0.0/0` route pointing to the NAT gateway in the same zone
* CF VMs have internet connectivity. You can ssh into the VMs and do e.g. `curl google.com`

### Reviewing
Not @mtekel or @combor who worked on this story.


terraform/cloudfoundry/variables.tf
```diff
@@ -36,6 +32,6 @@ variable "elb_idle_timeout" {
   default     = 300
 }
 
-variable "subnet0_id" {
-    description = "Subnet that is used to provision ELB"
+variable "internet_gateway" {
+  description = "Default internet gateway"
```
alext: What's this variable used for?

combor: It is leftover. Removed now.

terraform/concourse/elb.tf
```diff
@@ -1,6 +1,6 @@
 resource "aws_elb" "concourse" {
   name            = "${var.env}-concourse"
-  subnets         = ["${var.subnet0_id}"]
+  subnets         = ["${split(",", var.infra_subnet_ids)}"]
```
saliceti: Concourse is only deployed in the first infra subnet. Does it make sense to configure all subnets in the ELB?

saliceti:
Tested end-to-end successfully.

terraform/concourse/elb.tf
mtekel: Good point. We wanted to make current TF deployment a bit more general. With this config you can place concourse into any subnet without ELB reconfiguration (e.g. in zone failure).

I also think we should use concourse_subnet (like we use bosh_subnet) instead of subnet0 in both TF variable name and in concourse manifest. But that's a bit out of scope of this PR.

dcarley:
Good PR - small and descriptive commits, plus tidying up small pieces of related tech debt :+1:

PR #57, opened 2016-01-19
[#111760326] use custom s3 and semver resources with iam support

keymon:
[#111760326 Use custom resources for S3 & Semver using IAM profiles](https://www.pivotaltracker.com/story/show/111760326)

# What

We'd like to use custom resources for S3 and Semver because this will allow us to use IAM profiles for provisioning. We will use the images generated automatically in docker hub: https://hub.docker.com/r/governmentpaas/s3-resource/builds/ and https://hub.docker.com/r/governmentpaas/semver-resource/builds/, which where setup in https://github.com/alphagov/paas-s3-resource/pull/2 and https://github.com/alphagov/paas-semver-resource/pull/3

We will also stop using static AWS credentials in the pipelines for these resources.

# Context

 * I want to add the new resources in concourse lite, using a shell provisioner to change the required configuration. Before that, we refactor how post-deployment scripts are executed in vagrant to make it easier to add/remove steps.
 * I want to configure the new resources in the "deployer" concourse, by changing the manifest to include the required configuration.
 * I want to remove the credentials used by this resources in all the pipelines, and not even pass it as variables. The only exception is the `deploy-cloudfoundry` pipeline, which needs the credentials to setup S3 access for the cloud-controller.

# How to test this PR

Follow the steps described in the README.md deploying **from scratch** (at least, destroy and recreate vagrant, as concourse deployer will be redeployed).

Remember to specifying the current branch by exporting or passing the `$BRANCH` variable: 

```
```

It is not required to wait until the cloudfoundry deploy job finish, as the CF deployment has not being changed.

# Who? 

Anyone but @dcarley or @keymon

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -17,65 +17,51 @@ resources:
       branch: {{branch_name}}
 
   - name: build-all-trigger
-    type: semver
+    type: semver-iam
     source:
       bucket: {{state_bucket}}
       region_name: {{aws_region}}
-      access_key_id: {{aws_access_key_id}}
-      secret_access_key: {{aws_secret_access_key}}
       key: build-all-trigger
 
   - name: bucket-terraform-state
-    type: s3
+    type: s3-iam
     source:
       bucket: {{state_bucket}}
       region_name: {{aws_region}}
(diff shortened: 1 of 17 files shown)


keymon:
Rebased after #51

PR #58, opened 2016-01-20
Refactor pipeline scripts

dcarley:
### What

Refactor the pipeline scripts to reduce duplication and make them more consistent. See individual commits for more details.

### How to test

Deploy all of the pipelines to the Bootstrap and Deployer Concourse machines. You probably don't need to run them, so long as they are created and un-paused successfully.
```
./vagrant/deploy.sh && \
export FLY_TARGET=$DEPLOY_ENV && \
./concourse/scripts/create-microbosh.sh && \
./concourse/scripts/deploy-cloudfoundry.sh && \
./concourse/scripts/destroy-cloudfoundry.sh && \
./concourse/scripts/destroy-microbosh.sh
```

### Who can review

Anyone but @dcarley

### DIFF
--- a/concourse/scripts/create-deployer.sh
+++ b/concourse/scripts/create-deployer.sh
@@ -2,9 +2,6 @@
 set -e
 
 SCRIPT_DIR=$(cd $(dirname $0) && pwd)
-ATC_URL=${ATC_URL:-"http://192.168.100.4:8080"}
-FLY_TARGET=${FLY_TARGET:-$ATC_URL}
-FLY_CMD=${FLY_CMD:-fly}
 
 env=${DEPLOY_ENV:-$1}
 pipeline="create-deployer"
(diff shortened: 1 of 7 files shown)


saliceti:
We could simplify even more by removing ATC_URL completely. It is not documented and only used as a default for FLY_TARGET.

We could set FLY_TARGET to reasonable defaults like "$DEPLOY_ENV-bootstrap" in *-deployer.sh scripts and "$DEPLOY_ENV" in the other ones. Ex in create-deployer.sh:

```
export FLY_TARGET=${FLY_TARGET:-$DEPLOY_ENV-bootstrap}
   "${env}" "${pipeline}" "${config}" <(generate_vars_file)
```

And we should check that FLY_TARGET was properly set in deploy-pipeline.sh

dcarley:
Agree with all those suggestions :+1: 

But I'll do it in a separate PR after this, because it's more complex/controversial than these changes.

dcarley:
I've also updated the PR description to include testing instructions - bad me.

saliceti:
OK tested successfully.

PR #59, opened 2016-01-20
[#108441228] Add pipelines to automatically delete cloudfoundry deployment and concourse-lite at night

keymon:
[#108441228 Automatically shutdown dev environments at night / weekends](https://www.pivotaltracker.com/story/show/108441228)

## Dependencies

The PR https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/14 should be merged first and we should update the container used in `./concourse/pipelines/ticking-time-bomb.yml`

## What

Given the size of a deployed Cloud Foundry environment and the associated cost of the VMs we want to shut down all development environments at night and over weekends to minimise the overall cost of infrastructure. 

It is acceptable to leave a small number of machines running (e.g. BOSH and Concourse) if this expedites the recovery of the development environment in the morning.

## Context

 * I want to add a job to delete the cloudfoundry deployment:
   * it will be triggered at 20:00 GMT
   * will be ready to run whenever we deploy CF
   * it will try to delete the deployment regardless if the last deployment was successful or not.

 * I want to add a pipeline to terminate the concourse-lite and triggered at 20:00 GMT

## How to test this PR?

Vagrant:

 1. Start a vagrant instance as described in the `README.md`: `./vagrant/deploy.sh <env>`
 2. Change the times in the resource `destroy-timer` in `./concourse/pipelines/ticking-time-bomb.yml`, to a window that includes the current time.
 3. Rerun `./vagrant/deploy.sh <env>` to see the changes.
 4. Go to http://localhost:8080/pipelines/ticking-time-bomb and observe how concourse-lite commits suicide :smile: 

For Cloud Foundry deployment:

 1. Follow the guidelines in the `README.md` to fully deploy CF.
     * It does not need to finish successfully
     * you can even cancel the deploy tasks during the deployment to save time if you want.
 2. Change the times in the resource `delete-timer` in `./concourse/pipelines/deploy-cloudfoundry.yml`, to a window that includes the current time.
 3. Upload the pipeline again: 
     ```
export FLY_TARGET=$DEPLOY_ENV
./concourse/scripts/deploy-cloudfoundry.sh $DEPLOY_ENV
```
 4. The `delete` task should be triggered and delete the deployment.

# who?

Anyone but @keymon or @actionjack

### DIFF
--- a/concourse/pipelines/concourse-lite-self-terminate.yml
+++ b/concourse/pipelines/concourse-lite-self-terminate.yml
@@ -1,0 +1,31 @@
+resources:
+  - name: delete-timer
+    type: time
+    source:
+      start: 20:00 -0000
+      stop: 21:00 -0000
+      interval: 24h
+
+jobs:
+  - name: delete
+    serial: true
+    plan:
+    - get: delete-timer
+      trigger: true
+    - task: delete-deployment
+      config:
+        inputs:
+        - name: delete-timer
+        image: docker:///governmentpaas/awscli
+        run:
+          path: sh
+          args:
+          - -e
+          - -c
+          - |
+            export AWS_AVAILIABILITY_ZONE=$(curl -qs http://169.254.169.254/latest/meta-data/placement/availability-zone)
+            export AWS_DEFAULT_REGION="${AWS_AVAILIABILITY_ZONE%%[a-z]}"
+            export AWS_INSTANCE_ID=$(curl -qs http://169.254.169.254/latest/meta-data/instance-id)
+
+            aws ec2 terminate-instances --instance-ids ${AWS_INSTANCE_ID}
+
--- a/concourse/scripts/concourse-lite-self-terminate.sh
+++ b/concourse/scripts/concourse-lite-self-terminate.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -e
+set -u
+
+SCRIPT_DIR=$(cd $(dirname $0) && pwd)
+FLY_CMD=${FLY_CMD:-fly}
+
+env=${DEPLOY_ENV:-$1}
+pipeline="self-terminate"
+config="${SCRIPT_DIR}/../pipelines/concourse-lite-self-terminate.yml"
+
+[[ -z "${env}" ]] && echo "Must provide environment name" && exit 100
+
+generate_vars_file() {
+   set -u # Treat unset variables as an error when substituting
+   cat <<EOF
+---
+deploy_env: ${env}
+log_level: ${LOG_LEVEL:-}
+EOF
+}
+
+generate_vars_file > /dev/null # Check for missing vars
+
+bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
+   "${env}" "${pipeline}" "${config}" <(generate_vars_file)
+
+$FLY_CMD -t "${FLY_TARGET}" unpause-pipeline --pipeline "${pipeline}"
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by saliceti** [comment]
The 24h interval is risky because everyday we may delay a little bit more and eventually be out of the 2h window. I think it makes more sense to use in a way similar to this: https://github.com/concourse/time-resource#trigger-on-an-interval-within-time-range
In our case the window would be something like 8pm - 6am and the interval 1h. It would retry bosh delete every hour but there is no harm.

**Comment by saliceti** [comment]
`ticking-time-bomb` is not descriptive enough. Please change to `self-destruct` or `nightly-self-destruct`.

File: concourse/scripts/concourse-lite-self-terminate.sh [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+log_level: ${LOG_LEVEL:-}
+EOF
+}
+
+generate_vars_file > /dev/null # Check for missing vars
+
+bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
+   "${env}" "${pipeline}" "${config}" <(generate_vars_file)
+
+$FLY_CMD -t "${FLY_TARGET}" unpause-pipeline --pipeline "${pipeline}"
```
dcarley: This isn't needed after 9afd5757ee3551ab5d74da01306f9b58feb3972e

concourse/scripts/deploy-cloudfoundry.sh
```diff
@@ -32,3 +35,21 @@ generate_vars_file > /dev/null # Check for missing vars
...
 bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
    "${env}" "${pipeline}" "${config}" <(generate_vars_file)
+
+fly -t "$FLY_TARGET" unpause-pipeline --pipeline "${pipeline}"
+
+if [ ! "${DISABLE_AUTODELETE:-}" ]; then
+   bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
+	  "${env}" "${pipeline_autodelete}" "${config_autodelete}" <(generate_vars_file)
+
+   fly -t "$FLY_TARGET" unpause-pipeline --pipeline "${pipeline_autodelete}"
```
dcarley: Also not needed after 9afd5757ee3551ab5d74da01306f9b58feb3972e

dcarley:
> ticking-time-bomb is not descriptive enough. Please change to self-destruct or nightly-self-destruct.

:+1: this. Fun names are fun, but they make it harder for new members of the team to understand what something does.

concourse/pipelines/concourse-lite-self-terminate.yml
```diff
@@ -0,0 +1,31 @@
+resources:
+  - name: delete-timer
+    type: time
+    source:
+      start: 20:00 -0000
+      stop: 21:00 -0000
```
dcarley: Out of interest, why is the `stop` time for this earlier than the `stop` time in `autodelete-cloudfoundry.yml`?

keymon: The autodelete-cloudfoundry will be running every 2h until the morning, trying to delete the deployment.

But this one we run it once because once is run, the machine is gone.  

But on the other hand, one can argue that maybe the job fails because some reason (docker registry down, AWS timeout, etc...) in which case it would make sense to retry every 1-2h.... If you want, I can make it the same.

PR #60, opened 2016-01-20
[#111758188] Bosh cli helpers

mtekel:
### What
A set of helper scripts to:
* set up tunnel to bosh via deployer concourse, log in and configure bosh deployment
* enable easy bosh ssh to deployed VMs

### Testing
Run init script and then test bosh cli locally, e.g. try running an errand (`bosh run errand smoketests`).
Source ssh helper script and try doing `bosh ssh` into one of the CF VMs.

### Who
not @mtekel or @combor

### DIFF
--- a/bosh/bosh_ssh.sh
+++ b/bosh/bosh_ssh.sh
@@ -1,0 +1,21 @@
+#!/bin/bash -e
+SCRIPT_DIR=$(cd "$(dirname $_)" && pwd)
+BOSH_PW=${BOSH_PW:-"abc"}
+BOSH=$(which bosh)
+
+if [ -z "${DEPLOYER_CONCOURSE}" ]; then
+  echo "Must provide DEPLOYER_CONCOURSE ip as env. var"
+else
+  bosh() {
+  if [[ $1 == "ssh" ]]; then
+       shift
+       $BOSH ssh --default_password "${BOSH_PW}" \
+                 --gateway_host "${DEPLOYER_CONCOURSE}" \
+                 --gateway_user vcap \
+                 --gateway_identity_file "${SCRIPT_DIR}"/session/id_rsa \
+                 "$@"
+  else
+       $BOSH "$@"
+  fi
+  }
+fi
--- a/bosh/init.sh
+++ b/bosh/init.sh
@@ -1,0 +1,41 @@
+#!/bin/bash -e
+env=${DEPLOY_ENV:-$1}
+DEPLOYER_CONCOURSE=${DEPLOYER_CONCOURSE:-$2}
+SCRIPT_DIR=$(cd "$(dirname $0)" && pwd)
+BOSH_PORT=${BOSH_PORT:-"25555"}
+
+[ -z "${env}" ]                && echo "Must provide DEPLOY_ENV as \$1 or env. var"     && exit 101
+[ -z "${DEPLOYER_CONCOURSE}" ] && echo "Must provide DEPLOYER_CONCOURSE ip as env. var" && exit 102
+
+pushd ${SCRIPT_DIR}
+
+  # TODO: check if you really want/need to re-create the session
+  #rm session/* && rmdir session
+  [ ! -d session ] && mkdir session
+  pushd session
+    # Get the key
+    ../../concourse/scripts/s3get.sh ${env}-state id_rsa > /dev/null
+    chmod 600 id_rsa
+
+    # Create ssh tunnel to concourse if not present yet
+    if [[ ! $(netstat -an | grep 127.0.0.1.25555) ]]; then
+      ssh -i id_rsa -L ${BOSH_PORT}:10.0.0.6:25555 -fN vcap@${DEPLOYER_CONCOURSE} > /dev/null
+    fi
+
+    # Get bosh admin PW
+    ../../concourse/scripts/s3get.sh ${env}-state bosh-secrets.yml > /dev/null
+
+    boshpw=$(awk '/bosh_admin_password/ {print $2}' bosh-secrets.yml)
+
+    # Login with bosh cli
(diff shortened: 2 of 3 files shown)


bosh/init.sh
```diff
@@ -0,0 +1,44 @@
...
+  # TODO: check if you really want/need to re-create the session
+  #rm session/* && rmdir session
+  [ ! -d session ] && mkdir session
+  pushd session
+    # Get the key
+    ../../concourse/scripts/s3get.sh ${env}-state id_rsa > /dev/null
+    chmod 600 id_rsa
+
+    # Create ssh tunnel to concourse if not present yet
+    if [[ ! $(netstat -an | grep 127.0.0.1.25555) ]]; then
```
keymon: If you check in https://github.com/alphagov/paas-cf/blob/master/vagrant/deploy.sh#L21 , using the control socket feature is a nice way to check if the tunnel is running. Maybe it is worth it use it?

alext:
It's probably worth adding some details of how to use these to the README.

bosh/bosh_ssh.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
keymon: Can we add a `set -e` to fail if something is missing?

bosh/init.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
keymon: Can we add a set -e to fail if something is missing?

bosh/init.sh
```diff
@@ -0,0 +1,44 @@
...
+BOSH_PORT=${BOSH_PORT:-"25555"}
+
+[ -z "${env}" ]                && echo "Must provide DEPLOY_ENV as \$1 or env. var"                        && exit 101
+[ -z "${DEPLOYER_CONCOURSE}" ] && echo "Must provide DEPLOYER_CONCOURSE ip or hostname as \$2 or env. var" && exit 102
+
+pushd ${SCRIPT_DIR}
+
+  # TODO: check if you really want/need to re-create the session
+  #rm session/* && rmdir session
+  [ ! -d session ] && mkdir session
```
keymon: We should either store the session in `$HOME/.paas-cf/session/<env>/` or at least ignore this directory in `.gitignore`

bosh/init.sh
```diff
@@ -0,0 +1,44 @@
...
+    # Select main CF deployment
+    uuid=$(bosh status --uuid)
+    ../../concourse/scripts/s3get.sh ${env}-state cf-manifest.yml > /dev/null
+   sed -i -e "s/^director_uuid:.*$/director_uuid: ${uuid}/" cf-manifest.yml
+   bosh deployment cf-manifest.yml
+  popd
+popd
+
+echo "If you need to enable bosh ssh please source $SCRIPT_DIR/bosh_ssh.sh"
+echo ". $SCRIPT_DIR/bosh_ssh.sh"
```
keymon: Why do we need to source this? Is it not better to just give it execution permissions.

bosh/bosh_ssh.sh
```diff
@@ -0,0 +1,21 @@
...
+SCRIPT_DIR=$(cd "$(dirname $_)" && pwd)
+BOSH_PW=${BOSH_PW:-"abc"}
+BOSH=$(which bosh)
+
+if [ -z "${DEPLOYER_CONCOURSE}" ]; then
+  echo "Must provide DEPLOYER_CONCOURSE ip or hostname as \$1 or env. var"
```
keymon: `DEPLOYER_CONCOURSE` is not read from $1, only as environment variable. this message is misleading

keymon:
The idea is good, but I summarise my suggestions here:

 * Probably there will be more scripts. I propose put these scripts in a folder called `./scripts`. There are other scripts we can move there (e.g. vagrant ones)
 * Can we store the state in `$HOME/.paas-cf/<envname>` or in `.session/<envname>` or something like that?
 * Can we make the script `init.sh` automatically pull the concourse IP from the `concourse.tfstate` outputs?

bosh/bosh_ssh.sh
```diff
@@ -0,0 +1,21 @@
...
+else
+  bosh() {
+  if [[ $1 == "ssh" ]]; then
+       shift
+       $BOSH ssh --default_password "${BOSH_PW}" \
```
keymon: Is `default_password` required?

bosh/init.sh
mtekel: With experience from running ansible, there can be a lot of "dead" control sockets which can instead block execution of the script...

keymon: In my experience this implementation can easily end with ssh running but not working, and I am not sure if greping the `netstat` output  will work fine on macosx and on systems with IPv6. 

I did not have bath experience with control sockets, unless you delete the socket. In that case, it would be the same than not using the socket at all. The control socket gives you a communication channel to a running SSH.

Anyway, it is worth it to `echo` some messages saying that the tunnel was created (or not) and how to kill it.

bosh/bosh_ssh.sh
mtekel: yes. for some reason bosh doesn't ask anymore for password interactively.... Also, it saves a lot of time.

mtekel:
Fixed some of the issues, but closing as this is just a for-now helper script.

keymon:
I think the idea is good, but we start to have a lot of random scripts doing small things around the project (e.g. `./vagrant/environment.sh`, `deploy-pipeline.sh`)... we should revisit all of them.

bosh/bosh_ssh.sh
keymon: Got it, it is the password for the user to do `sudo` in the target machine.

PR #61, opened 2016-01-20
[#111523754] Cloudfoundry instance profiles

saliceti:
# What
Avoid requiring AWS access keys in Cloud Foundry and use IAM profile instead. This applies to the cloud controller so it can read and write from S3 buckets.

Also fixes an issue with the bucket names in the CF manifest.

# How to review
Deploy Cloud Foundry from scratch. It should deploy successfully and be able to read and write to S3 buckets. For example you can check it has created content in bucket `<environment>-cf-buildpacks`.

You should be able to reuse an existing Cloud foundry. Delete the buckets first and run the full pipeline, including Terraform.

# Who can review
Anyone but @alext or @saliceti

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ export AWS_ACCESS_KEY_ID=XXXXXXXXXX
 export AWS_SECRET_ACCESS_KEY=YYYYYYYYYY
 ```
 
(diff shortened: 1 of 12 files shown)


keymon:
LGTM! :+1:

PR #62, opened 2016-01-21
[#112135729] README: Restrict Travis CI badge to master

dcarley:
At the moment it lists the last build, which may be a development branch
that is failing. We only really care about the stability of `master`.

screen shot 2016-01-21 at 13 53 51


combor:
Looks good.

PR #63, opened 2016-01-21
[#112135729] Refactor use of FLY_TARGET

dcarley:
### What

As suggested by @saliceti in:

- https://github.com/alphagov/paas-cf/pull/58#issuecomment-173311797

This makes the default `FLY_TARGET` within `deploy-pipeline.sh` match `env`
(same as `DEPLOY_ENV` in other scripts), which is the Deployer Concourse for
your environment. By doing this we no longer need to instruct people to
`export FLY_TARGET` before running each of the pipeline scripts.

This is overridden by `vagrant/deploy.sh` which needs to deploy pipelines to
the Bootstrap Concourse machine. I've moved the variable from
`vagrant/environment.sh` so that it's not exported to someones shell when
they follow the README instructions to source this file.

`ATC_URL` is no longer needed. The default value points to Vagrant
concourse-lite running in VirtualBox, whereas we now run it in EC2.

## How to review

If you want to simulate from-fresh:
```
unset FLY_TARGET DEPLOY_ENV
```

Follow the README instructions. tl;dr
```
./vagrant/deploy.sh <ENV>
$(./vagrant/environment.sh <ENV>)
./concourse/scripts/create-microbosh.sh && \
./concourse/scripts/deploy-cloudfoundry.sh && \
./concourse/scripts/destroy-cloudfoundry.sh && \
./concourse/scripts/destroy-microbosh.sh
```

## Who can review

Anyone but @dcarley


saliceti:
Much cleaner, thanks.

PR #64, opened 2016-01-21
[#108441288] Remove unnecesary unpause in setup pipeline scripts

keymon:
# What

As per #58, the script deplo-pipeline.sh will do the unpause for us, so it is no needed for the self-terminate concourse lite or
the deploy-cloudfoundry.

Also we are sure that deploy-cloudfoundry uses th right fly command path
for deleteing the autodelete pipeline.

# How to test this

Check #59 for instructions

# who? 

Anyone but @keymon

THIS PR IS ON FIRE!!!


dcarley:
Thanks :+1:

PR #65, opened 2016-01-21
[#112135729] README: Replace `fly` with `${FLY_CMD}`

dcarley:
### What

In 6b0ef65 and 2e1cc94 we stopped recommending that people download the
`fly` binary to `/usr/local/bin`. Instead `vagrant/deploy.sh` downloads the
binary into your PWD and sets `FLY_CMD`. This command in the README is the
last reference to a `fly` binary on your `PATH`.

### How to test

Run the commands in the modified codeblock.

### Who can review

Anyone but @dcarley


keymon:
LGTM

PR #67, opened 2016-01-21
[#111758188] Enable smoketests

mtekel:
### What
Clean up CF manifest system and app domains a bit. Ensure app domain is different from system domain (else you can deploy e.g. `api` called app, which then gets traffic meant for CF api). Add configuration for smoketests.

### Testing
Deploy or apply against existing evironment. After deployment, `bosh run errand smoke_tests`. They should pass. You can also run tests, `cd manifests/cf-manifest; bundle exec rspec`. They should also pass.

### Who
not @mtekel or @combor

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -1,6 +1,8 @@
 meta:
   environment: ~
 
+  api_domain: (( concat "api." properties.domain ))
+
   release:
     name: cf
 
@@ -21,7 +23,7 @@ meta:
       component: CloudController
     port: (( grab properties.cc.external_port ))
     uris:
-    - (( concat "api." properties.domain ))
+    - (( grab meta.api_domain ))
 
   api_templates:
   - name: cloud_controller_ng
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -912,3 +912,12 @@ properties:
   acceptance_tests:
 
   smoke_tests:
+    api: (( grab meta.api_domain ))
+    apps_domain: (( grab properties.app_domains[0] ))
+    user: "admin"
+    password: (( grab secrets.uaa_admin_password ))
+    org: "SMOKE_TESTS"
+    space: "SMOKE_TESTS"
+    use_existing_org: false
+    use_existing_space: false
+    skip_ssl_validation: true


manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -1,6 +1,9 @@
 meta:
   environment: ~
 
+  api_domain: (( concat "api." properties.domain ))
+  apps_domain: (( concat "apps." properties.domain ))
```
dcarley: Do we need to purchase and configure two separate wildcards certificates for the router?

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -910,3 +913,13 @@ properties:
   acceptance_tests:
 
   smoke_tests:
+  smoke_tests:
```
dcarley: This key duplicates the line above.

manifests/cf-manifest/deployments/030-cf-jobs.yml
mtekel: It all depends. Maybe you want to offload cert management to app owners. Last time I have checked certificates with 2 * records were non trivial. Our solution was to use asterisk for apps and then add SAN for api, uaa, login, loggregator etc.

manifests/cf-manifest/deployments/030-cf-jobs.yml
mtekel: Somehow slipped in during rebasing.

mtekel:
Updated.

manifests/cf-manifest/deployments/030-cf-jobs.yml
dcarley: Do we need to change it now to get the smoke tests working? It feels like something that deserves it's own story and a bit of consideration. Especially as we have other SSL stories in-flight at the moment. /cc @bleach

mtekel: The story was to test environment being "basically sound" and fix small issues found. So I have fixed this as I consider it quite important that app and system domains are different.

dcarley: I agree, it's important, which means that we shouldn't slip it in with the more innocuous smoke test work. Please rebase it out and ask Rory/Brett to raise a separate story. You can push it to a separate branch and reference that in the story so that we don't lose it.

dcarley:
> After deployment, bosh run errand smoke_tests. They should pass.

Could you add that to the message of the commit which adds the errand configuration? That would make it easier for someone that isn't familiar with errands to understand the purpose of that commit.

mtekel:
closing until I finish testing with the new configuration

mtekel:
Had to force push, hence can't reopen. New PR: https://github.com/alphagov/paas-cf/pull/68

PR #69, opened 2016-01-22
Add documentation on failures with many ssh keys

Jonty:
This just bit me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ ssh-add $(pwd)/id_rsa
 ssh vcap@<concourse_ip>
 ```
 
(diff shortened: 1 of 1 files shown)


mtekel:
This is true, even though I think it's even better not to add these temporary environment keys in to ssh agent at all and instead use st. like helper scripts in https://github.com/alphagov/paas-cf/tree/bosh-cli

dcarley:
This is quite wordy. I agree with @mtekel. Also @keymon and I talked about adding the file to `.gitignore` or moving it outside of the current working directory, which we might not need to worry about if we used a helper script. However this is a good improvement for now and likely to save people time.

PR #70, opened 2016-01-25
[#111509456] - Lint checks for shell scripts

jimconner:
## What

Use shellcheck to check all our shell scripts for common errors. This also fixes all the errors that were reported.

## Details

This includes some refactorings:

The `bosh-manitest/build_manifest.sh` has been refactored to be more consistent
with the cloudfoundry one and take the paths to output files in environment
variables. This was done to avoid some issues with array variables in POSIX
scripts that were being flagged up by shellcheck (bash isn't available in
alpine based containers). See 9138c37

The travis build has been refactord to be a single job instead of a matrix
build as it's more efficient this way. This also included creating a
Makefile which will make it easier to run locally. (see 2538f91)

## How to test

Run the tests localy - `make test` (these will require shellcheck > 0.4.0 to be
installed locally).

On Mac - `brew install shellcheck`

On Linux:
   git clone git@github.com:koalaman/shellcheck.git
   sudo apt-get install cabal-install ghc
   cd shellcheck
   cabal update
   cabal install # installs into ~/.cabal/bin

Confirm that the deployment process still works after the various refactorings.

## Who can review

Anyone but @alext  or myself.

##Notes : We can re-enable the SC2086 checks in the `s3init.sh` and `s3get.sh` scripts after Shellcheck 0.4.3 is available for everyone.


dcarley:
This looks great.

I'm in the process of testing the pipeline changes.

I got bored of waiting for `brew` to compile Haskell. What do you think about allowing people to override the `shellcheck` command so that you can use a Docker container? e.g.
```diff
diff --git a/Makefile b/Makefile
index c7f6094..f26845a 100644
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .PHONY: test spec lint_terraform lint_shellcheck

+SHELLCHECK=shellcheck
+
 test: spec lint_terraform lint_shellcheck

 spec:
@@ -12,4 +14,4 @@ lint_terraform:
        find terraform -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 -n 1 -t terraform graph > /dev/null

 lint_shellcheck:
-       find . -name '*.sh' -print0 | xargs -0 shellcheck
+       find . -name '*.sh' -print0 | xargs -0 $(SHELLCHECK)
```
With this container:
```Dockerfile
FROM debian:sid

RUN apt-get -yq update && apt-get -yq upgrade
RUN apt-get -yq install shellcheck

ENTRYPOINT ["shellcheck"]
CMD []
```
And run like:
```
➜  paas-cf git:(shellcheck) ✗ make test SHELLCHECK="docker run -v $(pwd):/mnt -w /mnt shellcheck"
…
find . -name '*.sh' -print0 | xargs -0 docker run -v /Users/dcarley/projects/paas/paas-cf:/mnt -w /mnt shellcheck
```

jimconner:
I think that's a superb idea - Will get it implemented tomorrow.

mtekel:
When building a new container, we should probably build the latest version.  That apt package is not the most recent....

combor:
0.4.3 is already in brew

```
brew info shellcheck
shellcheck: stable 0.4.3 (bottled), HEAD
Static analysis and lint tool, for (ba)sh scripts
http://www.shellcheck.net
/usr/local/Cellar/shellcheck/0.4.1 (6 files, 17.2M)
  Poured from bottle
/usr/local/Cellar/shellcheck/0.4.2 (6 files, 17.6M)
  Poured from bottle
/usr/local/Cellar/shellcheck/0.4.3 (6 files, 17.6M) *
  Poured from bottle
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/shellcheck.rb
```

but using shellcheck from a container sounds like a better idea.

dcarley:
The container matches what's used in Travis. Using a later version would allow us to re-enable the check SC2086, but makes the installation in Travis much more ghetto. I think we should hold off upgrading for the moment.

dcarley:
:+1:

dcarley:
Note: we talked about not documenting the `Dockerfile` approach for now. I'm going to try getting `shellcheck` working again from a Homebrew bottle. If enough people find that they need to run the tests locally and having trouble installing specific versions of shellcheck then we can review it again.

PR #72, opened 2016-01-25
[#109035334] IAM for api_workers as well

mtekel:
api_worker instances need access to AWS as well, as they are essentially
part of CF API and need to do same tasks as api instances. Moreover,
some tasks (like handling `cf copy-source`) are exclusive to workers and
will fail without IAM role. Thus change instance type to same as API.

`cf copy-source` is used in acceptance tests, so without this PR they fail on that test.

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -383,7 +383,7 @@ jobs:
   - name: api_worker_z1
     templates: (( grab meta.api_worker_templates ))
     instances: 1
-    resource_pool: small_z1
+    resource_pool: api_z1
     persistent_disk: 0
     networks:
       - name: cf1
@@ -396,7 +396,7 @@ jobs:
   - name: api_worker_z2
     templates: (( grab meta.api_worker_templates ))
     instances: 1
-    resource_pool: small_z2
+    resource_pool: api_z2
     persistent_disk: 0
     networks:
       - name: cf2


saliceti:
We may need it on cloud_controller_clock as well, since it's a cloud controller too.

keymon:
I would rather remove the cloud_controller_clock job, and merge it with the main api_z1. In general I think we can group several services in our current deployment.

But that is a different story, without having tested it, this PR looks OK for me.

alext:
What is the thing that needs access to the S3 buckets? I'm not sure I understand why these machines would need direct access to the buckets.

(The only thing the IAM profile grants is access to the CF S3 buckets).

mtekel:
@saliceti clock is a very specific job that mostly just triggers CF jobs and doesn't do any work by itself. For now it doesn't need IAM role
@keymon indeed. We did collocate in alpha. We should however discuss if we really want to do this (due to e.g. workload isolation, possibly different security groups/IAM profiles etc., see [currently last comment in the story](https://www.pivotaltracker.com/n/projects/1275640/stories/109035334)
@alext at least `cf copy-source` needs the bucket access, maybe some more tasks as well

saliceti:
OK the clock tasks are here: https://github.com/cloudfoundry/cloud_controller_ng/blob/757997c5193772c0c13fee533d13c998051d806f/lib/cloud_controller/scheduler.rb

It looks like they only do cleanup in the DB.

mtekel:
@saliceti Wow. And you need a whole separate VM just to do that... (and ensure that there's only one running).

PR #73, opened 2016-01-26
[#111946739] Parameterise zone ID, zone name, and ELB cert ARN

Jonty:
## What

This PR moves the `zone_id`, `zone_name` and `elb_whateveritis` parameters to being specified separately for both the "development" and "production" accounts. They're hardcoded in terraform `.tfvar` files for convenience.

Due to terraform quirks we need to declare the variables up front for cases where we're not actually using the `.tfvars` as it will throw errors - we use `dummy_default` as a placeholder, as it will be overridden by the `production.tfvars` or `development.tfvars` later.

The credentials for production are currently dummy values until [111676128](https://www.pivotaltracker.com/n/projects/1275640/stories/111676128) is complete.

## How to test

Spin up a concourse and an environment and it should deploy correctly, defaulting to the "development" credentials. Switching to production is achieved by defining the `ACCOUNT` environment variable as `production`, however testing this will currently explode in a hilarious fashion - the credentials are dummy values until the story to generate them has been finished.

## Who can review this

Anybody but @mtekel and @jonty


alext:
In [terraform-iam](https://github.gds/government-paas/terraform-iam), we've named the accounts dev, ci, stage and prod. It would be good to be consistent here.

terraform/globals.tf
```diff
@@ -57,13 +57,13 @@ variable "vagrant_cidr" {
...
-  default     = "Z3SI0PSH6KKVH4"
+  description = "Amazon Route53 DNS zone identifier. Different per account."
+  default     = "dummy_default"
 }
 
 variable "dns_zone_name" {
-  description = "Amazon Route53 DNS zone name"
-  default     = "cf.paas.alphagov.co.uk"
+  description = "Amazon Route53 DNS zone name. Differs per account."
+  default     = "dummy_default"
```
alext: Do we want these default values here? I'm concerned that there's a risk these could leak into a real environment.

The commit message mentions running terraform with a limited scope, and therefore not having to specify these. I believe terraform will still update the outputs when running with a limited scope, and therefore these could end up in the outputs, and get written into the S3 bucket.

mtekel: These values are safe dummies that will break TF deployment if they leak. When you run TF with limited scope, without defaults it would still ask you for the value of these variables regardless if it needs to use them. TF will update some related outputs (and can lose some, see [this](https://github.com/alphagov/paas-cf/commit/164f954789c5e53b9e8c47a4e6cd758f4451e95d)) when running on limited scope.

vagrant/deploy.sh
```diff
@@ -17,6 +17,7 @@ vagrant up
 export VAGRANT_IP=$(vagrant ssh -- curl -qs http://169.254.169.254/latest/meta-data/public-ipv4)
 export CONCOURSE_URL=http://localhost:8080
 export FLY_TARGET=${DEPLOY_ENV}-bootstrap
+export ACCOUNT=${ACCOUNT:-development}
```
alext: I wonder whether it would be better to call this env var `AWS_ACCOUNT` to avoid any confusion about which account we're referring to.

Jonty:
Have addressed @alext's issues with AWS_ACCOUNT and renaming environments in two new commits.

terraform/globals.tf
dcarley: Can either of you describe what changes would be required to omit the defaults? So that I can understand the impact of implementing this either way.

mtekel: If you don't include default values then terraform will ask for a value every time you run it, e.g. in s3init when creating the initial bucket. You would have to provide some value (regardless if variable is used or not) in every single terraform use case.

alext: To me, that's an argument that these shouldn't be in globals - they should only be included in the configs that use them (concourse and cloudfoundry).

mtekel: Yes. But you would still need to add defaults. So to avoid duplication, we moved them to globals. There are other variables in globals that are not used by _all_ our TF jobs.

dcarley:
Sorry, as predicted, this needs rebasing against the changes merged in #77.

I presume that @alext will continue to review after.

mtekel:
I will wait for https://github.com/alphagov/paas-cf/pull/78 as well...

dcarley:
> I will wait for #78 as well...

Good point. I'm happy for this to go first though and I'll refactor #78 after. This is more important.

PR #74, opened 2016-01-27
[#112135729] Simplify README pt2

dcarley:
## What

Simplify the README by standardising the sections for each component.

## How to review

You can either review the:

- complete diff
- individual commits

Either way you should also double check the rendered `README.md` in this branch to confirm that it looks visually correct and that the links work.

## Who can review

Not @dcarley

## Notes

There may be one remaining PR in this story to refactor the pipeline scripts so that we have a single entry point to deploy the pipelines to either Bootstrap Concourse or Deployer Concourse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,6 @@ This pipeline will:
 
 ### Deploy and destroy cloudfoundry with microbosh
 ```
-# Optionally pass the current branch for the git resources
-export BRANCH=$(git rev-parse --abbrev-ref HEAD)
-
 ./concourse/scripts/deploy-cloudfoundry.sh $DEPLOY_ENV
 ```
 
--- a/README.md
+++ b/README.md
@@ -10,18 +10,26 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
 [BOSH]: https://bosh.io/
 [CloudFoundry]: https://www.cloudfoundry.org/
 
-## Glossary
+## Overview
 
-- Environment: a single Cloud Foundry installation and its supporting
-  infrastructure.
-- Bootstrap Concourse: is responsible for creating or destroying a *Deployer
-  Concourse* to an *environment*. You don't need to keep this once you
-  have a *Deployer Concourse* running.
-- Deployer Concourse: is responsible for deploying everything else within
-  its *environment*, e.g. BOSH and CloudFoundry. It should be kept running
-  while that *environment* exists.
+The following components needs to be deployed in order. They should be
+destroyed in reverse order so as not to leave any orphaned resources:
+
+1. [Bootstrap Concourse](#bootstrap-concourse)
+1. [Deployer Concourse](#deployer-concourse)
+1. [MicroBOSH](#microbosh)
+1. [CloudFoundry](#cloudfoundry)
+
+The word *environment* is used herein to describe a single Cloud Foundry
+installation and its supporting infrastructure.
 
 ## Bootstrap Concourse
+
+This runs outside an environment and is responsible for creating or
+destroying a [Deployer Concourse](#deployer-concourse) in an environment.
+You don't need to keep this running once you have the Deployer Concourse,
+and you can create it again when the Deployer Concourse needs to be modified
+or destroyed.
 
 ### Prerequisites
 
@@ -69,6 +77,11 @@ Run the following script, with the name of your existing environment:
 
 ## Deployer Concourse
 
+This runs within an environment and is responsible for deploying everything
+else to that environment, such as [MicroBOSH](#microbosh) and
+[CloudFoundry](#cloudfoundry). It should be kept running while that
+environment exists.
+
 ### Prerequisites
 
 You will need a working [Bootstrap Concourse](#bootstrap-concourse).
@@ -100,6 +113,9 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
 Run the `destroy-deployer` pipeline from your *Bootstrap Concourse*.
 
 ## MicroBOSH
+
+MicroBOSH is responsible for deploying [CloudFoundry](#cloudfoundry) and
+supporting services for the platform.
 
 ### Prerequisites
 
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ override the working branch for development and review:
 export BRANCH=$(git rev-parse --abbrev-ref HEAD)
 ```
 
-#### SSH to deployed Concourse and microbosh
+## SSH to deployed Concourse and microbosh
 
 In the `create-deployer` pipeline when creating the initial VPC,
 a keypair is generated and uploaded to AWS to be used by deployed instances.
--- a/README.md
+++ b/README.md
@@ -187,17 +187,12 @@ used within the pipeline. This is useful for development and code review:
 BRANCH=$(git rev-parse --abbrev-ref HEAD) <script>.sh
 ```
 
-## SSH to Bootstrap Concourse and sharing
+## Sharing your Bootstrap Concourse
 
-Instead of allowing a non secure HTTP connection via the internet to the
-*Bootstrap Concourse*, we create a SSH tunnel to redirect the port 8080 to
-localhost.
+If you need to share access to your *Bootstrap Concourse* with a colleague
+then you will need to reproduce some of the work that Vagrant does.
 
-After running vagrant/deploy.sh "daemonized" SSH redirection will be
-running listening in the port localhost:8080.
-
-If the machine needs to be shared with a coworker, and additional SSH
-public key can be added with:
+Add their SSH public key:
 
 ```
 cd vagrant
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -188,27 +198,17 @@ echo "ssh-rsa AAAA... user" | \
...
+ssh ubuntu@<bootstrap_concourse_ip> -L 8080:127.0.0.1:8080 -fN
 ```
 
-To learn the public IP of the created concourse, simply run:
+To learn the public IP of your *Bootstrap Concourse*, simply run:
 
 ```
 cd vagrant
 vagrant ssh-config
 ```
```
jimconner: Perhaps swapping the ordering of the 'learn the public IP' and 'ssh to the public IP' steps could be reversed here so that the IP is learned before you need to use it for the ssh tunnel.

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
    ${FLY_CMD} -t ${DEPLOY_ENV} login -k -c "https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk"
 ```
```
jimconner: Do we need to explain how to get the fly command? ... and setting the FLY_CMD variable isn't covered in the readme (although arguably that is pretty obvious if you know what you're doing)

README.md
dcarley: Good point. Swapped in 8e708ed

README.md
dcarley: I don't think so - it's downloaded by `vagrant/deploy.sh` which is listed as a prerequisite and the environment variable is set by the preceding `./vagrant/environment.sh` command.

README.md
```diff
@@ -228,16 +228,16 @@ For example:
 chmod 400 id_rsa && \
 ssh-add $(pwd)/id_rsa
 
-ssh vcap@<concourse_ip>
+ssh vcap@<deployer_concourse_ip>
```
jimconner: Given that we're assigning DNS names to the deployer_concourse, might it be easier to use the name rather than the IP here?

dcarley: Since 48db141d73c7222bf07866e6a58b023de64aaa8d the hostname points at an ELB instead of the instance's EIP. We had a similar problem with the instructions in 8e03fd1aeb5e118d2679a83c1f2938b76b9fcd49. We could consider adding TCP/22 to the ELB and removing the EIP (is it still needed?).

dcarley: To clarify - I don't intend to change the ELB as part of this PR. But we could do it later on.

README.md
```diff
@@ -10,19 +10,24 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
...
+destroyed in reverse order so as not to leave any orphaned resources:
+
+1. [Bootstrap Concourse](#bootstrap-concourse)
+1. [Deployer Concourse](#deployer-concourse)
+1. [MicroBOSH](#microbosh)
+1. [CloudFoundry](#cloudfoundry)
 
 ## Bootstrap Concourse
 
+This runs outside an environment and is responsible for creating or
```
saliceti: I think we should keep the definition of an environment. Inline here or in a specific section above.

README.md
saliceti: I think we should create a wrapper login script and get rid of this. But this is for another PR.

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
    ${FLY_CMD} -t ${DEPLOY_ENV} login -k -c "https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk"
 ```
 
-## BOSH and Cloud Foundry
+### Destroy
 
-### Microbosh deployment from concourse bootstrap
+Run the `destroy-deployer` pipeline from your *Bootstrap Concourse*.
 
-These pipelines will deploy/destroy a microbosh using bosh-init. It will use the IAM profile `deployer-concourse` that was given to the Concourse instance.
```
saliceti: We depend a lot on the IAM profiles but we don't describe their names nor the permissions. This should be documented somewhere as without it this project is unusable.

I think the minimum would be a list of the profile names and just a summary of the permissions, like "read-write access to S3". Ideally, we could link to a separate doc that has all the details.

alext: I'd be nervous about linking to a doc for this as it's bound to be out of date. Maybe it would be enough to link to the terraform repo that specifies all of these.

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
...
- * A VPC is setup and there is a bucket for the state files.
- * The `fly` command is installed and available
- * A running concourse configured in the CLI, with target name `$DEPLOY_ENV`
+You will need a working [Deployer Concourse](#deployer-concourse).
 
-#### Pipelines: create-microbosh and destroy-microbosh
+### Deploy
 
-To setup the pipeline `create-microbosh`:
+Run the `create-microbosh` pipeline which can be setup by running:
```
saliceti: It would look more correct with "set up" when using the verb ("setup" for a noun). Ref: http://www.future-perfect.co.uk/grammar-tip/is-it-setup-set-up-or-set-up/

README.md
saliceti: What about a summary so the reader can at least get the idea?

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
...
 
-This pipeline will
+# Additional notes
 
- * Connect to microbosh and delete CF deployment
- * Use Terraform to destroy all the resources created by `deploy-cloudfoundry`
+## Optionally override the branch used by pipelines
 
-# Additional notes
+All of the pipeline scripts honour a `BRANCH` environment variable which
```
saliceti: The pipeline scripts AND the vagrant deploy.sh script.

README.md
dcarley: I agree with Alex. I am a firm believer in trying not duplicate code/config in documentation because people *will* forget to maintain it and out-of-date documentation is worse than no documentation at all.

The profile names are detailed in code. We can link to the terraform-iam repo near the top of the README, where we talk about access keys and instance profiles - but it'll only be of use to people within GDS.

I guess it depends who we're documenting for - people on our project, or people external to GDS wanting to re-use the project. I think we should optimise for the former at the moment.

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
...
+BRANCH=$(git rev-parse --abbrev-ref HEAD) <pipeline_script>.sh
+```
 
-## SSH tunnel to vagrant concourse-lite and share access to instance
+## SSH to Bootstrap Concourse and sharing
 
 Instead of allowing a non secure HTTP connection via the internet to the
-concourse-lite AWS instance, we create a SSH tunnel to redirect the port
-8080 to localhost.
+*Bootstrap Concourse*, we create a SSH tunnel to redirect the port 8080 to
```
saliceti: redirect the *remote* port 8080 ?

README.md
dcarley: `deploy.sh` uses the pipeline scripts. Do you think it's worth explicitly calling out? I could add "(including those deployed by the Bootstrap Concourse)"?

README.md
```diff
@@ -87,91 +108,80 @@ echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
...
+localhost.
 
 After running vagrant/deploy.sh "daemonized" SSH redirection will be
 running listening in the port localhost:8080.
```
saliceti: listening *on* the port?

README.md
dcarley: It's both: "the local port 8080 to the remote port 8080". Maybe we should omit this altogether?

README.md
saliceti: I mean there are 2 use cases:

* `BRANCH=... vagrant/deploy.sh [env]` to update the bootstrap concourse
* `BRANCH=... <pipeline_script>.sh [env]` to update a specific pipeline on the deployer concourse

README.md
saliceti: Yes I would omit it or find a way to express this better.

README.md
dcarley: Added a link in 191c30f.

dcarley:
@saliceti think I've addressed all the comments. Could you have another look?

README.md
dcarley: I've rebased this back into the Overview section in 7c8fba00c203c70f85635850cdf11146ccf673a8.

Jonty:
This is substantially better and should probably be merged as-is, with further work pulled into another documentation story.

Two comments, both of which are probably out of scope for this story:
* I can't help but feel that a small diagram explaining what each concourse deploys and how they relate to one another would make things a lot clearer - this is a lot of text and it can be hard to keep track.
* Any time we have a multi-line code block it should be either turned into a script or perhaps a Makefile - it makes this look substantially more complicated and introduces potential for user error.

dcarley:
> This is substantially better and should probably be merged as-is, with further work pulled into another documentation story.
> 
> Two comments, both of which are probably out of scope for this story:

:+1:, when merged I'll collect up any of the suggestions that were outside the scope of this PR and put them in Pivotal so that we can follow up on them.

PR #75, opened 2016-01-27
[#109035334] Implement acceptance and smoke tests

keymon:
[#109035334 Implement acceptance tests on the CF deployment](https://www.pivotaltracker.com/story/show/109035334)

What
----

The objective is to implement a set of acceptance tests to ensure that the deployment was successful and that the platform functions as expected. Given that we have started with a vanilla install, we should be able to use the existing CF acceptance tests and we will build on these as we add functionality.

Context
-------

The PR should be reviewed in this context:

 * I want to add the required configuration to run the basic acceptance tests.
 * I want to add a job that runs the smoke tests after deployment and is triggered
   automatically after any successful deployment.
 * I want to add a job that runs the acceptance tests after deployment, altough
   not triggered automaticaly.

How to test it?
---------------

Apply the pipelines from this branch and deploy CF as described in the README.md.

Remember set the variable BRANCH:

```
```

Run the jobs `smoke-tests` and `acceptance-tests`.

`acceptance-tests` might have some tests failing due timeouts, but it is not
in the scope of this story fix that.

Who?
----

Anyone but @keymon or @combor

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -910,6 +910,18 @@ properties:
   collector:
 
   acceptance_tests:
+    api: (( grab meta.api_domain ))
+    apps_domain: (( grab properties.app_domains[0] ))
+    system_domain: (( grab properties.system_domain ))
+    admin_user: "admin"
+    admin_password: (( grab secrets.uaa_admin_password ))
+    skip_ssl_validation: true
+    backend: "diego"
+    skip_diego_unsupported_tests: true
+    include_route_services: true
+    include_tasks: true
+    client_secret: (( grab secrets.uaa_clients_gorouter_secret ))
+    include_v3: true
 
   smoke_tests:
     api: (( grab meta.api_domain ))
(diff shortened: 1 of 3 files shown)


manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -910,6 +910,18 @@ properties:
...
+    system_domain: (( grab properties.system_domain ))
+    admin_user: "admin"
+    admin_password: (( grab secrets.uaa_admin_password ))
+    skip_ssl_validation: true
+    backend: "diego"
+    skip_diego_unsupported_tests: true
+    include_route_services: true
+    include_tasks: true
+    client_secret: (( grab secrets.uaa_clients_gorouter_secret ))
+    include_v3: true
```
mtekel: What about
```
include_sso: true
include_operator: true
include_logging: true
include_security_groups: true
include_internet_dependent: true
include_services: true
include_v3: true
include_routing: true
include_routing_api: true
include_route_services: true
include_diego_docker: false   # deployed diego version doesn't support this yet
```

Shall we run those as well?

mtekel:
This works, both smoke and acceptance tests. Additional tests will be enabled as issues making them fail are being fixed...

PR #77, opened 2016-01-27
[#111676128] Auto generate ssl cert for deployer concourse

saliceti:
# What
Story: https://www.pivotaltracker.com/n/projects/1275640
We generate a self signed certificate for each environment and set up the Concourse ELB with it. It is based on the environment name and the terraform variable `dns_zone_name`.

The certificate is stored in S3. If it is already present in S3, then it will use this one. This will be useful for manually generated certificates.

# How to review
* Create a new deployer Concourse
* Check it uses a newly generated certificate.
* Check if you run the second time the certificate is not changed
* Check the certificate is removed from IAM when you destroy the environment

# Note
We have observed a bug that looks like a race condition when we tried to delete the certificate from S3 to force the generation. Sometimes Concourse uses the cached resource and will re-upload it.

The workaround is to wait at least 5 min between removing the cert and running the pipeline again.

# Who can review
Anyone but @saliceti or @timmow


terraform/concourse/elb.tf
```diff
@@ -1,3 +1,13 @@
+resource "aws_iam_server_certificate" "concourse" {
+  name = "${var.env}-concourse"
+  certificate_body = "${file("concourse.crt")}"
+  private_key = "${file("concourse.key")}"
+
+  provisioner "local-exec" {
+    command = "sleep 10"
```
dcarley: From the log in the commit message:
```
* aws_elb.concourse: [WARN] Error creating ELB Listener with SSL Cert, retrying: CertificateNotFound: Server Certificate not found for the key: arn:aws:iam::xxxx:server-certificate/xxxx
	status code: 400, request id: xxxx
```
Am I right in thinking that the Terraform run doesn't fail (non-zero exit code) when run without the `sleep`? It's a warning, rather than an error, which indicates that it had to be retried. If that's the case then I don't think we should add the `sleep` to mask the behaviour.

saliceti: Unfortunately terraform does fail and the pipeline stops. If you restart the job, then it works.
The sleep allows to make it work the first time.

dcarley:
It works and code looks good :+1: 

Two things we need to address before merging:

- I think we should raise a new upstream issue for the race condition. The referenced issue has been closed and it doesn't actually appear to be retrying correctly/successfully. It also deserves a comment in our Pivotal, so that we create a tech debt story to remove the `sleep` in future.
- Can you add more detail about "why" to the commit messages? It's clear what they do, but somebody reading the commits in the future would struggle to understand *why* we're now generating the cert in the pipeline, which is just as important as *what*.

dcarley:
In fact, is that the right error?

When I hijack the container, modify the config to remove the sleep, and destroy the two resources with:
```
terraform destroy -target aws_elb.concourse -target aws_iam_server_certificate.concourse ${terraform_params} -state=concourse.tfstate paas-cf/terraform/concourse
```

Then recreate them with:
```
terraform apply -target aws_elb.concourse -target aws_iam_server_certificate.concourse ${terraform_params} -state=concourse.tfstate paas-cf/terraform/concourse
```

It occasionally, but not always, fails with a different error:
```
1 error(s) occurred:

* aws_elb.concourse: Failure adding new or updated ELB listeners: DuplicateListener: A listener already exists for dcarley-concourse with LoadBalancerPort 443, but with a different InstancePort, Protocol,
 or SSLCertificateId
        status code: 400, request id: 7432ccc6-c5ca-11e5-aa67-2b07be61a112
```

saliceti:
No I've never experienced that error. It looks like the ELB wasn't completely destroyed, or some inconsistent state?

saliceti:
When trying to reproduce the error, it failed every time with the `DuplicateListener` error now. However, the workaround is the same.

I have raised issue https://github.com/hashicorp/terraform/issues/4880 and changed the commit message.

dcarley:
Thanks :+1:

PR #49, opened 2016-01-15
[#111428934] Move Diego to new Repos

mtekel:
### What
[Move Diego to new Repos](https://www.pivotaltracker.com/projects/1275640/stories/111428934)

Move diego commits from [vanilla manifest branch in cf-terraform](https://github.com/alphagov/cf-terraform/tree/vanilla_cf_prototype). Add fixes and changes necessary to make deployment start.

### Testing
Run the cf deploy pipeline. Watch diego CF being deployed. Destroy. All should pass successfully.

### Who
not @mtekel or @keymon

PR #79, opened 2016-01-28
[#111428592] Remove unnused DEA configuration in CF manifest

keymon:
What
----

In #49 we replaced DEA with diego. We don't need anymore the properties
and templates defined in the manifest.

How to test this?
-------------------

Try to redeploy the manifest with bosh, bosh client should report that
there are no be changes to the cluster.

Who?
----

Anyone but @keymon or @timmow


mtekel:
Indeed no updates to cluster.

PR #82, opened 2016-01-29
[#112581575] skip release download

combor:
# What

We would like to remove unnecessary downloads of releases and stemcell during every run of **deploy** job.   Pivotal story: [Implement Workaround for caching issue with download of concourse resources](https://www.pivotaltracker.com/story/show/112581575) 

# Details

All download tasks are rewritten to call bosh directly and check whether release/stemcell is already uploaded. The helper script is provided to accomplish this. If release is not present, another bosh call triggers download directly on bosh VM so we also skip upload phase. 

Additionally, I have decided to remove bosh-deployment resource as it can't be used in case all releases and stemcells are already uploaded. It is replaced by deployment task. 

# How to test

- Upload updated pipeline by executing:
```
BRANCH=112581575_skip_release_download concourse/scripts/deploy-cloudfoundry.sh <env_name>
```
- Trigger the build manually from Concourse GUI

# Who can review

Anyone but @combor

### DIFF
--- a/concourse/pipelines/deploy-cloudfoundry.yml
+++ b/concourse/pipelines/deploy-cloudfoundry.yml
@@ -1,13 +1,4 @@
 resources:
-  - name: cf-deployment
-    type: bosh-deployment
-    source:
-      target: https://10.0.0.6:25555
-      username: admin
-      password: {{bosh_password}}
-      deployment: {{deploy_env}}
-      ignore_ssl: true
-
   - name: paas-cf
     type: git
     source:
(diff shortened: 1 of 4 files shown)


concourse/scripts/bosh_ensure_uploaded.sh
```diff
@@ -0,0 +1,31 @@
...
+url="${URL:-$4}"
+bosh_secrets="${BOSH_SECRETS:-$5}"
+
+[[ -z "${type}" ]] && echo "Must provide artifact type: release|stemcell" && exit 100
+[[ -z "${name}" ]] && echo "Must provide artifact name" && exit 100
+[[ -z "${version}" ]] && echo "Must provide artifact version" && exit 100
+[[ -z "${url}" ]] && echo "Must provide download url" && exit 100
+[[ -z "${bosh_secrets}" ]] && echo "Must provide bosh secrets file path" && exit 100
+
+bosh_password=$(awk '$1~/bosh_admin_password/ {print $2}' "${bosh_secrets}")
```
keymon: As we were discussing in person, I am not convince that we do this `awk` on the `bosh-secrets.yml` in this script itself.

I think that we should do it outside of the script, and only do the  `awk` in the pipeline. This way we decouple the method we use to retrieve the password from this script.

dcarley:
Out of interest, how many minutes has this saved off the deploy to a new and existing environment?

concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -6,16 +6,25 @@ resources:
       stop: 6:00 -0000
       interval: 2h
 
+  - name: bosh-secrets
```
keymon: And now, I will be evil and ask you to raise a different PR for this last commit (942f5e0f129ae21378db73c63fc4d5ad6dacaead), referring to the story [#112670769](https://www.pivotaltracker.com/story/show/112670769)  :laughing: 

It is easy, only do:
```
git checkout 112581575_skip_release_download
git checkout -b 112670769_get_bosh_password_in_pipeline # create a new branch
git push

git checkout 112581575_skip_release_download
git reset HEAD^1 --hard # drop last commit
git push -f origin 112581575_skip_release_download

```

And do the PR for 112670769_get_bosh_password_in_pipeline

keymon:
The PR works  OK so far :)

concourse/scripts/bosh_ensure_uploaded.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/sh
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+
+type="${TYPE:-$1}"
+name="${NAME:-$2}"
+#shellcheck disable=SC2153
+version="${VERSION:=$3}"
+#shellcheck disable=SC2153
```
dcarley: We should only disable checks in exceptional cases and describe the reason why in the commit message.

From reading [SC2153](https://github.com/koalaman/shellcheck/wiki/SC2153) am I right in thinking that we've disabled this because the variable names are too alike? In which case, I think it's probably right to complain, since it's confusing to have upper and lowercased variables of the same name.

It looks like we only use this script in the pipeline with environment variables, not by passing arguments. In which case it'd be preferable to use the variables directly, remove the reassignments/defaults, and we could probably replace the validation with `set -u` to exit if any are not set.

concourse/scripts/bosh_ensure_uploaded.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/sh
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+
+type="${TYPE:-$1}"
+name="${NAME:-$2}"
+#shellcheck disable=SC2153
+version="${VERSION:=$3}"
```
dcarley: Is the `:=` (if not set, set to default) instead of `:-` (if not set, use default) intentional?

Might not be relevant though, depending on the comment below.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -234,118 +225,173 @@ jobs:
     serial: true
     plan:
       - aggregate:
+        - get: paas-cf
+          passed: ['generate-manifest']
+        - get: cf-manifest
+          passed: ['generate-manifest']
+          trigger: true
+        - get: bosh-secrets
```
keymon: This should be removed

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -234,118 +225,173 @@ jobs:
...
+              VERSION: {{stemcell-version}}
+              NAME: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
+              URL: https://bosh.io/d/stemcells/"${NAME}"?v="${VERSION}"
+              BOSH_PASSWORD: {{bosh_password}}
+              TYPE: stemcell
+            image: docker:///concourse/bosh-deployment-resource
             platform: linux
+            inputs:
+            - name: paas-cf
+            - name: bosh-secrets
```
keymon: Remove this as mention in https://github.com/alphagov/paas-cf/pull/82/files#r51403781

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -234,118 +225,173 @@ jobs:
...
+              NAME: cf
+              URL: https://bosh.io/d/github.com/cloudfoundry/cf-release?v="${VERSION}"
+              BOSH_PASSWORD: {{bosh_password}}
+              TYPE: release
+            image: docker:///concourse/bosh-deployment-resource
             platform: linux
-            image: docker:///governmentpaas/curl-ssl
+            inputs:
+            - name: paas-cf
+            - name: bosh-secrets
```
keymon: Remove this, as mention in https://github.com/alphagov/paas-cf/pull/82/files#r51403781

keymon:
Now is good :) :+1:

dcarley:
Thanks :tada:

PR #83, opened 2016-01-29
[#112610295] Revert "Avoid new cert / ELB error"

dcarley:
## What

This reverts commit f46ff52b653b80b1ef538207121bfd20192bfde7.

When alphagov/paas-docker-terraform#10 is merged to upgrade Terraform to
0.6.10 then we no longer need this workaround.

## How to test

alphagov/paas-docker-terraform#10 should be merged first. But if you want to test both together than you can temporarily change all of your pipelines to use the container built from the branch (be sure to undo this later with `git checkout -p`):
```
gsed -ri 's/(docker-terraform)/\1#terraform_0.6.10/' concourse/pipelines/*
```

Deploy the pipelines to the Bootstrap Concourse machine to use this BRANCH:
```
BRANCH=112610295-terraform_elb_sleep ./vagrant/deploy.sh
```

Run both the `create-deployer` and `destroy-deployer` pipelines and confirm that they complete successfully. It's the `concourse-terraform` task that we're changing here.

## Who can review

Not @dcarley

### DIFF
--- a/terraform/concourse/elb.tf
+++ b/terraform/concourse/elb.tf
@@ -2,10 +2,6 @@ resource "aws_iam_server_certificate" "concourse" {
   name = "${var.env}-concourse"
   certificate_body = "${file("concourse.crt")}"
   private_key = "${file("concourse.key")}"
-
-  provisioner "local-exec" {
-    command = "sleep 10"
-  }
 }
 
 resource "aws_elb" "concourse" {


mtekel:
Got
```
Error applying plan:

1 error(s) occurred:

* aws_elb.concourse: Failure adding new or updated ELB listeners: CertificateNotFound: Server Certificate not found for the key: arn:aws:iam::988997429095:server-certificate/mt2-concourse
	status code: 400, request id: 868d0bb1-c697-11e5-b5b1-eda6a77e5b4a
```
in concourse-terraform. I've hijacked container to verify that terraform is 0.6.10 and it was. On 2nd run it works fine.

I am not sure what proper logic should be here, as it is AWS that takes a while since cert upload to make it available via other APIs...

mtekel:
So now I get invalid cert in the ELB and it refuses to accept connections. 
image
However the instance in ELB is marked as healthy.
image

mtekel:
For some reason, in the 2nd run it generated new cert. Afterwards it's not generating certificates anymore. TF is not updating the ELB (as there's no changes). I tried re-adding the instance, which is again marked as healthy, but ELB still won't accept connections.

mtekel:
Changed cert by hand to one called like my deployment. Now all works and TF doesn't want to update it. Not sure what happened the 1st time and what was the cert applied to ELB. Could be also AWS issue.

dcarley:
Sad times :cat: 

I'll do some repeat runs and more investigation on Monday morning.

mtekel:
I want to complete env. deletion and try to deploy from scratch. This could be some other TF/AWS issue which happens either only randomly or when you try to update existing cert. In that case we don't need to revert this.

dcarley:
I've run `create-deployer` and `destroy-deployer` a few times this morning and haven't been able to reproduce the same error as you. Can you try again?

mtekel:
This works well when building from scratch, but when you are changing the cert (you had env. build from previous master that didn't generate certs yet or e.g. you lost the cert from your bucket), it fails:
```
aws_iam_server_certificate.concourse: Destroying...
Error applying plan:

1 error(s) occurred:

* aws_iam_server_certificate.concourse: timeout while waiting for state to become '[success]'
```

This is a bit different from previous error, as here TF is trying to destroy the cert that it put previously on ELB, so it knows about it and tries to destroy it first...

dcarley:
> This works well when building from scratch, but when you are changing the cert (you had env. build from previous master that didn't generate certs yet or e.g. you lost the cert from your bucket), it fails

Can you tell if this PR changes that behaviour in any way? If not, then we should merge it as-is.

I'm not sure that we need to do anything about the upgrade path (while we're still in early development and destroying environments frequently) or losing state files (what circumstances would this happen? I'm not sure any of the pipelines are resilient to state files being accidentally deleted).

mtekel:
I can test if that sleep affects that and if not then I'll merge this...

dcarley:
Cool, thanks

mtekel:
Same behaviour. Updating certs simply isn't supported. We should think about how we do this (e.g. on production, when the certs expire).

PR #84, opened 2016-01-29
Reorder the pipelines in the bootstrap concourse

alext:
Concourse lists the pipelines in the side menu in the order they were
added. It also shows the first pipeline in the list when you visit the
root URL.  Given the first pipeline that's normally used is the
create-deployer one, it makes sense for this one to be first in the
list.


timmow:
Deployed my vagrant machine and pipelines loaded as expected

PR #88, opened 2016-02-01
[#112130611] Enable ssl on bbs traffic

actionjack:
[Use hard-coded SSL certs on Diego Components for everything except etcd](https://www.pivotaltracker.com/story/show/112130611)

## What

When we installed Diego we disabled SSL on some components. This story closes this out by enabling SSL.

**Background**
The original purpose of this story was to enable SSL on all the components unfortunately, we had issues with the new Diego etcd cluster's configuration conflicting with the existing one that prevented the existing etcd cluster (which is the etcd cluster used by loggregator and routing-API) from starting.

This PR has been written to the following Narrative:
* I want to:
  * Add the new SSL certificates and key assets for [BBS](https://github.com/cloudfoundry-incubator/bbs) generated using [certstap](https://github.com/square/certstrap) to the SSL certificate store.
  * I want to enable SSL-secured communication of BBS traffic between Diego components by providing the correct `SSL` information and enabling it.
  * I want our tests to run without any errors.

## How to test this PR

* This PR can be tested by creating an environment using the remote concourse deployer i.e. 
```
BRANCH=112130611/enable-ssl-on-bbs-traffic FLY_TARGET=remote CONCOURSE_ATC_PASSWORD=${MY_ATC_PASSWORD) concourse/scripts/deploy-cloudfoundry.sh ${MY_DEPLOY_ENV}
```
* Triggering the `deploy-cloudfoundry` concourse task
* `bosh ssh` onto the `colocated_z1 0` node can do a `tcpdump` on the BBS ports:
```
$ bosh ssh colocated_z1 0
$ sudo su -
$ openssl s_client -connect localhost:8889
```

## Who merge this PR

Anyone but @Jonty, @jimconner and @timmow and @actionjack


manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -257,19 +257,19 @@ properties:
...
 
     bbs:
       active_key_label: keylabel1
       encryption_keys:
         - label: keylabel1
-          passphrase: "a secure passphrase"
-      ca_cert: (( grab secrets.consul_ca_cert ))
-      server_cert: (( grab secrets.consul_server_cert ))
-      server_key: (( grab secrets.consul_server_key ))
+      passphrase: "a secure passphrase"
```
keymon: I think this `passphrase` has been unindented by accident. 

Anyway, do we use the `encription_keys` at all? More info [in the spec](https://github.com/cloudfoundry-incubator/diego-release/blob/develop/jobs/bbs/spec#L35) and [the template](https://github.com/cloudfoundry-incubator/diego-release/blob/7045d0f81845ff1e105c75163d9c901f2d4067ef/jobs/bbs/templates/bbs_ctl.erb#L85)

manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -279,16 +279,20 @@ properties:
...
+      require_ssl: true
 
     converger:
       bbs:
         api_location: bbs.service.cf.internal:8889
         ca_cert: (( grab secrets.consul_ca_cert ))
-        client_cert: (( grab secrets.consul_agent_cert ))
-        client_key: (( grab secrets.consul_agent_key ))
+        client_cert: (( grab secrets.bbs_client_cert  ))
+        client_key: (( grab secrets.bbs_client_key  ))
```
keymon: Although I think is fine for the current story use one certificate for all clients, in the future we should create a certificate for each client.

manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -279,16 +279,20 @@ properties:
...
+        require_ssl: true
+      ca_cert: (( grab secrets.bbs_ca_cert ))
+      server_cert: (( grab secrets.bbs_server_cert ))
+      server_key: (( grab secrets.bbs_server_key ))
```
keymon: Here we are enabling SSL for the service [converger](https://github.com/cloudfoundry-incubator/converger), not only BBS, but using the same certificates than BBS.

Is that intended? Shall we create a different certificate for converger?

manifests/cf-manifest/outputs/cf-ssl-certificates.yml
```diff
@@ -1,125 +1,116 @@
 secrets:
-  router_ssl_cert:  |
+  bbs_ca_cert: |
```
dcarley: This first commit is quite difficult to read because it appears to have introduced new items and re-ordered existing items at the same time. Could you split it into two separate commits?

It would also be good to add the `certstrap` commands that you used to the commit message for future reference.

manifests/cf-manifest/outputs/cf-ssl-certificates.yml
```diff
@@ -228,3 +249,397 @@ secrets:
...
+    9lcnAX2y7/ypULbH/0Fs2d1lSAm8jH5hViV+/nf5WqH4bUeyHeIrhfvT+C5PDR1p
+    N/iHm2YExxha/Jk5Mi+rd0jVXH8I1SQ0ynxSv61Nm5vaxQaKHnVKVjJuYBR2LJwA
+    7El7SUR/aKffkXNF83aV7lYOZo5A1r4Fqus4YaJ+4pIbTw0FzOGInQjRrMh74YmM
+    LzLhGip9CpBqfQ36KR01uuzZ3Y5p14LGY+ZDwstwLkV/aik0T0sVFIAfUVeQXCpt
+    x4fjef8CFMgwdKrWWoXaEq9zUav7D329bjsaTqW6lYbhEO42UxVSpPu1h/UpeCIc
+    tKWqXFSjsaddXy3UsNOfsNRcW+EiHayhb2FmWYKYTKzjbsztQOKt8vI1UyXYT43O
+    kdtMmjGxvHMBp0Oy2MRPkHQTwi4cSDy6hWr+UgF/nJ5AC2I50LAM478r88s89zR3
+    Dyo5mgTvYM3lA9hHdK6nnQ6ZX2uLNLA2ilCHTejqYUejdFhC
+    -----END CERTIFICATE-----
+  etcd_ca_cert: |
```
dcarley: I don't think these should be added yet if this PR only relates to BBS.

manifests/cf-manifest/deployments/040-cf-properties.yml
keymon: Actually, checking out the [spec of converger](https://github.com/cloudfoundry-incubator/diego-release/blob/v0.1441.0/jobs/converger/spec), it seems that these options `diego.converger.require_ssl`, `diego.converger.ca_cert`, `diego.converger.server_cert` and `diego.converger.server_key` are not being used :-m.

I will double check if that is the case and they can be removed.

keymon:
Despite the comments this deploys and smoke test pass. I will run acceptance tests and do some additional checks.

manifests/cf-manifest/deployments/040-cf-properties.yml
keymon: I can confirm that these options are not used at all. I removed them from my manifest and deployed again, no changes where applied to the deployed vms.

I think we can remove them safely.

keymon:
This deploys OK and passes the smoke and acceptance tests. 

If the comments above are fixed, I will be happy to merge:
 * Do not include not used certs (etcd)
 * Add in commit comment the commands used to generate the certificates.
 * Remove the unnecessary properties `diego.converger.require_ssl`, `diego.converger.ca_cert`, `diego.converger.server_cert` and `diego.converger.server_key` 
 * Fix the indentation (or remove) of the property `bbs.encryption_keys[label=keylabel1].passphrase`

actionjack:
@keymon updated.

actionjack:
@dcarley: I split it all out into separate commits so you can see what goes in and what get moved.

keymon:
LGTM, I am deploying and will run the tests after. I am testing in a branch rebased with master, to be sure that will work after merging.

actionjack:
@keymon  I've cleaned up the converger commit.

manifests/cf-manifest/deployments/040-cf-properties.yml
actionjack: @keymon: I have removed the value.

keymon:
:+1:

PR #89, opened 2016-02-03
[#108966132] Use a separate domain for deployed apps.

alext:
## What

In order to prevent apps from potentially overwriting the routes for system components we want apps to be served from a different domain.

## Details

This uses the new domain (`cf-apps.paas.alphagov.co.uk`) for hosting apps deployed to Cloudfoundry. This domain has already been created in the trial account, and had DNS delegated.

This PR does not include changing SSL certs used to serve apps (they'll still use the self-signed `*.cf.paas.alphagov.co.uk` cert). This will be addressed in another PR.

As part of this change we've also changed the hostname for the deployer-concourse to `deployer.<env>.cf.paas.alphagov.co.uk` to bring it inline with other components and disambiguate from the bootstrap-concourse.

## How to review

Deploy a complete environment from this branch, and verify that it all works correctly. Push an app to the deployment, and verify that it is served correctly from the new domain.

Verify that it's possible to push an app called "api" and that it won't affect the system functionality.

Verify that it's not possible to create an app on the system domain when not in the system CF org (called "GDS").

## Who can review

Anyone but @timmow and myself.

### DIFF
--- a/terraform/cloudfoundry/outputs.tf
+++ b/terraform/cloudfoundry/outputs.tf
@@ -14,6 +14,10 @@ output "cf_root_domain" {
   value = "${var.env}.${var.system_dns_zone_name}"
 }
 
+output "cf_apps_domain" {
+  value = "${var.env}.${var.apps_dns_zone_name}"
+}
+
 output "elb_name" {
   value = "${aws_elb.router.name}"
 }
--- a/terraform/cloudfoundry/outputs.tf
+++ b/terraform/cloudfoundry/outputs.tf
@@ -11,7 +11,7 @@ output "ssh_elb_name" {
 }
 
 output "cf_root_domain" {
-  value = "${var.env}.${var.dns_zone_name}"
+  value = "${var.env}.${var.system_dns_zone_name}"
 }
 
 output "elb_name" {
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ When complete you should:
 1. Access the UI from a browser with the same credentials as your
   *Bootstrap Concourse*.
 
-  - `https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk/`
+  - `https://deployer.${DEPLOY_ENV}.cf.paas.alphagov.co.uk/`
 
 1. Add a new target to the `fly` CLI utility:
 
@@ -115,7 +115,7 @@ DEPLOY_ENV=<deploy-env>
 $(./vagrant/environment.sh $DEPLOY_ENV) # get the credentials
 
 echo -e "admin\n${CONCOURSE_ATC_PASSWORD}" | \
-   ${FLY_CMD} -t ${DEPLOY_ENV} login -k -c "https://${DEPLOY_ENV}-concourse.cf.paas.alphagov.co.uk"
+   ${FLY_CMD} -t ${DEPLOY_ENV} login -k -c "https://deployer.${DEPLOY_ENV}.cf.paas.alphagov.co.uk"
 ```
 
 ### Destroy
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -246,7 +246,7 @@ jobs:
             . vpc-terraform-outputs/tfvars.sh
             openssl req -x509 -newkey rsa:2048 -keyout concourse.key \
               -out concourse.crt -days 365 -nodes -subj \
-              "/C=UK/ST=London/L=London/O=GDS/CN={{deploy_env}}-concourse.${TF_VAR_dns_zone_name}"
+              "/C=UK/ST=London/L=London/O=GDS/CN={{deploy_env}}-concourse.${TF_VAR_system_dns_zone_name}"
             tar czvf concourse-cert.tar.gz concourse.crt concourse.key
 
     - task: terraform-apply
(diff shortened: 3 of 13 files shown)


mtekel:
I think you should add functionality/changes from https://github.com/alphagov/paas-cf/commit/d42baa55ed13ce44c07ce9e0e7e18e7e4621a451

* a test that checks that the app and system domains are different
* the fix for manifests where app domain is defined in multiple places
* refactor manifests to explicitly use app domain wherever system/root/shared domain was used previously, just because it was the same as app domain

alext:
> a test that checks that the app and system domains are different

I looked at adding that test, but I don't think it adds much value as it actually only ends up testing that the values in the test fixtures are different.

> the fix for manifests where app domain is defined in multiple places

We looked at refactoring that, but settled for leaving it like this. The way things are defined at the moment the `terraform_outputs` are only referenced by manifests under the `aws` directory, which seems correct. We additionally wanted to keep the setting of the `apps_domains` consistent with the setting of the `system_domain`.

> refactor manifests to explicitly use app domain wherever system/root/shared domain was used previously, just because it was the same as app domain

I'm not sure I follow what you mean by this. The `apps_domain` is only used in one property in the manifest. Can you point to an example of what you mean here.

mtekel:
I should be able to run tests against real manifest, not only against one generated using fixtures. Then the test is useful.

In 030, the app domain is defined again. This is quite confusing, as it gets overwritten in 999. You should remove definition from 030. 

Then there's properties.domain and properties.system_domain. Currently both the same. Yet they are referenced multiple times in different places and sometimes one is used and sometimes other....

alext:
> In 030, the app domain is defined again. This is quite confusing, as it gets overwritten in 999. You should remove definition from 030.

I've removed the duplicated defintions from `030-jobs.yml`, because I agree that could lead to confusion.

alext:
> I should be able to run tests against real manifest, not only against one generated using fixtures. Then the test is useful.

We are running the tests against a real manifest.  It's the terraform outputs that use fixtures because they're specific to an individual deployment.

The logic in this manifest it merely pulling the two values from the outputs. There's nothing here that enforces that they be different. It therefore doesn't make sense for the unit tests to cover this because its not logic implemented in the unit being tested.

mtekel:
We have a test for e.g. IP address uniqueness, why not domain uniqueness?

mtekel:
Also, we still have properties.domain and properties.system_domain in 999 cf stub. Some jobs reference properties domain (e.g. doppler and loggregator in 035), but some other configuration references properties.system_domain (in 900 tests). And we have a test to explicitly require system_domain being set. Are properties.domain and properties.system_domain intended to be different and is it just current coincidence that they are set to the same value in 999 stub?

terraform/cloudfoundry/outputs.tf
```diff
@@ -11,7 +11,11 @@ output "ssh_elb_name" {
 }
 
 output "cf_root_domain" {
```
combor: Why we are outputting cf_root_domain here if in terraform/vpc/outputs.tf we are outputting system_dns_zone_name with the same value?

dcarley:
> The logic in this manifest it merely pulling the two values from the outputs. There's nothing here that enforces that they be different. It therefore doesn't make sense for the unit tests to cover this because its not logic implemented in the unit being tested.

Is it useful to have a test to confirm that the app domain and system domain aren't a sub-domain of one another? Even if it's an acceptance test..

terraform/cloudfoundry/outputs.tf
alext: They're not the same value.  `cf_root_domain` includes the environment component.

combor: Ah, yes, indeed.

alext:
> Also, we still have properties.domain and properties.system_domain in 999 cf stub.

There are both required properties. See http://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/cf-release&version=225#p=domain and http://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/cf-release&version=225#p=system_domain. I'm not 100% sure what the difference between the 2 is. From reading those docs and the default manifests, there's some contradiction here. According to the docs, domain is used by the api, and system_domain is used by all the other system components (uaa etc), but from reading the default manifests it would seem that domain is used by all the system components, and system_domain is only used to reserve a domain in the router to prevent it being used by other CF orgs.

Either way, given this story is about creating a separate app domain, I think refactoring the system_domain stuff is out of scope.

mtekel:
I agree about the scope, but it's a simple fix. We want them to be the same, so let's define system_domain and then simply say that properties.domain = (( grab properties.system_domain )). All other references in manifest stubs should only use properties.domain. We should add test that ensures both are the same.

alext:
> Is it useful to have a test to confirm that the app domain and system domain aren't a sub-domain of one another? Even if it's an acceptance test..

I've created https://www.pivotaltracker.com/story/show/112988983 to capture this.

combor:
Seems to be working fine:
```
$ cf push api -d cf.paas.alphagov.co.uk  -b staticfile_buildpack
Creating app api in org test / space test as test...
OK

FAILED
Domain cf.paas.alphagov.co.uk not found

$ cf create-domain test cf.paas.alphagov.co.uk
Creating domain cf.paas.alphagov.co.uk for org test as test...
FAILED
Server error, status code: 400, error code: 130001, message: The domain is invalid: name overlapping_domain
```

PR #90, opened 2016-02-03
[#112503409] separate subnet for cell

combor:
# What 

This PR adds separation for Diego Cell. It creates new Cell subnet and moves its components from colocated VM to dedicated one. 

# Details

- new cell subnet per AZ, ranges from 10.0.20.0 to 10.0.22.0
- separation of rep, garden and rootfs from collocated VM to dedicated one
- new route table association as we need internet to access API and for apps
- extension of tests to include new networks
- Terraform version bump to reflect version in TF container

# How to test 

Use this branch to deploy new CF instance. Run smoke and acceptance tests.

Note that some acceptance tests may fail as they are known to be flaky.

# Who can test 
Anyone but @combor and @jimconner

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.6.8"
+    - TF_VERSION="0.6.10"
     - SPRUCE_VERSION="0.13.0"
 
 addons:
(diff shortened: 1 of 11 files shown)


actionjack:
:+1: Looks good to me!

.travis.yml
```diff
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.6.8"
+    - TF_VERSION="0.6.10"
```
dcarley: I can guess why you've done this and it's good, but can you split this out to a separate commit that describes why?

combor: Done.

PR #92, opened 2016-02-03
[#111428758] Generate certs for cloudfoundry in each deployment

keymon:
[Auto-generate SSLs for all internal components](https://www.pivotaltracker.com/story/show/111428758)

## What

We want to generate the certificates for all the components, signed by a CA which is trusted by all the related VMs. We will generate the certificates for all the current SSL certificates used in the deployment, as the ones hardcoded [here](https://github.com/alphagov/paas-cf/blob/7b649d96e6b730bfaf9980717c3ed08bbc20899d/manifests/cf-manifest/outputs/cf-ssl-certificates.yml) 
 
## Dependencies

This PR must be merged first https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/16
 and the a dockerhub automated build container added and used in this PR.

## Context

 * I want to generate a CA certificate to sign the rest of certificates and that is setup as a [trusted certificate](https://bosh.io/docs/trusted-certs.html) in all deployed VMs.
 * I want to generate and use the certificates for the gorouter with the deployment entry domain.
 * I want to to pass the generated CA to consul and BBS services.
 * I want to generate the signing key for UAA tokens.
 * I want to remove any hardcoded certificate or key.

Additionally:
 * I want to make it easy to add new YAML stubs to the bosh and CF manifest.
 * I want to make it easy to generate and add new certificates, even new CAs.
 * I want all tests to pass.

## Notes

 * In the future we should use a real SSL certificate for the router, but for the time being is OK to use one generated by us.
 * In this PR we are, *again* abusing of the script s3init.  Because that, the generation jobs are not connected to the consumer. To solve this, we suggest  that in the future we refactor the certificates as resources, with this logic for `in`: *"If the certificate is in a S3 bucket and will not expire, download it from S3. If not, generate it and upload it to the S3 bucket"*

## How to test this?

 1. Deploy the system as usual. Note that **you must redeploy bosh** as, a new config is required, and  redeploy all CF vms **from scratch** so they get the trusted certificate.
     * If the deployment is OK, the certificate work for BBS, consul.
 2. Run `smoke_tests`, they should pass.
 3. Run `acceptance_tests`, but note that some tests as flacky and might fail.

You can also check if the gorouter certificate is accepted by any other VM by SSHing  to one server and doing `https://api.${DEPLOY_ENV}.cf.paas.alphagov.co.uk/v2/info`

Example, from one cell:

```
bosh_z2lictl6z@52cfb1f0-bd9b-4b90-9463-2a8ac748da70:~$ curl https://api.michael.cf.paas.alphagov.co.uk/v2/info
{"name":"vcap","build":"2222","support":"http://support.cloudfoundry.com","version":2,"description":"Cloud Foundry sponsored by Pivotal","authorization_endpoint":"https://login.michael.cf.paas.alphagov.co.uk","token_endpoint":"https://uaa.michael.cf.paas.alphagov.co.uk","min_cli_version":null,"min_recommended_cli_version":null,"api_version":"2.43.0","app_ssh_endpoint":"ssh.michael.cf.paas.alphagov.co.uk:2222","app_ssh_host_key_fingerprint":null,"app_ssh_oauth_client":"ssh-proxy","routing_endpoint":"https://api.michael.cf.paas.alphagov.co.uk/routing","logging_endpoint":"wss://loggregator.michael.cf.paas.alphagov.co.uk:443","doppler_logging_endpoint":"wss://doppler.michael.cf.paas.alphagov.co.uk:443"}
```

But from our personal computer:
```
$ curl https://api.michael.cf.paas.alphagov.co.uk/v2/info
curl: (60) SSL certificate problem: unable to get local issuer certificate
```

# who?

Anyone but @keymon or @mtekel


concourse/pipelines/create-microbosh.yml
```diff
@@ -131,6 +136,42 @@ jobs:
...
+      trigger: true
+      passed: [ init ]
+    - get: paas-cf
+      passed: [init]
+    - get: bosh-CA
+
+    - try:
+        task: generate
+        config:
+          image: docker:///keymon/certstrap
```
saliceti: This should be an official governmentpaas image

Jonty:
I'm still reviewing this, however I feel that 26 commits is way too many for one PR - it looks like many of these should be squashed into logical chunks?

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -168,13 +232,99 @@ jobs:
...
+  - name: generate-cf-certs
+    plan:
+      - aggregate:
+        - get: paas-cf
+          passed: ['init']
+          trigger: true
+        - get: bosh-CA
+      - task: generate-cf-CAs
+        config:
+          image: docker:///keymon/certstrap#withcurl
```
Jonty: This should also be an official governmentpaas image.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -168,13 +232,99 @@ jobs:
...
+                  tar -cvzf ../${ca}.tar.gz ${ca}.*
+                  ../paas-cf/concourse/scripts/s3init.sh {{state_bucket}} ${ca}.tar.gz ../${ca}.tar.gz
+                  cd ..
+                else
+                  echo "CA ${ca} is already generated, skipping."
+                fi
+              done
+      - task: generate-cf-certs
+        config:
+          image: docker:///keymon/certstrap#withcurl
```
Jonty: This should also be an official governmentpaas image.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -182,6 +332,49 @@ jobs:
...
+                cn=${cert%%-cert}
+                yaml_key_name=$(echo ${cn} | tr A-Z- a-z_)
+
+                tar -xzf ${cert}/${cert}.tar.gz
+                paas-cf/concourse/scripts/file-to-yaml.sh secrets ${yaml_key_name}_key ${cn}.key  > ${cn}_key.yml
+                paas-cf/concourse/scripts/file-to-yaml.sh secrets ${yaml_key_name}_cert ${cn}.crt  > ${cn}_cert.yml
+              done
+      - task: uaa_jwt_verification_key
+        config:
+          image: docker:///keymon/certstrap#withcurl
```
Jonty: This should also be an official governmentpaas image.

saliceti:
I don't see the point of using more than one CA. After all, we only use one Verisign?
We need to think about securing the private key, managing one intermediate CA, etc. Doing that for 3 different CAs would add even more complexity.

keymon:
@Jonty About squashing this: yes and no. It depends on the coding style and the one we are following is encouraging small commits with small changes.

mtekel:
Regarding multiple CAs: There are tradeoffs in each approach you can take. We have also considered the main CA and bbs, consul and bosh being intermediate CAs. We chose to have independent CAs to simplify management and updates in case only one component is affected (e.g. bbs). This way you have to re-create only the affected certs and only update affected components. If you use bosh CA everywhere, in case of compromise you need to re-deploy all the VMs and re-generate all the certs.

concourse/pipelines/deploy-cloudfoundry.yml
keymon: Yes, this must be merged first and the dockerhub image setup: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/16

keymon:
After https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/16 has been merged, I updated the docker name.

dcarley:
> @Jonty About squashing this: yes and no. It depends on the coding style and the one we are following is encouraging small commits with small changes.

There are a lot of commits. Though there's no concrete rule about what is too much or too little. Looking over them I can't see a lot of commonality, besides generating certificates, but combining all of the pipeline changes might be more confusing. I think it's OK to leave as-is, unless anyone has specific suggestions. I do think it's worth discussing though and people shouldn't be afraid to ask/suggest.

> Regarding multiple CAs: There are tradeoffs in each approach you can take. We have also considered the main CA and bbs, consul and bosh being intermediate CAs. We chose to have independent CAs to simplify management and updates in case only one component is affected (e.g. bbs). This way you have to re-create only the affected certs and only update affected components. If you use bosh CA everywhere, in case of compromise you need to re-deploy all the VMs and re-generate all the certs.

We should go with simple for now. Multiple CAs are more difficult to manage, especially if there isn't a single parent. We'll need to refactor them at a later date anyway to generate/store them in a more secure way. I'm also not sure how confident we can be about the best way to split them up at this stage. Can you refactor them into a single CA, please?

keymon:
I changed it to use bosh-CA for everything.

Jonty:
Can confirm this works everywhere after the key fixes. That was not fun to review!

PR #93, opened 2016-02-04
Enable parallel job updates

mtekel:
### What
[Enable parallel job deployment](https://www.pivotaltracker.com/projects/1275640/stories/112501849).

In this PR we:

* Remove per-job bosh update directives and use global settings.
* Set 2 parallel updates per job (using max_in_flight bosh directive). We should update this later based on actual number of instances per job we will have in production.
* Disable serial updates. This enables bosh to deploy jobs in parallel.
* We also disable canaries. This way it's only max_in_flight that controls how many instances of a job are updated in parallel.
* Still deploy consul_z1 job first (both instances in parallel). This is to enable other jobs to cleanly connect to consul immediately when starting deployment. The jobs that need to do this would have to wait for consul server to become available anyway, so deploying it as a 1st job is not a slowdown.

### Testing

Deploy CF or make a change that requires update of all jobs & apply. Observe that consul_z1 job instances get deployed/updated in parallel first and then the rest of the jobs being all updated/deployed in parallel. Smoketests should pass after deploy/update. Optionally, you can also run acceptance tests, but some of them fail randomly (there are some race conditions, mainly for CF security groups test).

### Who
Anyone from our team except @saliceti or @mtekel

### DIFF
--- a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
@@ -25,8 +25,8 @@ compilation:
   reuse_compilation_vms: true
 
 update:
-  canaries: 1
-  max_in_flight: 1
+  canaries: 0
+  max_in_flight: 2
   canary_watch_time: 30000-600000
   update_watch_time: 5000-600000
-  serial: true
+  serial: false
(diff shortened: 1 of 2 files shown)


actionjack:
:+1: Last update deploy was 16 minutes!

PR #94, opened 2016-02-04
[#112788839] Consolidate etcd clusters

actionjack:
[Reconcile etcd config/clusters](https://www.pivotaltracker.com/story/show/112788839)

## What

We have two different types of etcd jobs, which appear to be part of the same cluster but use a different configuration. We should reconcile these into a single cluster of the same configuration.

* I want to
  * Add a consul agent to the existing etcd cluster so I can resolve it as `etcd.service.cf.internal`
  * I want to apply the correct base configuration to the etcd configuration so it will start up cleanly
  * I want to remove the legacy etcd cluster from the co-located machines
  * I want to bump the version of etcd to 18 so it's [compatible](https://github.com/cloudfoundry-incubator/diego-cf-compatibility/blob/master/compatibility-v2.csv) with Diego and the CF release at the moment we are using the version of etcd bundled with CF release 225 (which is 14).

## How to review this PR

* To review this PR deploy a new environment and run smoke tests against it,
* You can also ssh into an etcd node and run `dig etcd.service.cf.internal` to see if it resolves correctly and check the etcd service is running using `monit summary` and check the etc logs under `/var/vcap/sys/log`

**tip!** you can cherry pick @alext 's parallel VM commit to deploy faster `git cherry-pick 9b56c7798874d6f9d006900e8019bc0011a767bc`

## Who should review this PR

Anyone except @keymon and @actionjack

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -6,6 +6,7 @@ meta:
   release:
     name: cf
 
+
   consul_servers: (( grab jobs.consul_z1.networks.cf1.static_ips jobs.consul_z2.networks.cf2.static_ips ))
 
   nfs_client_ranges:
@@ -65,9 +66,9 @@ meta:
 
   etcd_templates:
   - name: etcd
-    release: (( grab meta.release.name ))
+    release: etcd
   - name: etcd_metrics_server
-    release: (( grab meta.release.name ))
+    release: etcd
   - name: metron_agent
     release: (( grab meta.release.name ))
   - name: consul_agent
@@ -226,8 +227,6 @@ jobs:
         agent:
           services:
             etcd: {}
-    update:
-      max_in_flight: 1
 
   - name: etcd_z2
     templates: (( grab meta.etcd_templates ))
@@ -244,8 +243,6 @@ jobs:
         agent:
           services:
             etcd: {}
-    update:
-      max_in_flight: 1
 
   - name: stats_z1
     templates: (( grab meta.stats_templates ))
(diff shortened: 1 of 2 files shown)


dcarley:
@actionjack Could you add some more detail to the first three commit messages about why we're performing each change? It'd be difficult for someone reading those in the future to understand the good rationale for doing this.

saliceti:
Tested successfully with smoke tests, plus loooking at loggregator because it relies on etcd.

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -231,6 +241,12 @@ jobs:
     properties:
       metron_agent:
         zone: z2
+      consul:
+        agent:
+          services:
+            etcd: {}
+    update:
+      max_in_flight: 1
```
mtekel: We should not be adding that back in.

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -219,6 +223,10 @@ jobs:
     properties:
       metron_agent:
         zone: z1
+      consul:
+        agent
```
mtekel: should be `agent:`

mtekel:
It looks like database_z1 and database_z2 are still configured to deploy etcd. Even though they use the same domain name `etcd.service.cf.internal` - meaning they will connect to existing etcd cluster, which is the same relaase, so that should work. Is this intended? Maybe the database services are connected to use local ETCD instances to speed up access? Should we then run ETCD on the collocated machines as well?

mtekel:
After long struggle with ETCD, which seems to be related to randomness in parallel deployment, all smoketests pass. Merging.

PR #98, opened 2016-02-08
[#113257145] Fix etcd cluster

saliceti:
# What
We observed failures in etcd clustering that were due to a race condition caused by  https://github.com/alphagov/paas-cf/pull/94 and https://github.com/alphagov/paas-cf/pull/93 together.
We now force one etcd node to build before the others. Also, we standardise on DNS resolution via consul to connect to etcd.

# How to review
Build a new cloud foundry. It should pass smoke tests.

You can also test if the etcd cluster is complete. On an etcd node, run:

```
# /var/vcap/jobs/etcd/packages/etcd/etcdctl member list
3bb81497205219e5: name=etcd-z1-0 peerURLs=http://10.0.10.20:7001 clientURLs=http://10.0.10.20:4001
426443662bae1c80: name=etcd-z1-1 peerURLs=http://10.0.10.35:7001 clientURLs=http://10.0.10.35:4001
f2b470f31458c77f: name=etcd-z2-0 peerURLs=http://10.0.11.19:7001 clientURLs=http://10.0.11.19:4001
```
It should return 3 nodes.

# Who can review
Anyone but @saliceti or @mtekel

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -6,6 +6,7 @@ meta:
   release:
     name: cf
 
+  etcd_consul_service: ["etcd.service.cf.internal"]
 
   consul_servers: (( grab jobs.consul_z1.networks.cf1.static_ips jobs.consul_z2.networks.cf2.static_ips ))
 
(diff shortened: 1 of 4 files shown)


dcarley:
@saliceti @timmow: Does this close out https://www.pivotaltracker.com/story/show/113280629 or do we need to review the rest of the jobs in the manifests?

saliceti:
This makes consul available on all jobs. It doesn't mean the jobs are using it.
Here only etcd was changed to reference the DNS name. The other story should still be played to check the other services.

dcarley:
> This makes consul available on all jobs. It doesn't mean the jobs are using it.
> Here only etcd was changed to reference the DNS name. The other story should still be played to check the other services.

Thanks :ok_hand:

dcarley:
How easy was this to reproduce before? Do I need to deploy more than once to confirm that it's fixed?

Could you include a description in the commit message that it's the addition of `update: serial: true` to `etcd_z1`, which is now the second job in the manifest, that prevents any other jobs from happening until that has completed?

I don't know if everyone knows that, but I hadn't understood that it blocked other jobs until @mtekel explained it recently.

> everyone thinks he’s the leader

etcd doesn't have a gender. While you're there could you change the commit message to "everyone thinks they are the leader"? :blush:

saliceti:
Yesterday before the fix I ran it once and it worked, but the second time it failed smoke test. I guess if you do it 2 or 3 times that should cover it.
I will change the commit messages.

saliceti:
Commit messages updated.

dcarley:
Thanks :+1:

Acceptance tests passed on my environment. Waiting for Travis to pass on the rebased commits.

PR #99, opened 2016-02-09
[#112503901] Implement Custom Acceptance Test Framework

alext:
# What

We need to put in place a framework to run our own set of acceptance tests and fit it into the pipeline for deploying an environment. These are in addition to the upstream acceptance-tests.

## Details

These are implemented using [ginkgo](http://onsi.github.io/ginkgo/) so that we can reuse the [upstream test helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers).

At the moment, the test only include a POC test that pushed a simple static app, and verifies that it works. This test can be removed once we've added some real tests here.

## How to review

Deploy cloudfoundry from this branch and verify that the deployment completes successfully, and that these acceptance tests are triggered, and pass.

## Who can review

Anyone but @actionjack and myself.

### DIFF
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -1,0 +1,39 @@
+# Custom acceptance tests
+
+These are custom acceptance tests for the CF deployment that are run in
+addition to the [upstream acceptance
+tests](https://github.com/cloudfoundry/cf-acceptance-tests). These are
+automatically run from the deploy-cloudfoundry concourse pipeline after the
+concourse deploy.
+
+## Dependencies
+
+The tests have the following dependencies:
+* a recent [Go compiler](https://golang.org/dl/)
+* the [CF CLI](https://github.com/cloudfoundry/cli#downloads).
+* `curl`
+* [`godep`](https://github.com/tools/godep)
+
+When running on concourse, these use our [cf-cli
+container](https://hub.docker.com/r/governmentpaas/cf-cli/).
+
+## Running the tests
+
+You must set an environment variable `$CONFIG` which points to a JSON file that
+contains several pieces of data that will be used to configure the acceptance
+tests, e.g. telling the tests how to target your running Cloud Foundry
+deployment.
+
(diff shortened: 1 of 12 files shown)


alext:
:rotating_light: **Note:** This depends on https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/19 being merged and built on dockerhub before the tests will run successfully.

dcarley:
All of the vendored files makes the diff hard to read but probably makes sense :panda_face:

Should 0a94263 be squashed into 345caf6? Or both of them into 1f12478? Do we need to document how to update the dependency versions and vendored files? (this surprised me recently when working on Terraform)

dcarley:
> All of the vendored files makes the diff hard to read but probably makes sense

Discussed this with @alext and @combor. Right now, ensuring that we have the right versions is more important than ensuring availability. We'll have other stories later to look at our dependencies on third-parties like GitHub. Alex is going to test using Godep without vendoring to see what the code looks like and how long it adds to the run time.

One other solution that was discussed was having the acceptance tests in a separate repo. We should keep them in the same repo, at least for now, because we're likely to make changes to the tests in-step with other config changes (e.g. manifests) and co-ordinating those across repos is hard.

alext:
I've created a separate branch that uses godep to download the dependencies on each run - https://github.com/alphagov/paas-cf/compare/acceptance_tests_no_vendor (which depends on https://github.com/alphagov/paas-docker-cloudfoundry-tools/compare/add_godep).

Does that look like a better approach?  If so I'll get the PRs created/updated.

alext:
I've updated this PR with the non-vendored approach.

This currently depends on https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/20, and the last commit here will need to be removed once that's merged.

alext:
I've removed the temporary commit now that the container PR has been merged.

dcarley:
LGTM

PR #100, opened 2016-02-09
[#110953182] Replace s3get with aws-cli

dcarley:
## What

This is some "gardening" preparation for the referenced story and to prevent other re-use of this script. I'm not doing the other work described in the story.

#### README: Replace use of s3get with aws-cli

We can use the official aws-cli tool to do the same job as `s3get.sh`. If we
also replace the use of the script in the pipeline, then we can remove this
script from our repo.

I've included a very brief description (with some links) about how to
install and configure it.

#### create-deployer: Replace s3get with aws-cli

Use the official aws-cli utility instead of our custom script. We hope to
replace this task with s3-resource, if we can add support for it to error
instead of block when the bucket doesn't exist:

- https://www.pivotaltracker.com/story/show/110953182

In the meantime though, replacing this means that we have less code to
maintain. This is especially important as we don't currently have any tests
for this functionality.

We originally created the aws-cli container for 5448cee.

#### try-fetch-bucket-state: explicit error handling

Add the `-e` flag to the `try-fetch-bucket-state` task to exit early on
errors, as used in all other tasks. Then add `|| true` to the command which
attempts to fetch the state from the bucket and we expect to fail on the
very first run. This has the same effect but makes error handling more
explicit to read.

## How to test

1. Destroy your existing Deployer Concourse (and BOSH/CF) or choose a new environment name, so that you can test this from scratch
1. Run the `create-deployer` pipeline and check that it completes successfully
1. Run the `create-deployer` pipeline again and check that it completes successfully

## Who can review

Not @dcarley


combor:
It would be good to squash https://github.com/alphagov/paas-cf/commit/7e3d126acf281c738503ab20d7425d456e2d22b7 and https://github.com/alphagov/paas-cf/commit/94fcf4e6f931a33276f34fcf9706daa86895a6a1 as it's a modification of the same line.

dcarley:
I've rebased against master to fix the merge conflict.

> It would be good to squash 7e3d126 and 94fcf4e as it's a modification of the same line.

I disagree, because it's the same line but the purpose is different. I called it out as a separate commit because the original behaviour surprised me. I had assumed that all of our tasks were using `set -e` and that `s3get.sh` was swallowing errors because it didn't have `set -e`. Which was incorrect, because the `curl` is the last command in the script and is what produced the exit code, we were just ignoring it in the pipeline.

combor:
This works OK. Merging...

PR #101, opened 2016-02-09
[#112876059] Deployment deletion failures

Jonty:
This PR makes deployment auto-deletion more robust in two ways:

* By deterministically delaying the deletion based on the environment name we ensure that fewer environments will not be auto-deleting at the same time, thus making us less likely to hit the AWS per-account rate limit.
* By using `--force` we cause bosh to attempt to remove the deployment even when some conditions are not met.

# How to test
Run the `autodelete-cloudfoundry` pipeline on a deployed environment - you should see `Sleeping for X seconds before deletion..` appear in the log, and the deletion will proceed X many seconds later.

# Who can review
Not @jonty or @timmow, who thinks this story was too easy and would like harder ones.

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -31,6 +31,15 @@ jobs:
           args:
           - -e
           - -c
+          # The following sleep monstrosity deterministically sleeps for a
+          # period of time between 0-20mins in order to prevent all our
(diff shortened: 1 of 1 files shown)


mtekel:
Applied to my deployment. I'll let it auto-delete tonight and see how it went tomorrow...

mtekel:
Works well. Didn't get any throttling on the last env. delete.

dcarley:
:clap:

PR #102, opened 2016-02-09
[#112503399] Implement isolation between system components and runners - ACLs for components

jimconner:
# What

We need to ensure that Diego cells only have access to the minimum set of components that they need, to further secure the deployed environment and to ensure that if other security mechanisms fail we do not fail open. We also implemented ACLs from the Infrastructure network as part of this story.

Additionally we have renumbered CF networks to fit into 20 bit subnets to make ACLs more readable and easier to implement.   

# How

This PR adds AWS Network ACLs to the CF components subnets to restrict the ports allowed access from the cell and infrastructure networks to the CF network. This is implemented as a number of Terraform manifests to create ACLs and rules within ACLs.

### How to test

* Deploy an environment as normal by following the instructions in the README.md 
* Confirm that smoke tests complete successfully during deployment.
* Confirm that Network ACLs have been created by checking in the AWS gui (In VPC->Network ACLS)
* Try pushing an app to be extra-sure that things are OK.
* Run the Acceptance Tests for extra brownie points.

### Notes

You may notice that some of the ACL rules are super-sets of other rules. (i.e. we have a rule which allows a wide port range, but also rules which allow individual ports within that range) - This is because the gorouter component will issue requests from a random port between 1024-65535, and the responses from the diego cell back to these ports must be permitted). The intention is to later move the gorouter component to it's own dedicated subnet, at which point these rules can be modified more easily than having to capture all of the individual ports once again.

### Who can review

Anyone on the team except @combor or myself


terraform/cloudfoundry/acl_bbs.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_network_acl_rule" "60_bbs" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = "tcp"
+    rule_number = 60
+    rule_action = "allow"
+    from_port = 8889
+    to_port = 8889
+    cidr_block = "10.0.32.0/20"
```
dcarley: Can you move all references to `10.0.32.0/20` to a named variable such as `cell_cidr_all`? It'll make it easier to understand the meaning of the value and reduce the chance of us making a typo.

terraform/cloudfoundry/acl_cf_local_in.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_network_acl_rule" "90_cf_local_in" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = -1
+    rule_number = 90
+    rule_action = "allow"
+    cidr_block = "10.0.16.0/20"
```
dcarley: Ditto the comment in https://github.com/alphagov/paas-cf/pull/102/files#r52454687 - but called `cf_cidr_all`?

terraform/cloudfoundry/acl_bosh.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_network_acl_rule" "110_bosh" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = "tcp"
+    rule_number = 110
+    rule_action = "allow"
+    from_port = 0
+    to_port = 65535
+    cidr_block = "10.0.0.6/32"
```
dcarley: Ditto the comment in https://github.com/alphagov/paas-cf/pull/102/files#r52454687 - we have already have a variable for this: https://github.com/alphagov/paas-cf/blob/7a95a56745577c580ca189ac380d528da17ed82a/terraform/globals.tf#L79-L82

terraform/cloudfoundry/acl_cell_local_in.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_network_acl_rule" "90_local_in" {
+    network_acl_id = "${aws_network_acl.cell.id}"
+    protocol = -1
+    rule_number = 90
+    rule_action = "allow"
+    cidr_block = "10.0.0.0/16"
```
dcarley: Ditto the comment in https://github.com/alphagov/paas-cf/pull/102/files#r52454687 - we can use https://github.com/alphagov/paas-cf/blob/7a95a56745577c580ca189ac380d528da17ed82a/terraform/globals.tf#L14-L17

manifests/cf-manifest/deployments/900-cf-tests.yml
```diff
@@ -41,6 +41,8 @@ properties:
     include_operator: true
     include_services: true
     include_route_services: false
+    default_timeout: 500
+    nodes: 6
```
dcarley: Was this committed by accident? No mention in the commit message and I think it should be handled in [#113369495](https://www.pivotaltracker.com/story/show/113369495). Could you leave a comment there about whether this proved to be effective?

dcarley:
I'm in two minds about having the ACL resources split across multiple files. I understand why it was done, because there are a lot of them, but I wonder if it makes it harder to understand them in context. Especially when the rule IDs are significant to whether something is blocked or allowed.

Anyone else have any thoughts on this?

terraform/cloudfoundry/acl_internet.tf
```diff
@@ -0,0 +1,47 @@
...
+    egress = true
+}
+
+resource "aws_network_acl_rule" "130_internet_cf_in" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = -1
+    rule_number = 130
+    rule_action = "allow"
+    cidr_block = "0.0.0.0/0"
+    egress = false
```
dcarley: There's a missing closing brace here. Not sure why `terraform graph` considers that legal, though.

terraform/cloudfoundry/acl_router.tf
```diff
@@ -0,0 +1,32 @@
...
+    cidr_block = "10.0.0.0/20"
+    egress = false
+}
+
+resource "aws_network_acl_rule" "22_gorouter_replies" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = "tcp"
+    rule_number = 22
+    rule_action = "allow"
+    from_port = 1024
```
dcarley: @bleach and I had a chat about this. The ACLs are only really valuable if we have a plan to move the routers to a different (either the cell subnet or a completely new subnet) very shortly after, but I'm not sure what other implications that would have or when we'd do it. He's going to think about it and we'll talk about it together soon.

Could you hold off making changes for the other comments in the meantime?

dcarley:
Unassigning myself from this while the comments are addressed, because I seem to be assigned to all the PRs :smile_cat:

terraform/cloudfoundry/acl_internet.tf
dcarley: Reported to and fixed by upstream: https://github.com/hashicorp/hcl/issues/88

manifests/cf-manifest/deployments/900-cf-tests.yml
combor: This was left here by accident. When it comes to effectiveness than ```default_timeout``` definitely helped for some tests but ```nodes: 6``` doesn't seem to have big effect on tests speed as it took 50 minutes to finish.

terraform/cloudfoundry/acl_router.tf
combor: Fixed.

combor:
All comments are addressed now.

dcarley:
> I'm in two minds about having the ACL resources split across multiple files. I understand why it was done, because there are a lot of them, but I wonder if it makes it harder to understand them in context. Especially when the rule IDs are significant to whether something is blocked or allowed.
> 
> Anyone else have any thoughts on this?

We decided that a reasonable balance would be to group the rules into four files:

- CF in
- CF out
- Cell in
- Cell out

I'll do this so as not to pull Piotr off the story he's currently working on.

dcarley:
Important to note that this change won't apply to an existing CF environment because Terraform needs to destroy and recreate the subnet resources in order to re-number them, which it can't because BOSH has provision VMs within the subnets:
```
Error applying plan:

4 error(s) occurred:

* aws_subnet.cell.0: Error deleting subnet: timeout while waiting for state to become '[destroyed]'
* aws_subnet.cell.1: Error deleting subnet: timeout while waiting for state to become '[destroyed]'
* aws_subnet.cf.0: Error deleting subnet: timeout while waiting for state to become '[destroyed]'
* aws_subnet.cf.1: Error deleting subnet: timeout while waiting for state to become '[destroyed]'
```

dcarley:
@combor I've pushed a commit to regroup them. Can you see what you think?

combor:
Sorry for delay, I forgot to answer. It looks good. Additionally, commit messages are fixed. Can you please check?

dcarley:
LGTM.

I warned everyone in Slack yesterday afternoon that this change is not forwards compatible for an existing deployment. However this will apply cleanly to an environment where the deployment has been automatically deleted overnight, like this morning.

Just before merging it occurred to me that the commits were missing some details about the non-forwards compatible change and the importance of the router ACL. I've rebased to add some additional comments to this commits - hope you don't mind, it didn't seem worth delaying this PR further to ask you to do it.

We've prioritised [#113458409](https://www.pivotaltracker.com/story/show/113458409) to move the router to it's own subnet immediately after this.

PR #103, opened 2016-02-10
Fix concourse terraform state version pinning.

alext:
It's important that every terraform run uses the latest state file so
that it has an accurate reflection of what's been created etc.

Previously, the `cf.tfstate` used in the deploy-cloudfoundry pipeline was
pinned to the version that has passed the `init` step. This meant that
if a terraform run failed for any reason (eg AWS rate limit), and the
job was restarted, it would not use the updated statefile that included
the partial changes made in the previous run.

This changes the pipeline to use a non-pinned version of the statefile
to ensure it always uses the latest.

It also stops the `init` job from `put`ing the resource (leaving
`s3init.sh` to do the writing). This makes this pipeline consistent with
the others.

## How to review

Deploy cloudfoundry using this pipeline, and verify that it works. Trigger a second run of the terraform job (without triggering the init job), and verify that it uses the latest statefile, and doesn't attempt to make any changes.

## Who can review

Anyone but myself

### DIFF
--- a/concourse/pipelines/deploy-cloudfoundry.yml
+++ b/concourse/pipelines/deploy-cloudfoundry.yml
@@ -121,10 +121,6 @@ jobs:
               paas-cf/concourse/scripts/s3init.sh {{state_bucket}} cf.tfstate paas-cf/concourse/init_files/terraform.tfstate.tpl
           inputs:
           - name: paas-cf
-      - put: cf-tfstate
-        params:
-          file: init/cf.tfstate
-
   - name: terraform
     serial_groups: [ deploy ]
     serial: true
@@ -132,12 +128,11 @@ jobs:
       - aggregate:
         - get: paas-cf
           passed: ['init']
+          trigger: true
         - get: vpc-tfstate
         - get: concourse-tfstate
         - get: bosh-tfstate
         - get: cf-tfstate
-          passed: ['init']
-          trigger: true
       - task: terraform-variables
         config:
           image: docker:///ruby
--- a/concourse/pipelines/deploy-cloudfoundry.yml
+++ b/concourse/pipelines/deploy-cloudfoundry.yml
@@ -1,4 +1,12 @@
 resources:
+
+  - name: pipeline-trigger
+    type: semver-iam
+    source:
+      bucket: {{state_bucket}}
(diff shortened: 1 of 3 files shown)


mtekel:
This will only trigger when paas-cf has changed (that is on commits). This will probably trigger on the 1st run, but likely not afterwards. That put on statefile made always a new version and hence triggered next job. You probably want to add some trigger to replace that.

dcarley:
> You probably want to add some trigger to replace that.

Like the `pipeline-trigger` semver used in `create-microbosh`?

Incidentally, for pipelines that are using that semver pattern, is it expected that if you run the destroy pipeline the semver gets reset then you have to run the create pipeline multiple times until it has bumped the semver to a value greater than it was before? (or delete the pipeline and recreate)

alext:
> Like the pipeline-trigger semver used in create-microbosh?

That was my thought.  I'm going to add that in to trigger this step.

> Incidentally, for pipelines that are using that semver pattern, is it expected that if you run the destroy pipeline the semver gets reset then you have to run the create pipeline multiple times until it has bumped the semver to a value greater than it was before? (or delete the pipeline and recreate)

I think this problem happens when the state bucket gets deleted (effectively resetting things), but concourse still has cached versions of the resource in some places.

alext:
I've updated this with a pipeline trigger.  I've also renamed the trigger in the destroy-cloudfoundry pipeline to be consistent with the others.

mtekel:
You should only get to a point where you delete the init bucket when you destroy everything (including deployer). You can then destroy and re-create your vagrant box and all should be fine. If you keep re-using the vagrant box and same environment name then you'll have the problem of concourse knowing that for that semver version it has already triggered and won't trigger again. Yet due to new bucket you'll start from 0.0.1. Interrestingly, if you delete and create bucket with the same name quickly in the same account, you are less likely to have issue with it, than when you try to create it in the different account after deletion...

dcarley:
Tested several times while doing something else. This does solve the issue I've seen previously whereby manually triggering the Terraform job (not the first job in the pipeline) caused Terraform to recreate all resources, some of which failed because of duplicates but some succeeded and required manual intervention to tidy up.

Thanks :+1:

PR #104, opened 2016-02-11
[#113432925] Ensure correct domain is used in Generated SSL Certs.

jimconner:
# What

Replace hardcoded domain name used for router certificate generation with dynamic names based upon cf_root_domain and cf_apps_domain terraform variables.
In order to simplify the concourse pipelines, certificates are now all stored within a single tarball which is extracted by subsequent pipeline stages.

### How to test

* Deploy an environment as normal by following the instructions in the README.md 
* Confirm that smoke tests complete successfully during deployment.
* Try pushing an app to be extra-sure that things are OK.
* Examine the certificate details for https://api.<env>.dev.paas.alphagov.co.uk and confirm that both cf_root_domain and cf_apps_domain are listed as subject-alternative-name fields within the certificate.

### Who can review

Anyone on the team except @alext or myself

### DIFF
--- a/concourse/pipelines/deploy-cloudfoundry.yml
+++ b/concourse/pipelines/deploy-cloudfoundry.yml
@@ -233,37 +233,11 @@ jobs:
             - |
               . terraform-variables/cf.tfvars.sh
 
-              CERTS_TO_GENERATE="
-              bosh-CA:bbs_server,bbs.service.cf.internal
-              bosh-CA:bbs_client,
-              bosh-CA:router_ssl,*.${TF_VAR_cf_root_domain},*.${TF_VAR_cf_apps_domain}
-              bosh-CA:uaa_jwt_signing,
-              bosh-CA:consul_server,server.dc1.cf.internal,server.dc2.cf.internal
-              bosh-CA:consul_agent,
-              "
-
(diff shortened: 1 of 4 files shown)


concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -232,49 +198,80 @@ jobs:
...
             args:
             - -e
             - -c
             - |
+              . terraform-variables/cf.tfvars.sh
+
+              CERTS_TO_GENERATE="
+              bosh-CA:bbs_server,bbs.service.cf.internal
+              bosh-CA:bbs_client,
+              bosh-CA:router_ssl,*.${TF_VAR_cf_root_domain},*.${TF_VAR_cf_apps_domain}
```
dcarley: This made me think that we should add `set -u` to this task in case the variable/output names change and we create a version containing empty strings. But that also made me think that this inline script is far too long. What do you think about moving this out to a separate script, that we'd also run shellcheck on and do away with some of the weird interpolation workarounds?

alext: Good call. I'll pull this out.

PR #107, opened 2016-02-12
[#112411703] Fix Rate Limit exceeded error on AWS when running BOSH

actionjack:
[Follow up PR for 107477038 - Rate Limit exceeded on AWS when running BOSH](https://www.pivotaltracker.com/story/show/112411703)

## What

**Acceptance Criteria**
We have upgrade to latest bosh release, and test that we are not hitting rate limit.

* I want to:
  * Use a newer version of the [bosh-aws-cpi](https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/releases/tag/v44) release that exposes the `AWS` `max_retries` option
  * Use a newer version of the [bosh release](https://bosh.io/releases/github.com/cloudfoundry/bosh?version=253) release that has  v2.1.1 (tag v44) of bosh-aws-release gem that exposes the `AWS` `max_retries` option
  * Set max retries so that the cloud provider and bosh-init will both try 8 times before failing
  * Update reference file for rspec tests so they won't fail.

## How this PR should be tested

* **Nota bene** [The PR against Docker container for bosh-init](https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/21) needs to be merged in before this is PR is tested.
* Update your microbosh using the `create-microbosh` job on your deployer concourse installation,
* Create and delete you cloud foundry deployment several times to test that you do not receive the `Error 100: Unknown CPI error 'Unknown' with message 'Request limit exceeded.'` error

## Who should review this PR

* Any leprechaun, pixie or sprite if we can not find any legendary creature from Irish folklore then any member of the core tean will do with the exception of @combor or @actionjack who worked on this PR.

### DIFF
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -22,8 +22,8 @@ name: bosh
 
 releases:
 - name: bosh
-  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=236
-  sha1: 88dd60313dbd7dd832faa44c90493ffa6cd85448
+  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=253
+  sha1: 940956a23b642af3bb24b3cac37c4da746d6f9a9
 
 
 disk_pools:
--- a/manifests/bosh-manifest/deployments/aws/010-bosh-aws-cpi.yml
+++ b/manifests/bosh-manifest/deployments/aws/010-bosh-aws-cpi.yml
@@ -3,8 +3,9 @@ releases:
 # bosh-init container must also be updated so that the cached CPI compile
 # will be used.
 - name: bosh-aws-cpi
-  url: https://bosh.io/d/github.com/cloudfoundry-incubator/bosh-aws-cpi-release?v=36
-  sha1: db2a6c6cdd5ff9f77bf083e10118fa72e1f5e181
+  url: https://bosh.io/d/github.com/cloudfoundry-incubator/bosh-aws-cpi-release?v=44
+  sha1: a1fe03071e8b9bf1fa97a4022151081bf144c8bc
+
 
 jobs:
 - name: bosh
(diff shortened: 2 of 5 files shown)


dcarley:
You were having problems with this upgrade last week and it's not clear from the commit messages what changed the situation. Could you add some more detail to help people in the future?

Does the first commit ("Increment bosh-aws-cpi release to version 44") only affect `bosh-init`? If so, is the second commit ("Increment Bosh release to 253") necessary to make the change also available to MicroBOSH?

Are gem v2.1.1 and tag v44 actually the same? Does that version of the gem, which the BOSH release uses, contain the `max_retries` change? It seems cloudfoundry-incubator/bosh-aws-cpi-release@870d0a8cc3ff412e00dd91c9b91fbd07aad1b0d1cut bumped the gem at v2.1.1 at tag v40.

Could you add the reason that we're increasing `max_retries` to the commit message that bumps it - that we're frequently hitting the AWS API rate limit, which causes BOSH to abort all operations and leave the deployment in an inconsistent state.

combor:
> Does the first commit ("Increment bosh-aws-cpi release to version 44") only affect bosh-init?

According to http://bosh.io/docs/init-aws it doesn't affect bosh-init only but it is deployed to microbosh as well. There is a cpi job defined for director: https://github.com/alphagov/paas-cf/blob/affbb9e20063e1c998aa5f2fbe903ce49b67774a/manifests/bosh-manifest/deployments/010-bosh-template.yml#L88

> If so, is the second commit ("Increment Bosh release to 253") necessary to make the change also available to MicroBOSH?

This release comes with gem 2.1.1 that may be necessary for max_retires option. Frankly speaking it is uncertain how it works internally and how microbosh determines which gem to use. Release 44 of bosh_aws_cpi that contains max_retires fix comes with v 2.1.1 as well so it's better to stay consistent. 

> Are gem v2.1.1 and tag v44 actually the same? 

No, gem 2.1.1 was introduced earlier, before release 44.  Release 44 contains specification fixes for max_retries option that can now be passed to underlying library.

actionjack:
@dcarley I've redone the commits with your requested changes.

dcarley:
After several runs of `destroy-cloudfoundry` and `deploy-cloudfoundry` I haven't seen the rate limit error at all :+1: 

I'm unsure about the exact retry numbers and concerned that cloudfoundry-incubator/bosh-aws-cpi-release#24 hasn't been merged yet, but this is a definite improvement and I think it's better that err on the side of retrying more times and never failing.

NB: You might notice that MicroBOSH takes longer to deploy if you're using an existing Deployer Concourse because it won't pick up the updated `bosh-init` container from alphagov/paas-docker-cloudfoundry-tools#21. You'll need to restart or redeploy your Deployer Concourse to benefit from the new cached CPI.

captainblackjack:
\o/

PR #109, opened 2016-02-15
[#113262177] Lint checks for duplicate YAML keys

dcarley:
## What

Enable lint checks to find duplicate keys in static and dynamically generated YAML documents. Please see the individual commit messages for more information.

## How to review

1. Check whether Travis is passing
1. Install the dependencies and run locally (see README and would welcome feedback on how easy this is)
1. Try adding a duplicate key, run the tests again, and see if they fail

## Who can review

Not @dcarley


.travis.yml
```diff
@@ -12,6 +12,8 @@ addons:
       - debian-sid
     packages:
       - shellcheck
+      - pwgen
+      - whois
```
alext: It's probably worth a comment here explaining why we need this package.

dcarley: In addition to the commit message? Do you think somebody would find it, if necessary, with `git log -G whois`?

manifests/cf-manifest/spec/secret_generation_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'tempfile'
+
+RSpec.describe "secret generation" do
+  describe "generate-cf-secrets" do
+    let(:script) {
+      File.expand_path("../../scripts/generate-cf-secrets.sh", __FILE__)
+    }
+
+    specify "it should produce lint-free YAML" do
+      pending "No mkpasswd(1) on Mac OS X" if Gem::Platform.local.os == 'darwin'
```
alext: `skip` would be better than `pending` here. `pending` causes rspec to still run the test, and actually flag it as a failure if it passes (ie, by using `pending` you're effectively adding an assertion that the test fails). `skip` on the other hand, just skips this test. See http://www.relishapp.com/rspec/rspec-core/v/3-4/docs/pending-and-skipped-examples for more details.

.travis.yml
alext: I'm on the fence actually... I was wondering whether this is surprising enough to be worth a comment here as well.

manifests/cf-manifest/scripts/generate-cf-secrets.sh
```diff
@@ -37,7 +35,6 @@ cat <<EOF
     uaa_clients_cloud_controller_username_lookup_secret: $(simplepass)
     uaa_clients_gorouter_secret: $(simplepass)
     uaa_clients_firehose_password: $(simplepass)
-    uaa_admin_password: $(simplepass)
```
alext: Given the number of duplicate entries there were in here, I wonder whether alphabetising this would be a good idea.

README.md
```diff
@@ -171,6 +171,12 @@ Run the `destroy-cloudfoundry` pipeline.
 
 # Additional notes
 
+## Running tests locally
+
+You will need to install some dependencies to run the unit tests on your own
+machine. The most up-to-date reference for these is the Travis CI
+configuration in [`.travis.yml`](.travis.yml).
+
```
alext: I wonder whether we need a little more detail around installing things from pip. It took me a while to figure out where `pip install --user` had put `yamllint` because it wasn't on my path by default.

alext:
A few minor comments, otherwise this looks good to me.

We could usefully enable some of the other lint checks as I've noticed some inconsistencies with the formatting of our YAML (specifically around indentation of array elements). That that's something for a separate PR/story.

manifests/cf-manifest/scripts/generate-cf-secrets.sh
dcarley: I'm reluctant to do so unless/until we have a means of enforcing the order for subsequent changes.

.travis.yml
dcarley: Added an inline comment - the type I dislike the least, because it will be deleted with the line it affects.

manifests/cf-manifest/spec/secret_generation_spec.rb
dcarley: Done!

README.md
dcarley: Fair point. I've added a comment to the commit message about why `--user` is required. Not sure how we should address the documentation. Any suggestions?

manifests/cf-manifest/scripts/generate-cf-secrets.sh
alext: Fair enough...

README.md
dcarley: Alex and I agreed to leave this as-is for now. We can add more detail later, at the cost of duplication, if people are having problems.

dcarley:
Rebased to fix merge conflicts.

PR #110, opened 2016-02-16
[#112908839] Workflow for accessing Bosh using CLI

alext:
## What

We want to be able to easily use bosh cli commands (including `bosh ssh`) against our deployed µBOSH.

## Details

This adds a script that pushes a one-off task to concourse that sets up bosh credentials, target and deployment. It then intercepts (hijacks) that container to provide an interactive session with the bosh CLI available.

We've had to increase the `idle_timeout` for the concourse ELB to 600 seconds to prevent sessions being dropped after 60 seconds of inactivity.

This also updates all the pipelines that use the bosh CLI to use the new docker container instead of using the bosh deployment resource container in a way that wasn't intended.

## How to review

Re-run the create-deployer pipeline to apply the ELB timeout change. Deploy cloudfoundry.

Run the bosh cli script:
```
./concourse/scripts/bosh-cli.sh $DEPLOY_ENV
```
The bosh cli should be ready to use. It should already be logged in etc. so you can run `bosh vms`, and also `bosh ssh` to vms (See note in README about the blank sudo password)

## Who can review

### DIFF
--- a/concourse/pipelines/bosh-cli/bosh-cli.yml
+++ b/concourse/pipelines/bosh-cli/bosh-cli.yml
@@ -4,6 +4,7 @@ platform: linux
 image: docker:///governmentpaas/bosh-cli
 
 inputs:
+  - name: paas-cf
   - name: cf-manifest
   - name: bosh-secrets
 run:
@@ -12,9 +13,8 @@ run:
   - -c
   - -e
   - |
-    bosh_password=$(awk '/bosh_admin_password/ { print $2 }' bosh-secrets/bosh-secrets.yml)
-    bosh -t 10.0.0.6 login admin "${bosh_password}"
-    bosh target 10.0.0.6
+    ./paas-cf/concourse/scripts/bosh_login.sh bosh-secrets/bosh-secrets.yml
+
     uuid=$(bosh status --uuid)
     sed -e "s/^director_uuid:.*$/director_uuid: ${uuid}/" cf-manifest/cf-manifest.yml > cf-manifest-with-uuid.yml
     bosh deployment ./cf-manifest-with-uuid.yml
--- a/concourse/pipelines/bosh-cli/bosh-cli.yml
+++ b/concourse/pipelines/bosh-cli/bosh-cli.yml
@@ -1,0 +1,20 @@
+---
+platform: linux
+
+image: docker:///governmentpaas/bosh-cli
+
+inputs:
+  - name: cf-manifest
+  - name: bosh-secrets
+run:
+  path: sh
+  args:
+  - -c
+  - -e
+  - |
+    bosh_password=$(awk '/bosh_admin_password/ { print $2 }' bosh-secrets/bosh-secrets.yml)
+    bosh -t 10.0.0.6 login admin "${bosh_password}"
+    bosh target 10.0.0.6
+    uuid=$(bosh status --uuid)
+    sed -e "s/^director_uuid:.*$/director_uuid: ${uuid}/" cf-manifest/cf-manifest.yml > cf-manifest-with-uuid.yml
+    bosh deployment ./cf-manifest-with-uuid.yml
--- a/concourse/scripts/bosh_login.sh
+++ b/concourse/scripts/bosh_login.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+set -eu
+
+bosh_secrets_file=$1
+bosh_ip=${BOSH_IP-"10.0.0.6"}
+
+bosh_password=$(awk '/bosh_admin_password/ { print $2 }' "${bosh_secrets_file}")
+bosh -t "${bosh_ip}" login admin "${bosh_password}"
+bosh target "${bosh_ip}"
--- a/README.md
+++ b/README.md
@@ -226,6 +226,21 @@ handled by `vagrant/deploy.sh`:
 ssh ubuntu@<bootstrap_concourse_ip> -L 8080:127.0.0.1:8080 -fN
 ```
 
+## Using the bosh cli and `bosh ssh`
+
+There's a script that starts an interactive session on the deployer concourse
+to allow running bosh CLI commands targeting MicroBOSH:
+
+```
+./concourse/scripts/bosh-cli.sh $DEPLOY_ENV
+```
+
+This connects you to a one-off task in concourse that's already logged into
+bosh and has the deployment set using the CF manifest.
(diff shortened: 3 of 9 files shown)


concourse/pipelines/bosh-cli/bosh-cli.yml
```diff
@@ -0,0 +1,20 @@
...
+  path: sh
+  args:
+  - -c
+  - -e
+  - |
+    bosh_password=$(awk '/bosh_admin_password/ { print $2 }' bosh-secrets/bosh-secrets.yml)
+    bosh -t 10.0.0.6 login admin "${bosh_password}"
+    bosh target 10.0.0.6
+    uuid=$(bosh status --uuid)
+    sed -e "s/^director_uuid:.*$/director_uuid: ${uuid}/" cf-manifest/cf-manifest.yml > cf-manifest-with-uuid.yml
```
dcarley: Are these two lines necessary? Doesn't `deploy-cloudfoundry/deploy`, which this pipeline takes inputs from, do this and store the result already?

- https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/deploy-cloudfoundry.yml#L498

concourse/pipelines/bosh-cli/bosh-cli.yml
```diff
@@ -0,0 +1,20 @@
...
+  - name: cf-manifest
+  - name: bosh-secrets
+run:
+  path: sh
+  args:
+  - -c
+  - -e
+  - |
+    bosh_password=$(awk '/bosh_admin_password/ { print $2 }' bosh-secrets/bosh-secrets.yml)
+    bosh -t 10.0.0.6 login admin "${bosh_password}"
```
dcarley: We have this pattern - get password, set target, login - in quite a few places. Do you think it's worth refactoring into a separate script?
```
➜  paas-cf git:(master) ✗ ack 'bosh .* target'
concourse/pipelines/deploy-cloudfoundry.yml
496:              bosh -u admin -p "${bosh_password}" target https://10.0.0.6:25555
526:            bosh -u admin -p "${bosh_password}" target https://10.0.0.6:25555
552:            bosh -u admin -p "${bosh_password}" target https://10.0.0.6:25555

concourse/scripts/bosh_ensure_uploaded.sh
6:bosh -u admin -p "${BOSH_PASSWORD}" target https://10.0.0.6:25555 >/dev/null
```

concourse/pipelines/bosh-cli/bosh-cli.yml
alext: It does do this, but the resulting resource isn't written anywhere outside of the specific task. It doesn't appear to be possible to use a task from a pipeline as an input when using `fly execute`. When I tried it it errors with "error: missing required input `cf-deploy`".

dcarley: Alex and I discussed this. It looks like [it is `put` after the task](https://github.com/alphagov/paas-cf/blob/adac9f91590f86d950745c7c28d8838abef93c14/concourse/pipelines/deploy-cloudfoundry.yml#L502-L504), but `--inputs-from` takes the state from *before*. Nevermind.

concourse/pipelines/bosh-cli/bosh-cli.yml
dcarley: Alex is going to see what this looks like and decide whether it's worth doing now.

concourse/pipelines/bosh-cli/bosh-cli.yml
```diff
@@ -0,0 +1,20 @@
+---
+platform: linux
+
+image: docker:///governmentpaas/bosh-cli
```
dcarley: The `bosh` commands generate a lot of warnings about OpenSSL. It looks like these come from [`httpclient` because of the OpenSSL` version](https://github.com/nahi/httpclient/blob/85b3fcde9a0378e9a219765b0b79b465162b173f/lib/httpclient/ssl_config.rb#L441-L452) in the [`governmentpaas/bosh-cli` container](https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/cc3240fad4ed606bb1b5173e32f05db25e32b3c4/bosh-cli/Dockerfile):
```
bosh deployment ./cf-manifest-with-uuid.yml

RSA 1024 bit CA certificates are loaded due to old openssl compatibility
Logged in as `admin'
RSA 1024 bit CA certificates are loaded due to old openssl compatibility
Target set to `my-bosh'
RSA 1024 bit CA certificates are loaded due to old openssl compatibility
RSA 1024 bit CA certificates are loaded due to old openssl compatibility
Deployment set to `/tmp/build/beda48a9-3deb-40e6-54a1-ed5cd322f94e/cf-manifest-with-uuid.yml'
succeeded
```

Is that reason enough to move to the smaller Alpine based image which has a later version of OpenSSL?
```
➜  paas-cf git:(bosh_cli_task) ✗ docker run --rm -ti ruby:2.2-alpine openssl version
OpenSSL 1.0.2f  28 Jan 2016
```

dcarley:
It works and I like the approach - means we don't need to pull any state back to the local machine, worry about local dependencies, or scripting around SSH tunnels. :+1:

concourse/pipelines/bosh-cli/bosh-cli.yml
alext: We've created an updated container based on alpine in https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/23, and also updated the bosh-cli script to specify the shell (now that bash will no longer be available).

concourse/pipelines/bosh-cli/bosh-cli.yml
alext: This has now been extracted into a separate script.

dcarley:
Thanks, I'll retest with the new container!

concourse/scripts/bosh_login.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
```
dcarley: Should we set `-u` as well to prevent this from being run without an `bosh_secrets_file` argument?

alext: Good call, I'll update.

alext: updated.

dcarley:
:+1: :+1:

PR #113, opened 2016-02-17
[#113265403] Optionally trigger builds from paas-cf master

Jonty:
This will cause builds to automatically trigger from paas-cf if the environment variable `ENABLE_AUTO_DEPLOY=1` is set. We can be sure that the tests have already passed as we have master protection turned on, so the Travis build must have already passed.

# How to test
Deploy your development environment on this branch, then `fly get-pipeline deploy-cloudfoundry`. The jobs section should look like this:
````
jobs:
  - name: init
    serial_groups: [ deploy ]
    serial: true
    plan:
      - get: paas-cf
        trigger: false
      - task: init
        config:
...
````
Specifically you're looking for the `trigger: false` on `paas-cf`.

# Next steps
After this PR is merged the `ci` pipelines need to be updated using `./concourse/scripts/pipelines-cloudfoundry.sh` with `ENABLE_AUTO_DEPLOY=1` set.

# Who can review this
Not @jonty.

### DIFF
--- a/concourse/scripts/pipelines-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-cloudfoundry.sh
@@ -23,6 +23,12 @@ debug: ${DEBUG:-}
 EOF
 }
 
+generate_manifest_file() {
+   # This exists because concourse does not support boolean value interpolation by design
+   enable_auto_deploy=$([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
+   sed -e "s/{{auto_deploy}}/${enable_auto_deploy}/" \
+       < "${SCRIPT_DIR}/../pipelines/${ACTION}-cloudfoundry.yml"
+}
 
 for ACTION in deploy destroy; do
   trigger_file="${ACTION}-cloudfoundry.trigger"
@@ -30,7 +36,7 @@ for ACTION in deploy destroy; do
 
   bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
     "${env}" "${ACTION}-cloudfoundry" \
-    "${SCRIPT_DIR}/../pipelines/${ACTION}-cloudfoundry.yml" \
+    <(generate_manifest_file) \
     <(generate_vars_file)
 done
 
--- a/concourse/pipelines/deploy-cloudfoundry.yml
+++ b/concourse/pipelines/deploy-cloudfoundry.yml
@@ -82,6 +82,7 @@ jobs:
     serial: true
     plan:
       - get: paas-cf
+        trigger: {{auto_deploy}}
       - task: init
         config:
           image: docker:///governmentpaas/curl-ssl

### CODE REVIEW

File: concourse/scripts/pipelines-cloudfoundry.sh [hunk 1]
```diff
@@ -11,9 +11,11 @@ config_autodelete="${SCRIPT_DIR}/../pipelines/autodelete-cloudfoundry.yml"
 
 generate_vars_file() {
    set -u # Treat unset variables as an error when substituting
+   local aws_account=${AWS_ACCOUNT:-dev}
    cat <<EOF
 ---
-aws_account: ${AWS_ACCOUNT:-dev}
+aws_account: ${aws_account}
+auto_deploy: $([ "${aws_account}" != "dev" ] && echo "true" || echo "false")
```
dcarley: I'm unsure about us keying feature against environment names. It reminds me of this from 12factor.net:

> Another aspect of config management is grouping. Sometimes apps batch config into named groups (often called “environments”) named after specific deploys, such as the development, test, and production environments in Rails. This method does not scale cleanly: as more deploys of the app are created, new environment names are necessary, such as staging or qa. As the project grows further, developers may add their own special environments like joes-staging, resulting in a combinatorial explosion of config which makes managing deploys of the app very brittle.

Maybe we should default this to the development case and have a named environment variable to enable it? Like we do for `DISABLE_AUTODELETE`.

Jonty: I was equally uncomfortable with the name keying on the AWS account, but was trying to avoid adding further configuration environment variables with defaults - they always seem to get accidentally omitted at some point leading to environments that are not configured as expected. In the case of this, we'd have environments that were silently not updating.

I'm happy to pull this back into being an env var however I'd really like to ensure that we're storing the settings used to generate our production environments somewhere - right now the only record of the environment variables used to generate an environment is the resulting changes to configuration inside the environment itself.

Perhaps this is a job for the proposed Makefile?

dcarley: Agree with all of that :+1: 

We agreed that we'd change this to a new environment variable name and raise a new story to create `Makefile` targets to group these together.

@Jonty can you check that [#114080971](https://www.pivotaltracker.com/story/show/114080971) makes sense?

Jonty:
Paging @dcarley to the pull request update aisle.

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -14,6 +14,7 @@ generate_vars_file() {
    cat <<EOF
 ---
 aws_account: ${AWS_ACCOUNT:-dev}
+auto_deploy: $([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
```
dcarley: I got an error when trying to deploy the pipeline using either option:
```
➜  paas-cf git:(enable_build_from_master_113265403) ✗ BRANCH=$(git rev-parse --abbrev-ref HEAD) ./concourse/scripts/pipelines-cloudfoundry.sh
Concourse API target dcarley
Deployment dcarley
Pipeline deploy-cloudfoundry
Config file /Users/dcarley/projects/paas/paas-cf/concourse/scripts/../pipelines/deploy-cloudfoundry.yml
failed to parse configuration file: yaml: unmarshal errors:
  line 85: cannot unmarshal !!str `false` into bool
➜  paas-cf git:(enable_build_from_master_113265403) ✗ ENABLE_AUTO_DEPLOY=1 BRANCH=$(git rev-parse --abbrev-ref HEAD) ./concourse/scripts/pipelines-cloudfoundry.sh
Concourse API target dcarley
Deployment dcarley
Pipeline deploy-cloudfoundry
Config file /Users/dcarley/projects/paas/paas-cf/concourse/scripts/../pipelines/deploy-cloudfoundry.yml
failed to parse configuration file: yaml: unmarshal errors:
  line 85: cannot unmarshal !!str `true` into bool
```

My local shell is zsh, if that turns out to matter.

Jonty:
I've pushed a version that works around the problem with concourse quoting all interpolations, making boolean variables impossible. It's horrible.

This is the problem: https://github.com/alphagov/paas-cf/pull/113

Ideally concourse would [remove the quoting](https://github.com/concourse/fly/blob/master/template/template.go#L25), however we could also work around this by using spruce to merge the variables instead of passing them to concourse if we end up doing this more than once.

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -29,6 +29,11 @@ debug: ${DEBUG:-}
 EOF
 }
 
+generate_manifest_file() {
+   # This exists because concourse does not support boolean value interpolation by design
+   local enable_auto_deploy=$([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
```
dcarley: shellcheck didn't like this:
```
In ./concourse/scripts/pipelines-cloudfoundry.sh line 34:
   local enable_auto_deploy=$([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
         ^-- SC2155: Declare and assign separately to avoid masking return values.
```

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -29,6 +29,11 @@ debug: ${DEBUG:-}
...
+   cat "${SCRIPT_DIR}/../pipelines/${ACTION}-cloudfoundry.yml" | sed -e "s/{{auto_deploy}}/${enable_auto_deploy}/"
```
dcarley: shellcheck didn't like this:
```
In ./concourse/scripts/pipelines-cloudfoundry.sh line 35:
   cat "${SCRIPT_DIR}/../pipelines/${ACTION}-cloudfoundry.yml" | sed -e "s/{{auto_deploy}}/${enable_auto_deploy}/"
       ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

Jonty: This is what you get for disabling your vim shellcheck integration.

Jonty:
@dcarley Commit message updated and rebase rebased.

PR #114, opened 2016-02-18
Fix autodelete pipeline.

alext:
The delete task was missing the paas-cf input, so couldn't find the
bosh-login script. This was missed in d282c21.


mtekel:
works

PR #115, opened 2016-02-18
[#110652158] Specify release/stemcell version etc. in the manifest

alext:
## What

Prior to upgrading CF and diego we want to consolidate how we're specifying the versions of components.

## Details

Newer versions of the bosh CLI allow us to specify the version, url and sha1 for releases and stemcells in the manifest, and will automatically upload them to bosh if they're not already present. This enables us to simplify things considerably. It also means that all the details relating to these are specified in a single place.

The downside of this change is that it no longer parallelises the
downloads for a new deployment, but I consider that cost worth it in
exchange for the simplification.

## How to review

Deploy cloudfoundry using a newly created bosh instance. Verify that the deploy completes successfully. Do a second deploy, and verify that it doesn't download the releases and stemcells a second time.

## Who can review


mtekel:
Indeed bosh now supports specifying stemcell and release urls directly in the manifest and it will fetch them for you (and not download when it already has them).

dcarley:
:+1: :clap:

PR #116, opened 2016-02-18
[#113458409] create dedicated router subnet

actionjack:
[Move Router to it's own Subnet](https://www.pivotaltracker.com/story/show/113458409)

## What

Move the router to a subnet separate from the core CF and cell subnets, and update the ACLs to:
- be more restrictive about traffic going into the CF subnet (remove the allow 1024-61000 rule)
- allow the router to talk to components that it needs to in the CF subnet
- allow the router to talk to high order ports in the cell subnet
- allow return communication back into the router subnet as required

The PR has been written in the following context:
I would like to:

* Create a dedicated router subnet
* Setup routing between the net router subnet to the other Cloud Foundry infrastructure Subnets
* Added ACL for router and additional tagging to the other subnets so they can be easily identified in the AWS Console
* Allow  traffic from the `router` network to the  `Cell` network
* Allow  traffic from the `router` network to the  `CF` network
* Allow  established traffic from the `CF` network to come back from `router` network
* Move routers to new router subnet

## How to review this PR

* Deploy new environment and this new environment should pass the smoke tests and a [test app](https://github.com/cloudfoundry-samples/spring-music) and try and communicate with it

## Who should review this PR

* Anyone who thinks Batfink was truly the greatest cartoon ever created Except @mtekel and @actionjack even though they are great fans of [batfink](https://en.wikipedia.org/wiki/Batfink) and his wings which were like a shield of steel, they did also work on this story.


terraform/cloudfoundry/network_acls_router_in.tf
```diff
@@ -0,0 +1,63 @@
...
+    rule_number = 32
+    rule_action = "allow"
+    cidr_block = "${var.vpc_cidr}"
+    icmp_type = -1
+    icmp_code = -1
+    egress = false
+}
+
+resource "aws_network_acl_rule" "33_ssh" {
+    network_acl_id = "${aws_network_acl.cf.id}"
```
dcarley: It looks like this resource should either be in `network_acls_cf_in.tf` or the value should be `"${aws_network_acl.router.id}"`.

terraform/cloudfoundry/network_acls_cf_in.tf
```diff
@@ -172,6 +139,15 @@ resource "aws_network_acl_rule" "110_bosh" {
     egress = false
 }
 
+resource "aws_network_acl_rule" "115_router_in" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = -1
+    rule_number = 115
+    rule_action = "allow"
+    cidr_block = "${var.router_cidr_all}"
```
dcarley: Can we be more specific about the ports allowed here? Do we know what the router needs to talk to?

terraform/cloudfoundry/network_acls_router_in.tf
```diff
@@ -0,0 +1,63 @@
...
+    egress = false
+}
+
+resource "aws_network_acl_rule" "21_elb_in" {
+    network_acl_id = "${aws_network_acl.router.id}"
+    protocol = "tcp"
+    rule_number = 21
+    rule_action = "allow"
+    from_port = 1024
+    to_port = 65535
```
dcarley: Can you explain why this is so broad? Does the ELB need to initiate to anything except TCP/80 and TCP/443?

terraform/cloudfoundry/network_acls_router_in.tf
```diff
@@ -0,0 +1,63 @@
+resource "aws_network_acl_rule" "20_cf_in" {
+    network_acl_id = "${aws_network_acl.router.id}"
+    protocol = "tcp"
+    rule_number = 20
+    rule_action = "allow"
+    from_port = 1024
+    to_port = 65535
```
dcarley: Can you explain why this is so broad?

actionjack: We need to create a broad range for the  ELB since they  will try to communicate to the routers using an arbitary high port to communicate with the router

terraform/cloudfoundry/network_acls_router_in.tf
actionjack: We need to create a broad range for the  ELB since they  will try to communicate to the routers using an arbitary high port to communicate with the router

dcarley: I'm not sure about that. It'll be easier to talk about it in-person though, so I'll catch you this afternoon.

terraform/cloudfoundry/network_acls_cf_in.tf
```diff
@@ -172,6 +139,28 @@ resource "aws_network_acl_rule" "110_bosh" {
...
+    protocol = -1
+    rule_number = 115
+    rule_action = "allow"
+    from_port = 1024
+    to_port = 65535
+    cidr_block = "${var.router_cidr_all}"
+    egress = false
+
+
+    resource "aws_network_acl_rule" "116_router_consul_in" {
```
dcarley: Can you fix the indentation and missing close brace?

keymon:
I think it makes sense if we also move the ELBs of the routers to the router subnet, by changing `${split(",", var.infra_subnet_ids)}` with `aws_subnet.router.*.id` [here](https://github.com/alphagov/paas-cf/blob/113458409-create-route-subnet/terraform/cloudfoundry/routers.tf#L3--1) and [here](https://github.com/alphagov/paas-cf/blob/113458409-create-route-subnet/terraform/cloudfoundry/routers.tf#L33)

dcarley:
> I think it makes sense if we also move the ELBs of the routers to the router subnet, by changing ${split(",", var.infra_subnet_ids)} with aws_subnet.router.*.id here and here

:+1: 

Will that change the rules at all, e.g. allow us to tighten them?

manifests/cf-manifest/deployments/aws/000-cf-infrastructure.yml
```diff
@@ -241,14 +241,14 @@ jobs:
   - name: router_z1
     instances: 1
     networks:
-      - name: cf1
-        static_ips: (( static_ips(5, 6, 15, 16, 17, 18, 19, 20) ))
+      - name: router1
+        static_ips: (( static_ips(1) ))
```
keymon: I would suggest that we allocate more static IPs for the routers, rather than only one.

The reason is that in the future, during a normal operation, we might want to increase the number of routers to cope with the load. That can be really easy if there are IPs already allocated (just change the # of instances), but if the IPs are not allocated, we need to re-render the manifest and apply it again.

I think it is fairly safe to set this to something like `static_ips(1, 2, 3, 4, 5, 6, 7, 8)`, as there are plenty of IPs and we have only the routers. Also, this way we keep the same setup than before.

I am open to consider that this should be out of this PR, but the change is small.

keymon:
> >   I think it makes sense if we also move the ELBs of the routers to the router subnet

> Will that change the rules at all, e.g. allow us to tighten them?

I don't think so,  basically because in the last commit we get rid of all the rules and simply allow any traffic to/from routers subnet.

But it will make it easier to handle security, as the "router ELBs" will be in the same network as the gorouters and this way we don't need to add any config allowing the traffic from ELBs and gorouters. 

We can assume the ELBs are trustworthy and there is no need to configure ACLs.

keymon:
> I think it makes sense if we also move the ELBs of the routers to the router subnet

As discussed in person, this might be more tricky than what initially seems. You need to attach only public subnets to your ELB, no private. 

If we want to change this we should change the router network to be public and routed to internet, and then either:
 * Add a ephemeral public ip to the routers, instead of using nat.
 * Or remove any internet access for the gorouters, not given them a public IP.

Other option is just leave it as it is and revisit in the future the configuration of the ELBs. It makes sense to create a dedicated subnet for these ELBs, rather than using the "infra" subnet where bosh and concourse are.

dcarley:
> Other option is just leave it as it is and revisit in the future the configuration of the ELBs. It makes sense to create a dedicated subnet for these ELBs, rather than using the "infra" subnet where bosh and concourse are.

Let's leave it for now then. Can someone note in the story summary that we'd like a new story for the above?

mtekel:
done

keymon:
LGTM

PR #120, opened 2016-02-22
[#113996323] Add Logsearch

mtekel:
## What
[Log aggregation platform](https://www.pivotaltracker.com/n/projects/1275640/stories/113996323)

Add default logsearch deployment to our CF. We have decided to make this a part of main CF deployment for the sake of simplicity. Otherwise we'd have to update concourse pipelines and template generation scripts to deploy multiple releases, which was out of scope of this story.

## How to review

Deploy or update your existing deployment. Create ssh tunnel via deployer concourse to kibana:
`ssh -L 5601:10.0.16.49:5601 -i session/id_rsa vcap@$DEPLOYER_CONCOURSE`. You should be able to browse logs and see them appearing in real time.

## Who can review

The [Scarlet Pimpernel](https://en.wikipedia.org/wiki/The_Scarlet_Pimpernel) because _They seek him here, they seek him there_ if he can not be found any member of the core team can merge this PR with the exception of @actionjack and @mtekel

### DIFF
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -275,7 +275,6 @@ jobs:
   release: logsearch
   templates:
   - {name: parser, release: logsearch}
-  - {name: elasticsearch, release: logsearch}
   resource_pool: parser
   instances: 1
   networks:
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -1,0 +1,323 @@
+releases:
+- name: logsearch
+  url: https://bosh.io/d/github.com/logsearch/logsearch-boshrelease?v=200.0.0
+  version: 200.0.0
+  sha1: 527c35db31cd66810accf128ceffee4f5fde80c3
+
+disk_pools:
+- name: elasticsearch_master
+  disk_size: 102400
+  cloud_properties:
+    type: gp2
+
+- name: elasticsearch_data
+  disk_size: 102400
+  cloud_properties:
+    type: gp2
+
+- name: queue
+  disk_size: 102400
+  cloud_properties:
+    type: gp2
+
+- name: cluster_monitor
+  disk_size: 102400
+  cloud_properties:
+    type: gp2
+
+resource_pools:
+- name: elasticsearch_master
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.micro
+
+- name: elasticsearch_data
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.micro
+
+- name: ingestor
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.micro
+
+- name: queue
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.small
+
+- name: parser
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.micro
+
+- name: kibana
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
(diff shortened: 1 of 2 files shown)


combor:
Travis build fails with 
```
yamllint -c yamllint.yml .
./manifests/cf-manifest/deployments/050-logsearch.yml
  119:3     error    duplication of key "update" in mapping  (key-duplicates)
  248:3     error    duplication of key "update" in mapping  (key-duplicates)
```

mtekel:
This is tricky. Update is a special property that can be repeated multiple times in the manifest and controls how BOSH updates jobs. The two conflicts are also present in the logsearch release manifest, where the yaml was taken from: https://github.com/logsearch/logsearch-boshrelease/blob/develop/templates/logsearch-jobs.yml

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -0,0 +1,326 @@
...
+      elasticsearch_port: 9200
+      purge_logs:
+        retention_period: 7
+    elasticsearch_config:
+      elasticsearch:
+        host: 127.0.0.1
+        port: 9200
+      bulk_data_files: []
+      templates:
+      - shards-and-replicas: "{ \"template\" : \"*\", \"order\" : 99, \"settings\" : { \"number_of_shards\" : 1, \"number_of_replicas\" : 0 } }"
```
dcarley: Is this a default in the upstream logsearch release? Why are shards and replicas disabled?

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -0,0 +1,326 @@
...
+      - index_template: /var/vcap/packages/logsearch-config/default-mappings.json
+    logstash_parser:
+      filters:
+      - metrics: /var/vcap/packages/logsearch-config/logstash-filters-metric.conf
+    nats_to_syslog:
+      nats:
+        subject: ">"
+        user: nats
+        password: nats-password
```
dcarley: Is this password valid? Shouldn't it come from secrets instead?

dcarley:
> Update is a special property that can be repeated multiple times in the manifest and controls how BOSH updates jobs.

Can it be repeated at the same level? Wouldn't a YAML parser discard one or more of them?

dcarley:
> the yaml was taken from: https://github.com/logsearch/logsearch-boshrelease/blob/develop/templates/logsearch-jobs.yml

Could you add this to the commit message with the link updated to a specific version/tag? That might save some of the questions in this PR about why some of the config is as it is.

manifests/cf-manifest/deployments/050-logsearch.yml
mtekel: yes, this is default. replicas likely disabled due to logs being considered "ephemeral"

manifests/cf-manifest/deployments/050-logsearch.yml
mtekel: Good catch. Yet this component is not enabled in default config. For this story we're just getting _some_ logs into ELK. I'll update this nevertheless to reference real NATS config.

dcarley:
@mtekel @actionjack I know some of the comments may be outside the scope of this PR and story. However, since we're committing a lot of default config and might not go back to look at it in the detail, do you think it's helpful for people to review these defaults and we can capture some of the changes as new stories?

manifests/cf-manifest/deployments/050-logsearch.yml
mtekel: Also, as it turns out, logstash controls shards for its data separately:
```
curl localhost:5601/elasticsearch/_cat/shards
.kibana             0 p STARTED     2  8.7kb 10.0.16.26 elasticsearch_data/0 
.kibana             0 r STARTED     2  8.7kb 10.0.16.27 elasticsearch_data/1 
logstash-2016.02.22 2 p STARTED 81420 91.9mb 10.0.16.26 elasticsearch_data/0 
logstash-2016.02.22 2 r STARTED 81420 91.9mb 10.0.16.27 elasticsearch_data/1 
logstash-2016.02.22 3 r STARTED 81326 92.4mb 10.0.16.26 elasticsearch_data/0 
logstash-2016.02.22 3 p STARTED 81326 92.4mb 10.0.16.27 elasticsearch_data/1 
logstash-2016.02.22 1 r STARTED 81148 91.7mb 10.0.16.26 elasticsearch_data/0 
logstash-2016.02.22 1 p STARTED 81148 91.8mb 10.0.16.27 elasticsearch_data/1 
logstash-2016.02.22 4 p STARTED 81720 92.3mb 10.0.16.26 elasticsearch_data/0 
logstash-2016.02.22 4 r STARTED 81720 92.3mb 10.0.16.27 elasticsearch_data/1 
logstash-2016.02.22 0 p STARTED 81643 92.4mb 10.0.16.26 elasticsearch_data/0 
logstash-2016.02.22 0 r STARTED 81643 92.5mb 10.0.16.27 elasticsearch_data/1
```

combor: There are two levels of number of shards and replicas setting. First is global and defaults to 5 shards and one replica per shard. This seems to be the case here. Second is per index setting and you can override global value by specifying number of shards and replicas during index creation.

saliceti:
Working as designed.

PR #121, opened 2016-02-22
[#1106592863] Deploy etcd and consul across 3 availability zones

alext:
## What

This adds subnets in a 3rd AZ, and configures etcd and consul to be deployed across all 3 zones.

## How to review

It's necessary to run a complete deployment because this is adding new subnets etc. Run all 3 pipelines in order - create-deployer, create-microbosh and deploy-cloudfoundry. Verify that the deployment works. Verify that the etcd and consul clusters are healthy, and are indeed spread across all 3 availability zones:

* etcd: `/var/vcap/packages/etcd/etcdctl member list` (on one of the etcd servers)
* consul: `/var/vcap/packages/consul/bin/consul members | grep server` (on one of the consul servers)

We have seen issues with the etcd cluster not being healthy after upgrading an existing deployment resulting on the loggregator_trafficcontroller jobs failing with an error contacting etcd. This will require either manual intervention to fix the cluster, or a re-deploy from scratch.

## Who can review

### DIFF
--- a/manifests/cf-manifest/spec/networks_spec.rb
+++ b/manifests/cf-manifest/spec/networks_spec.rb
@@ -3,6 +3,7 @@ RSpec.describe "networks" do
   CF_NETWORK_NAMES = %w(
     cf1
     cf2
+    cf3
     cell1
     cell2
     router1
--- a/manifests/cf-manifest/deployments/020-cf-resource-pools.yml
+++ b/manifests/cf-manifest/deployments/020-cf-resource-pools.yml
@@ -9,6 +9,11 @@ resource_pools:
     stemcell: (( grab meta.stemcell ))
     env: (( grab meta.default_env ))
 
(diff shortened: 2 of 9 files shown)


manifests/cf-manifest/spec/networks_spec.rb
```diff
@@ -3,6 +3,7 @@
   CF_NETWORK_NAMES = %w(
     cf1
     cf2
+    cf2
```
saliceti: cf3?

alext: Doh!...  Fixed and rebased.

saliceti:
All looking fine.

```
$ curl localhost:8500/v1/status/peers
["10.0.17.37:8300","10.0.16.37:8300","10.0.18.37:8300"]
```

PR #123, opened 2016-02-25
[#113824903] Work around terraform diffs bug

alext:
## What

Terraform currently throws an error with this specific combination of
interpolation functions when the EIPs' public IPs can't be known ahead of
time (ie when they're also being created in the same terraform run).
Splitting these into separate ingress blocks avoids runing into this
issue.  See hashicorp/terraform#5301 for more details.

This also includes @saliceti's change to use the new terraform container. This is so that I can upgrade this to use terraform 0.6.12. On earlier versions terraform will update the security group on every run when it's configured with multiple ingress blocks with matching port ranges. 

## How to review

Run the cloudfoundry terraform job against a clean environment (ie either a brand new deployment, or one that's had the cloudfoundry terraform destroy job run), and verify that it succeeds first time.

Additionally, point the terraform container at the branch with the updated version (`docker:///governmentpaas/terraform#bump_terraform`), and verify that the security group doesn't get updated on every run (it will be updated on the first run with the new version, but shouldn't be on subsequent runs).
## Who can review

Anyone but @dcarley or myself.

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -98,7 +98,7 @@ jobs:
 
     - task: create-init-bucket
       config:
-        image: docker:///governmentpaas/docker-terraform
+        image: docker:///governmentpaas/terraform
         params:
           TF_VAR_env: {{deploy_env}}
           AWS_DEFAULT_REGION: {{aws_region}}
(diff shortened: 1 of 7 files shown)


mtekel:
Indeed this works cleanly now.

PR #124, opened 2016-02-26
[#113910983] Prevent network ACL's flapping

Jonty:
## What

The mapping from `-1` to `all` is broken in terraform acl rules and only works one-way. This means that whenever terraform polls AWS for changes it finds that `-1` (aws) != `all` (statefile). Changing this to `all` in our configuration stops this from occurring.

Open terraform issue is here:
https://github.com/hashicorp/terraform/issues/5226

## How to review

Redeploy a platform, the network acl's should NOT be destroyed and recreated. Do it a couple of times to be sure.

If the PR functions the apply should look like this:

`
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
`

## Who can review

Anybody except @actionjack and @jonty, who have gazed inside terraform and will never be the same again.


mtekel:
Works as described.

PR #125, opened 2016-02-29
[#114193485] Use RDS Postgres DB in CF

saliceti:
## What

Story: [Implement Postgres using RDS for System Components](https://www.pivotaltracker.com/story/show/114193485)

This removes the previous postgres job, creates an RDS postgres instance, 2 databases (API and UAA), creates security groups, assigns security groups to API and UAA and let API and UAA use the new databases.

## How to review

Create a new CF deployment and run tests. To test for production settings (multi az, backups, final snapshot, restore), try the production settings in dev.tfvars.

## Who can review

Anyone but @mtekel, @keymon or me.


terraform/cloudfoundry/variables.tf
```diff
@@ -77,3 +77,31 @@ variable "cf_subnet_count" {
...
+  description = "Master password for CF database"
+}
+
+variable "secrets_cf_db_api_password" {
+  description = "API password for CF database"
+}
+
+variable "secrets_cf_db_uaa_password" {
+  description = "UAA password for CF database"
+}
```
alext: Are these 2 passwords used by terraform? If not, can we remove them.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -159,40 +205,59 @@ jobs:
...
+
+              for db in api uaa; do
+                # Create database
+                psql ${connect_str} -d postgres -l | grep -q " ${db} " || \
+                  psql ${connect_str} -d postgres -c "CREATE DATABASE ${db} OWNER ${db}"
+                # Enable extensions
+                for ext in citext pgcrypto pg_stat_statements; do
+                  psql ${connect_str} -d ${db} -c "CREATE EXTENSION IF NOT EXISTS ${ext}"
+                done
+              done
```
alext: I think this is long enough to merit being pulled out into a separate script.  That way it'll get shellchecked etc..

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -108,10 +109,50 @@ jobs:
...
+          image: docker:///governmentpaas/curl-ssl
+          inputs:
+            - name: paas-cf
+            - name: generate-cf-secrets
+          run:
+            path: paas-cf/concourse/scripts/s3init.sh
+            args:
+              - {{state_bucket}}
+              - cf-secrets.yml
+              - generate-cf-secrets/cf-secrets.yml
```
alext: This will be an issue when upgrading a deployment with existing secrets in S3 because the new secrets won't get written back to S3. The only way round this at the moment is to manually delete the secrets.yml from S3.

terraform/cloudfoundry/variables.tf
saliceti: Good point, they are not in use anymore. I will remove them.

concourse/pipelines/deploy-cloudfoundry.yml
saliceti: I believe this is a general issue, as we use this pattern for other secrets and certs.
We should think of a better way.

mtekel: Yes, this is known. Moreover, if you remove old secrets, the job will regenerate _all_ secrets, which would mean a big update to CF deployment (likely all VMs needing update). We should look at this more closely. Some secrets are easier to rotate than others...

alext: For the certs, the job operates on a cert-by-cert basis, and only generates ones that don't already exist in the tarball. It should be possible to do something similar for these secrets.

Having said that, this is probably out of scope for this PR.

concourse/pipelines/create-deployer.yml
```diff
@@ -413,6 +413,7 @@ jobs:
     serial: true
     plan:
     - get: paas-cf
+      passed: [vpc]
```
alext: Why is this being added here?  I'm not sure I see what it has to do with using RDS.

mtekel: See commit message & http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule

We got hit by that issue during deployment (it was preventing us doing our work), so we fixed it.

alext: Actually, thinking about this further, this is bad. A terraform run must always use the latest version of the state file (ie not have it pinned). See https://github.com/alphagov/paas-cf/pull/103 for more details, but the summary is that if the state file is pinned, it can potentially run with an older statefile that won't match the actual state of AWS.

mtekel: This pins the git repo. So that the last TF runs with the same code as VPC TF. Without this, if you commit meanwhile (because you work on your branch), the last step runs with different code version and can break.

alext: Doh, yes, misread that detail.

Having said that, I still don't think it belongs in this PR.

PR #126, opened 2016-03-01
[#114193485] Pin expunge vagrant to paas-cf version

saliceti:
## What
Use the same paas-cf version as was used in vpc terraform step.
This is because expunge vagrant step is directly related to vpc step, as it is just removing vagrant IP address from SGs, which was added there in the vpc job.

## How to review

* Spin up concourse lite, it can be an existing environment
* Pause the `concourse-prepare-deploy` job
* Start the pipeline
* Make a change in terraform that will break it, eg syntax error
* Push to the branch
* Unpause the job
* `expunge-vagrant` should be able to run because it is pinned to paas-cf with the working terraform

## Who can review
Anyone but @mtekel or me

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -413,6 +413,7 @@ jobs:
     serial: true
     plan:
     - get: paas-cf
+      passed: [concourse-prepare-deploy]
     - get: build-all-trigger
       trigger: true
       passed: [concourse-deploy]


concourse/pipelines/create-deployer.yml
```diff
@@ -413,6 +413,7 @@ jobs:
     serial: true
     plan:
     - get: paas-cf
+      passed: [vpc]
```
alext: Since this job actually runs both the vpc and concourse terraform runs, it should probably pin to concourse-terraform.

saliceti: I would even pin it to concourse-prepare-deploy because it's the last job in the pipeline that requires paas-cf, and I want to run expunge on the same version. What do you think?

alext: Yeah, makes sense.  And makes for a neater pipeline...

PR #127, opened 2016-03-01
[#114186797] move smoke and acceptance test to concourse

combor:
## What

We want to run the smoke tests and acceptance tests directly in a container in Concourse, but also keep the original templates from the CF release.

## Dependencies

 * https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/28 must be merged and the container rebuild before merging/testing this.

## Scope

 - Change pipeline to use a container to run both tests pulling the right version of the test for the cf-release.
 - Render the configuration from the cf-release, with the manifest settings. We add render code that is tested.

## Notes 

 * https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/28 must be merged and the container rebuild before merging/testing this.

 * We had to remove some quotes using sed in the code in a call to the curl command due [the issue described here](https://github.com/cloudfoundry/cf-acceptance-tests/pull/90). That was simplier than backporting a modern version of curl or rebuilding all the container from scratch.

 * we had to do a workaround for non working tag property for the git resource to ensure that the test version is the same as the one that was pinned to the release tag. We manually checkout the release tag after resource cloned tests code. This should be removed in the future after concourse upgrade. 

 * Also, one of the logging tests is disabled as it tries to create local syslog drain inside the container and forward app messages to it.

## How to review

Upload the pipeline and run smoke and acceptance-tests - they should pass.

```
BRANCH=114186797_move_smoke_and_accfeptance_test_to_concourse concourse/scripts/pipelines-cloudfoundry.sh env_name
```

In paas-cf root directory run:

```
bundle install
make test
```
 All tests should pass.

## Who can review

Anyone but @keymon and @combor


keymon:
I will rebase to make more clear the changes.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -441,18 +447,24 @@ jobs:
...
+      - get: cf-release
+        params:
+          submodules:
+          - src/smoke-tests
       - get: paas-cf
         passed: ['deploy']
       - get: cf-manifest
         passed: ['deploy']
-        trigger: true
       - get: bosh-secrets
```
alext: I don't think this uses `bosh-secrets` any more (neither do the acceptance tests).

keymon: True, you are right.

I will do a PR.

keymon:
@alext https://github.com/alphagov/paas-cf/pull/134

PR #128, opened 2016-03-02
[#113996483] Ensure all components have consolidated logging

actionjack:
[Ensure all components have consolidated logging](https://www.pivotaltracker.com/story/show/113996483)

## What

Acceptance criteria:

if it logs and it is a system component
then its logs should appear in logsearch
or there should be a story to implement the logs

After reviewing our logsearch implementation we found that we are not logging all events most notably we are missing the all the logs under `/var/vcap/*/*.log` 

This PR 
Installs and configures [logsearch-shipper-boshrelease](https://github.com/logsearch/logsearch-shipper-boshrelease)
to ensure all the logs under `/var/vcap/*/*.log` are sent to logsearch
ingestor. This commit is fairly large because it is non-breaking.

I've drafted this PR to be reviewed in the following context:

* I would like to:
  * Install the logsearch-shipper bosh release
  * Add the logsearch-shipper to each node
  * Configure the logsearch-shipper to send logs to the logsearch ingestor node
  * Disable Metrics collection and metrics field tagging because we are not going to use them 

## How to Review this PR

Deploy or update your existing deployment. Create ssh tunnel via deployer concourse to kibana:
ssh -L 5601:10.0.16.49:5601 -i $session/id_rsa vcap@$DEPLOYER_CONCOURSE. You should be able to browse logs and see them appearing in real time.
You can also connect to one of the nodes and verify that the logs are being parsed by the logsearch-shipper process.

## Who can review this PR

This PR can be reviewed by a android who is dreaming of electric sheep if a product of the Tyrell corporation can not be found then any member of the corp team will do with the exception of @actionjack and @tmower

### DIFF
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -326,5 +326,7 @@ properties:
     port: 2514
     transport: relp
   logsearch:
+    metrics:
+      enabled: false
     logs:
       server: (( concat jobs.ingestor.networks.cf1.static_ips.[0] ":5514" ))
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -3,6 +3,10 @@ releases:
   url: https://bosh.io/d/github.com/logsearch/logsearch-boshrelease?v=200.0.0
   version: 200.0.0
   sha1: 527c35db31cd66810accf128ceffee4f5fde80c3
+- name: logsearch-shipper
+  url: https://bosh.io/d/github.com/logsearch/logsearch-shipper-boshrelease?v=4
+  version: 4
+  sha1: c92ce357c5e1e77a4fb7944d58a1971c7aee06b0
 
 disk_pools:
 - name: elasticsearch_master
@@ -179,6 +183,7 @@ jobs:
     logstash_parser:
       filters:
       - metrics: /var/vcap/packages/logsearch-config/logstash-filters-metric.conf
+
     nats_to_syslog:
       nats:
         subject: ">"
@@ -320,3 +325,6 @@ properties:
     address: (( grab jobs.ingestor.networks.cf1.static_ips.[0] ))
     port: 2514
     transport: relp
+  logsearch:
+    logs:
+      server: (( concat jobs.ingestor.networks.cf1.static_ips.[0] ":5514" ))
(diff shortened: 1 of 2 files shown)


keymon:
Rebased with master

keymon:
I am concern that this PR setups collectd to send the metrics to ElasticSearch as log events.

After discussing with @bleach  and @alext, we agreed that it makes sense to disable the metrics feature from collectd [by setting `logsearch.metrics.enabled=false`](https://github.com/logsearch/logsearch-shipper-boshrelease/blob/master/jobs/logsearch-shipper/spec#L94).

That option set to `false` will prevent collectd to start. It does not matter that the package and config is installed, and it won't conflict with other deployments.

If we disable metrics, probably we can get rid of that filter in `logstash_parse.filters`

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -177,8 +181,24 @@ jobs:
       - shards-and-replicas: "{ \"template\" : \"*\", \"order\" : 99, \"settings\" : { \"number_of_shards\" : 1, \"number_of_replicas\" : 0 } }"
       - index_template: /var/vcap/packages/logsearch-config/default-mappings.json
     logstash_parser:
-      filters:
```
keymon: If we do not enable the metrics, do we need to add this filter?

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -176,9 +180,7 @@ jobs:
       templates:
       - shards-and-replicas: "{ \"template\" : \"*\", \"order\" : 99, \"settings\" : { \"number_of_shards\" : 1, \"number_of_replicas\" : 0 } }"
       - index_template: /var/vcap/packages/logsearch-config/default-mappings.json
-    logstash_parser:
```
keymon: I think you added this change by accident. You don't need to remove this to disable metrics collection.

keymon:
I have found out that [there is a issue](https://github.com/logsearch/logsearch-shipper-boshrelease/pull/15
) in the upstream bosh release which makes it not work disabling the collectd service. 

I do think that we can live with this issue for the time being until they fix it upstream.

keymon:
LGTM, merging

keymon:
/cc @timmow @actionjack 

There are some things I missed in the PR and that I noticed now:
 1. Not all the jobs have the logshipper template defined. Is that intended?
 2. We parametrized the job name rather than the release name (https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/deployments/030-cf-jobs.yml#L50). It should be:

```
  - name: logsearch-shipper
    release: (( grab meta.logsearch_shipper.release.name ))
```

not 

```
  - release: "logsearch-shipper"
    name: (( grab meta.logsearch_shipper.release.name ))
```

PR #129, opened 2016-03-02
[#114080971] New makefile targets per environment 

combor:
## What

We want our Makefile to support a target per environment so for instance it would be possible to do `make dev` to upload DEV pipeline. Additionally we want help to be available as a default target. 

## How to review

Check whether prod,stage,ci and dev targets deploy pipelines with following config:

- Prod
  - multi AZ
  - AWS account PROD
- Staging
  - multi AZ
  - AWS account  STG
- Ci
 - AWS account CI
 - auto trigger on paas-cf
- Dev
  - AWS account DEV
  - autodelete

## Who can review

Anyone but @actionjack or @combor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,9 @@
-.PHONY: test spec lint_yaml lint_terraform lint_shellcheck set_aws_count set_auto_trigger disable_auto_delete check-env-vars dev ci stage prod
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck set_aws_count set_auto_trigger disable_auto_delete check-env-vars dev ci stage prod
+
+.DEFAULT_GOAL := help
+
+help:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,3 +27,32 @@ lint_terraform:
...
+	$(eval export AWS_ACCOUNT=prod)
+
+set_auto_trigger:
+	$(eval export ENABLE_AUTO_DEPLOY=true)
+
+disable_auto_delete:
+	$(eval export DISABLE_AUTODELETE=1)
+
+deploy_cloudfoundry_pipeline:
+	concourse/scripts/pipelines-cloudfoundry.sh
```
alext: I think this should deploy all the pipelines (bosh and CF), and therefore be named `deploy_pipelines`.

combor: Fixed.

PR #130, opened 2016-03-02
[#114778351] Make secret generation robust to changes

alext:
## What

Update the secret generation mechanism so that it reads in the existing secrets, and only generates secrets that aren't already present. It also removes values that are no longer required. This allows us to add and remove secrets (eg in the recent RDS story), and have those changes take effect.

This also changes the pipelines so that the secret generation job gets and puts the resource, meaning that the downstream jobs can be pinned to the secret generation job to ensure they use the up-to-date version.

To achieve this, I've reimplemented the secret generation in ruby (I didn't fancy parsing YAML with bash/awk etc.). This now means that secrets can contain `-` and `_` in addition to alphanumeric characters. These are safe to use in a URL though, so I don't anticipate that being a problem.

## How to review

If you have a deployment from before the RDS merge, this is a perfect opportunity to test this.  On current master, this will error with missing secrets, with this branch, it should complete successfully.

Running this against an existing deployment should not change the secrets (for either bosh or CF).

If a secret is manually removed from the secrets file in S3, a redeployment should generate a new value for that secret, but leave all others unchanged.

## Who can review

Anyone but myself.


alext:
Note: when this is merged, I intend to remove the https://hub.docker.com/r/governmentpaas/mksecrets/ container (both from dockerhub, and from the repo) as it's no longer needed.

combor:
I had missing password so I took this great opportunity and tested this change and it works OK.

keymon:
quite good code! :)

Good for me, running the latest tests and will merge in some minutes :)

keymon:
In the `create-microbosh` there is a race condition: The bosh-secrets job might fail if the `bootstrap-s3-state` does not initialise the s3 files first.

Does not make sense to add a dependency? Or move the `s3init` execution to the init job, as it is done in the deploy-cloudfoundry.

alext:
> In the create-microbosh there is a race condition:

I don't think it actually will race. The bosh-secrets resource won't be available until the bootstrap-s3-state job has run, so the job won't be triggered (if it were a problem it would affect the terraform and generate-CA-certs jobs as well).

Having said that, I can see that it's a bit confusing at the moment.  I'll change it so that the bootstrapping is done in the init job to make it consistent with the other pipelines.

alext:
I've added a commit to move the bootstrapping to the init job.

keymon:
You are right, likely the job will wait until the resource is ready. But good that you changed it to look like the other pipeline.

keymon:
Last changes LGTM, merging

dcarley:
:+1: :+1: :+1:

PR #132, opened 2016-03-03
[#114385219] Update deployer concourse

mtekel:
## What

[Upgrade Concourse](https://www.pivotaltracker.com/n/projects/1275640/stories/114385219)

This PR updates deployer concourse to version 0.74 (this includes updates to stemcell and garden linux). In the new concourse version, each task has to have platform defined. Moreover, you can't input whole task any more, but have to define task outputs explicitly. This PR also contains minimal changes to pipelines to satisfy these new requirements. The pipeline update commits have been split into logical chunks, so that it is not a single huge commit: update bosh pipelines, update CF delete pipelines and update main CF pipeline.

## How to review

Switch to this branch, apply against your existing environment (update works fine and preserves all your changes/pipelines from previous concourse). Or you can deploy new. After deployment, you should be able to admire the shiny new mustard-yellow/pea green concourse theme. Run the pipelines and verify they still work like before upgrading. I suggest testing the create pipelines first and destroy last (preferably, destroy CF first and then destroy bosh) - this saves time.

## Who can review

Someone else than me.

### DIFF
--- a/concourse/pipelines/create-microbosh.yml
+++ b/concourse/pipelines/create-microbosh.yml
@@ -141,9 +141,12 @@ jobs:
     - try:
         task: generate
         config:
+          platform: linux
           image: docker:///governmentpaas/certstrap
           inputs:
           - name: bosh-CA
+          outputs:
+          - name: generate
           run:
             path: sh
             args:
@@ -154,6 +157,7 @@ jobs:
                 certstrap init --passphrase "" --common-name bosh-CA
                 cd out
                 tar -cvzf ../bosh-CA.tar.gz bosh-CA.*
+                cp ../bosh-CA.tar.gz ../generate
               else
                 echo "The CA cert already exists, skipping generation..."
                 exit 1
@@ -177,6 +181,7 @@ jobs:
     - get: bosh-tfstate
     - task: terraform-variables
       config:
+        platform: linux
         image: docker:///ruby#2.2.3-slim
         run:
           path: sh
@@ -198,6 +203,7 @@ jobs:
         - name: terraform-variables
     - task: terraform-apply
       config:
+        platform: linux
         image: docker:///governmentpaas/terraform
         params:
           DEPLOY_ENV: {{deploy_env}}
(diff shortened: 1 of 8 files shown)


concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -616,11 +638,13 @@ jobs:
           - -e
           - -c
           - |
+            ls -lR
```
saliceti: Debug code?

mtekel: yes

saliceti:
On a brand new deployment, the generate-bosh-manifest job is pending. It is stuck on `waiting for a suitable set of input versions`

```
preparing build
checking pipeline is not paused
checking job is not paused
discovering any new versions of bosh-CA
discovering any new versions of bosh-secrets
discovering any new versions of bosh-tfstate
discovering any new versions of paas-cf
discovering any new versions of pipeline-trigger
discovering any new versions of vpc-tfstate
waiting for a suitable set of input versions
checking max-in-flight is not reached
```

concourse/pipelines/create-microbosh.yml
```diff
@@ -120,11 +122,14 @@ jobs:
           - -e
           - -c
           - |
-            ./paas-cf/manifests/bosh-manifest/scripts/generate-bosh-secrets.sh > bosh-secrets.yml
-            ls -l bosh-secrets.yml
+            ./paas-cf/manifests/bosh-manifest/scripts/generate-bosh-secrets.sh > generate/bosh-secrets.yml
+            ls -l generate/bosh-secrets.yml
+        outputs:
+        - name: generate
```
alext: I think it's easier to follow if the `outputs` are listed before the `run` because the `run` step write to directories that are created by the `outputs` definitions.  What do you think?

alext: Also, is there the potential for confusion if an output is named the same as a task?

mtekel: We have 2 patters how we do outputs, I went with the pattern that's used locally. Meaning in bosh pipelines it's inputs, script, outputs; in CF it's in, out, script. We should pick one pattern and refactor all pipelines to use it. Likely when merging. I wanted to do only minimal changes for this PR - and this is also why the outputs are named same as task, as this is how they were imported previously. If I change this name, I would have to also change all the tasks that import it. The story here is to update concourse and not to refactor pipeline. But I think this is also one of the things we should fix.

combor: At the other hand the behaviour before update was to implicitly create output from task with a task name assigned to it. Right now we are following this patter but we specify output name explicitly.

mtekel:
The blockage was due to missing outputs in generate-CA task, which was failing to upload generated certs. Fixed now.

PR #135, opened 2016-03-04
[#114186797] Remove smoke/acceptance test errand defintions.

alext:
## What

These are no longer needed now that we're running these tests in
concourse, and not in an errand. Left behind after #127.

## How to review

Deploy this branch, verify that the smoke and acceptance tests still run.

## Who can review

Anyone but myself

### DIFF
--- a/manifests/cf-manifest/deployments/900-cf-tests.yml
+++ b/manifests/cf-manifest/deployments/900-cf-tests.yml
@@ -1,25 +1,3 @@
-jobs:
-  - name: acceptance_tests
-    templates:
-    - name: acceptance-tests
-      release: (( grab meta.release.name ))
-    instances: 1
-    resource_pool: small_errand
-    lifecycle: errand
-    networks:
-      - name: cf1
-
-  - name: smoke_tests
-    templates:
-    - name: smoke-tests
-      release: (( grab meta.release.name ))
-    instances: 1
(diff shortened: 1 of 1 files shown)


mtekel:
These come from the CF release manifest. If we remove them, I guess we stop tracking changes made to these - which would show if anything has changed about the tests... (e.g. new properties, needs bigger instance type etc.)

alext:
This is only removing the errand definitions, not the properties. We'll still want to track changes to the properties, but I don't see why we'd need to track changes to the errands as we're not using errands any more.

mtekel:
We're not using errands, but it's the official way to run the tests. There might be some changes about errands perhaps (instance types? networks? jobs?), not very likely, but who knows...

alext:
I don't think it's necessary to maintain these errands here to be able to track any changes to upstream. When playing the most recent upgrade story, the easiest way to see theupstream changes was to look at the diff between the 2 tags in the upstream manifests. That will hilight any changes in these errand definitions that might be relevant to us.

dcarley:
> I don't think it's necessary to maintain these errands here to be able to track any changes to upstream. When playing the most recent upgrade story, the easiest way to see the upstream changes was to look at the diff between the 2 tags in the upstream manifests. That will highlight any changes in these errand definitions that might be relevant to us.

:+1: this

PR #136, opened 2016-03-04
[#114233261] Automated failure testing for key components

actionjack:
[Automated failure testing for key components](https://www.pivotaltracker.com/story/show/114233261)

## What

 Acceptance criteria:
 
- Should remove component one by one
- Should test the most essential functions still work
- Tests should be green at the end of the story
- If there are important components that don't produce green tests if killed, then separate stories to fix these should be spun off

This PR creates a separate failure testing pipeline that will delete core component vm's and test if the service still operates and passes the smoke tests.

I've drafted this PR to be reviewed in the following context:

* I would like to:
  * Create a `concourse` pipeline that terminates the following job nodes and runs smoketests:
     *  cloud-controller
     *  colocated
     *  nats
     *  router
     *  etcd
     *  consul
     *  cell
  *  We have added sleeps to ensure that the Virtual machines have time to terminate and for the platform to heal itself before running smoke tests
  * Create a shell script to create the `failure-testing` pipeline
  * Add the failure tests to `deploy_pipelines` Makefile goal

## How to Review this PR

Deploy or update your concourse deployer and trigger the the `failure-testing` pipeline.

## Who can review this PR

This PR can be reviewed by a team of inebriated [satyrs](https://en.wikipedia.org/wiki/Satyr) if troup of tipsy legendary creatures can not be found in a nearby public house then any sober member of the core team will do with the exception of @actionjack and @combor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,3 +61,4 @@ disable_auto_delete:
 deploy_pipelines:
 	concourse/scripts/pipelines-microbosh.sh
 	concourse/scripts/pipelines-cloudfoundry.sh
+	concourse/scripts/pipelines-failure-testing.sh
--- a/concourse/pipelines/failure-testing.yml
+++ b/concourse/pipelines/failure-testing.yml
@@ -1,0 +1,257 @@
+resources:
+  - name: paas-cf
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-cf.git
+      branch: {{branch_name}}
+
+  - name: pipeline-trigger
+    type: semver-iam
+    source:
+      bucket: {{state_bucket}}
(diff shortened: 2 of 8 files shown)


mtekel:
I think it would be nice to remove some of the repetition by including tasks like kill-instance, generate-test-config and run-test (including recover) from a file.

mtekel:
I think it would be nice to remove some of the repetition by including tasks like kill-instance, generate-test-config and run-test (including recover) from a file.

combor:
This is a first step in failure testing pipeline lifecycle. Next step should be auto-generation of this pipeline as the only difference between tasks is VM_NAME.

PR #137, opened 2016-03-04
[#114385219] Upgrade Concourse - bootstrap concourse

timmow:
**What**

A new version of Concourse has been released, we should update to latest version.

Destroy any existing deployment or use a new environment name

Check this branch out and run the `vagrant/deploy.sh` script, and run the `create-deployer` and `destroy-deployer` pipelines and check all is working.

To make this work, we had to upgrade the kernel on the concourse vm, as the AMI concourse provide uses an older kernel version that doesnt support the new version of the garden linux release included. This means we also need to make the additional disk mount persist across reboots. There are more details on this in the individual commit messages.

Concourse are planning on [upgrading the AMIs](https://www.pivotaltracker.com/n/projects/1059262/stories/114892073) however its not clear when they will be upgraded.

**Who should review this PR**

* Anyone  except me or @mtekel

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -79,7 +79,8 @@ jobs:
     - get: paas-cf
     - task: try-fetch-bucket-state
       config:
-        image: docker:///governmentpaas/awscli
+        platform: linux
+        image: docker:///ubuntu
         params:
           AWS_DEFAULT_REGION: {{aws_region}}
         inputs:
@@ -98,6 +99,7 @@ jobs:
 
     - task: create-init-bucket
       config:
+        platform: linux
         image: docker:///governmentpaas/terraform
         params:
           TF_VAR_env: {{deploy_env}}
@@ -105,6 +107,8 @@ jobs:
         inputs:
           - name: paas-cf
           - name: bucket-state
+        outputs:
+          - name: create-init-bucket
         run:
           path: sh
           args:
@@ -113,7 +117,7 @@ jobs:
             if [ -f bucket-state/bucket.tfstate ]; then
               cp bucket-state/bucket.tfstate .
             fi
-            terraform apply -state=bucket.tfstate -var-file=paas-cf/terraform/{{aws_account}}.tfvars \
+            terraform apply -state=bucket.tfstate -state-out=create-init-bucket/bucket.tfstate -var-file=paas-cf/terraform/{{aws_account}}.tfvars \
               paas-cf/terraform/bucket
       on_success:
         put: bucket-terraform-state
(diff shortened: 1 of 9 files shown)


vagrant/destroy.sh
```diff
@@ -22,3 +22,4 @@ fi
 
 # 3. Destroy bootstrap concourse
 vagrant destroy -f
+rm .kernel-updated
```
alext: This won't cover the case when the machine is deleted by the `self-terminate` pipeline.

concourse/pipelines/create-deployer.yml
```diff
@@ -79,7 +79,8 @@ jobs:
     - get: paas-cf
     - task: try-fetch-bucket-state
       config:
-        image: docker:///governmentpaas/awscli
+        platform: linux
+        image: docker:///ubuntu
```
dcarley: There was a comment on an individual commit that hasn't been addressed yet. Making a note here, so we don't forget it: https://github.com/alphagov/paas-cf/commit/c30bb834dd01efeea29b47aa165e57dcac4aa499#diff-5eac5d6fc041d1da10144e10c0fa2ed0L82

vagrant/post-deploy.d/20-setup-basic-auth.sh
```diff
@@ -2,6 +2,7 @@
 set -u
 set -e
 echo "Setting up concourse basic auth as ${CONCOURSE_ATC_USER}: ${CONCOURSE_ATC_PASSWORD}"
+sed "s/--external-url.*/--external-url \'\' \\\/" -i /var/vcap/jobs/atc/bin/atc_ctl # Remove hardcoded auth redirect IP
```
dcarley: This sounds like a bug from the description in the commit message. Can you create an issue on the upstream project?

mtekel: _what_ upstream project? We only think this AMI is coming from https://github.com/concourse/concourse-lite

Apparently they have some jenkins job that builds the AMI. They have said earlier that AMI is not supported. I am not sure the same issue is present in the lite version - probably not, as this IP could be coming from there.

mtekel: But still https://github.com/concourse/concourse-lite/issues/4

dcarley: Thanks.

For the benefit of anyone else following along: they have explicitly stated in that issue that they don't intend to build the AMI any longer.

dcarley:
We decided not to proceed with this because the workarounds required outweighed the benefit of upgrading the Bootstrap Concourse, where we aren't using as many features and don't use as frequently. It has been superseded by [#115322155](https://www.pivotaltracker.com/story/show/115322155) to make sure that we use the correct version of `fly` against each Concourse. We'll need another story, in the near future, to look at alternatives to using the upstream AMI which is no longer supported.

PR #138, opened 2016-03-04
[#114854977] Merge bosh and CloudFoundry creation pipelines

alext:
## What

This replaces the `create-microbosh` and `deploy-cloudfoundry` pipelines with a single `create-bosh-cloudfoundry` pipeline. This will allow us to add locking around this so that we can auto-upgrade pipelines.

As part of this work, I've tried to apply some consistent patterns across this pipeline:

* Naming: tasks should be named with verb-like names, inputs and outputs should have noun-like names.
* Task ordering: Task params are ordered `platform`, `image`, `inputs`, `outputs`, `params`, `run`. This makes it easier to follow the flow of a task and see what it operates on.
* Triggering: everything now triggers from the `pipeline-trigger` resource alone.
* Indentation of lists: I've updated the indentation to be consistent and follow the examples in the YAML spec (see cfd8ba4)

## How to review

Run the new pipeline against an existing deployment. It should make no changes.

Destroy CF and Bosh, then run the new pipeline again. It should successfully redeploy both. It should also parallelise the middle section of the pipeline (deploy-bosh, and the CF config and terraform bits).

## Who can review

Anyone but myself.


combor:
Alex, can you rebase please ?

README.md
```diff
@@ -141,17 +141,17 @@ supporting services for the platform.
 
 ### Deploy
 
-Run the `create-microbosh` pipeline.
+Run the `create-bosh-cloudfoundry` pipeline (which will also deploy CloudFoundry - see below).
```
combor: I think we can merge CF and BOSH subsections now as we've got just one pipeline.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+      tag_filter: {{cf-release-version}}
+
+jobs:
+  - name: init
+    serial_groups: [bosh-deploy]
+    serial: true
+    plan:
+      - get: paas-cf
+        trigger: {{auto_deploy}}
+      - task: bootstrap-s3-state
```
combor: Do you think it's worth to split this into several tasks based on what we init and aggregate them. This should look better in UI and speed up initialisation.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+        config:
+          platform: linux
+          image: docker:///governmentpaas/curl-ssl
```
combor: There is new way of defining docker resources, e.g: https://github.com/concourse/concourse/blob/master/ci/atc-go.yml#L4

I't might me good to take opportunity now and migrate our image definitions.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+                ruby paas-cf/concourse/scripts/extract_tf_vars_from_terraform_state.rb \
+                < vpc-tfstate/vpc.tfstate \
+                > terraform-variables/vpc.tfvars.sh
+                ruby paas-cf/concourse/scripts/extract_tf_vars_from_terraform_state.rb \
+                < concourse-tfstate/concourse.tfstate \
+                > terraform-variables/concourse.tfvars.sh
+      - task: terraform-apply
+        config:
+          platform: linux
+          image: docker:///governmentpaas/terraform
```
combor: I wonder whether it would be good to use tags here like for `docker:///ruby#2.2-slim`. We don't have versions now but I guess we want to have.

alext: I'm not sure about this. IMO, specifying a version is saying we want to use something other than the default version (ie `latest`). I think if we start formally versioning our containers, then that's the time we should update things in the pipelines.

alext:
> Alex, can you rebase please ?

Done.

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: I'm not sure that would gain us much. I think the overhead of spinning up a separate container for each task would outweigh the gain from aggregating these. Once the container has been downloaded, these items are very quick to run.

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: We have a spike [#113285957](https://www.pivotaltracker.com/story/show/113285957) to investigate it

README.md
alext: Good call.  Updated and rebased.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+              ./terraform-outputs/bosh.terraform-outputs.yml
+              ./terraform-outputs/vpc.terraform-outputs.yml
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                ./paas-cf/manifests/bosh-manifest/build_manifest.sh $BOSH_MANIFEST_STUBS > bosh-manifest/bosh-manifest.yml
+                /bin/cat bosh-manifest/bosh-manifest.yml
```
combor: Do we want to do it here? Manifest contains sensitive data.

alext: That's a valid question, but this was present in the existing pipeline, so I left it.

mtekel:
There were 2 approaches in outputs from tasks: some people like no output, some others like to see what's going on (e.g. ls certs/other created files to show that they exits. You can then compare size between job runs). Do we have consistency on this?

alext:
> There were 2 approaches in outputs from tasks: some people like no output, some others like to see what's going on

I think I definitely prefer tasks to have at least some output (although `cat`ing the whole manifest if too much IMO).

Having said that, I think changing the level of output from the various tasks is probably beyond the scope of this PR.

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: I'll take a look at that.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+                cp bosh-init-state/bosh-manifest-state.json bosh-manifest/bosh-manifest-state.json
+                bosh-init deploy bosh-manifest/bosh-manifest.yml
+                cp bosh-manifest/bosh-manifest-state.json updated-bosh-init-state/bosh-manifest-state.json
+        ensure:
+          put: bosh-init-state
+          params:
+            file: updated-bosh-init-state/bosh-manifest-state.json
+
+  - name: generate-cf-secrets
+    serial_groups: [cf-deploy]
```
combor: The name of serial group is different for BOSH and CF part. Does it mean that we can run both parts in parallel? I have an impression that this can lead to race conditions.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+                ln -snf $(pwd)/test-config/run /var/vcap/jobs/smoke-tests/bin/run
+                ln -snf /usr/local/go /var/vcap/packages/golang1.4
+                ln -snf $(pwd)/cf-release/src/smoke-tests /var/vcap/packages/smoke-tests/src/github.com/cloudfoundry/cf-smoke-tests
+                /var/vcap/jobs/smoke-tests/bin/run
+
+  - name: acceptance-tests
+    plan:
+      - aggregate:
+          - get: pipeline-trigger
+            passed: [ 'cf-deploy']
```
combor: We can remove the trigger here, It was committed by mistake and we don't want to trigger acceptance tests automatically.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+            params:
+              submodules:
+                - src/github.com/cloudfoundry/cf-acceptance-tests
+          - get: paas-cf
+            passed: ['cf-deploy']
+          - get: cf-manifest
+            passed: ['cf-deploy']
+          - get: bosh-CA
+
+      - task: generate-test-config
```
combor: We've got a task file for this so we can share the code between pipelines: https://github.com/alphagov/paas-cf/blob/master/concourse/tasks/generate-test-config.yml

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,901 @@
...
+
+                chmod +x ./test-config/run
+
+                ./paas-cf/tests/bosh-template-renderer/render.rb \
+                  ./cf-release/jobs/acceptance-tests/templates/config.json.erb \
+                  ./cf-release/jobs/acceptance-tests/spec \
+                  ./cf-manifest/cf-manifest.yml \
+                    > ./test-config/config.json
+
+      - task: run-tests
```
combor: Same as for generate-config task: https://github.com/alphagov/paas-cf/blob/master/concourse/tasks/run-tests.yml

concourse/pipelines/create-deployer.yml
```diff
@@ -208,7 +208,7 @@ jobs:
 
     - task: vpc-terraform-outputs-to-sh
       config:
-        image: docker:///ruby#2.2.3-slim
+        image: docker:///ruby#2.2-slim
```
combor: Just out of curiosity...why not 2.2.3 ?

alext: See the commit message for full details, but ruby uses a semver style versioning scheme.

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: Maybe, but if that's desirable, it belongs in a separate PR.

As an aside, I'm not sure it necessarily is desired. The reason (or one of the reasons) we weren't triggering before was because it would conflict with the smoke tests. Now that these are no longer using errands, they don't conflict any more.

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: I think any refactoring to use separate task definitions is out of scope for this PR. It's not something that's been discussed amongst the team, and needs a bit of thought (eg at the moment, some of the naming isn't very clear).

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: I've had an experiment with this. It's not working for me at the moment. The jobs fail with error "fetching the image did not produce a volume" after downloading the fs layers and building the docker image.

I think we should leave that for a future enhancement.

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: That's definitely not desired behaviour on DEV environment. We can parametrize it like triggering on paas-cf resource https://github.com/alphagov/paas-cf/blob/master/Makefile#L55

keymon:
I like the description of the patterns and style guides. I think we should write them down and apply them across all pipelines. Maybe even do a linter?

alext:
>  I think we should write them down and apply them across all pipelines.

I had in mind to start some styleguides for the team - maybe as part of the team manual.

keymon:
As commented in person: It will be great if you change the commits so that there is:
 - one commit (or more) adding the bosh stuff into `deploy-cloudfoundry.yml`
 - a final commit renaming `deploy-cloudfoundry.yml` to `create-bosh-cloudfoundry.yml`

The reason is that this way we keep most of the history of changes (specially the CF ones) and also will make it easier to rebase other PR with this.

alext:
@keymon Rebased with that change.

concourse/pipelines/create-bosh-cloudfoundry.yml
alext: That's intentional to allow the middle section of the pipeline to run in parallel (the `bosh-deploy` run at the same time as the CF config generation and terraform jobs).

keymon:
I am happy  to merge. There are several minor things I would like to solve, but there are not really related to this change.

PR #139, opened 2016-03-07
[#113749465] Deploy Graphite and Grafana

saliceti:
## What
Story: [Make Cloud Foundry Metrics Available](https://www.pivotaltracker.com/story/show/113749465)

We already had graphite deployed in trial so we imported the code along the commit history and based our new deployment upon it.

Changes:

* Deploy graphite in a single VM with a static IP
* Add ACL to access graphite and grafana from infra subnet
* Ruby script using the bosh cli library to create and upload release
* Add graphite nozzle that plugs into the firehose to read metrics and upload to statsd on the graphite server. Currently deployed as a CF app.
* Extend consul ACL to upload data from nozzle to statsd
* Helper script to read values from manifest
* Deploy grafana alongside graphite

## Requirements

This PR is based on #138 which must be merged first

## How to review

* Run deployment in a new or existing environment.
* Access graphite and grafana by running tunnels to the graphite_z1 VM on port 3001 and 3000 respectively
* You should see metrics populated
* You can play by sending metrics from the cell subnet with:

```
echo "deploys.test.myservice:1|c" | nc -w 1 -u 10.0.16.20 8125
```

## Who can review
Anyone but @keymon or me


dcarley:
I'm not sure about the inclusion of the commits from alphagov/cf-terraform. A lot of the commit messages don't make sense in the context of this repo (e.g. modifying a provisioning script, changing retention rates for soak tests, registering ELBs, etc) and the content is subsequently rewritten in one large commit. Are you able to restructure these whilst retaining some of the important information (e.g. increasing the persistent disk size and blacklisting smoketests apps)?

keymon:
@dcarley hope you will find the history more clear and relevant now.

BTW, I am rebasing with #138

dcarley:
> @dcarley hope you will find the history more clear and relevant now.

Perfect! Thanks for taking the time do it.

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -464,6 +507,117 @@ jobs:
...
+        outputs:
+        - name: config
+        params:
+        image: docker:///ruby#2.2-slim
+        run:
+          path: sh
+          args:
+          - -e
+          # FIXME
+          - -x
```
combor: Fix me please...

concourse/pipelines/deploy-cloudfoundry.yml
```diff
@@ -464,6 +507,117 @@ jobs:
...
+            GRAPHITE_SERVER=$($VAL_FROM_YAML jobs.graphite_z1.networks.cf1.static_ips.0 cf-manifest/cf-manifest.yml)
+
+            for var_name in CF_ADMIN CF_PASS FIREHOSE_USER FIREHOSE_PASS API_ENDPOINT UAA_ENDPOINT DOPPLER_ENDPOINT GRAPHITE_SERVER; do
+              echo export "${var_name}"=\"$(eval echo \$${var_name})\"
+            done > config/config.sh
+
+            # FIXME
+            cat config/config.sh
+
+    - task: deploy-graphite-nozzle
```
combor: Why don't we use CF resource to deploy an app ? https://github.com/concourse/cf-resource

terraform/cloudfoundry/network_acls_cf_in.tf
```diff
@@ -172,6 +172,17 @@ resource "aws_network_acl_rule" "117_kibana" {
     egress = false
 }
 
+resource "aws_network_acl_rule" "118_graphite" {
+    network_acl_id = "${aws_network_acl.cf.id}"
+    protocol = "tcp"
+    rule_number = 118
```
combor: It's going to be rule clash: https://github.com/alphagov/paas-cf/blob/113996375_ha_for_logsearch/terraform/cloudfoundry/network_acls_cf_in.tf#L178

concourse/pipelines/deploy-cloudfoundry.yml
keymon: There are several reasons:
  1. We configure security groups, although it is true that it could be done separately.
  2. We generate the `manifest.yml`. Although we could fork the repo and commit it.
  3. We need to pass several environment variables, so we need to pass that to concourse. But the values come from values from the `cf-manifest.yml` and `secrets.yml`. When we initially create the pipeline, we don't know these values.
  4. We need to pass the CloudFoundry credentials to the resource, but again, we don't them and they come from `cf-secrets.yml`

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -0,0 +1,1047 @@
...
+              done > config/config.sh
+
+    - task: deploy-graphite-nozzle
+      config:
+        platform: linux
+        inputs:
+          - name: graphite-nozzle
+          - name: config
+        params:
+        image: docker:///governmentpaas/cf-acceptance-tests
```
alext: Should we create a separate cf-cli container for this?  We don't need the Go runtime, or godep etc for this task.

keymon: True, I will. Thx

terraform/cloudfoundry/network_acls_cf_in.tf
Jonty: I'll bump the other branch to 119 when it's finished review.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -705,6 +747,111 @@ jobs:
...
+                  STATSD_ENDPOINT: "${GRAPHITE_SERVER}:8125"
+                  FIREHOSE_USERNAME: "${FIREHOSE_USER}"
+                  FIREHOSE_PASSWORD: "${FIREHOSE_PASS}"
+                  SUBSCRIPTION_ID: "firehose"
+                  STATSD_PREFIX: "cfstats."
+                  SKIP_SSL_VALIDATION: "true"
+                  PREFIX_JOB: "true"
+              EOF
+
+              cf push
```
keymon: /cc @Jonty 
> Rather than cf-pushing on every run, we could use something like cf apps | grep graphite-nozzle to only push the app if it's not already deployed? It's really not good to be pushing on every deploy as it'll restart the app and we'll have metric gaps.

Yes, but that won't update the app if there is an update. Other approach would be put the git hash in the name of the app, and delete the old one if required. 

But as we comment in the commit, this is a temporary solution, we will deploy this app with a release. Unless you are strongly against it, I suggest to leave it as it is.

concourse/scripts/val_from_yaml.rb
```diff
@@ -0,0 +1,60 @@
...
+    property_tree = PropertyTree.load_yaml(STDIN.load())
+  end
+
+  val = property_tree[key]
+  if val.is_a? Array or val.is_a? Hash
+    puts YAML.dump(val)
+  else
+    puts val unless val.nil?
+  end
+end
```
keymon: /cc @Jonty 

> I assume this script has been grabbed from somewhere online, would be nice to have the link in the commit message for future reference.

Err, yes and no. I started with https://github.com/alphagov/cf-terraform/blob/master/scripts/val_from_yaml.rb (which I wrote back then), but I re-factored it so much, and I added a spec that I don't know if it is even worth it to refer to it.

concourse/scripts/bosh_create_and_upload_release.rb
```diff
@@ -0,0 +1,78 @@
...
+      puts "Release version already exists locally"
+    else
+      raise e
+    end
+  end
+
+end
+
+def upload_release(name, version, dir)
+  release_filename = "dev_releases/#{name}/#{name}-#{version}.tgz"
```
Jonty: dev_releases is only mentioned here - should it be hardcoded?

concourse/scripts/val_from_yaml.rb
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env ruby
```
Jonty: I assume this script has been grabbed from somewhere online, would be nice to have the link in the commit message for future reference.

concourse/pipelines/create-bosh-cloudfoundry.yml
Jonty: Rather than `cf-push`ing on every run, we could use something like `cf apps | grep graphite-nozzle` to only push the app if it's not already deployed? It's really not good to be pushing on every deploy as it'll restart the app and we'll have metric gaps.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -705,6 +747,111 @@ jobs:
...
+              EOF
+              cf create-security-group graphite-nozzle graphite-nozzle.json
+              cf bind-staging-security-group graphite-nozzle
+              cf bind-running-security-group graphite-nozzle
+
+              cd graphite-nozzle
+
+              # The sleep 1 is to avoid logs being eaten if the configuration is wrong
+              # due to this bug: https://github.com/cloudfoundry/loggregator/issues/86
+              echo "web: sleep 1 && graphite-nozzle" > Procfile
```
Jonty: The `sleep` is no longer needed since we've switched to Diego.

concourse/scripts/bosh_create_and_upload_release.rb
keymon: /cc @Jonty 
> dev_releases is only mentioned here - should it be hardcoded?

`Bosh::Cli::Command::Release::CreateRelease` creates a "development release" unless it is specified as "final". Dev releases are always drop in that directory to be uploaded manually. 

Final releases are not drop in a local directory, they must be uploaded to the blobstore configured in `config/final.yml`

concourse/scripts/val_from_yaml.rb
keymon: /cc @Jonty 

> I assume this script has been grabbed from somewhere online, would be nice to have the link in the commit message for future reference.

Err, yes and no. I started with https://github.com/alphagov/cf-terraform/blob/master/scripts/val_from_yaml.rb (which I wrote back then), but I re-factored it so much, and I added a spec that I don't know if it is even worth it to refer to it.

concourse/pipelines/create-bosh-cloudfoundry.yml
keymon: Sure, I will test that.

Jonty:
I'm happy with this now, although I've not been able to deploy it to test it actually works yet as I'm on other tasks.

jimconner:
All seems to work as advertised. Merging.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -42,6 +44,24 @@ resources:
...
+  - name: graphite-statsd-boshrelease
+    type: git
+    source:
+      uri: https://github.com/alphagov/graphite-statsd-boshrelease.git
+      tag_filter: {{cf_graphite_version}}
+
+  - name: grafana-boshrelease
+    type: git
+    source:
+      uri: https://github.com/keymon/grafana-boshrelease.git
```
alext: Did we mean to leave this pointing at Hector's fork?

manifests/cf-manifest/deployments/055-graphite.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: statsd
+      release: graphite
+    - name: grafana
+      release: grafana
+
+releases:
+- name: graphite
+  version: commit-db49809b0f339a45ae377fc162f01b0ec49c87dc
+- name: grafana
+  version: commit-44564533c9d4d656bdcd5633b808f0bf6fb177ae
```
alext: This tag doesn't exist in the repo referenced in the manifest.

PR #140, opened 2016-03-07
[#113996375] HA for Logsearch

Jonty:
## What

The full write-up is available [on pivotal](https://www.pivotaltracker.com/projects/1275640/stories/113996375).

This PR ensures that we have one instance of each Logsearch component in both Z1 and Z2 with ELBs to balance across each of the relevant services. It also splits ES across Z1, Z2, and Z3, with one master and one data node in each zone.

## How to review

Spin up the cluster, then load kibana on `http://kibana.[$ENV].dev.paas.alphagov.co.uk`. You'll need to add a Kibana filter of `*`, then hit the Discover button. You should see data rolling in.

Head to the EC2 dashboard and kill an Ingestor node, you should still see data coming in. Do the same for a Kibana node and an Elasticsearch node. Everything should keep working.

If you kill a Parser node you'll still be getting data, but 50% of it will be missing unless the Ingestor is also killed. This is documented in the write-up.

## Who can review

Not @jonty or @jimconner, as Jim is on holiday and Jonty is taking one after resolving so many merge conflicts.

### DIFF
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -8,6 +8,17 @@ releases:
   version: 4
   sha1: c92ce357c5e1e77a4fb7944d58a1971c7aee06b0
 
+resource_pools:
+- name: ingestor_z1
+  network: cf1
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: eu-west-1a
+    instance_type: t2.micro
+    elbs:
+      - (( grab terraform_outputs.ingestor_elb_name ))
+
 disk_pools:
 - name: elasticsearch_master
   disk_size: 102400
@@ -91,7 +102,7 @@ jobs:
   templates:
   - {name: ingestor_syslog, release: logsearch}
   - {name: ingestor_relp, release: logsearch}
-  resource_pool: small_z1
+  resource_pool: ingestor_z1
   instances: 1
   networks:
   - name: cf1
--- a/manifests/cf-manifest/deployments/050-logsearch.yml
+++ b/manifests/cf-manifest/deployments/050-logsearch.yml
@@ -10,12 +10,22 @@ releases:
 
 resource_pools:
 - name: ingestor_z1
-  network: cf1
+  network: (( grab resource_pools.medium_z1.network ))
   stemcell: (( grab meta.stemcell ))
   env: (( grab meta.default_env ))
   cloud_properties:
-    availability_zone: eu-west-1a
-    instance_type: t2.micro
+    availability_zone: (( grab resource_pools.medium_z1.cloud_properties.availability_zone ))
+    instance_type: (( grab resource_pools.medium_z1.cloud_properties.instance_type ))
+    elbs:
+      - (( grab terraform_outputs.ingestor_elb_name ))
+
+- name: ingestor_z2
+  network: (( grab resource_pools.medium_z2.network ))
+  stemcell: (( grab meta.stemcell ))
+  env: (( grab meta.default_env ))
+  cloud_properties:
+    availability_zone: (( grab resource_pools.medium_z2.cloud_properties.availability_zone ))
+    instance_type: (( grab resource_pools.medium_z2.cloud_properties.instance_type ))
     elbs:
       - (( grab terraform_outputs.ingestor_elb_name ))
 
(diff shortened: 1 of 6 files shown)


terraform/cloudfoundry/network_acls_cf_in.tf
```diff
@@ -123,7 +123,7 @@ resource "aws_network_acl_rule" "100_ssh" {
     rule_number = 100
     rule_action = "allow"
     from_port = 22
-    to_port = 22
+    to_port = 9300
```
combor: SSH with a port range from 22 to 9300 ?

combor: Ok, I just read commit message. That's terrible.

Jonty: Enormously. Unsure what other choices we have.

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -8,106 +8,190 @@ releases:
...
-    availability_zone: eu-west-1a
-    instance_type: t2.micro
+    availability_zone: (( grab resource_pools.medium_z1.cloud_properties.availability_zone ))
+    instance_type: (( grab resource_pools.medium_z1.cloud_properties.instance_type ))
+    elbs:
+      - (( grab terraform_outputs.elastic_master_elb_name ))
 
-- name: elasticsearch_data
-  network: cf1
+- name: elastic_master_z2
```
dcarley: > This also introduces a master into each zone as it is required for getting quorum if a zone dies.

Are the `allow_master` and `allow_data` options mutually exclusive? Could we combine these into a single job to halve our instance count?

Jonty: To quote from the pivotal writeup:
> It is recommended that in high-volume ES clusters (and in fact any cluster with more than a couple of nodes) you should split the Master-eligable nodes from the Data-eligable nodes as one single Master is elected at any one time and placed under significantly more load. Logsearch does this by default in their upstream release.

terraform/cloudfoundry/network_acls_cf_in.tf
combor: Hm, you are creating `118_ingestor_syslog` as well that allows TCP traffic from infra on ports 2514-5514. We can extend this to be 2514-9300. It's still terrible but a little bit less.

combor: Or....it is possible to raise the rule limit to 40.

Jonty: Personally I'd prefer the limit being raised, but yes changing the syslog one is a much less terrible solution! Good call.

dcarley: We're going to have to open up more than we need, but I don't think we should modify this particular rule because:

- the description of the rule is now misleading
- the port range is huge - it makes some of the other rules redundant and covers most of the <1024 privileged ports
- the source address range covers all of the other subnets (including router and cell), rather than just the infra subnets (where the ELBs currently live)

It might be more appropriate to refactor one or both of the `117_kibana` and `118_ingestor_syslog` rules, which apply to the correct source subnet and have higher `from_port` values.

dcarley: Oh, snap.

combor: Note that in `118_ingestor_syslog` allows TCP traffic from `"${var.infra_cidr_all}"` while  SSH rule allows from 10.0.0.0/16 ( not sure why).

dcarley:
I'm testing this on a branch that has been rebased against `master` to include the merged pipelines in #138. When you get a change could you rebase this PR?

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -8,106 +8,190 @@ releases:
...
+  network: (( grab resource_pools.medium_z1.network ))
```
dcarley: Should this be `resource_pools.medium_z2.network`?

manifests/cf-manifest/deployments/050-logsearch.yml
```diff
@@ -8,106 +8,190 @@ releases:
...
-    instance_type: t2.micro
+    availability_zone: (( grab resource_pools.medium_z2.cloud_properties.availability_zone ))
+    instance_type: (( grab resource_pools.medium_z2.cloud_properties.instance_type ))
+    elbs:
+      - (( grab terraform_outputs.elastic_master_elb_name ))
 
-- name: ingestor
-  network: cf1
+- name: elastic_master_z3
+  network: (( grab resource_pools.medium_z1.network ))
```
dcarley: Should this be `resource_pools.medium_z3.network`?

Jonty: Yup, good catch.

manifests/cf-manifest/deployments/050-logsearch.yml
Jonty: Yup, good catch, again.

terraform/cloudfoundry/routers.tf
```diff
@@ -51,3 +51,93 @@ resource "aws_elb" "ssh-proxy-router" {
...
+    unhealthy_threshold = "${var.health_check_unhealthy}"
+  }
+  listener {
+    instance_port = 9200
+    instance_protocol = "tcp"
+    lb_port = 9200
+    lb_protocol = "tcp"
+  }
+  listener {
+    instance_port = 9300
```
dcarley: Does the ELB need to have the transport port open? I think it's only used by other ES nodes.

https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-transport.html

Jonty:
@dcarley Rebased.

terraform/cloudfoundry/routers.tf
```diff
@@ -51,3 +51,93 @@ resource "aws_elb" "ssh-proxy-router" {
...
+  }
+}
+
+resource "aws_elb" "kibana_elb" {
+  name = "${var.env}-cf-kibana-elb"
+  subnets = ["${split(",", var.infra_subnet_ids)}"]
+  idle_timeout = "${var.elb_idle_timeout}"
+  cross_zone_load_balancing = "true"
+  security_groups = [
+    "${aws_security_group.web.id}",
```
dcarley: Do we want to re-use the `web` security group here? I think it will allow CF apps to access Kibana via NAT, which they probably shouldn't. I'm also concerned that when we want to open up the routers to real tenants during beta then we'll also give them access to other things that they shouldn't (like unuathed Kibana).

terraform/cloudfoundry/routers.tf
```diff
@@ -51,3 +51,93 @@ resource "aws_elb" "ssh-proxy-router" {
...
+    target = "TCP:5601"
+    interval = "${var.health_check_interval}"
+    timeout = "${var.health_check_timeout}"
+    healthy_threshold = "${var.health_check_healthy}"
+    unhealthy_threshold = "${var.health_check_unhealthy}"
+  }
+  listener {
+    instance_port = 5601
+    instance_protocol = "tcp"
+    lb_port = 80
```
dcarley: Should this be HTTPS if we're exposing it externally when it was previously behind an SSH tunnel? Are people able to modify/destroy data via Kibana? Are there any passwords stored in logs?

manifests/cf-manifest/deployments/050-logsearch.yml
dcarley: From meatspace: we've agreed to combine them for now on the basis that:

- we don't want to significantly increase the deploy times for dev environments right now
- we want parity with dev and production environments
- we don't expect *that* much data
- however we do need to turn off debug logging, in a separate story

terraform/cloudfoundry/routers.tf
dcarley: From meatspace: it's significant more work to generate a new cert and upload it to IAM in the pipeline, so we've decided to not introduce the ELB in this story and will continue accessing it using an SSH tunnel. I'm a bit indifferent about whether we keep two Kibana instances or not - Jonty can decide. We'll create another story to replace the SSH tunnels for Logs and Metrics.

dcarley:
Stopped the following and it still seems to be working fine*:
```
/tmp/build/e55deab7 # bosh vms | grep stopped
Acting as user 'admin' on 'my-bosh'
| elasticsearch_data_z1/0 (3d61c47d-bb54-4c4e-9028-711ed58e60d9)            | stopped | n/a | medium_z1         | 10.0.16.26 |
| elasticsearch_master_z1/0 (34e8bfa1-309b-4cdb-8741-e28d93e6fbac)          | stopped | n/a | elastic_master_z1 | 10.0.16.10 |
| ingestor_z1/0 (00e981e8-904f-4853-aecc-9a00500c5a6a)                      | stopped | n/a | ingestor_z1       | 10.0.16.12 |
```

* the histogram with auto-refresh enabled is still showing the same amount of data

Jonty:
Have resolved all comments, and pulled the Kibana HA work into [a different branch](https://github.com/alphagov/paas-cf/tree/make_kibana_ha) to be merged when it has been SSL enabled.

dcarley:
Comments resolved and re-tested. As mentioned in Slack, I've added three commits for something that Jonty ran out of time to address - I'd like someone else to look at them (not necessarily test) and confirm that they make sense before I merge this.

dcarley:
> Jonty Wareing [16:27] 
> dcarley: They look good to me!

PR #142, opened 2016-03-08
[#113159997] Generate NATS passwords

saliceti:
## What
Story: [Generate mbus passwords for Deployer Concourse and BOSH](https://www.pivotaltracker.com/story/show/113159997)

BOSH and Concourse were using default passwords for NATS. Now the passwords are generated and used in the manifests.

Concourse manifest files are slightly refactored to look more like bosh and CF.

## How to review
Build a new environment, with focus on building concourse deployer and microbosh. It should all work and build cloud foundry.
Check in the generated manifests that they don't use default passwords.

## Who can review
Anyone but me


actionjack:
:+1: LGTM

PR #145, opened 2016-03-09
[#115322155] Use correct version of fly for Bootstrap/Deployer Concourse

alext:
## What

We should use the correct version of fly for each Concourse. Assuming that the Bootstrap Concourse and Deployer Concourse will be using different versions, either now or again in the future.

As part of this we've taken the opportunity to refactor how the environment is setup for these scripts. See commit messages for more details.

## How to review

Use this branch to push pipelines to a bootstrap concourse (`./vagrant/deploy.sh`). Push pipelines to the deployer concourse (`make dev`). This should create and use 2 different fly commands in the `bin` directory. This should not touch any `fly` binaries that happen to be on your path.

Check that the `bosh-cli` script works correctly.

## Who can review

Anyone but @timmow or myself.


Makefile
```diff
@@ -61,5 +61,4 @@ disable_auto_delete:
 	$(eval export DISABLE_AUTODELETE=1)
 
 deploy_pipelines:
-	concourse/scripts/pipelines-bosh-cloudfoundry.sh
-	concourse/scripts/pipelines-failure-testing.sh
+	concourse/scripts/push-deployer-pipelines.sh
```
saliceti: I'd rather avoid creating another wrapper script if possible. I've made a simple attempt in branch fly_versions_2, isn't it good enough?

concourse/scripts/fly_sync_and_login.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "Downloading fly command..."
+  curl "$FLY_CMD_URL" -L -f -k -o "$FLY_CMD" -u "${CONCOURSE_ATC_USER}:${CONCOURSE_ATC_PASSWORD}"
+  chmod +x "$FLY_CMD"
+fi
+
+echo "Doing fly login & sync"
+echo -e "${CONCOURSE_ATC_USER}\n${CONCOURSE_ATC_PASSWORD}" | \
+  $FLY_CMD -t "${FLY_TARGET}" login -k --concourse-url "${CONCOURSE_URL}"
+
+$FLY_CMD -t "${FLY_TARGET}" sync
```
saliceti: This always downloads the fly binary. Can we add an option like FLY_SYNC=false to disable it when we're developing?

alext: Good call. Added.

dcarley: If everyone starts using this habitually, are we going to be bitten when we next upgrade Concourse?

How about checking the modified time on the file and only updating if it's N hours/days old?

alext: > How about checking the modified time on the file and only updating if it's N hours/days old?

I like it.  I'll look at doing that.  What do you think is a reasonable value for `N`?

dcarley: A day?

alext: I've now implemented this, and rebased.

Makefile
alext: This has now been refactored so that is has a single environment script.

alext:
I think I've now addressed all the comments here. Ready to be re-tested.

saliceti:
Tested OK
Thanks! 👍

PR #146, opened 2016-03-09
[#113749465] Fix urls of the graphite and grafana bosh releases

keymon:
What?
-----

The bosh releases for graphite and grafana had wrong urls merged in #139.

We want them to point to the alphagov github forks.

How to test?
------------

Check that the resources point to the right repo and are able to pull the code

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -47,13 +47,13 @@ resources:
   - name: graphite-statsd-boshrelease
     type: git
     source:
-      uri: https://github.com/alphagov/graphite-statsd-boshrelease.git
+      uri: https://github.com/alphagov/paas-graphite-statsd-boshrelease.git
       tag_filter: {{cf_graphite_version}}
 
   - name: grafana-boshrelease
     type: git
     source:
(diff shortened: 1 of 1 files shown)


alext:
There are currently no tags in https://github.com/alphagov/paas-grafana-boshrelease, so that `tag_filter` on the grafana release will mean that it's constantly waiting. Do you need to push the tags?

keymon:
@alext tag added

Actually I know what happened. I cloned to my personal space because there was nobody to clone it in alphagov. My personal one had the tag pushed.

alext:
The tag doesn't match the version in the manifest (https://github.com/alphagov/paas-cf/blob/113749465_fix_graphite_git_resource_urls/manifests/cf-manifest/deployments/055-graphite.yml#L18)

The tag is `commit-3c7023ab15b159996f427e4cdb72a5ec9bd62bca`, whereas the manifest has `commit-44564533c9d4d656bdcd5633b808f0bf6fb177ae`.

keymon:
:sob: 

Can you try now? I added the other tag.

The commit-3c7023ab15b159996f427e4cdb72a5ec9bd62bca includes the feature of configure the datasource, which is handy.

PR #150, opened 2016-03-10
[#114308825] Enable RDS secret rotation

mtekel:
## What

Enable updating of RDS passwords, so that we can rotate these secrets. Previously 'api' and 'uaa' roles were only created once and never changed aftewards, which skipped updating passwords if you have changed them in secrets.

## How to review

Deploy. Update `cf_db_api_password` and `cf_db_uaa_password` secrets (edit cf-secrets.yml, upload to bucket, apply - only UAA and CC jobs (cc, worker, clock) should be updated). Run smoke and acceptance tests. All should keep working.

## Who can review

not @mtekel or @combor


saliceti:
Tested successfully on new environment, and changing passwords on existing environment.

PR #151, opened 2016-03-10
[#114308825] Enable NATS password rotation

mtekel:
## What

Enable updating of NATS password, so that we can rotate this secret. Do this by ensuring that the NATS jobs get updated (and deployed) concurrently with ETCD. This lets ETCD jobs connect to NATS using the new password. Otherwise they get stuck waiting for NATS to be updated, but which was configured to update after ETCD has finished updating.

## How to review

Deploy. Update `nats_password` (edit cf-secrets.yml, upload to bucket, apply). Run smoke and acceptance tests. All should keep working.

## Who can review

not @mtekel or @combor

### DIFF
--- a/manifests/cf-manifest/spec/jobs_spec.rb
+++ b/manifests/cf-manifest/spec/jobs_spec.rb
@@ -1,0 +1,71 @@
+RSpec.describe "the global update block" do
+
+  let(:manifest) { manifest_with_defaults }
+
+  describe "in order to run parallel deployment by default" do
+    it "has serial false" do
+      expect(manifest["update"]["serial"]).to be false
+    end
+  end
+
+end
+
+RSpec.describe "the jobs definitions block" do
+
+  let(:jobs) { manifest_with_defaults["jobs"] }
+
+  def get_job(job_name)
+    jobs.select{ |j| j["name"] == job_name}.first
+  end
+
+  def is_serial(job_name)
+    job = get_job(job_name)
+    job["update"]["serial"]
+  end
+
+  def ordered(job1_name, job2_name)
+    i1 = jobs.index{ |j| j["name"] == job1_name }
+    i2 = jobs.index{ |j| j["name"] == job2_name }
+    i1 < i2
+  end
+
+  describe "in order to enforce etcd dependency on NATS" do
+    it "has etcd_z1 serial" do
+      expect(is_serial("etcd_z1")).to be true
+    end
+
+    it "has nats_z1 before etcd_z1" do
+      expect(ordered("nats_z1", "etcd_z1")).to be true
+    end
+
+    it "has nats_z2 before etcd_z1" do
+      expect(ordered("nats_z2", "etcd_z1")).to be true
+    end
+  end
+
+  describe "in order to start one etcd master for consensus" do
+    it "has etcd_z1 serial" do
+      expect(is_serial("etcd_z1")).to be true
+    end
+
+    it "has etcd_z1 before etcd_z2" do
+      expect(ordered("etcd_z1", "etcd_z2")).to be true
+    end
+
+    it "has etcd_z1 before etcd_z3" do
+      expect(ordered("etcd_z1", "etcd_z3")).to be true
+    end
+  end
+
+  describe "in order to start one consul master for consensus" do
+    it "has consul_z1 serial" do
+      expect(is_serial("consul_z1")).to be true
+    end
+
+    specify "has consul_z1 first" do
+      expect(jobs[0]["name"]).to eq("consul_z1")
+    end
+  end
+
+
+end
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -243,6 +243,28 @@ jobs:
       metron_agent:
         zone: z2
 
+  - name: nats_z1
+    templates: (( grab meta.nats_templates ))
+    instances: 1
+    resource_pool: medium_z1
+    networks:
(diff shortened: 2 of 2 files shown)


dcarley:
The code changes for this seem subtle and like we could break this in the future without meaning to. Do you have any ideas how we might be able to have an automated test for this? /cc @bleach

saliceti:
We could add tests for password rotation?

mtekel:
I think the only real test is to rotate passwords and test if all is fine. We can perhaps create a CI pipeline for this. The catch with testing in other way is that different CF releases can change how components behave. E.g. when ETCD fixes their bugs and makes new release, and when diego supports that release, we no longer depend on NATS being updated before ETCD. However, that might introduce other dependencies.

The problem in general is that these dependencies are implicit and not clearly defined/described anywhere. There was a story to fix up the way we do updates/deploy so that we don't create outage on production. Now another factor could be not to create outage during password rotation. This all impacts how we arrange updates and deployments and I think this can be changing almost every release...

bleach:
We had an in-person conversation about it:

- since we've changed the ordering within the manifests, it's necessary to test that we can still deploy a new CF rather than changing an existing one
- for now we're thinking that a spec that checks the ordering of jobs would be a reasonable interim solution and one we might want to keep for fast feedback even if we have an automated test
- the spec alone won't give us feedback if things stop working if some other behaviour changes eg. the way BOSH deploys or a component retries
- so we agree that it's probably a good idea to have an automated test that changes the passwords and makes sure everything is OK, it's one for another story though

dcarley:
:+1: 

Out of interest, do we need to test the order (NATs somewhere before etcd) or proximity (NATs immediately before etcd)? Is there a difference? Does the `serial: true` on etcd matter? I thought that meant that it will deploy NATs and then etcd, rather than both concurrently (mentioned in the commit message).

mtekel:
Updated commit message. As this is a spike I did not want to add anything more to the PR. I think we should have a story to do proper testing. It's not only about order (or, the order itself is not enough to guarantee one component updating before other), but also about `serial: true` definition in jobs. Essentially, you need to simulate bosh update logic AND define the right order of updates - which can change between releases, due to new/changed dependencies.

saliceti:
I've pushed a commit with manifest validation tests for this use case. I've included consul as well since it's a similar issue.

This should be reviewed in the future as the current issues may be fixed and we may not require this ordering.

@dcarley and @mtekel have a look and let me know what you think.

mtekel:
This test is fine for this PR. What I had in mind was some more generic solution, where we specify update dependencies and the test checks if manifest complies. But that's for a different story.

manifests/cf-manifest/spec/jobs_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    job["update"]["serial"]
+  end
+
+  def ordered(job1_name, job2_name)
+    i1 = jobs.index{ |j| j["name"] == job1_name }
+    i2 = jobs.index{ |j| j["name"] == job2_name }
+    i1 < i2
+  end
+
+  specify "has etcd_z1 serial" do
```
dcarley: It'd be good to group some of these related tests together and describe the reason that we care about the specific behaviour, e.g.
```ruby
describe "etcd depends on NATS" do
  specify "has nats_z1 before etcd_z1" do
    expect(ordered("nats_z1", "etcd_z1")).to be true
  end

  specify "has nats_z2 before etcd_z1" do
    expect(ordered("nats_z2", "etcd_z1")).to be true
  end
end

describe "etcd needs to start with one master node for concensus" do
  specify "has etcd_z1 serial" do
    expect(is_serial("etcd_z1")).to be true
  end

  specify "has etcd_z1 before etcd_z2" do
    expect(ordered("etcd_z1", "etcd_z2")).to be true
  end

  specify "has etcd_z1 before etcd_z3" do
    expect(ordered("etcd_z1", "etcd_z3")).to be true
  end
end
```

saliceti:
I have now made the grouping so it makes natural sentences in rspec output:

```
the jobs definitions block
  in order to start one consul master for consensus
    has consul_z1 first
```

manifests/cf-manifest/spec/jobs_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    job["update"]["serial"]
+  end
+
+  def ordered(job1_name, job2_name)
+    i1 = jobs.index{ |j| j["name"] == job1_name }
+    i2 = jobs.index{ |j| j["name"] == job2_name }
+    i1 < i2
+  end
+
+  describe "in order to control jobs order relative to etcd_z1" do
```
dcarley: From meatspace: we'll duplicate this in each of the "in order to" blocks, because it belongs to both and doesn't currently describe why we care about `serial: true`.

manifests/cf-manifest/spec/jobs_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it "has etcd_z1 before etcd_z2" do
+      expect(ordered("etcd_z1", "etcd_z2")).to be true
+    end
+
+    it "has etcd_z1 before etcd_z3" do
+      expect(ordered("etcd_z1", "etcd_z3")).to be true
+    end
+  end
+
+  describe "in order to control job order relative to consul_z1" do
```
dcarley: Ditto https://github.com/alphagov/paas-cf/pull/151/files#r55987243

saliceti:
Updated specs.

dcarley:
Sorry for the delay, was waiting for smoke tests. Thanks @mtekel @saliceti :+1:

PR #152, opened 2016-03-10
[#115403129] Build from tag in staging and production

saliceti:
## What
Story: [Enable build from tag option](https://www.pivotaltracker.com/story/show/115403129)

Adds the tag_filter option on the paas-cf git resource in all the pipelines and creates predefined filters for staging and production.
This is controlled via make commands or can be set via the variable `$PAAS_CF_TAG_FILTER`

## How to review
* Build an environment, including bootstrap with different values for variable `$PAAS_CF_TAG_FILTER`
* Empty variable means no filter and all commits will pass
* To test for staging and prod, you can execute the targets `set_stage_tag_filter` or `set_prod_tag_filter` within the `dev` target.
* Push tags ([annotated tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging#Annotated-Tags) preferred). Patterns are [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html) compatible
* Observe the paas-cf resource only fetches the required commits
* Observe the bosh-cf pipeline is triggered

## Who can review
Anyone but me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck set_aws_count set_auto_trigger disable_auto_delete check-env-vars dev ci stage prod
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck set_aws_count set_auto_trigger disable_auto_delete check-env-vars
 
 .DEFAULT_GOAL := help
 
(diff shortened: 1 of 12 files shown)


bleach:
I'm unable to review this at the moment due to us having reached the max number of ELBs in the dev account. (We're going to fix that, but it might take a while).

Makefile
```diff
@@ -36,11 +36,22 @@ lint_shellcheck:
 
 dev: check-env-vars set_aws_account_dev deploy_pipelines ## Deploy Pipelines to Dev Environment
 
+dev-bootstrap: check-env-vars set_aws_account_dev vagrant-deploy
```
combor: This Makefile has auto-documenting feature. `help` target parses target comment and generates help message. You are missing comments on the new targets.

Makefile
```diff
@@ -36,11 +36,22 @@ lint_shellcheck:
...
+
 ci: check-env-vars set_aws_account_ci set_auto_trigger disable_auto_delete deploy_pipelines  ## Deploy Pipelines to CI Environment
 
-stage: check-env-vars set_aws_account_stage disable_auto_delete set_auto_trigger deploy_pipelines  ## Deploy Pipelines to Staging Environment
+ci-bootstrap: check-env-vars set_aws_account_ci vagrant-deploy
```
combor: Do we want to add all new targets to `.PHONY` ?

combor:
+1

dcarley:
We should update the README to refer to the new bootstrap targets in the `Makefile`.

saliceti:
Comments addressed

dcarley:
Tested OK. I'll have look over the updated commits on Monday :+1:

README.md
```diff
@@ -62,11 +62,12 @@ state files).
 
 ### Deploy
 
-Run the following script with the name of your environment:
+Create the bootstrap Concourse with `make`. Select the target based on which AWS account you want to work with. For instance for a DEV bootstrap:
 
 ```
-./vagrant/deploy.sh <deploy_env>
+make dev-bootstrap
```
alext: There's no mention of setting `DEPLOY_ENV` prior to this, which is required when using the make target. It would be good to cover this detail.

saliceti: Rebased and updated.

dcarley:
:+1:

PR #153, opened 2016-03-11
Readme update

tobwilk:
## What

While deploying the PaaS environment I had to ask colleges a number of questions to get it running. I have added some material into the readme that would have helped me during the process.

## How to review

its plaintext markdown, just have a skim and see what you think.

## Who can review

Anyone who understands how the deployment process works and can write good readme files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ deployment.
 ### Connecting to your environment
 
 Your PaaS environment is accessed via the cloudfoundry command line tool. This can be
-downloaded from hhttps://github.com/cloudfoundry/cli/releases. The cloudfoundry
+downloaded from https://github.com/cloudfoundry/cli/releases. The cloudfoundry
 web interface is disabled.
 
 Log into your newly configured environment by running the `cf` command:
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ The password will be different to the ones used previously. Ask a friendly colle
 where you can find it.
 
 If you do not have valid certificates for your PaaS environment you can use the
-`--skip-ssl-validation` switch to bypass the warnings`.
+`--skip-ssl-validation` switch to bypass the warnings.
 
 
 # Additional notes
--- a/README.md
+++ b/README.md
@@ -191,6 +191,9 @@ where you can find it.
 If you do not have valid certificates for your PaaS environment you can use the
 `--skip-ssl-validation` switch to bypass the warnings.
 
+Also to note, it takes a while for deployer-concourse to create your environment.
+You can still log in and make use of it while the tests take place.
+
 
 # Additional notes
 
--- a/README.md
+++ b/README.md
@@ -169,11 +169,11 @@ NB: If the `destroy-microbosh` pipeline is run without first cleaning up
 CloudFoundry, it will be necessary to manually clean up the CloudFoundry
 deployment.
 
-### Connecting to your environment.
+### Connecting to your environment
 
-PaaS currently is accessed via the cloudfoundry command line tool which can be
-downloaded from https://github.com/cloudfoundry/cli. PaaS currently has the web
-UI disabled.
+Your PaaS environment is accessed via the cloudfoundry command line tool. This can be
+downloaded from hhttps://github.com/cloudfoundry/cli/releases. The cloudfoundry
+web interface is disabled.
 
 Log into your newly configured environment by running the `cf` command:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,31 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
...
+The high level process for deploying the PaaS environment is as follows. For a
+step by step guide, follow the more detailed instructions further down.
+
+1. Use the vagrant AWS plugin to deploy the "bootstrap-concourse" into your
+AWS environment. Once its deployed, vagrant will port forward access to the
+web UI for the bootstrap concourse to localhost over SSH. Use this to log into
+the bootstrap concourse UI from your web browser.
+
+2. Inside the bootstrap-concourse web UI, there are a number of 'pipelines' that
+you may run. Specifically, the `create-deplyer` pipeline is used that will then
```
mtekel: typo: create-deployer

README.md
```diff
@@ -12,7 +12,31 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
...
+create a new "deployer-concourse" into that AWS environment. Once this is complete,
+you will be given the URL, username and password you can use to log into the
```
mtekel: This is a bit confusing/misleading. You know the URL and PW upfront. Pipeline/bootstrap concourse doesn't give you anything when it finishes running pipeline.

README.md
```diff
@@ -12,7 +12,31 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
...
+'deployer concourse' web UI.
+
+3. Through the "deployet-concourse" Web UI, a number of new pipelines exist that
```
mtekel: Just a note here: the pipelines don't exist, you need to `make dev` to upload them.

README.md
```diff
@@ -12,7 +12,31 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
...
+you can run. Use the "create-bosh-cloudfoundry" pipeline to deploy BOSH and the
+cloudfoundry environments into the AWS environment.
+
+Note: Overnight, both the "bootstrap-concourse" and the cloudfoundry environment
+will auto delete itself to save unnecessary AWS costs. You will need to re running the
+"create-bosh-cloudfoundry" pipeline each day.
```
dcarley: I'm unsure about the additional level of detail in this section (from the top of this line). It duplicates the information further down the README and we've had trouble in the past keeping duplicated information accurate (we did some refactoring in #71 and #74). I'm interested in other peoples opinions on this though.

README.md
```diff
@@ -141,6 +169,32 @@ NB: If the `destroy-microbosh` pipeline is run without first cleaning up
 CloudFoundry, it will be necessary to manually clean up the CloudFoundry
 deployment.
 
+### Connecting to your environment
```
dcarley: This is a really good section to include. It can be easy to forget that the end goal is making Cloud Foundry work!

It might be worth renaming the title though to something like "Connecting to Cloud Foundry" or "Connecting to Cloud Foundry in your environment".

README.md
```diff
@@ -12,7 +12,31 @@ and [BOSH][] manifests that allow provisioning of [CloudFoundry][] on AWS.
 
 ## Overview
 
-The following components needs to be deployed in order. They should be
+The high level process for deploying the PaaS environment is as follows. For a
+step by step guide, follow the more detailed instructions further down.
+
+1. Use the vagrant AWS plugin to deploy the "bootstrap-concourse" into your
+AWS environment. Once its deployed, vagrant will port forward access to the
```
saliceti: Once *it's* deployed (or once deployed)

README.md
saliceti: Yes we shouldn't duplicate. Or just link to [Overnight deletion of environments](https://github.com/tobwilk/paas-cf/blob/readme-update/README.md#overnight-deletion-of-environments)

saliceti:
Good doc, thanks!

README.md
```diff
@@ -141,6 +169,32 @@ NB: If the `destroy-microbosh` pipeline is run without first cleaning up
...
+
+```
+cf -a <url of environment> -u admin
+```
+
+The url will be the same that was given to you when you created the deployer-concourse,
+the only difference being the subdomain will be "http://api.yourdomain" instead of
+"http://deployer.yourdomain.
+
+The password will be different to the ones used previously. Ask a friendly colleague
```
keymon: We should not rely on asking people. 

We should either say that the password is in the `cf-secrets.yml`. Or we should provide a generic way (script or something) to retrieve it.

keymon:
Can we rebase this to: 
 * Merge commits that fix grammar.
 * Change the author to the one that must be used.

keymon:
What shall we do with this PR? It has conflicts and several comments to fix. There is not a card related, so we are neglecting merge it.

dcarley:
I've extracted and refactored the section about accessing Cloud Foundry in #203. I think there's consensus that the changes to "overview" duplicate other sections of the README and we should leave it as-is for now. Sorry for taking so long to get round to this. Thanks for taking the time do this, Toby.

PR #156, opened 2016-03-14
[#114000923] Implement self-update pipelines for microbosh and cloudfoundry

keymon:
[#114000923 self-updating pipelines](https://www.pivotaltracker.com/story/show/114000923)

What?
-----

In order to ensure that we promote changes from paas-cf into environments in a consistent way, we need to ensure that changes to concourse pipelines are applied automatically.

To do so, we will add a initial task in each pipeline to update the pipeline itself using the code from the repository.

Notes:

 - The "environment class" specific variables will be set in `Makefile` file, with some sensible defaults defined in code.  In `dev` environments some parameters can be optionally overwritten using environment variables.

 - Self-update will be enabled by default. 

   To disable it export the variable `SELF_UPDATE_PIPELINES=false make dev` when uploading the pipleines. This way one can apply changes without having to commit and push the code.

 - We won't add logic to lock the pipeline as it is out of the scope for this story.

 - Following the Boy Scouts rule, we try simplify the environment class variable definition in the Makefile.

Dependencies
---------------------

 - docker container for https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/35  must be merged before this one.
 - The previous container must be created in government-paas dockerhub account and the reference changed in this PR. (https://hub.docker.com/r/governmentpaas/self-update-pipelines/)

How to test this?
---------------------

Deploy the environment, but enabling self-update (which is disabled in dev by default), and commit some change to the pipeline. 

I recommend use a temporary branch for this, and pause the job after `init`:

```
git checkout 114000923_self_updating_pipelines_v2
git checkout -b 114000923_self_updating_pipelines_review_$(USER) # Create a temporary branch

BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev
```

Deploy the environment and observe that in the job `init` the self-update pipeline works and do not do any change.

Then add some changes in the pipeline and commit and push the changes. Rerunning the `init` job should show that the pipeline is updated.

You can do the same for:
 - `destroy-microbosh`, `destroy-cloudfoundry` and `failure-testing`
 - using `ci`, `staging` and `prod` targets (in that case export the variable `URSE_URL=https://deployer.hector.dev.paas.alphagov.co.uk`)

Who?
------

Anyone but @saliceti, @keymon or @HenryTK


dcarley:
Sorry, will have merge conflicts with the recently merged #152

concourse/pipelines/destroy-cloudfoundry.yml
```diff
@@ -47,10 +47,48 @@ resources:
...
+                if [ "$SELF_UPDATE_PIPELINE" != "true" ]; then
+                  echo "Self update pipeline is disabled (set SELF_UPDATE_PIPELINE=true to enable)"
+                else
+                  echo "Self update pipeline is enabled (set SELF_UPDATE_PIPELINE=false to disable)"
+
+                  VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+                  export CONCOURSE_ATC_PASSWORD=$($VAL_FROM_YAML jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
+
+                  make -C ./paas-cf ${MAKEFILE_ENV_TARGET}
+                fi
```
dcarley: Can we move this to a script in `concourse/scripts/` to reduce duplication in each pipeline and make it easier to maintain? When we do, it should have `set -eu` at the top of it to abort on errors and unbound variables.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -18,6 +19,9 @@ cf_grafana_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.grafana.version "
 generate_vars_file() {
    cat <<EOF
 ---
+pipeline_name: ${pipeline_name}
+makefile_env_target: ${MAKEFILE_ENV_TARGET:-dev}
+self_update_pipeline: ${SELF_UPDATE_PIPELINE:-false}
```
dcarley: Should this default to `true` and be explicitly disabled for the dev environments in the `Makefile`?

> Self-update will be disabled by default in the dev environments, as we want to be able to apply changes without having to commit and push the code.

This is definitely a use case that we need to allow. Should it be opt-in though?

It's not very clear to newcomers which changes they can make with/without pushing a branch. For people that know what they're doing and want to speed up testing they could do: `SELF_UPDATE_PIPELINE=false make dev`

dcarley:
This is really good. Much cleaner that I could have possibly hoped for. Thanks @keymon :+1: :+1:

concourse/pipelines/destroy-cloudfoundry.yml
keymon: Moved to a new script

concourse/scripts/pipelines-bosh-cloudfoundry.sh
keymon: Now self update is enabled by default in all environments, unless you specify `SELF_UPDATE_PIPELINE=false`.

Shall I add a note in the README?

keymon:
@dcarley rebased with master and added the suggested changes.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
dcarley: I don't mind either way. I'll leave you to decide, or if anyone else has an opinion.

keymon: The script itself prints how to disable the feature, so I will not put it in the README.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -160,6 +167,25 @@ jobs:
...
+          params:
+            DEPLOY_ENV: {{deploy_env}}
+            BRANCH: {{branch_name}}
+            MAKEFILE_ENV_TARGET: {{makefile_env_target}}
+            SELF_UPDATE_PIPELINE: {{self_update_pipeline}}
+            PIPELINES_TO_UPDATE: {{pipeline_name}}
+          run:
+            path: bash
+            args:
+              - ./paas-cf/concourse/scripts/self-update-pipeline.sh
```
alext: Any reason to run this via bash here as opposed to just calling the script directly?

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -8,7 +8,8 @@ export TARGET_CONCOURSE=deployer
 $("${SCRIPT_DIR}/environment.sh" "$@")
 "${SCRIPT_DIR}/fly_sync_and_login.sh"
 
-env=${DEPLOY_ENV}
+env="${DEPLOY_ENV}"
+pipelines_to_update="${PIPELINES_TO_UPDATE:-create-bosh-cloudfoundry destroy-microbosh destroy-cloudfoundry autodelete-cloudfoundry failure-testing}"
```
alext: I think `destroy-cloudfoundry` should come before `destroy-microbosh` as that's the order they should be run in.

concourse/pipelines/create-bosh-cloudfoundry.yml
keymon: No reason really, I will change that

concourse/scripts/self-update-pipeline.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# Optional variables:
+# - BRANCH
+# - PIPELINES_TO_UPDATE
+
+set -u
+set -e
+
+if [ ! -d "./paas-cf" -o ! -f "concourse-manifest/concourse-manifest.yml" ]; then
+	echo "Resources paas-cf and concourse-manifest must be checkout"
```
alext: The indentation here is inconsistent with other scripts (we're using 2 spaces elsewhere).

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -39,35 +43,40 @@ generate_manifest_file() {
...
+	"${env}" "${pipeline_name}" \
+	<(generate_manifest_file) \
+	<(generate_vars_file)
+    ;;
+    autodelete-cloudfoundry)
+      if [ ! "${DISABLE_AUTODELETE:-}" ]; then
+	bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
+		"${env}" "${pipeline_name}" \
+	  "${SCRIPT_DIR}/../pipelines/${pipeline_name}.yml" \
+	  <(generate_vars_file)
```
alext: The indentation is a little odd here as well. There's a mixture of tabs and spaces, which makes it a little hard to read.

concourse/pipelines/create-bosh-cloudfoundry.yml
keymon: Actually my motivations where that it is easier to add arguments using this structure, and it is obvious it is a shell script. But I will change it anyway.

terraform/concourse/elb.tf
```diff
@@ -47,7 +47,10 @@ resource "aws_security_group" "concourse-elb" {
     from_port       = 443
     to_port         = 443
     protocol        = "tcp"
-    cidr_blocks     = ["${split(",", var.office_cidrs)}"]
+    cidr_blocks     = [
+      "${split(",", var.office_cidrs)}",
+      "${aws_eip.concourse.public_ip}/32"
+    ]
```
alext: This is triggering the "Terraform diffs didn't match" bug when creating an environment from scratch (ie when the EIP is created in the same terraform run as this security group).  See https://github.com/alphagov/paas-cf/pull/123 for a workaround.

PR #157, opened 2016-03-15
[#114380365] Use Correct Domains for Environments

alext:
## What

Now that we have the real domains setup and DNS delegated, we want to use them everywhere.

While looking at this, I also noticed that we were using the wrong attribute for the concourse DNS name, resulting in a trailing `.` in the URLS.

## How to review

Deploy an environment, and verify that everything is using a `cloudpipeline.digital` or `cloudpipelineapps.digital` domain.

Updating an existing environment should also work, although it will be necessary to delete the certs stored in the state bucket so that new certs will be generated for the new domains.
There's also sometimes a known issue with updating the cert for the concourse ELB, so it may be necessary to manually delete this ELB first.

## Who can review

Anyone but myself.


keymon:
Note: This PR will mean a lot of changes in the existing deployments. Once it is merged into master, it is recommended that everybody destroys all the environment and redeploys again.

If you don't want to destroy everything, this is the procedure to update your environment with the minimum changes:
 1. Remove the old certificates for concourse and CF in your state bucket:
     ```
aws s3 rm s3://${DEPLOY_ENV}-state/concourse-cert.tar.gz --region eu-west-1 
aws s3 rm s3://${DEPLOY_ENV}-state/cf-certs.tar.gz --region eu-west-1 
```
 2. Redeploy the "deployer" concourse:
     1. Run `make dev-bootstrap`
     2. Manually change the certificate in the concourse ELB to something other certificate. This is because terraform cannot update the certificate if it is being used by a ELB. Terraform will update the ELB to the certificate when it runs.

 3. Pause the acceptance tests, and redeploy CF.

 4. Before run acceptance tests:
     1. Remove the old domain: `cf delete-shared-domain hector.dev-apps.paas.alphagov.co.uk`, if not the `services suite` will fail because the test service broker helper does not set the domain (fixed [here](https://github.com/cloudfoundry/cf-acceptance-tests/commit/2f680ef4d619fad7eb0cf6155ed36ecdfc3be918#diff-9b7c5136c43c607c6a6b8f5b15d96834L113)).
     2. Remove the application `CATS-persistent-app` from `CATS-persistent-org:CATS-persistent-space`. This app is not removed between tests and will have the old domain:
        ```
cf target -o CATS-persistent-org -s CATS-persistent-space
cf delete CATS-persistent-app 
```

keymon:
I tested the change in dev and LGTM, although it will break everybody's environment as described.

I will merge as I get green lights.

PR #158, opened 2016-03-15
[#115251665] ELB for access to metrics

actionjack:
[ELB for access to metrics](https://www.pivotaltracker.com/story/show/115251665)

## What

Currently to access metrics we are using SSH tunnels. We should change this to allow access via ELB over HTTPS and limited to whitelisted IP ranges.

## How to Review This PR

This PR has been written with the following narrative:

* I would like to:
	* I create a new security group for:
     * Grafana and;
     * Graphite
 	* Create a new ELB for 
     * Grafana listening on port 443 and forwarding to port 3000 on all registered grafana nodes;
     * Graphite listening on port 3001 and forwarding to port 3001 on all registered grafana/graphite nodes;
     * Create Terraform outputs for graphana so I can use them in cloud foundry manifests
     * Fix ACL port range on rule 118 to allow ELB's to talk to grafana and graphite node.
     * Create new resources for metrics-certificates and vpc terraform state
			* The metrics-cert is a resource for the metrics certificates
			* The vpc-state is to get the terraform variables for the VPC in order
  to correctly name the certificate
     * Generate and use metrics SSL certificates

## How to test this PR

* Deploy a new environment
* For Grafana:
  * Browse to https://metrics.$DEPLOY_ENV.dev.paas.alphagov.co.uk 
* For Graphite:
  * Browse to https://metrics.$DEPLOY_ENV.dev.paas.alphagov.co.uk:3001 

  
## Who can merge this PR

[Don't cry because this story is over. Smile because it happened.](http://www.brainyquote.com/quotes/authors/d/dr_seuss.html) Though @jimconner, @actionjack and @henrytk aren't allowed to

PR #159, opened 2016-03-15
[#115318745] Ability to deploy without environment name

combor:
## What

[Story link: https://www.pivotaltracker.com/story/show/115318745](https://www.pivotaltracker.com/story/show/115318745)
We want to be able to deploy CF environment without a specific development name.
In staging and production we don't have a need for the "environment name" because there will only be one environment under that "domain name". We need to optionally omit that "thing" from the naming, so that we can create DNS entries such as:
- api.cloud.service.gov.uk
- api.staging.cloudpipeline.digital

And not:
- api.production.cloud.service.gov.uk
- api.staging.staging.cloudpipeline.digital

## How to review

```
BRANCH=no-env-name-115318745 make dev
```
Deploy BOSH and CF (smoke test should pass ) and test destruction as well.
If you are not following master branch closely and your env is pre https://github.com/alphagov/paas-cf/pull/157 you will have to start from scratch and create deployer concourse as well. 
If you are FTA please test stage and prod as well.

## Who can review

Anyone but @combor or @keymon

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,9 @@ spec:
 lint_yaml:
 	$(YAMLLINT) -c yamllint.yml .
 
-lint_terraform:
+lint_terraform: set_env_class_dev
+	$(eval export TF_VAR_system_dns_zone_name=$SYSTEM_DNS_ZONE_NAME)
+	$(eval export TF_VAR_apps_dns_zone_name=$APPS_DNS_ZONE_NAME)
 	find terraform -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 -n 1 -t terraform graph > /dev/null
 
 lint_shellcheck:
@@ -64,6 +66,8 @@ prod-bootstrap: check-env-vars set_env_class_prod vagrant-deploy  ## Start Produ
 set_env_class_dev:
 	$(eval export MAKEFILE_ENV_TARGET=dev)
 	$(eval export AWS_ACCOUNT=dev)
+	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
+	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
 
 .PHONY: set_env_class_ci
 set_env_class_ci:
@@ -72,6 +76,8 @@ set_env_class_ci:
 	$(eval export ENABLE_AUTO_DEPLOY=true)
 	$(eval export DISABLE_AUTODELETE=1)
 	$(eval export TAG_PREFIX=stage-)
+	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.ci.cloudpipeline.digital)
+	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.ci.cloudpipelineapps.digital)
 
 .PHONY: set_env_class_stage
 set_env_class_stage:
@@ -81,6 +87,8 @@ set_env_class_stage:
 	$(eval export DISABLE_AUTODELETE=1)
 	$(eval export PAAS_CF_TAG_FILTER=stage-*)
 	$(eval export TAG_PREFIX=prod-)
+	$(eval export SYSTEM_DNS_ZONE_NAME=staging.cloudpipeline.digital)
+	$(eval export APPS_DNS_ZONE_NAME=staging.cloudpipelineapps.digital)
 
 .PHONY: set_env_class_prod
 set_env_class_prod:
@@ -88,6 +96,8 @@ set_env_class_prod:
 	$(eval export AWS_ACCOUNT=prod)
 	$(eval export ENABLE_AUTO_DEPLOY=true)
 	$(eval export PAAS_CF_TAG_FILTER=prod-*)
+	$(eval export SYSTEM_DNS_ZONE_NAME=cloud.service.gov.uk)
+	$(eval export APPS_DNS_ZONE_NAME=cloudapps.digital)
 
 .PHONY: vagrant-deploy
 vagrant-deploy:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck set_aws_count set_auto_trigger disable_auto_delete check-env-vars
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck check-env-vars
 
 .DEFAULT_GOAL := help
 
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -240,7 +240,8 @@ jobs:
         image: docker:///governmentpaas/curl-ssl
         inputs:
         - name: concourse-cert
-        - name: vpc-terraform-outputs
+        params:
+          SYSTEM_DNS_ZONE_NAME: {{system_dns_zone_name}}
         run:
           path: sh
           args:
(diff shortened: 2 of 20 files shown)


Makefile
```diff
@@ -79,6 +89,10 @@ set_env_class_stage:
 	$(eval export ENABLE_AUTO_DEPLOY=true)
 	$(eval export DISABLE_AUTODELETE=1)
 	$(eval export PAAS_CF_TAG_FILTER=stage-*)
+	$(eval export ROOT_DNS_ZONE=cloudpipeline.digital)
+	$(eval export APPS_ROOT_DNS_ZONE=cloudpipelineapps.digital)
+	$(eval export ENV_DNS_ZONE=${AWS_ACCOUNT}.${ROOT_DNS_ZONE})
```
alext: We can't use `AWS_ACCOUNT` here.  `AWS_ACCOUNT` is set to `stage`, but the domain needs to be `staging.cloudpipeline.digital`

Makefile
```diff
@@ -86,6 +100,10 @@ set_env_class_prod:
 	$(eval export AWS_ACCOUNT=prod)
 	$(eval export ENABLE_AUTO_DEPLOY=true)
 	$(eval export PAAS_CF_TAG_FILTER=prod-*)
+	$(eval export ROOT_DNS_ZONE=cloud.service.gov.uk)
+	$(eval export APPS_ROOT_DNS_ZONE=cloudapps.digital)
+	$(eval export ENV_DNS_ZONE=${AWS_ACCOUNT}.${ROOT_DNS_ZONE})
```
alext: We shouldn't be including `AWS_ACCOUNT` in the domains for production, they should be using the root domains directly.

concourse/pipelines/create-deployer.yml
```diff
@@ -256,7 +258,7 @@ jobs:
             . vpc-terraform-outputs/tfvars.sh
             openssl req -x509 -newkey rsa:2048 -keyout concourse.key \
               -out concourse.crt -days 365 -nodes -subj \
-              "/C=UK/ST=London/L=London/O=GDS/CN=deployer.{{deploy_env}}.${TF_VAR_system_dns_zone_name}"
+              "/C=UK/ST=London/L=London/O=GDS/CN=deployer.${ENV_DNS_ZONE_NAME}"
```
alext: With this change, we no longer need the `vpc-terraform-outputs` in this task. Would be good to remove them.

terraform/dev.tfvars
```diff
@@ -1,8 +1,6 @@
 aws_account = "dev"
-system_dns_zone_id = "Z1QGLFML8EG6G7"
-system_dns_zone_name = "dev.cloudpipeline.digital"
+env_dns_zone_id = "Z1QGLFML8EG6G7"
```
alext: I'm not sure I understand why this variable was renamed? I'm not sure that "env" in the new variable name is very clear. When reading through this, I initially assumed that "env" referred to environment variables.

Is there a reason to move away from `system_dns_zone*` and `apps_dns_zone*` as names for the system components and deployed apps respectively?

concourse/scripts/environment.sh
```diff
@@ -9,10 +9,6 @@ hashed_password() {
   echo "$1" | shasum -a 256 | base64 | head -c 32
 }
 
-fetch_system_dns_zone(){
-  awk -F' *= *|"' '$1=="system_dns_zone_name" { print $3}' "${PROJECT_DIR}/terraform/${AWS_ACCOUNT}.tfvars"
-}
```
alext: It's good to see this gone...

Makefile
combor: Good catch. I've noticed this as well but it was somehow left unmodified.

combor: I introduced new variable `$ENV_TYPE` do differentiate fro `AWS_ACCOUNT` as it now always matches : (

keymon:
#158 will conflict with this PR

terraform/dev.tfvars
combor: We have removed system dns zone completely and introduced new variable that consists of old  var.env + var.system_dns_zone_name. Example: https://github.com/alphagov/paas-cf/pull/159/commits/b4078cc63d130bf1c8d6b3ed41f429a5b0935d2f#diff-88a715cd562c7412a9c5bf1909c04b56R22. To be consistent we have renamed terraform variables to match Makefile and Concourse variables. When it comes to naming, as you know it's hard :) . I picked this one as it refers to entire environment not just root or system domain.

Makefile
alext: See my comment below about naming of variables. I think `ENV` is somewhat ambiguous here. It can be interpreted to mean values that come from environment variables.

I'm not sure we need the extra layer of indirection. We can just set the apps and system domain names here.  Do we even need the `ROOT_DNS_ZONE` variable? I think it's worth avoiding having more variables here than we need.

concourse/pipelines/create-deployer.yml
combor: Correct. Removed.

Makefile
combor: I explained below why it's called `ENV` but we can simplify this as well. I'm open minded about it.

terraform/dev.tfvars
dcarley: I agree with @alext, I think we should keep the naming "system" and "apps".

> I picked this one as it refers to entire environment not just root or system domain.

I don't quite understand this though. Aren't they the same thing, and different from the app domain?

Makefile
dcarley: I am also unsure about `ENV_TYPE` because it doesn't describe the usage (DNS prefix?). It looks like it's only used to construct `ENV_APPS_DNS_ZONE`, in which case do we need a variable at all?

Makefile
```diff
@@ -64,21 +66,36 @@ prod-bootstrap: check-env-vars set_env_class_prod vagrant-deploy  ## Start Produ
...
+	$(eval export ROOT_DNS_ZONE=cloudpipeline.digital)
+	$(eval export APPS_ROOT_DNS_ZONE=cloudpipelineapps.digital)
+	$(eval export ENV_DNS_ZONE=${DEPLOY_ENV}.${AWS_ACCOUNT}.${ROOT_DNS_ZONE})
+	$(eval export ENV_APPS_DNS_ZONE=${DEPLOY_ENV}.${ENV_TYPE}.${APPS_ROOT_DNS_ZONE})
 
 .PHONY: set_env_class_stage
 set_env_class_stage:
 	$(eval export MAKEFILE_ENV_TARGET=stage)
 	$(eval export AWS_ACCOUNT=stage)
+	$(eval export ENV_TYPE=stageing)
```
dcarley: Typo `s/stageing/staging`. Though see the comment below about `ENV_TYPE` first.

concourse/pipelines/create-deployer.yml
```diff
@@ -256,7 +257,7 @@ jobs:
             . vpc-terraform-outputs/tfvars.sh
```
alext: This line needs to be removed as well.

concourse/scripts/environment.sh
combor: Me too.

dcarley:
Looks good. Thanks for the addressing the comments. Ready for review again when @alext is back.

combor:
I added one more commit to this PR as we have discovered that running bosh-cli.sh script standalone is no longer possible. It uses environment.sh script that needs SYSTEM_DNS_ZONE_NAME to be set and this one is being created by Makefile.

alext:
> I added one more commit to this PR as we have discovered that running bosh-cli.sh

This means that it's now not possible to run this against a non-dev deployment.

combor:
I dropped last commit so this card wont block other cards for longer than necessary. We can create a follow up story.

dcarley:
> I dropped last commit so this card wont block other cards for longer than necessary. We can create a follow up story.

I've created [#115845521](https://www.pivotaltracker.com/story/show/115845521).

PR #161, opened 2016-03-16
Bump terraform in Travis.

alext:
## What

We upgraded terraform in our concourse containers to 0.6.12 a while ago.
We should therefore be using the same version in Travis.

## How to review

Verify that the Travis build downloads 0.6.12, and passes

## Who can review

Anyone but myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.6.10"
+    - TF_VERSION="0.6.12"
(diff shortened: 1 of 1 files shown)


dcarley:
I wish we didn't have to maintain these version numbers separately. But, until we can [re-use our existing containers on Travis](https://docs.travis-ci.com/user/docker/) [without disabling builds on their container-based infrastructure](https://docs.travis-ci.com/user/workers/container-based-infrastructure/), I don't think we have a choice.

PR #162, opened 2016-03-16
[#115414833] Ensure orphaned volumes aren't left behind in EC2

alext:
https://www.pivotaltracker.com/story/show/115414833

## What

Bosh doesn't delete persistent disks imediately when deleting the
corresponding VM, instead there's a scheduled job that deletes them N
days after their VM was deleted (5 days by default).

This PR does 2 things to prevent this.
* It sets the `max_orphaned_age_in_days` parameter to 0, which means bosh will clean them up within 30 mins (the cleanup job runs every 30 mins by default)
* It adds a task to run a `bosh cleanup --all` to the `destroy-microbosh` pipeline to ensure that all orphaned disks are cleaned up before bosh is destroyed.

## How to review

Deploy from this branch. Then delete the CF deployment. Note that there are 14 unused volumes in the EC2 console. Verify that these are automatically removed within 30 mins (the cleanup job runs on the hour and at 30 mins past the hour).

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/destroy-microbosh.yml
+++ b/concourse/pipelines/destroy-microbosh.yml
@@ -98,6 +98,22 @@ jobs:
           - get: bosh-secrets
           - get: bosh-init-state
           - get: bosh-manifest
+      - task: cleanup-orphaned-disks
+        config:
+          platform: linux
+          image: docker:///governmentpaas/bosh-cli
+          inputs:
+            - name: paas-cf
+            - name: bosh-secrets
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                ./paas-cf/concourse/scripts/bosh_login.sh bosh-secrets/bosh-secrets.yml
+
+                bosh cleanup --all
       - task: bosh-init-microbosh
         config:
           platform: linux
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -85,6 +85,8 @@ jobs:
       address: 127.0.0.1
       name: my-bosh
       db: (( grab meta.postgres ))
+      disks:
+        max_orphaned_age_in_days: 0
       cpi_job: (( grab cloud_provider.template.name ))
       ignore_missing_gateway: "false"
       user_management:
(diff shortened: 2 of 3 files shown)


concourse/pipelines/destroy-microbosh.yml
```diff
@@ -98,6 +98,23 @@ jobs:
...
+            - name: bosh-secrets
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                ./paas-cf/concourse/scripts/bosh_login.sh bosh-secrets/bosh-secrets.yml
+
+                # Cleanup ensures any orphaned disks are deleted before bosh itself is deleted
```
dcarley: Does `bosh cleanup` do anything more than disks? I think this comment could be omitted.

alext: It does. According to the inline help it cleans up unused releases and stemcells. Cleaning up orphaned disks is a more recent addition.

dcarley: That's good, because it's stuff that we also want to happen. But it means the comment is factually inaccurate and I don't think it adds much (compared to looking at the commit history to see exactly why we added it).

keymon: `bosh cleanup` help in the bosh cli is also misleading :):

https://github.com/cloudfoundry/bosh/blob/d2159dca24cacbf2731ecd6b4d0d5545b16c678e/bosh_cli/lib/cli/commands/maintenance.rb#L9

alext: > That's good, because it's stuff that we also want to happen.

In this instance, there's no need to clean the releases and stemcells up because we're about to delete the machine. It's just the volumes we care about. I had initially written a script to do just that, but it did them in serial, so it was actually quicker to do the full cleanup (and is one less script to maintain).

keymon:
Everything works for me. I can merge.

concourse/pipelines/destroy-microbosh.yml
alext: I've removed this comment, and added more detail to the commit message.

dcarley: Thanks :shipit:

keymon:
LGTM

PR #164, opened 2016-03-17
[#113266439] Updates to allow staging deployment

alext:
## What

There were a few small changes necessary to allow deploying to staging:

* We needed to populate the subnet and security group for the bootstrap concourse in the `Vagrantfile`
* We needed to disable the creation of the prod tags - we're not ready for this yet, there are stories to implement this in the backlog.

While doing this, I noticed that `DISABLE_AUTODELETE` had not been enabled for production. I've therefore updated this flag so that autodelete defaults to disabled unless explicitly enabled, and updated the Makefile so that it's enabled for dev. This felt safer, and less likely that autodelete would accidentally be enabled in a production environment.

## How to review

There's not much review that can be done here beyond verifying that the changes look good, and that the pipeline scripts still setup the autodelete pipeline correctly.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/scripts/pipelines-bosh-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-bosh-cloudfoundry.sh
@@ -57,7 +57,7 @@ update_pipeline() {
         <(generate_vars_file)
     ;;
     autodelete-cloudfoundry)
-      if [ ! "${DISABLE_AUTODELETE:-}" ]; then
+      if [ -n "${ENABLE_AUTODELETE:-}" ]; then
         bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
           "${env}" "${pipeline_name}" \
           "${SCRIPT_DIR}/../pipelines/${pipeline_name}.yml" \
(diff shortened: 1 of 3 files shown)


concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -57,7 +57,7 @@ update_pipeline() {
         <(generate_vars_file)
     ;;
     autodelete-cloudfoundry)
-      if [ ! "${DISABLE_AUTODELETE:-}" ]; then
+      if [ "${ENABLE_AUTODELETE:-}" = "true" ]; then
```
dcarley: :+1: inverting this.

Though I think the previous behaviour of accepting any set value is preferable. I realise it allows inconsistencies in what values can be set, but it does mean that people don't have to hunt out the conditional to see what a valid value is. And it would be consistent with the behaviour of `ENABLE_AUTO_DEPLOY`.

vagrant/Vagrantfile
```diff
@@ -13,6 +13,14 @@ AWS_ACCOUNT_DATA = {
     :subnet_id => "subnet-68bb7e30", # us-east-1a in default VPC, 172.31.16.0/20 range
     :security_group => "sg-23e1c15a", # "Bootstrap Concourse" security group
   },
+  "stage" => {
+    :subnet_id => "subnet-9034fbc8", # us-east-1c in default VPC, 172.31.16.0/20 range
```
dcarley: Out of interest, why the availability zone `1c`?

concourse/scripts/pipelines-bosh-cloudfoundry.sh
alext: I've updated this to accept any value.

vagrant/Vagrantfile
alext: No specific reason, just picked one out of the list.

dcarley:
:ok_hand:

PR #165, opened 2016-03-17
[#115845521] Move bosh-cli to Makefile

combor:
## What

This is a fix for `bosh-cli.sh` as https://github.com/alphagov/paas-cf/pull/159 introduced SYSTEM_DNS_ZONE_NAME variable. This var is set only by running `make` so we had to move bosh-cli.sh execution to Makefile. Additionally we re-arranged all targets so it's now possible to execute `make $ENV bootstrap`.  Read the updated README if you want more details. 

## How to review

Check if `make dev bosh-cli` works properly and connects to Concourse container. Also, run create-bosh-cloudfoundry pipeline and make sure that test-bosh-cli job succeeds .

## Who can review

Not @jimconner or @combor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,28 +39,28 @@ lint_shellcheck:
 	find . -name '*.sh' -print0 | xargs -0 $(SHELLCHECK)
 
 .PHONY: dev
-dev: check-env-vars set_env_class_dev deploy_pipelines ## Deploy Pipelines to Dev Environment
-
-.PHONY: dev-bootstrap
-dev-bootstrap: check-env-vars set_env_class_dev vagrant-deploy ## Start DEV bootsrap
+dev: check-env-vars set_env_class_dev ## Set environment to DEV
 
 .PHONY: ci
-ci: check-env-vars set_env_class_ci deploy_pipelines  ## Deploy Pipelines to CI Environment
-
-.PHONY: ci-bootstrap
-ci-bootstrap: check-env-vars set_env_class_ci vagrant-deploy  ## Start CI bootsrap
+ci: check-env-vars set_env_class_ci ## Set environment to CI
 
 .PHONY: stage
-stage: check-env-vars set_env_class_stage deploy_pipelines  ## Deploy Pipelines to Staging Environment
-
-.PHONY: stage-bootstrap
-stage-bootstrap: check-env-vars set_env_class_stage vagrant-deploy  ## Start Staging bootsrap
+stage: check-env-vars set_env_class_stage  ## Set Envirnoment to Staging
 
 .PHONY: prod
-prod: check-env-vars set_env_class_prod deploy_pipelines  ## Deploy Pipelines to Production Environment
+prod: check-env-vars set_env_class_prod ## Set Envirnoment to Production 
 
-.PHONY: prod-bootstrap
-prod-bootstrap: check-env-vars set_env_class_prod vagrant-deploy  ## Start Production bootsrap
+.PHONY: bootstrap
+bootstrap: ## Start bootsrap 
+	vagrant/deploy.sh
+
+.PHONY: bootstrap-destroy
+bootstrap-destroy: ## Destroy bootsrap 
+	./vagrant/destroy.sh
+
+.PHONY: bosh-cli
+bosh-cli: ## Create interactive connnection to BOSH container
+	concourse/scripts/bosh-cli.sh $(DEPLOY_ENV)
 
 .PHONY: set_env_class_dev
 set_env_class_dev:
@@ -97,10 +97,6 @@ set_env_class_prod:
 	$(eval export SYSTEM_DNS_ZONE_NAME=cloud.service.gov.uk)
 	$(eval export APPS_DNS_ZONE_NAME=cloudapps.digital)
 
-.PHONY: vagrant-deploy
-vagrant-deploy:
-	vagrant/deploy.sh
-
-.PHONY: deploy_pipelines
-deploy_pipelines:
+.PHONY: pipelines
+pipelines: ## Upload pipelines to Concourse
 	concourse/scripts/pipelines-bosh-cloudfoundry.sh
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ $ export DEPLOY_ENV=environment-name
 Create the bootstrap Concourse with `make`. Select the target based on which AWS account you want to work with. For instance for a DEV bootstrap:
 
 ```
-make dev-bootstrap
+make dev bootstrap
 ```
 `make help` will show all available options.
 
@@ -89,7 +89,7 @@ use to login.
 Run the following script:
 
 ```
-./vagrant/destroy.sh
+make dev bootstrap-destroy
 ```
 
 ## Deployer Concourse
@@ -110,7 +110,9 @@ Run the `create-deployer` pipeline from your *Bootstrap Concourse*.
 When complete you can access the UI from a browser with the same credentials as
 your *Bootstrap Concourse* on the following URL:
 
-`https://deployer.${DEPLOY_ENV}.dev.cloudpipeline.digital/`
+```
+https://deployer.${DEPLOY_ENV}.dev.cloudpipeline.digital/
+```
 
 ### Destroy
 
@@ -127,7 +129,7 @@ You will need a working [Deployer Concourse](#deployer-concourse).
 
 Deploy the pipeline configurations with `make`. Select the target based on which AWS accout you want to work with. For instance, execute: 
 ```
-make dev
+make dev pipelines
 ```
 if you want to deploy to DEV account.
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -39,28 +39,27 @@ lint_shellcheck:
...
-.PHONY: stage-bootstrap
-stage-bootstrap: check-env-vars set_env_class_stage vagrant-deploy  ## Start Staging bootsrap
+stage: check-env-vars set_env_class_stage  ## Set Envirnoment to Staging
 
 .PHONY: prod
-prod: check-env-vars set_env_class_prod deploy_pipelines  ## Deploy Pipelines to Production Environment
+prod: check-env-vars set_env_class_prod ## Set Envirnoment to Production 
+
+.PHONY: bootstrap
+bootstrap: vagrant-deploy ## Start bootsrap
```
dcarley: We don't need a separate target for `vagrant-deploy` anymore. The contents could be moved into this target like `bootstrap-destroy`.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1098,6 +1100,35 @@ jobs:
...
+          - name: paas-cf
+          - name: concourse-manifest
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
+              apk update && apk add bash
+              export CONCOURSE_ATC_PASSWORD=$(awk '$1~/basic_auth_password/ {print $2}' concourse-manifest/concourse-manifest.yml)
```
dcarley: Can you make this consistent with https://github.com/alphagov/paas-cf/blob/d6e7de881a2664f04dbd4e38c79adb21dca5f3fc/concourse/scripts/self-update-pipeline.sh#L26-L27 ?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1098,6 +1100,35 @@ jobs:
...
+          - get: pipeline-trigger
+            passed: ['cf-deploy']
+            trigger: true
+          - get: paas-cf
+            passed: ['cf-deploy']
+          - get: concourse-manifest
+      - task: connect-to-bosh-cli
+        config:
+          platform: linux
+          image: docker:///governmentpaas/curl-ssl
```
dcarley: Can we use https://hub.docker.com/r/governmentpaas/self-update-pipelines/ instead?

Makefile
```diff
@@ -103,6 +102,6 @@ set_env_class_prod:
 vagrant-deploy:
 	vagrant/deploy.sh
 
-.PHONY: deploy_pipelines
-deploy_pipelines:
+.PHONY: pipelines
+pipelines: ## Upload pipelines to Concourse
```
alext: I wonder whether we should have a guard here to prevent this being run on its own.

Perhaps we should rename the existing `check-env-vars` target to `check-deploy-env`, and then add a new `check-env-vars` that checks for the variables these jobs require. It's probably enough to just check for one of the vars set by the env targets (eg `MAKEFILE_ENV_TARGET`).

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: Done.

mtekel:
bosh-cli-test is nice, but does it really need to run on every build (it's being triggered)?

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: Well, we picked up ruby container as it's got ruby, bash and curl.

mtekel:
Also, given the amount of tests we have now, would it make sense to create a group for tests only?

Makefile
combor: Correct. Fixed.

dcarley:
> bosh-cli-test is nice, but does it really need to run on every build (it's being triggered)?

Would we know if it was broken if we didn't trigger it on every build?

mtekel:
Maybe we can just trigger as apart of review process?

combor:
> Also, given the amount of tests we have now, would it make sense to create a group for tests only?

Done.

dcarley:
> Maybe we can just trigger as apart of review process?

Nah, we should leave it automated. Otherwise we'll forget to run it and then there's no benefit. It shouldn't cause any harm running it each time - I believe it's in parallel to the other tests, so it won't slow anything down.

Makefile
jimconner: Check now added in https://github.com/alphagov/paas-cf/pull/165/commits/4dbb4e01b361f1582b37d3b3a436f8d0caf1797e

concourse/scripts/bosh-cli.sh
```diff
@@ -25,4 +25,4 @@ $FLY_CMD -t "${FLY_TARGET}" \
   intercept \
   --build="${BUILD_NUMBER}"\
   --step=one-off \
-  ${2:-sh}
+  "${2:-sh}"
```
mtekel: squash this in to previous commit

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -13,6 +13,7 @@ groups:
       - smoke-tests
```
mtekel: nothing wrong with this line, there's a typo in commit message 'out' instead of 'our'
github should allow commenting on PR messages as well

mtekel:
Would it be a better test to do some actual operation inside bosh-cli container, like bosh vms? Currently we just check if we can run that container at all... (and if it has 'true' command).

mtekel:
Apart from these last remarks it all works as described.

alext:
> Would it be a better test to do some actual operation inside bosh-cli container, like bosh vms?

I'm not sure that's necessary. This is intended to verify that the container orchestration works. to get this far it will have had to do a `bosh login`, so we've already verified that it can talk to bosh etc.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1098,6 +1106,36 @@ jobs:
...
+            - -e
+            - -c
+            - |
+              VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+              CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
+              export CONCOURSE_ATC_PASSWORD
+              ./paas-cf/concourse/scripts/bosh-cli.sh {{deploy_env}} true
+
+
   - name: tag-release
```
alext: Do we want to trigger this from the new test task as well?

combor:
All comments resolved.

mtekel:
Few things about bosh-cli test:
1. the commit message is wrong (contains description of curl-ssl)
2. what exactly is this trying to test? If the test wants to test that we (operators) can still get interactive and working bosh shell, then it should do the same steps as we do - "make dev bosh-cli" - and run a 'bosh vms' or 'bosh cck' command, to verify that the code of the (1) makefile, (2) bosh-cli script and (3) bosh-cli.yml are all working.
3. If it intends to test only a single aspect of the chain (ability to do fly hijack), then it should be renamed (e.g. test fly-intercept) and we should think about testing other parts of the bosh-cli chain

combor:
> 1. the commit message is wrong (contains description of curl-ssl)

Fixed.

alext:
> what exactly is this trying to test?

Following a discussion, I think it's reasonable to extend the test to call `bosh status`. This actually verifies that it can talk to bosh instead of the implicit reliance on the bosh login having happened.

combor:
I believe that all comments have been resolved.

README.md
```diff
@@ -163,12 +165,26 @@ All of the pipeline scripts (including `vagrant/deploy.sh`) honour a
...
+export AWS_ACCESS_KEY_ID=your_ci_access_key
+export AWS_SECRET_ACCESS_KEY=your_ci_secret_key
+make ci pipelines
+```
+
+Due to the isolation between AWS accounts, when switching accounts, it's
+necessary to start with a comletely new deployment.
+
+**Note:** Different AWS accounts use different DNS names, so it'll be necessary
+to adjust some of the instructions above accordingly.
```
alext: I don't think you meant to add this back in. These docs have moved into the separate linked doc.

combor: Probably it's a rebase leftover. I'll remove it.

combor: Removed.

mtekel:
works, merging

keymon:
Now self-update-pipelines does not work!!! :scream: 

It only calls `make dev` or whatever.

keymon:
PR to fix self-update https://github.com/alphagov/paas-cf/pull/172

PR #167, opened 2016-03-18
[#115320097] ELB for access to logs

HenryTK:
## What
Story: [ELB for access to logs](https://www.pivotaltracker.com/story/show/115320097)
We want to be able to access Logsearch via its web interface (Kibana), without the need for SSH tunneling. For this we provision an ELB and provide it with a self-signed certificate. 

An extension to the IAM policy allowing the uploading of the certificate has already been merged as a result of PR 25 in the aws-account-wide-terraform repository. 

There is an intermediate HAProxy because there is an intention to set up authentication for access to Kibana. This is already provided in the latest Logsearch releases. 

## How to review

- Deploy CF using branch `115320097-elb-access-to-logs-rebase`. 
- Check you can access Kibana on https://logsearch.[deploy_env].dev.cloudpipeline.digital
- View the certificate and check the Common Name is the same as the URL above. 

## Who can review

Anyone but @saliceti or myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -520,7 +520,7 @@ jobs:
                 fi
                 openssl req -x509 -newkey rsa:2048 -keyout metrics.key \
                   -out metrics.crt -days 365 -nodes -subj \
-                  "/C=UK/ST=London/L=London/O=GDS/CN=metrics.{{deploy_env}}.${SYSTEM_DNS_ZONE_NAME}"
+                  "/C=UK/ST=London/L=London/O=GDS/CN=metrics.${SYSTEM_DNS_ZONE_NAME}"
                 tar czvf generated-metrics-cert/metrics-cert.tar.gz metrics.crt metrics.key
         on_success:
           put: metrics-cert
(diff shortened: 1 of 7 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -107,6 +107,13 @@ resources:
       region_name: {{aws_region}}
       versioned_file: bosh-CA.tar.gz
 
+  - name: logsearch-cert
```
keymon: This commit message is wrong:

> Extract Terraform variables as certificate will use the system DNS zone name.

That was before, now you get the variable from the parameter, which is populated with the values from Makefile.

I would recommend avoid add in the comment what the code does in detail. Instead focus one explaining "why?", and overview of the "how" and maybe some caveats and links to references.

timmow: You've probably seen this, but the [team manual](https://government-paas-team-manual.readthedocs.org/en/latest/team/working_practices/#commit-messages) and [style guide](https://github.com/alphagov/styleguides/blob/master/git.md) expand on what @keymon is saying

HenryTK: Removed.

keymon:
Lovely, works as expected. Thx for fixing the issue from the previous PR.

PR #169, opened 2016-03-21
[#115709073] close port 80

mtekel:
## What

[Move public facing TLS terminations to ELBs](https://www.pivotaltracker.com/n/projects/1275640/stories/115709073)

Close port 80 as a part of leaving only secure endpoints open. This is 1st PR for the story.

## How to review

Apply to existing environment or build new. All tests should pass. You should not be able to connect to port 80 on any CF or app interfaces.

## Who can review

not @mtekel

### DIFF
--- a/tests/acceptance-tests/src/acceptance/http_closed_test.go
+++ b/tests/acceptance-tests/src/acceptance/http_closed_test.go
@@ -1,0 +1,44 @@
+package acceptance_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
+
+	"net"
+	"time"
+)
+
+var _ = Describe("Http is closed", func() {
+
+	var (
+		CONNECTION_TIMEOUT = 11 * time.Second
+	)
+
+	It("for apps", func() {
+		appName := generator.PrefixedRandomName("CATS-APP-")
+		Expect(cf.Cf(
+			"push", appName,
+			"--no-start",
+			"-b", config.StaticFileBuildpackName,
+			"-p", "../../example_apps/static_app",
+			"-d", config.AppsDomain,
+		).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+		Expect(cf.Cf("start", appName).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
+
+		uri := appName + "." + config.AppsDomain + ":80"
+		_, err := net.DialTimeout("tcp", uri, CONNECTION_TIMEOUT)
+		Expect(err.(net.Error).Timeout()).To(BeTrue())
+	})
+
+	It("for api", func() {
+		uri := "api." + config.SystemDomain + ":80"
+		_, err := net.DialTimeout("tcp", uri, CONNECTION_TIMEOUT)
+		Expect(err.(net.Error).Timeout()).To(BeTrue())
+
+	})
+
+})
--- a/manifests/cf-manifest/deployments/900-cf-tests.yml
+++ b/manifests/cf-manifest/deployments/900-cf-tests.yml
@@ -1,6 +1,6 @@
 properties:
   acceptance_tests:
-    api: (( grab meta.api_domain ))
+    api: (( concat "https://" meta.api_domain ))
     apps_domain: (( grab properties.app_domains[0] ))
     system_domain: (( grab properties.system_domain ))
     admin_user: "admin"
--- a/terraform/cloudfoundry/routers.tf
+++ b/terraform/cloudfoundry/routers.tf
@@ -15,12 +15,6 @@ resource "aws_elb" "router" {
     unhealthy_threshold = "${var.health_check_unhealthy}"
   }
   listener {
-    instance_port = 80
-    instance_protocol = "http"
-    lb_port = 80
-    lb_protocol = "http"
-  }
-  listener {
     instance_port = 443
     instance_protocol = "tcp"
     lb_port = 443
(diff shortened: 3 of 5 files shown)


Jonty:
Noting that this should not be merged until @dcarley has had time to think about it.

I am concerned about this change as the service manual specifies that a service _must_ be HTTPS and set [HSTS headers](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security), but it does _not_ specify that hosts should be registered in the [HSTS preload list](https://www.chromium.org/hsts).

HSTS via headers is a trust-on-first-use protocol meaning that future visits to the service will automatically use port 443 once the browser has seen the headers, but unless the service is in the preload list it will attempt to use port 80 on first connect, which obviously won't work.

The service manual also says that the service _must not_ be available on port 80.

There are two options available:

1. The service manual is updated to specify that services _must_ be registered in the HSTS preload list.
2. We re-open port 80 as a redirect-only port that simply 301's clients to the `https://` URL variant. The manual says the service must not be available on port 80, but I would argue that a redirector is not the service.

I would actually argue that we should play _both_ options. All services should be in the preload list to avoid MITM HTTPS downgrade attacks, and we should maintain a port 80 redirector for browsers/clients that do not use the preload list.

dcarley:
The advice in the Service Manual is very specific because users should always be arriving via a GOV.UK start page: https://www.gov.uk/service-manual/operations/operating-servicegovuk-subdomains#one-entry-point

@bleach and I agreed that it's alright to close TCP/80 for now without HSTS. We've got other stories to look at advertising/preloading HSTS and redirecting a bit later.

However we should add a custom acceptance test to in this PR to ensure that HTTP is no longer accessible. We'll need to do this for both the API and applications (even though they go through the same place at the moment).

mtekel:
Added the tests...

tests/acceptance-tests/src/acceptance/http_closed_test.go
```diff
@@ -0,0 +1,37 @@
...
+                        "push", appName,
+                        "--no-start",
+                        "-b", config.StaticFileBuildpackName,
+                        "-p", "../../example_apps/static_app",
+                        "-d", config.AppsDomain,
+                ).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
+                Expect(cf.Cf("start", appName).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+
+		uri := "http://" + appName + "." + config.AppsDomain
+        	_, err := exec.Command("curl", "-v", "-s", "--connect-timeout", "11", "--retry", "0", uri).CombinedOutput()
```
dcarley: 11 seconds seem like a very specific number. Could you explain in the commit message why this was chosen? I think it's also worth pulling out to a constant that describes what it is, e.g. `FIREWALL_TIMEOUT`.

tests/acceptance-tests/src/acceptance/http_closed_test.go
```diff
@@ -0,0 +1,37 @@
...
+                Expect(err).To(ContainSubstring("exit status 28"))
```
dcarley: The indentation is off here. Could you run `go fmt` over this file? https://blog.golang.org/go-fmt-your-code

tests/acceptance-tests/src/acceptance/http_closed_test.go
```diff
@@ -0,0 +1,37 @@
...
+        It("for apps", func() {
+                appName := generator.PrefixedRandomName("CATS-APP-")
+                Expect(cf.Cf(
+                        "push", appName,
+                        "--no-start",
+                        "-b", config.StaticFileBuildpackName,
+                        "-p", "../../example_apps/static_app",
+                        "-d", config.AppsDomain,
+                ).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
+                Expect(cf.Cf("start", appName).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
```
dcarley: Do you know why this uses `CF_PUSH_TIMEOUT`? Should it be `DEFAULT_TIMEOUT`?

tests/acceptance-tests/src/acceptance/http_closed_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+        It("for apps", func() {
+                appName := generator.PrefixedRandomName("CATS-APP-")
+                Expect(cf.Cf(
+                        "push", appName,
+                        "--no-start",
+                        "-b", config.StaticFileBuildpackName,
+                        "-p", "../../example_apps/static_app",
+                        "-d", config.AppsDomain,
+                ).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
```
dcarley: Should this use `CF_PUSH_TIMEOUT` instead?

tests/acceptance-tests/src/acceptance/http_closed_test.go
dcarley: Discussed with Michal. We decided that it would be better to test TCP alone, rather than TCP+HTTP, because we aren't 100% confident about the exit code when curl can TCP connect() but the underlying HTTP request times out (e.g. slow API server). We can use native Go for this:

- https://golang.org/pkg/net/#DialTimeout
- https://golang.org/pkg/net/#OpError

mtekel:
Updated

tests/acceptance-tests/src/acceptance/http_closed_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
+
+	"net"
+	"time"
+)
+
+var _ = Describe("Http is closed", func() {
+
+	var (
+		CONNETION_TIMEOUT = 11 * time.Second
```
dcarley: There's a typo here, `s/CONNETION/CONNECTION/`.

I wonder if the name `FIREWALL_TIMEOUT` is more descriptive though.

dcarley:
Tests out fine. Good to merge when the comment about the variable name is addressed.

mtekel:
Fixed the typo and IMO it should stay conn. timeout as the firewall doesn't need to be the only reason for failure to connect (e.g. could be ACLs, specific server config etc.). I also think fw_timeout could indicate that we are trying to test some specific timeout we set on the FW...

dcarley:
:ok_hand:

PR #170, opened 2016-03-21
[#115845521] Remove $DEPLOY_ENV from bosh-cli execution in Makefile

combor:
## What

As we have changed bosh-cli behaviour in previous PR we need to modify
Makefile to behave accordingly.

## How to review

run `make dev bosh-cli`

## Who can review

not @jimconner or @combor


mtekel:
thanks!

PR #171, opened 2016-03-21
[#115869215] Create tags for production. 

jimconner:
## What

When the staging environment completes deployment and all acceptance tests correctly, we want to tag that build with a prod-x.x.x tag, where x.x.x matches the version number of the stage-x.x.x tag which triggered the build in the staging environment.

## How to review this PR.

* Because this uses the AWS_ACCOUNT environment variable to ensure that the tag bumping only happens when running on the stage platform, it is necessary to test on the staging platform (or tweak your own environment so that it pretends that AWS_ACCOUNT is set to staging)
* Modify your Makefile and set SELF_UPDATE_PIPELINE=false in the set_env_class_prod section
* Optionally (if you're impatient) modify the acceptance tests job to perform a quick operation such as `/bin/true` before you push the pipelines.
* Destroy existing staging pipeline.
* Deploy the pipelines from this branch to staging.
* Kick off a deployment from Concourse by running the 'init' job.
* Make coffee
* Wait for acceptance tests to complete.
* Drink the coffe you made - Additional cups of coffee may be required depending upon deployment speed.
* Watch the tag-release box turn green in Concourse
* Check the git tags in github to see the new prod-x.x.x tag that was just created.

## Notes
After testing is complete, do not forget to destroy and re-push the pipelines into staging from the Master branch

## Who can review this PR.
Anyone on the team with the exception of @combor and @jimconner

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1249,6 +1249,7 @@ jobs:
             aws_account: {{aws_account}}
             deploy_env: {{deploy_env}}
             TAG_PREFIX: {{TAG_PREFIX}}
+            TAG_FILTER: {{paas_cf_tag_filter}}
           inputs:
           - name: paas-cf
           - name: release-version
@@ -1264,7 +1265,7 @@ jobs:
                 exit 0
               fi
               paas-cf/concourse/scripts/tag-release.sh \
-                ${TAG_PREFIX} ${aws_account} ${deploy_env}
+                ${TAG_PREFIX} ${aws_account} ${deploy_env} ${TAG_FILTER}
 
   - name: generate-git-keys
     plan:
--- a/concourse/scripts/tag-release.sh
+++ b/concourse/scripts/tag-release.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -eu
+set -eu -o pipefail
 
 TAG_PREFIX=$1
 AWS_ACCOUNT=$2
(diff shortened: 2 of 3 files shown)


concourse/scripts/tag-release.sh
```diff
@@ -20,17 +20,22 @@ Host github.com
   IdentityFile $(pwd)/git-key
 EOF
 
-version=$(cat release-version/number)
-tag="${TAG_PREFIX}${version}"
+cd paas-cf
+if git tag -l --contains HEAD | grep -E -q 'stage-\d\.\d\.\d'
```
saliceti: This script should be generic and not reference environment names directly. One simple approach is to create a new environment variable in the Makefile to control this behaviour, something like `GENERATE_RELEASE_VERSION`.
Or refactor the other variables `TAG_PREFIX` and `PAAS_CF_TAG_FILTER`.

concourse/scripts/tag-release.sh
```diff
@@ -20,17 +20,22 @@ Host github.com
...
 
 echo Configure Git
-cd paas-cf
 git config --global user.email "${GIT_EMAIL}"
 git config --global user.name "${GIT_USER}"
 git remote add ssh "${GIT_REPO_URL}"
 
 echo "Create tag ${tag}"
-git tag -a "${tag}" -m "Version ${version} passed ${AWS_ACCOUNT} \
+git tag -a "${tag}" -m "Tag ${tag} passed ${AWS_ACCOUNT} \
```
saliceti: The message with version instead of tag is nicer. Can you keep it?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -203,7 +203,7 @@ resources:
     source:
       bucket: {{state_bucket}}
       region_name: {{aws_region}}
-      key: release-version
+      key: {{release_version}}
```
saliceti: What is the issue with keeping simple "release-version" as key name? The semver resource lives in the state bucket and it is not shared with the other environments.

concourse/scripts/tag-release.sh
combor: I like the tag in the comment more as it is more explicit and you can see whether prod and or stage passed all the tests. And what we are passing is a tag here.

combor:
Ok, I believe we have resolved all comments.

concourse/scripts/tag-release.sh
```diff
@@ -1,15 +1,17 @@
...
-DEPLOY_ENV=$3
+TAG_PREFIX="${1}"
+AWS_ACCOUNT="${2}"
+DEPLOY_ENV="${3}"
 
 GIT_EMAIL="the-multi-cloud-paas-team+deployer-ci@digital.cabinet-office.gov.uk"
 GIT_USER="gov-paas-${AWS_ACCOUNT}"
 GIT_REPO_URL="git@github.com:alphagov/paas-cf.git"
 
+TAG_FILTER="${PAAS_CF_TAG_FILTER:-""}"`
```
dcarley: Travis build failed on this line with:
```
In ./concourse/scripts/tag-release.sh line 13:
TAG_FILTER="${PAAS_CF_TAG_FILTER:-""}"`
                                      ^-- SC1073: Couldn't parse this backtick expansion.
                                      ^-- SC1009: The mentioned parser error was in this backtick expansion.
```

saliceti:
All good now, thanks.

PR #172, opened 2016-03-22
[#115845521] Add target pipelines to self-update-pipelines.sh

keymon:
What?
-----

After #165 you must pass the target `pipelines` to update the pipelines,
but the `self-update-pipeline.sh` script was not updated.

Scope
-----

Just add the right target to the script

How to test?
------------

Self update should work as described in #156. See that PR for instructions
to test this.

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/scripts/self-update-pipeline.sh
+++ b/concourse/scripts/self-update-pipeline.sh
@@ -26,5 +26,5 @@ else
   CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
   export CONCOURSE_ATC_PASSWORD
 
-  make -C ./paas-cf "${MAKEFILE_ENV_TARGET}"
+  make -C ./paas-cf "${MAKEFILE_ENV_TARGET}" pipelines
 fi


dcarley:
Example on CI: https://deployer.master.ci.cloudpipeline.digital/pipelines/create-bosh-cloudfoundry/jobs/init/builds/32
```
Self update pipeline is enabled. Updating. (set SELF_UPDATE_PIPELINE=false to disable)
make: Entering directory '/tmp/build/d911d478/paas-cf'
make: Nothing to be done for 'ci'.
make: Leaving directory '/tmp/build/d911d478/paas-cf'
```

It's a shame that this failed silently. I can't think of a way to test it though. The only consolation is that this happened because we refactored a single target into two targets, which is seems we're less likely to do again in the future.

keymon:
We could add a parameter that keeps with the latest commit hash, and a task
after the update task that checks if the hash of that parameter is the same
than the current commit.

But not sure if it is worth it.

On 22 March 2016 at 12:42, Dan Carley <notifications@github.com> wrote:

> Example on CI:
> https://deployer.master.ci.cloudpipeline.digital/pipelines/create-bosh-cloudfoundry/jobs/init/builds/32
>
> Self update pipeline is enabled. Updating. (set SELF_UPDATE_PIPELINE=false to disable)
> make: Entering directory '/tmp/build/d911d478/paas-cf'
> make: Nothing to be done for 'ci'.
> make: Leaving directory '/tmp/build/d911d478/paas-cf'
>
> It's a shame that this failed silently. I can't think of a way to test it
> though. The only consolation is that this happened because we refactored a
> single target into two targets, which is seems we're less likely to do
> again in the future.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/paas-cf/pull/172#issuecomment-199793265>
>



-- 
Regards
Hector Rivas | GDS / Multi-Cloud PaaS

PR #173, opened 2016-03-23
[#115624569] Implement locking for pipelines with pool resources

keymon:
[#115624569 Implement locking for pipelines](https://www.pivotaltracker.com/story/show/115624569)

What?
----

We want to implement logic to lock the pipeline to prevent double executions in our deployments.

We decided
 * to use [the existing pool resource](https://github.com/concourse/pool-resource) which
uses git as a backend.
 * Use codecommit to create one git repo per user
 * Create one IAM git user to access that repo and upload one generated SSH key.
 * use IAM groups to grant access to the codecommit repository

Dependencies
--------

This PR depends in these other PR and changes:
 * `aws-account-wide-terraform/26` (already merged and applied to dev account)
 * https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/42 needs be merged and the container generated. After that, update the reference in this repo to the custom container.

Context
------

In this PR will cover the following context:

 * I want to create a codecommit repository
 * I want to create a IAM user
 * I want to generate a new SSH key for git access
 * I want to upload a SSH key to that IAM user
 * I want to setup a pool resource, which uses the created codecommit git resource via the created user with the generated SSH key.
 * I want to implement locking around all the pipeline.
 * I want to allow easily unlock the lock if it is required.

Notes
-----

We decided rename the job group "Release" to "Operator", as it groups different
operator level tasks.

How to review / test this?
-------------------------

You must re-run the bootstrap in order to create the git repository and user.

```
BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev bootstrap DEPLOY_ENV=__yourenvname__
```

Follow the README instructions. Once that is finished, upload the pipelines:

```
BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines DEPLOY_ENV=__yourenvname__
```

And trigger the job `pipeline-lock` from in `create-bosh-cloudfoundry`. You can
see a message saying that the lock is adquired.

Meanwhile the pipeline is running, try to rerun `pipeline-lock`. It should stop
trying to adquire the lock and print a bunch of dots `....`.

Now you can either:
 * let the pipeline finish and see that once `pipeline-unlock` runs, the other job is released.
 * or go to the job group "Operator" and trigger `pipeline-release-unlock`. You can see that
   the pipeline is released again.

who?
----

Anyone but @keymon or @mtekel or @Jonty

PR #177, opened 2016-03-29
[#115624569] Workaround terraform codecommit issue with default_branch property

keymon:
[#115624569 Implement locking for pipelines](https://www.pivotaltracker.com/story/show/115624569)

What?
----

We did not catch this issue in the PR #173.

Terraform codecommit resource[1] has a bug managing the default_branch
property: https://github.com/hashicorp/terraform/issues/5641

Although this property is optional, terraform will try to manage it.

If defined, codecommit will fail during creation:

```
BranchDoesNotExistException: refs/heads/master does not exist
```

Meanwhile if default_branch is defined, codecommit it will fail as soon
as a branch is pushed to the repository. The reason is that AWS enforces
having at least one branch as default branch, failing when terraform is
trying to set it empty:

```
  * aws_codecommit_repository.concourse-pool: Error Updating Default
  * Branch for CodeCommit Repository: InvalidParameter: 1 validation
    * errors:
     - field too short, minimum length 1: DefaultBranchName
```

Scope
----

This issue will be reported upstream, but meanwhile we will workaround it by
passing the default branch as a variable, and quering it using awscli.
If the respository does not exist or does not have default branch, the
variable will be set to empty string "".

We additionally remove the bump of the `pipeline-trigger` in the init job, which is not required anymore.

How to test
----------

Running from a environment without an associated codecommit git pool-resource
respository:

 1. Run the bootstrap process `SELF_UPDATE_PIPELINE=false BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev bootstrap DEPLOY_ENV=...`
 2. Run the jobs up to `concourse-terraform` (the rest are not relevant)
  * `concourse-terraform` should finnish OK
  * the `terraform apply` output for the resource `aws_codecommit_repository.concourse-pool` should not refer to any `default_branch`
 3. Rerun `concourse-terraform` it should end OK.
 4. Run the `lock-pipeline` job in `create-bosh-cloudfoundry`
    a. Upload the pipelines: `SELF_UPDATE_PIPELINE=false BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines DEPLOY_ENV=...`
    b. This will create a initial branch in the codecommit repo.
 5. Back in the bootstrap concourse (http://localhost:8080), rerun `concourse-terraform`
   * It should finish OK
   * terraform output should not report any change

Who?
---

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -312,8 +312,6 @@ jobs:
                 paas-cf/concourse/scripts/s3init.sh {{state_bucket}} cf.tfstate paas-cf/concourse/init_files/terraform.tfstate.tpl
                 paas-cf/concourse/scripts/s3init.sh {{state_bucket}} cf-certs.tar.gz paas-cf/concourse/init_files/empty.tar.gz
                 paas-cf/concourse/scripts/s3init.sh {{state_bucket}} git-keys.tar.gz paas-cf/concourse/init_files/empty.tar.gz
-      - put: pipeline-trigger
-        params: {bump: patch}
 
   - name: bosh-terraform
     serial_groups: [bosh-deploy]
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -270,6 +270,41 @@ jobs:
               "/C=UK/ST=London/L=London/O=GDS/CN=deployer.${SYSTEM_DNS_ZONE_NAME}"
             tar czvf concourse-cert.tar.gz concourse.crt concourse.key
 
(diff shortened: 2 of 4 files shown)


dcarley:
Haven't tested, but this look reasonable given the circumstances :ok_hand:

keymon:
I did a PR on terraform to fix the issue, although I am not sure if this is the right way to implement it: https://github.com/hashicorp/terraform/pull/5904

One alternative can be just use the patched version and bundle it in [the terraform container](https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/terraform), but it that case we will have a dependency on that branch.

mtekel:
works well, merging

PR #179, opened 2016-03-30
[#115709073] Move public facing TLS terminations to ELBs - Move router TLS termination to ELB

timmow:
**What**

We want to move TLS termination to the ELB for the router. As part of this, we have refactored the way SSL certificates are generated in the pipeline, as this was duplicated and not consistent.

Checkout this branch and deploy the changes. Run the tests and check they apply.

Check the ssl certificate contains the correct SAN for the router

`openssl s_client -connect api.$ENV.dev.cloudpipeline.digital:443 </dev/null | openssl x509 -noout -text | grep 'DNS:'`

**Who should review this PR**

* Anyone except me or @jonty


manifests/cf-manifest/scripts/generate-cf-certs.sh
```diff
@@ -18,6 +18,9 @@ router_ssl,${ROUTER_DOMAINS}
 uaa_jwt_signing,
 consul_server,server.dc1.cf.internal,server.dc2.cf.internal
 consul_agent,
+logsearch,logsearch.${SYSTEM_DNS_ZONE_NAME}
+metrics,metrics.${SYSTEM_DNS_ZONE_NAME}
+router,${ROUTER_DOMAINS}
```
alext: How is this different from the `router_ssl` cert above? Do we need both?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -667,9 +624,7 @@ jobs:
           - get: concourse-tfstate
           - get: bosh-tfstate
           - get: cf-tfstate
-          - get: metrics-cert
-          - get: logsearch-cert
-            passed: ['generate-cf-secrets']
+          - get: cf-certs
```
alext: This should probably be pinned to `generate-cf-secrets` to ensure it gets the up-to-date version.

mtekel:
TF behaves weirdly. In one case it reports 0 instances destroyed, whereas it is destroying old certs:
```
aws_iam_server_certificate.logsearch: Destroying...
aws_elb.metrics_elb: Modifications complete
aws_iam_server_certificate.metrics: Destroying...

Apply complete! Resources: 3 added, 6 changed, 0 destroyed.
```
In other instance it fails (& forgets about that resource):
```
aws_iam_server_certificate.concourse: Destroying...
Error applying plan:

1 error(s) occurred:

* aws_iam_server_certificate.concourse: DeleteConflict: Certificate: ASCAIL7OJZQWQCOHRWL3K is currently in use by arn:aws:elasticloadbalancing:eu-west-1:595665891067:loadbalancer/mt2-test-elb. Please remove it first before deleting it from IAM.
	status code: 409, request id: 137c1831-f67e-11e5-be43-4d43ee79d79c
```
The 2nd case looks like a possible race condition in creating & applying new cert and then quickly trying to destroy old cert which is in process of being removed from ELB. Even though I would have thought cert swap operation on ELB is atomic.

Jonty:
May help to read the summary here: https://www.pivotaltracker.com/projects/1275640/stories/115709073

manifests/cf-manifest/scripts/generate-cf-certs.sh
timmow: As discussed in person, its ok to have both, as we want to replace the external facing elb certficate in production environments, but not use this on the gorouter. But we should rename them.
However, the current logic [requires the certificate to be called `router_ssl`](https://github.com/alphagov/paas-cf/blob/move_public_facing_tls_termination_to_elbs_115709073/concourse/pipelines/create-bosh-cloudfoundry.yml#L811-L812) because that matches the required field in the manifest - we could special case it, but i'm not sure its worth it...

alext: I don't think that should be a problem.  Would we need to do anything other than update the keys the manifest is looking for here: https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/deployments/030-cf-jobs.yml#L879-L880

mtekel:
works as described...

PR #181, opened 2016-03-30
Remove unnecessary set_env_class_* Makefile targets

alext:
## What

Now that the makefile is structured so that we always do `make <env> <action>`, there's no need for the environment setup to be in separate targets, it can be in the env targets themselves.

## How to review

Verify that the makefile still works (eg verify that `make <env> showenv` outputs the expected values etc).

## Who can review

Anyone but myself.


mtekel:
works

PR #182, opened 2016-03-30
Fix erroneous tfstate pinning in bootstrap pipeline

alext:
## What

I added this in error in 5fc0230.

Jobs that run terraform must never pin the tfstate inputs. It's
important that they always run with the latest tfstate file, otherwise
they won't have an accurate picture of what actually exists in AWS.

## How to review

Verify that the bootstrap pipeline works.

## Who can review

Anyone but myself.


mtekel:
works...

PR #183, opened 2016-03-31
Fix CF destroy by unpacking certificates for tf

Jonty:
## What

This appears to have been broken since we added ELB's with certificates on them for metrics etc, but nobody noticed as we rarely totally destroy cf environments.

## How to review

Run the destroy pipeline. It should succeed.

## Who can review

Not @jonty or @timmow.


mtekel:
Destroy now works again...

PR #185, opened 2016-03-31
[#116047559] Notification when system unusable

HenryTK:
## What

We want to be able to run smoke tests continually in CI, Staging, and Production and be informed via email if any tests fail. We have left the existing smoke tests job in place as this is part of the end-to-end pipeline. We have moved the two tasks from the smoke-tests job to external task files in `paas-cf/concourse/pipelines/tasks` to avoid duplication of source code. 

To disable the continuous smoke tests in development we interpolate a boolean as the `trigger` value of the timer resource. 

To simplify the usage of AWS Simple Email Service we have used the AWS CLI tool. We did not use the email resource type as it does not support instance profiles. 

## How to review

You can change the email recipient in `concourse/scripts/smoke_tests_email.sh` to avoid spamming the team during tests. You need to log into AWS console SES service and [verify your email address](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/verify-email-addresses.html) first. 

The continuous smoke tests are enabled by default for CI, Staging, and Production. They are disabled during development by adding `$(eval export DISABLE_CONTINUOUS_SMOKE_TESTS=true)` to the `set_env_class_dev` target in the Makefile. Remove this line or change its value to `false` to review this Pull Request. 

* Using remote branch `116047559-system-notifications` set up the deployer Concourse. 
* Run the pipeline from the start and check:
	* The continuous-smoke-tests job runs after cf-deploy (it will not run concurrently with the `smoke-tests` job). 
	* The continuous-smoke-tests job runs repeatedly on a 5 minute interval.
	* The continuous-smoke-tests job does not send an email when the tests pass.
* Make the smoke tests fail and check an email is sent. This is a chance to be creative. 
* Check the contents of the email. It should contain:
	* A working hyperlink to the correct Concourse instance.
	* A log of the last commit in the paas-cf repository
	* The log from the smoke tests. 

## Who

Anyone but @saliceti or me

### DIFF
--- a/concourse/scripts/pipelines-bosh-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-bosh-cloudfoundry.sh
@@ -64,6 +64,7 @@ TAG_PREFIX: ${TAG_PREFIX:-}
 system_dns_zone_name: ${SYSTEM_DNS_ZONE_NAME}
 apps_dns_zone_name: ${APPS_DNS_ZONE_NAME}
 git_concourse_pool_clone_full_url_ssh: ${git_concourse_pool_clone_full_url_ssh}
+ALERT_EMAIL_ADDRESS: ${ALERT_EMAIL_ADDRESS:-}
 EOF
   echo -e "pipeline_lock_git_private_key: |\n  ${git_id_rsa//$'\n'/$'\n'  }"
 }
@@ -71,7 +72,9 @@ EOF
 generate_manifest_file() {
   # This exists because concourse does not support boolean value interpolation by design
   enable_auto_deploy=$([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
+  continuous_smoke_tests_trigger=$([ "${ALERT_EMAIL_ADDRESS:-}" ] && echo "true" || echo "false")
   sed -e "s/{{auto_deploy}}/${enable_auto_deploy}/" \
+    -e "s/{{continuous_smoke_tests_trigger}}/${continuous_smoke_tests_trigger}/" \
     < "${SCRIPT_DIR}/../pipelines/${pipeline_name}.yml"
 }
 
--- a/concourse/scripts/smoke_tests_email.sh
+++ b/concourse/scripts/smoke_tests_email.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+set -eu
+
+DEPLOY_ENV=$1
+SYSTEM_DNS_ZONE_NAME=$2
+
+TO=the-multi-cloud-paas-team@digital.cabinet-office.gov.uk
+FROM=the-multi-cloud-paas-team@digital.cabinet-office.gov.uk
+SMOKE_TEST_LOG=./smoke-tests-log/smoke-tests.log
+LAST_COMMIT_LOG=./smoke-tests-log/last-commit.log
+
+text_2_html(){
+  TEXT_FILE=$1
+  sed -e 's@$@<br/>@g' "${TEXT_FILE}" | tr -d '\n' | tr -d '"' | tr -d '\015'
+}
+
+write_message_json() {
+
+  cat <<EOF > message.json
+{
+  "Subject": {
+    "Data": "Smoke tests failed in ${DEPLOY_ENV}"
+  },
+  "Body": {
+    "Html": {
+      "Data": "The smoke tests have failed in environment <b>${DEPLOY_ENV}</b>. See \
+      <a href='https://deployer.${SYSTEM_DNS_ZONE_NAME}/pipelines/create-bosh-cloudfoundry?groups=health'>Concourse</a> \
+      for details<br/>\
+      <h3>Last commit:</h3><br/>$(text_2_html "${LAST_COMMIT_LOG}")<br/>\
+      <h3>Smoke test log:</h3><br/>$(text_2_html "${SMOKE_TEST_LOG}")"
+    }
+  }
+}
+EOF
+}
+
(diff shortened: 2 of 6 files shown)


concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -71,7 +71,9 @@ EOF
 generate_manifest_file() {
   # This exists because concourse does not support boolean value interpolation by design
   enable_auto_deploy=$([ "${ENABLE_AUTO_DEPLOY:-}" ] && echo "true" || echo "false")
+  [ "${DISABLE_CONTINUOUS_SMOKE_TESTS}" = "true" ] && continuous_smoke_tests_trigger="false" || continuous_smoke_tests_trigger="true"
```
dcarley: Nitpick: could you use the same style as the preceding line so that it's easier to read this function? (I *think* the following logic is correct)
```sh
continuous_smoke_tests_trigger=$([ "${DISABLE_CONTINUOUS_SMOKE_TESTS:-}" ] && echo "false" || echo "true")
```

dcarley:
> To simplify the usage of AWS Simple Email Service we have used the AWS CLI tool. We did not use the email resource type as it does not support instance profiles.

Channelling alphagov/paas-team-manual#23: can you include this in the commit message of 2d362d0?

concourse/scripts/pipelines-bosh-cloudfoundry.sh
saliceti: We chose that form because it only allows value "true" to make it true. Any other value makes it false.
The previous form makes it true whatever the value, as long as the variable is not empty.

mtekel:
Is it OK to have our (team) email address exposed in github in concourse/scripts/smoke_tests_email.sh? Also testing is a bit non-trivial as I have to change it and commit (to the PR i'm reviewing or create another branch off this branch), exposing my own mail address....

dcarley:
> Is it OK to have our (team) email address exposed in github in concourse/scripts/smoke_tests_email.sh?

That's OK.

> Also testing is a bit non-trivial as I have to change it and commit (to the PR i'm reviewing or create another branch off this branch), exposing my own mail address....

Would moving it to an environment variable that's passed by the Makefile make it easier to test and more apparent that it can be modified?

mtekel:
I think we might want to parametrize at least TO email address (ideally also FROM). I suspect we want to send production alerts perhaps to a different destination than staging and CI mails, or at least have different sender to be able to set up some mail filters. Address being a parameter also makes testing easier....

saliceti:
Should I replace the $DISABLE_CONTINUOUS_SMOKE_TESTS variable with an $ALERT_EMAIL (for example) variable? I could test if $ALERT_EMAIL is empty instead of testing if $DISABLE_CONTINUOUS_SMOKE_TESTS is true or false.

dcarley:
> Should I replace the $DISABLE_CONTINUOUS_SMOKE_TESTS variable with an $ALERT_EMAIL (for example) variable? I could test if $ALERT_EMAIL is empty instead of testing if $DISABLE_CONTINUOUS_SMOKE_TESTS is true or false.

Sounds like a good idea

concourse/scripts/pipelines-bosh-cloudfoundry.sh
dcarley: RE: accepting any value, I'm not sure what would be the least surprising behaviour. I don't really mind either way.

It could still be written as follows to match the existing style though:
```
continuous_smoke_tests_trigger=$([ "${DISABLE_CONTINUOUS_SMOKE_TESTS}" = "true" ] && echo "false" || echo "true")
```

mtekel:
Yes, that would be great. Also, you could add +{aws_account} to the mail address, e.g. 
`ALERT_EMAIL=the-multi-cloud-paas-team+staging@digital.cabinet-office.gov.uk` for staging, similarly for other environments.

saliceti:
Done

concourse/scripts/smoke_tests_email.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+set -eu
+
+DEPLOY_ENV=$1
+SYSTEM_DNS_ZONE_NAME=$2
+ALERT_EMAIL_ADDRESS=$3
+
+TO="${ALERT_EMAIL_ADDRESS}"
+FROM="${ALERT_EMAIL_ADDRESS}"
```
mtekel: this means I will receive alerts from 'me', when I set this to my own email address, which is a bit confusing...

How about setting FROM to st. like '{aws_account}-smoke-tests', e.g. 'staging-smoke-tests@digital.cabinet-office.gov.uk'

Apart from this, all else works fine...

mtekel: To make it more weird, I can actually see these emails in my 'sent' tab in gmail, even though they were sent by SES....

dcarley: > Or do we want this to be a real address that we can reply to?

It should be a real address. Not necessarily so that we can reply to it, but to make it easier to understand where the emails have come from. `@digital.cabinet-office.gov.uk` is used by a lot of people and projects.

saliceti: And the address has to be verified in SES so it must be real.

mtekel: Indeed the address has to be verified. Because of that, I'd say it's fine that the sender is the same as addressee. And because there are no other issues, merging.

PR #188, opened 2016-04-04
Fix external SSL cert docs.

alext:
## What

The cred-store paths were missing the `AWS_ACCOUNT` component. This
updates the docs to match the paths used in the `Makefile`.

## How to review

Verify that the cred-store paths in the docs match the paths used in the `Makefile`

## Who can review

Anyone but myself.

### DIFF
--- a/doc/non_dev_deployments.md
+++ b/doc/non_dev_deployments.md
@@ -42,9 +42,10 @@ for the public facing endpoints, instead of using self signed certificates.
 
 In that case, the operator must manually upload the certificates:
 
- 1. Store the certificates in the `credentials-high` password store, using
-    this naming convention for the key: `certs/${DEPLOY_ENV}/${CERT_NAME}.crt`
-    and `certs/${DEPLOY_ENV}/${CERT_NAME}.key`.
+ 1. Store the certificates in the `credentials-high` password store, using this
+    naming convention for the key:
+    `certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/${CERT_NAME}.crt` and
+    `certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/${CERT_NAME}.key`.
 
  2. After deploying the deployer with `create-deployer`, execute the make task
     `manually_upload_certs`. You indicate the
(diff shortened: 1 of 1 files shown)


doc/non_dev_deployments.md
```diff
@@ -42,9 +42,10 @@ for the public facing endpoints, instead of using self signed certificates.
 
 In that case, the operator must manually upload the certificates:
 
- 1. Store the certificates in the `credentials-high` password store, using
-    this naming convention for the key: `certs/${DEPLOY_ENV}/${CERT_NAME}.crt`
-    and `certs/${DEPLOY_ENV}/${CERT_NAME}.key`.
+ 1. Store the certificates in the `credentials-high` password store, using this
+    naming convention for the key:
+    `certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/${CERT_NAME}.crt` and
+    `certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/${CERT_NAME}.key`.
```
mtekel: It would be nice if these 2 were just below each other in rendered readme like they are here (currently they might end up on a single line, depending on your browser window width).

doc/non_dev_deployments.md
```diff
@@ -42,9 +42,10 @@ for the public facing endpoints, instead of using self signed certificates.
 
 In that case, the operator must manually upload the certificates:
 
- 1. Store the certificates in the `credentials-high` password store, using
-    this naming convention for the key: `certs/${DEPLOY_ENV}/${CERT_NAME}.crt`
-    and `certs/${DEPLOY_ENV}/${CERT_NAME}.key`.
+ 1. Store the certificates in the `credentials-high` password store, using this
```
combor: Cn you explain what is `credentials-high` password store?

doc/non_dev_deployments.md
dcarley: :+1:, can we use bullet points?

doc/non_dev_deployments.md
dcarley: It's https://github.gds/government-paas/credentials-high which is an alternative to https://github.gds/government-paas/credentials that only Civil Servants with Security Clearance have access to. It contains some more sensitive credentials, such as AWS and Pingdom root accounts.

It might be useful to make `credentials-high` a hyperlink. Even if people can't access it, they will be able to determine that it's a repo (rather than a script or service). Welcome any other suggestions too, @combor ?

dcarley:
@alext I've pushed two commits that address my comments. Could you ACK them and I'll merge?

PR #189, opened 2016-04-05
[#116603765] Enable concourse public read-only access

alext:
## What

Enable `publicly_viewable` for the deployer-concourse so that we can put the pipelines on the big screen without having to have it logged in. This concourse is still locked down by IP.

This allows read-only access to the concourse UI, but not to the output
of job runs (unless the job itself is defined as public). See
http://bosh.io/jobs/atc?source=github.com/concourse/concourse&version=0.74.0#p=atc.publicly_viewable
for more details.

This also includes a commit to make the Travis `before_install` tasks fail with better errors. 

## How to review

Run the create-deployer pipeline from this branch. Verify that the deployer-concourse can be accessed in read-only mode without any auth.

## Who can review

Anyone but myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,16 +18,19 @@ before_install:
   - mkdir ~/bin
   - export PATH=~/bin:$PATH
   - |
+    echo "Fetching Terraform"
+    set -e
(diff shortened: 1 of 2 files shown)


mtekel:
works as described...

PR #190, opened 2016-04-05
Test X-Forwarded-For and X-Forwarded-Proto

Jonty:
## What

This test validates that the X-Forwarded-For address being passed to applications originates outside our network (and thus is being passed through correctly) and that the X-Forwarded-Proto header is set to `https`, as all incoming connections are over SSL.

## How to review

Run the acceptance tests, both tests should be skipped in "Pending" state as we don't want to be running them until we fix the XFF and XFP issues.

## Who can review

Anybody but @jonty or @timmow.


tests/acceptance-tests/example_apps/print_request_headers/main.go
```diff
@@ -0,0 +1,29 @@
+package main
```
combor: I think `example_apps` directory was just an example. Shall we put this app in `apps` dir instead like here: https://github.com/cloudfoundry/cf-acceptance-tests/tree/master/apps ?

tests/acceptance-tests/example_apps/print_request_headers/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	jsonData, err := json.Marshal(r)
+	if err != nil {
+		http.Error(w, fmt.Sprint(err), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	fmt.Fprint(w, string(jsonData))
```
alext: Instead of this, you could just do `w.Write(jsonData)`.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+type ResponseJson struct {
+	Header struct {
+		X_Forwarded_For   []string `json:"X-Forwarded-For"`
+		X_Forwarded_Proto []string `json:"X-Forwarded-Proto"`
+	} `json:"Header"`
+}
+
+var headers ResponseJson
+
+var _ = BeforeEach(func() {
```
alext: This should be inside the `Describe` block below, otherwise it'll run before every test in the whole suite.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	jsonData := helpers.CurlAppRoot(appName)
+	headers = ResponseJson{}
+	json.Unmarshal([]byte(jsonData), &headers)
+})
+
+var _ = Describe("HTTP forwarded headers are set correctly", func() {
+
+	// Validate that the upstream IP is _not_ on our network
+	It("for X-Forwarded-For", func() {
```
dcarley: Could you restructure this so that it doesn't use comments? It will ensure that the intended behaviour appears in the test output and it makes it less likely that the comment will become out-of-sync with the code. For example:
```
    Describe("X-Forwarded-For", func() {
          It("should have an upstream IP that is not on our network", func() {
```

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+		remoteIP := strings.Split(headers.Header.X_Forwarded_For[0], ",")[0]
+		parsedRemoteIP := net.ParseIP(remoteIP)
+		_, infra_cidr, _ := net.ParseCIDR("10.0.0.0/8")
+
+		Expect(infra_cidr.Contains(parsedRemoteIP)).To(Equal(false))
+	})
+
+	// And that it is encrypted
+	It("for X-Forwarded-Proto", func() {
```
dcarley: Ditto https://github.com/alphagov/paas-cf/pull/190/files#r58683880

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+var _ = Describe("HTTP forwarded headers are set correctly", func() {
+
+	// Validate that the upstream IP is _not_ on our network
+	It("for X-Forwarded-For", func() {
+		remoteIP := strings.Split(headers.Header.X_Forwarded_For[0], ",")[0]
+		parsedRemoteIP := net.ParseIP(remoteIP)
+		_, infra_cidr, _ := net.ParseCIDR("10.0.0.0/8")
+
+		Expect(infra_cidr.Contains(parsedRemoteIP)).To(Equal(false))
```
dcarley: I think the following has the same effect but it slightly easier to read:
```go
Expect(infra_cidr).NotTo(ContainElement(parsedRemoteIP))
```

ref: https://onsi.github.io/gomega/#containelementelement-interface

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+	jsonData := helpers.CurlAppRoot(appName)
+	headers = ResponseJson{}
+	json.Unmarshal([]byte(jsonData), &headers)
+})
+
+var _ = Describe("HTTP forwarded headers are set correctly", func() {
+
+	// Validate that the upstream IP is _not_ on our network
+	It("for X-Forwarded-For", func() {
+		remoteIP := strings.Split(headers.Header.X_Forwarded_For[0], ",")[0]
```
combor: What will happen if headers.Header.X_Forwarded_For is `nil` or headers.Header.X_Forwarded_For[0] is `nil` ?

dcarley:
> both tests should fail as neither of these things work yet.

Should we merge this yet, if it's going to make all of our pipelines red?

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,62 @@
...
+		"push", appName,
+		"-b", config.GoBuildpackName,
+		"-p", "../../example_apps/print_request_headers",
+		"-d", config.AppsDomain,
+		"-c", "./bin/debug_app; sleep 1; echo 'done'",
+	).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+
+	jsonData := helpers.CurlAppRoot(appName)
+	headers = ResponseJson{}
+	json.Unmarshal([]byte(jsonData), &headers)
```
combor:
```
err := json.Unmarshal([]byte(jsonData), &headers)
if err != nil {
...
```
?

alext: Even better:
```go
Expect(json.Unmarshal([]byte(jsonData), &headers)).To(Succeed())
```

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
Jonty: The test will explode, which is great as that header being totally missing is very much a failure.

tests/acceptance-tests/example_apps/print_request_headers/main.go
Jonty: Yes, that's reasonable.

tests/acceptance-tests/example_apps/print_request_headers/main.go
Jonty: Seems reasonable to me.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
Jonty: I'd prefer separating `err` and `Expect(err).To(Succeed()) purely because you're getting into line noise territory there.

Jonty:
@dcarley: I think we should probably disable these tests and merge them rather than letting the branch diverge while we make them pass.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
alext: It will explode with a runtime panic, but that might not make it the easiest error to interpret. Adding some more checking here would make it much clearer in the case where it is missing.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
alext: Fair enough, although in that case I prefer `Expect(err).NotTo(HaveOccurred())`.

An alternative is to split it onto multiple lines:
```go
Expect(
    json.Unmarshal([]byte(jsonData), &headers)
).To(Succeed())
```

Jonty: Hah, yes of course. Either of those is reasonable really.

dcarley:
> @dcarley: I think we should probably disable these tests and merge them rather than letting the branch diverge while we make them pass.

:ok_hand:, we'll need a follow-up story for it. We have one for XFP but not necessarily for XFF.

I'm not sure whether "skip" or "pend" is the right thing to use. I don't know if Ginkgo has the same behaviour as rspec [described here](https://github.com/alphagov/paas-cf/pull/109#discussion_r53146105) which flags up when the test no longer fails.

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
Jonty: Unfortunately this isn't an option as `infra_cidr` isn't a array, slice, or map, which are the only types supported for that operation.

Jonty:
@alext @dcarley Ok, all issues addressed. Please cast an eye over it now.

The tests are marked as pending, however unfortunately ginkgo doesn't support the Rspec test inversion behaviour so we'll just have to remember to turn them back on!

tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+		jsonData := helpers.CurlAppRoot(appName)
+		headers = Headers{}
+
+		err := json.Unmarshal([]byte(jsonData), &headers)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should have an upstream IP that is not on our network", func() {
+		Expect(headers.X_Forwarded_For).Should(HaveLen(1))
```
combor: Just out of curiosity - why we use slice of strings if we expect it to have only one element always?

alext: Because that's how the Go http library behaves. I think it's because HTTP headers can appear multiple times, and this allows access all of them if there are more than one.

Jonty: @alext is correct - one of the odder quirks of HTTP is allowing multiple headers of the same type if the field is comma separated. Given that it's not well defined which fields _are_ comma separated, generally it's easier to just let the implementer decide. Yes, this means you nearly always get unexpected behaviour.

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma.

jimconner:
All working OK now. Passes all the tests which should be getting passed. Merging.

PR #192, opened 2016-04-07
Make the chrome extension userscript compatible

Jonty:
## What

This removes the chrome-specific JS from the injected payload, and adds userscript compatability headers (which will be ignored by chrome).

This means that you can either load the whole directory as a chrome extension, or if your browser supports userscripts (Firefox, Safari, Chrome with a greasemonkey extension) you can simply use the `inject.user.js` file as-is.

## How to review

Try the firefox instructions in firefox.

## Who can review

Anyone, really.


keymon:
can you call the script `inject.user.js` (and change it in the `.json`, ofc). That way, when I chlick on "Raw" it will trigger the greasemonkey plugins automatically and ask if it should be installed

That will enable the auto-update too.

Also, once done, can you mention that with a [Greasemonkey plugin for firefox](https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/) or [TamperMonkey for Chrome](https://chrome.google.com/webstore/detail/tampermonkey/dhdgffkkebhmkfjojejmpbldmpobfkfo?hl=en) you can install and keep track of update automatically.

concourse/chrome_plugin/src/inject.user.js
```diff
@@ -0,0 +1,26 @@
+// ==UserScript==
+// // @name        Remove concourse elements
```
keymon: Double comments prevent this from work :(

Jonty: DAMN YOU VIM

keymon:
:+1: works both as a chrome extension and userscript.

PR #193, opened 2016-04-08
[#116104189] Notification when platform unreachable

timmow:
**What**

As the PaaS team we would like to be notified when the system is unreachable.

Checkout this branch and deploy cloudfoundry. Run the make command to add a pingdom check
and verify the pingdom check appears - login details for pingdom are in paas-pass. You will need to update the check to use https to make it work

```
make dev pingdom DEPLOY_ENV=<ENV>
```

Test when the app is taken offline, pingdom recognises this

```
cf login
cf target -o testers -s healthcheck
cf stop healthcheck-app
```

**Who should review this PR**

* Anyone except me, @keymon, @HenryTK, @jimconner


mtekel:
Shellcheck complains a lot...

dcarley:
Could you capture some detail in the commit messages about the following:
- why is this run manually from the `Makefile` instead of in the pipeline?
- what manual steps do you currently have to perform after running the `Makefile` target?
- what version are the binaries of `terraform-provider-pingdom`?

timmow:
@mtekel i've fixed the shellcheck errors

timmow:
It might be helpful for someone running linux to test this PR to test the linux binaries work as expected

tests/example-apps/healthcheck-app/public/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html PUBLIC "-//IETF//DTD HTML 2.0//EN">
+<HTML>
+  <HEAD>
+    <meta http-equiv="Cache-Control" content="no-store" />
```
dcarley: I think we need to change this to an HTTP header instead. Only the browser will inspect the document contents, so it won't be honoured by any intermediaries like caching HTTP proxies or CDNs.

It might also be safer to use a combination of `max-age=0` and `private` because not all intermediary caches respect `no-store` or `no-cache` as you'd expect. For reference this is what GOV.UK does:
```
➜  paas-cf git:(master) curlh https://www.gov.uk/__canary__ | grep Cache-Control
Cache-Control: max-age=0,no-store,no-cache
```

PS: If you're really interested there is: https://github.com/alphagov/cdn-acceptance-tests

tests/example-apps/healthcheck-app/manifest.yml
```diff
@@ -0,0 +1,5 @@
+---
+applications:
+- name: healthcheck-app
+  memory: 64M
+  instances: 1
```
dcarley: Should we use more than one cell?

terraform/scripts/set-up-pingdom.sh
```diff
@@ -0,0 +1,57 @@
...
+terraform apply -state="${STATEFILE}" \
+	-var "env=${MAKEFILE_ENV_TARGET}" \
+	-var "pingdom_user=${PINGDOM_USER}" \
+	-var "pingdom_password=${PINGDOM_PASSWORD}" \
+	-var "pingdom_api_key=${PINGDOM_API_KEY}" \
+	-var "pingdom_account_email=${PINGDOM_ACCOUNT_EMAIL}" \
+	-var "apps_dns_zone_name=${APPS_DNS_ZONE_NAME}"
+
+# Copy statefile back to bucket and remove local copy
+aws s3 cp "${STATEFILE}" "s3://${DEPLOY_ENV}-state/${STATEFILE}"
```
mtekel: maybe set -e? otherwise if upload fails you'll remove the statefile...

mtekel:
Why not have pingdom extension in our TF container? Then make the alert adding script idempotent & you can make it a part of pipeline. Not sure if this is too much additional work though.

timmow:
@mtekel hopefully 8171326 explains in more detail why we cant use this approach.

mtekel:
Thanks, that explains it. Also, the pipeline part applies cleanly.

timmow:
@mtekel i've added `set -e`

@dcarley I've updated the number of instances to 2 for the healthcheck app and made it send a cache control header - I did this using PHP as it seemed to be the quickest way to do it...

Jonty:
LGTM, works. Merging.

Jonty:
While this LGTM I can't merge it as there are conflicts with master. @timmow @HenryTK can you rebase/resolve please?

PR #194, opened 2016-04-08
[#117021589] Performance Testing - Request timing

combor:
## What

Story link: https://www.pivotaltracker.com/story/show/117021589

We need to test whether the performance of our platform is not decreasing over time after code changes. To achieve this we have implemented a set of custom performance tests that will run after acceptance-tests and check if response time from deployed static app does not cross configured level. 

# Note

https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/44 needs to be merged first as it bumps Go lang version to 1.6 to avoid  [vegeta](github.com/tsenart/vegeta) bug.

## How to review
```
export RUN_PERFORMANCE_TESTS=true
```
Upload the pipeline from this branch and run it. All tests should be green.  

You can get all test results from `performance-tests` task output and also download them from $env-test-results-state 

## Who can review

anyone but @keymon and @combor


tests/performance-tests/generate_test_config.rb
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'yaml'
+
+manifest_file = ENV.fetch("CF_MANIFEST")
+manifest = YAML.load_file(manifest_file)
+
+admin_password = manifest.fetch("properties").fetch("performance_tests").fetch("admin_password")
+api_url = manifest.fetch("properties").fetch("cc").fetch("srv_api_uri")
```
mtekel: I think you can grab this from properties/acceptance_tests - as you are building on acceptance test framework.
You can then remove whole 'performance tests' part from manifest, as you don't need it - and nothing else uses it.

combor: Yes, it's possible to grab it from acceptance tests but in my opinion the fact that those values match is just coincidence, they don't have to. I would say it's better to separate performance tests properties as we may change them more flexibly in the future and add more. For instance right now we are hardcoding this https://github.com/alphagov/paas-cf/blob/117021589_request_performance_testing/tests/performance-tests/src/performance/load_test.go#L20 which is not perfect solution and should be retrieved from manifest. Unfortunately to do that we would have to extend this struct: https://github.com/cloudfoundry-incubator/cf-test-helpers/blob/master/helpers/config.go#L12 which means forking cf-test-helpers repo.

mtekel:
Regarding uploads to a bucket:
* why does the bucket name end in `-state` ? Looks like it's not storing any state...
* why not just have a "folder" in existing state bucket?
* why do we want to do this? If for possible later analysis, isn't that just unnecessary luggage to carry for now?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1222,15 +1238,82 @@ jobs:
...
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                export CF_MANIFEST=cf-manifest/cf-manifest.yml
+                ./paas-cf/tests/performance-tests/generate_test_config.rb \
+                  > test-config/config.json
+
```
mtekel: Why not just use acceptance-tests generated config? This perftest has the same requirements, uses same framework and runs against the same target...

combor: > why does the bucket name end in -state ? Looks like it's not storing any state...

We are only allowed to create buckets with a `-state` in the name :)

> why not just have a "folder" in existing state bucket?

Performance data is not a state. 

> why do we want to do this? If for possible later analysis, isn't that just unnecessary luggage to carry for now?

Yes, that was the idea. To keep it for analysis so we are able to graph historical data.

combor:
> Why not just use acceptance-tests generated config? This perftest has the same requirements, uses same framework and runs against the same target...

Ditto https://github.com/alphagov/paas-cf/pull/194#discussion_r59188808

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1222,15 +1238,82 @@ jobs:
...
+              - -c
+              - |
+                echo "Adding bosh-CA to root certificates"
+                tar -xf bosh-CA/bosh-CA.tar.gz -C /usr/local/share/ca-certificates bosh-CA.crt
+                update-ca-certificates
+
+                echo "Running tests"
+                export CONFIG="$(pwd)/test-config/config.json"
+                ./paas-cf/tests/performance-tests/run_tests.sh
+                GIT_COMMIT=$(cd ./paas-cf && git log --pretty=format:'%H' -n 1)
```
dcarley: I think `git rev-parse HEAD` is easier to read than `git log …`

You could also use `git -C paas-cf`, but that matters less.

mtekel:
I'd say you should decide what to do:
* either a proper fork & configurable variables (can you perhaps override the config function?)
* or admit that you are in fact using acceptance tests' config function & other functions of the framework - and use the same config. 

I'd add duplication (& flexibility) once we want to use it, for now we run with the same settings....

Regarding stats in the bucket - I'd make it optional, or think if we can perhaps add it later (given the added complication of doing it right - the access credentials).

tests/performance-tests/src/performance/init_test.go
```diff
@@ -0,0 +1,67 @@
...
+var (
+	DEFAULT_TIMEOUT      = 30 * time.Second
+	CF_PUSH_TIMEOUT      = 2 * time.Minute
+	LONG_CURL_TIMEOUT    = 2 * time.Minute
+	CF_JAVA_TIMEOUT      = 10 * time.Minute
+	DEFAULT_MEMORY_LIMIT = "256M"
+)
+
+const diegoUnsupportedTag = "{NO_DIEGO_SUPPORT} "
+const deaUnsupportedTag = "{NO_DEA_SUPPORT} "
```
dcarley: These two constants are introduced in 40c954dd47210e871a5d222d9d01bee26a6a4455 but I can't see how they're used. Can you explain?

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: > Yes, that was the idea. To keep it for analysis so we are able to graph historical data.

I'm unsure about storing it in an arbitrary format while we don't have a means to access/process it. We could store in Graphite, but I think it's OK not to store it anywhere for this spike.

tests/performance-tests/src/performance/load_test.go
```diff
@@ -0,0 +1,89 @@
...
+			"-m", DEFAULT_MEMORY_LIMIT,
+			"-p", "../../apps/static_app/",
+			"-d", config.AppsDomain,
+		).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+
+		Expect(cf.Cf("start", appName).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+	})
+
+	AfterEach(func() {
+		Expect(cf.Cf("delete", appName, "-f", "-r").Wait(DEFAULT_TIMEOUT)).To(Exit(0))
```
dcarley: Does the app need to be deleted after each test (rather than the suite)? Does that make the tests quite slow?

mtekel:
I'd perhaps add 2 more tests: with and without keepalive against pingdom app. This is supposed to be a long running app, also static (or as static as possible). The point here is to spot regressions for long lived apps and router records - e.g. if there isn't any performance degradation over time.

tests/performance-tests/src/performance/load_test.go
```diff
@@ -0,0 +1,89 @@
...
+	})
+
+	It("tests the load of app without Keep Alive ", func() {
```
dcarley: Could you make these a bit more descriptive? I think it should describe the fact that we're testing latency (by applying load). It might be worth separating the "keep alive" part into a context block, e.g.
```go
Context("with HTTP Keep-Alive", func() {
    It("has a 99th percentile response latency within our threshold", func() {
        …
    })
})

Context("without HTTP Keep-Alive", func() {
    It("has a 99th percentile response latency within our threshold", func() {
        …
    })
})
```
Although you could omit the "99th pc" part since it's in a function separate from the test.

dcarley:
It's unusual for a SPIKE to immediately result in a pull request and code to be merged. I think it's fine in this case, because you have something working and we haven't identified any other approaches. But we should treat decisions about comments (like whether we store state) as if this was a normal piece of work that we're now going to maintain (rather than just proving it works for now).

mtekel:
Apart from comments here, it deploys and works correctly....

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: Ok, removed.

tests/performance-tests/src/performance/load_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
+)
+
+const (
+	loadTestRate     = 100
+	loadTestDuration = 10 * time.Second
+	loadTestLatency  = 300
```
dcarley: From meatspace: @combor is going to add a unit to this (`* time.Millisecond` we think) so that it's easier to understand the value and we can remove the `/ 1000000` from the assertions further down.

tests/performance-tests/generate_test_config.rb
dcarley: @combor and I talked about this and we've agreed to re-use the same config the moment. After further discussion with @alext we will de-dupe `generate_test_config.rb`/`run_test.sh`/`README.md`, but leave `init_test.go`.

We can review at a later date, when we have more configuration, whether it's useful to put additional config in the manifest or load a separate configuration file.

tests/performance-tests/src/performance/init_test.go
combor: That was a unintentional leftover. Removed.

tests/performance-tests/src/performance/load_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Expect(time.Duration.Nanoseconds(latency)).To(BeNumerically("<", loadTestLatency))
+
+		})
+	})
+	Context("with HTTP Keep-Alive", func() {
+		It("has a response latency within our threshold", func() {
+			metrics, latency := loadTest(appName, loadTestRate, loadTestDuration, true)
+			generateJsonReport(metrics, "load-test-keep-alive.json")
+			vegeta.NewTextReporter(metrics).Report(GinkgoWriter)
+			Expect(time.Duration.Nanoseconds(latency) / 1000000).To(BeNumerically("<", loadTestLatency))
```
dcarley: Not sure if you've finished rebasing yet, but noting that ` / 1000000` needs to be removed from this line now that `loadTestLatency` is a `time.Duration`. I suspect the test will always pass because it makes `latency` a comparatively very small number.

combor: I haven't finished yet. I'll remove this.

tests/performance-tests/src/performance/load_test.go
combor: So we found this: https://github.com/onsi/ginkgo/issues/70 . It seems that it's currently not possible to run sth like BeforeAll/AfterAll. I can move BeforeEach code to BeforeSuite that is in `init_test.go` but that will make it load test specific one.

combor:
I believe that I have addressed all comments.

mtekel:
Re-tested. It all works fine. My final comment is that 'De-duplicate common bits from accept/perf tests' commit message doesn't really reflect what's done in that commit...

PR #195, opened 2016-04-11
[#117020631] Reduce default quota to 2GB

mtekel:
## What

We only want to allocate 2GB of "Freemium" memory in the default
quota. We also limit usage to 10 basic (free) services.

## How to review

Deploy from scratch (see https://github.com/cloudfoundry/bosh/issues/1201). View quotas (`cf quotas`). Try exceeding it - you should not be able.

## Who can review

not @mtekel

### DIFF
--- a/manifests/cf-manifest/deployments/040-cf-properties.yml
+++ b/manifests/cf-manifest/deployments/040-cf-properties.yml
@@ -2,9 +2,9 @@ meta:
   environment: ~
   default_quota_definitions:
     default:
-       memory_limit: 10240
-       total_services: 100
-       non_basic_services_allowed: true
+       memory_limit: 2048
+       total_services: 10
+       non_basic_services_allowed: false
        total_routes: 1000
   secrets:
     consul_ca_cert: (( grab secrets.bosh_ca_cert ))


manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -2,9 +2,9 @@ meta:
   environment: ~
   default_quota_definitions:
     default:
-       memory_limit: 10240
-       total_services: 100
-       non_basic_services_allowed: true
+       memory_limit: 2048
+       total_services: 10
+       non_basic_services_allowed: false
```
dcarley: This change isn't described in the commit message. Could you explain it (and add to the commit message)?

mtekel: > We also limit usage to 10 basic (free) services.
Or which change did you mean?

dcarley:
The commit message title `Reduce default quota to 2GB` is a bit misleading because it then goes on to say "We also limit … services". If I was looking through `git log` for changes to service quotas then I could overlook it. I doesn't warrant separate commits, but could you rename the title to something like `Reduce default quotas`?

manifests/cf-manifest/deployments/040-cf-properties.yml
dcarley: Line 7: `non_basic_services_allowed`

mtekel:
updated, still some wip left to do

manifests/cf-manifest/deployments/040-cf-properties.yml
dcarley: Ah, after re-reading it and thinking about it, I now understand. I hadn't grokked the distinction between:
- limit to 10 basic services which are free
- limit to 10 services which can only be basic/free

I think the updated commit now describes this better :+1:

mtekel:
Because of https://github.com/cloudfoundry/bosh/issues/1201 , you need to rebuild from scratch to apply the quota.

dcarley:
> you need to rebuild from scratch to apply the quota

Rebuilding from scratch isn't going to be appropriate for our CI->Staging->Prod pipeline.

Would you recommend that we use `cf update-quota` to deploy these changes? Perhaps a job in the pipeline which can source the values from the manifest and can be removed later?

Or should we workaround it by renaming the quota? Did you and @Rory80Hz talk about other suitable names?

mtekel:
@Rory80Hz mentioned he likes 'default'. The easiest workaround in our current situation is to rename quota. Hopefully BOSH/cc release issue is fixed upstream soon.

A job to read quotas from manifest and update them on every CF deploy is also a possible workaround, but order of magnitude more complicated than renaming the quota. Also, it slows pipeline down a bit.

dcarley:
Any suggestions for a new "default" name? @Rory80Hz?

Do we know what happens if you rename the default quota? Will existing applications still be associated with the quota that has a new name? Is `default` a special case?

mtekel:
due to further complications with 'renaming' - using BOSH we can currently only create new quotas, I'll add a job to set quota using cf cli

PR #196, opened 2016-04-12
[#116974545] Split apps and system services entry endpoints traffic

keymon:
[#116974545 Seperate IP Restrictions for Apps and System domains](https://www.pivotaltracker.com/story/show/116974545)

## What

As an operator I want to be able to restrict the IPs that can connect to the CF API endpoints without affecting public access to tenant deployed applications.

During kick-off we decided to solve this by using ELBs for each one of the exposed CF API services which are being handled by the routers.

## How to review

This story should be review in this context:

 * Create a new security group for CF related APIs, which is open to the office only.
 * Create dedicated ELBs for: CloudController, UAA, Loggregator and Doppler.
 * Configure these ELBs in BOSH so the instances get automatically registered on creation.
 * Configure the DNS for those services, dropping the wildcard DNS and adding specific ones explicitelly. 
 * Additionally (following [boy scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule)): I want to clean up, reorganise and rename the terraform config and resources related to ELBs and Security groups.

## How to test

 1. Deploy the environment and run the smoke tests. Smoke tests is enough, as it uses all the affected endpoints.
 2. Try to deploy an app manually from the office and operate it with `cf-cli`

## Notes

 * This change is disruptive and will stop the platform for working as being deployed.
 * You might end in a situation with negative DNS caching in concourse. Wait 5m if so.
 * The domains `*.uaa.${system_root_domain}` and `*.login.${system_root_domain}` did not seem to be used.

## Who can review

anyone but @keymon or @alext


terraform/cloudfoundry/logsearch_security_groups.tf
```diff
@@ -0,0 +1,87 @@
...
+      "${var.vpc_cidr}"
+    ]
+  }
+
+  tags {
+    Name = "${var.env}-logsearch-elastic-master-elb"
+  }
+}
+
+
```
mtekel: extra line - makes it harder to diff

terraform/cloudfoundry/cf_api_elb.tf
```diff
@@ -22,3 +22,29 @@ resource "aws_elb" "cf_cc" {
...
+  name = "${var.env}-cf-uaa"
+  subnets = ["${split(",", var.infra_subnet_ids)}"]
+  idle_timeout = "${var.elb_idle_timeout}"
+  cross_zone_load_balancing = "true"
+  security_groups = [
+    "${aws_security_group.cf_api_elb.id}",
+  ]
+
+  health_check {
+    target = "TCP:8080"
```
mtekel: I think if you hit any endpoint (/ or even /whatever) you get some html error page back - which indicates that the process is alive & it's an http response, so you can use http check. Not sure how much better than just TCP check it is...

manifests/cf-manifest/spec/fixtures/terraform/terraform-outputs.yml
```diff
@@ -22,6 +22,8 @@ terraform_outputs:
   elb_name: unit-test-cf-router-elb
   cf_cc_elb_name: stub_cf_cc_elb
   cf_uaa_elb_name: stub_cf_uaa_elb
+  cf_loggregator_elb_name: stub_cf_loggregator_elb
+  cf_doppler_elb_name: stub_cf_doppler_elb
```
mtekel: flip the two lines above so that the ELB name is always longer on the next line

terraform/cloudfoundry/cf_api_elb.tf
keymon: No, it returns 404 and ELB marks is as broken :(

terraform/cloudfoundry/logsearch_security_groups.tf
keymon: sure, fixed

keymon:
rebased with master

terraform/cloudfoundry/dns-records.tf
```diff
@@ -14,6 +14,22 @@ resource "aws_route53_record" "cf_cc" {
...
+  zone_id = "${var.system_dns_zone_id}"
+  name = "uaa.${var.system_dns_zone_name}."
+  type = "CNAME"
+  ttl = "60"
+  records = ["${aws_elb.cf_uaa.dns_name}"]
+}
+
+resource "aws_route53_record" "cf_login" {
+  zone_id = "${var.system_dns_zone_id}"
+  name = "login.${var.system_dns_zone_name}."
```
mtekel: route registrar also registers *.login....

terraform/cloudfoundry/dns-records.tf
```diff
@@ -14,6 +14,22 @@ resource "aws_route53_record" "cf_cc" {
   records = ["${aws_elb.cf_cc.dns_name}"]
 }
 
+resource "aws_route53_record" "cf_uaa" {
+  zone_id = "${var.system_dns_zone_id}"
+  name = "uaa.${var.system_dns_zone_name}."
```
mtekel: route registrar also registers *.uaa....

terraform/cloudfoundry/dns-records.tf
keymon: As commented, we were not able to find how and when *.uaa... and *.login... is used.

alext: They also won't match the SSL certs we're using.

mtekel:
works fine. application might require two runs of deploy as one of etcd jobs can break...

PR #198, opened 2016-04-12
[#117020631] Reduce default quota to 2GB

mtekel:
## What

We only want to allocate 2GB of "Freemium" memory in the default quota. We also limit usage to 10 basic (free) services. Because BOSH can't update values directly from manifest changes (https://github.com/cloudfoundry/bosh/issues/1201), add a step in post-deploy job to use cf-cli to update quotas. Add a small script to help extract quota settings from manifest.

## How to review

Deploy. View quotas (`cf quotas`) and verify default has been updated to new values. Try exceeding it - you should not be able to.

## Who can review

not @mtekel

### DIFF
--- a/concourse/scripts/extract_quota_settings.rb
+++ b/concourse/scripts/extract_quota_settings.rb
@@ -1,0 +1,12 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'yaml'
+
+manifest_file = ENV.fetch("CF_MANIFEST")
+manifest = YAML.load_file(manifest_file)
+default = manifest['properties']['cc']['default_quota_definition']
+puts "export QUOTA_DEFAULT=#{default}"
+
+manifest['properties']['cc']['quota_definitions'][default].each { |k,v|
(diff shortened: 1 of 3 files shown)


concourse/scripts/extract_quota_settings.rb
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'yaml'
+
+manifest_file = ENV.fetch("CF_MANIFEST")
+manifest = YAML.load_file(manifest_file)
+default = manifest['properties']['cc']['default_quota_definition']
+  puts "export QUOTA_DEFAULT=#{default}"
```
saliceti: whitespace

mtekel: on purpose, so that it aligns with puts 2 lines below

saliceti:
Works as designed.

This removes any update made manually by `cf update-quota`. But I've confirmed with @Rory80Hz that this is the expected behaviour.

PR #199, opened 2016-04-12
[#117039827] Remove network ACLs

combor:
## What

Following the security group spike, we should remove network ACLs, so we don't need to maintain them as we change the platform.

## How to review

Deploy from the branch and check if acceptance-tests pass.

## Who can review

Not @combor


mtekel:
work correctly...

PR #201, opened 2016-04-14
Disable Travis CI email notifications on success

dcarley:
## What

I get a "build passed" email whenever a tag is created for a PR that I've
merged. These emails are cluttering up my mailbox and aren't terribly
useful, because I don't need to take any action when a build has succeeded.
I only need to know if it has failed.

Prevent this from happening by never sending email notifications for
successful builds. The default for `on_failure` is `always` so the behaviour
of failed builds remains unchanged.

ref: https://docs.travis-ci.com/user/notifications/#Email-notifications

Part of the problem here is that Travis CI doesn't really need to build
tags. But turning that off is currently a hack: travis-ci/travis-ci#1532

## How to review

This isn't very easy to test. I guess you could confirm that the failure path still works, by branching from this PR, pushing a commit which fails the build, and confirming that it still sends you an email.

## Who can review

Not @dcarley

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,3 +34,7 @@ before_install:
 
 script:
   - make test
+
+notifications:
+  email:
+    on_success: never


mtekel:
I never got any email from travis, regardless of me merging, creating new PRs or branches or pushing anything...

dcarley:
Strange. I seem to get quite a few to my personal address (which is what I use for git commits):

screen shot 2016-04-14 at 10 51 09
screen shot 2016-04-14 at 10 51 32

keymon:
I disagree on disabling this, I like to get them after the build has failed. Maybe we can we set it up to change:     `change: send a notification when the build status changes.`, but not sure if this would prevent the first mail from being sent.

mtekel:
As a part of testing merge commit script, I have logged in to travis for the 1st time, adding my own repos. Now I get all the spam from travis for all the repos I am part of (e.g. paas-cf) including successfull builds - for many branches. Some documentation on travis side: https://docs.travis-ci.com/user/notifications/#Email-notifications

In short, there's no way for me to go to travis and tell it I don't want emails. The setting is global (per repo) and travis automatically determines recipients from commit authors and repo owners...

mtekel:
I removed Travis from my OAuth apps in github, let's see if that disables the mails again. If so, that can be solution for individuals who wish not to receive any mails...

keymon:
I think it is easy to fix by simply filtering the mails in your mail
account.

On 27 April 2016 at 10:23, mtekel <notifications@github.com> wrote:

> I removed Travis from my OAuth apps in github, let's see if that disables
> the mails again. If so, that can be solution for individuals who wish not
> to receive any mails...
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/paas-cf/pull/201#issuecomment-215024667>
>



-- 
Regards
Hector Rivas | GDS / Multi-Cloud PaaS

dcarley:
The default for `on_success` is `change`. @alext suggested that the reason it's doing this for tags is that it might be considering the creation of a tag as change of state from "nothing".

I still want to receive emails about failed builds. How do you both feel about disabling builds for tags (branches which match the regex `/^v\d+\.\d+\.\d+$/`)?

keymon:
I am happy to disable builds for tags.

dcarley:
I'll open a new PR to disable builds for tags.

PR #202, opened 2016-04-14
[#117522857] Add a script testing framework based on rspec & aruba

keymon:
[#117522857 Setup a test framework for our scripts](https://www.pivotaltracker.com/story/show/117522857)

## What

We want a testing framework for helper scripts, such as those used for Concourse and the manifests. RSpec and [Aruba](https://github.com/cucumber/aruba) were chosen to remain consistent with existing tests. Tests have been written for the `concourse/scripts/tag-release.sh` and `concourse/scripts/file-to-yaml.sh`. Helper methods included in `concourse/scripts/spec` can be re-used when this pattern is used elsewhere in the repository.

## How to review

* This PR is dependent on [paas-docker-cloudfoundry-tools - PR 47](https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/47), so make sure that is merged first.
* Run `make test` and check the tests were executed and were successful. 
* Check the same tests pass in Travis.

## Who can review

Anyone but @keymon or @henrytk


concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,100 @@
...
+require "pathname"
+
+def find_git_repository_path()
+  path = __FILE__
+  begin
+    if File.exists? File.join(path, '.git')
+      return path
+    end
+    path = File.realpath(File.join(path, '..'))
+  end while path != "/"
```
dcarley: I found this quite difficult to read. I think `git rev-parse --show-toplevel` could do the same thing.

concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,100 @@
...
+  raise "Cannot find a git repository parent of '#{File.dirname(__FILE__)}'"
+end
+
+def repostory_name()
+  File.basename(find_git_repository_path())
+end
+
+def prepare_aruba_config()
```
dcarley: Could this be moved to a `support` file or `spec_helper` so that we don't need to call the method in each test?

ref: https://github.com/alphagov/paas-cf/blob/62aba6b02de37f21cf957d08b6f4b7ba7248c7d4/concourse/scripts/spec/spec_helper.rb#L102

concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,100 @@
...
+      relative_workdir = Pathname.new(@workdir).relative_path_from(Pathname.pwd)
+
+      Aruba.configure do |config|
+        config.working_directory = relative_workdir
+        config.home_directory = @workdir
+        @repo_root = find_git_repository_path()
+
+        config.command_search_paths = [
+          File.join(@repo_root, "concourse", "scripts"),
+          File.join(@repo_root, "manifests", "scripts")
```
dcarley: I guess we could exclude manifests for the moment because they already have a test framework. Maybe make the path relative to `__FILE__` for the moment?

concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+    end
+    config.after(:all) do
+      # Ensure we clean up after
+      def rmdir(dir)
+        FileUtils.rmdir(dir) if File.exists? dir
+      end
+      rmdir @workdir
+      rmdir @aruba_orig_workdir
+      rmdir @aruba_orig_workdir
```
dcarley: Dupe :cop:

concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,100 @@
...
+  }
+end
+
+# Helper to run a bash script in verbose mode and capturing
+# all output together
+#
+# Good for scripts and initialise stuff
+#
+def bash_c(script)
+  run_simple("bash -x -c 'exec 2>&1 # Send Stderr to Stdout\n\n#{script}'")
```
dcarley: Should these helpers also run with `set -euo pipefail` so that we know if the setup has failed?

concourse/scripts/spec/tag-release_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    before(:each) do
+      bash_c("cd paas-cf && git checkout previous-0.0.2")
+    end
+
+
+    it("should promote the same tag version for 'next-*'") do
+      run("./tag-release.sh next- test_aws_account test_env \"previous-*\"")
+      expect(last_command_started).to have_exit_status(0)
+
+      bash_c_quiet("cd paas-cf && echo a && git tag -l \"next-*\" --sort version:refname | tail -n 1")
```
dcarley: Would this be more succinct if it asserted that `previous-0.0.2` and `next-0.0.2` point at the same commit?

concourse/scripts/spec/tag-release_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+end
+
+def setup_cloned_repository()
+    bash_c(%q{
+      mkdir origin_repo
+      cd origin_repo
+      git init .
+      git commit --allow-empty -m "First commit"
+
+      git tag previous-0.0.1
```
dcarley: I think the tag creation might belong in the individual `context` blocks rather than the outer `before` block, so that it's easier to understand what the individual dependencies of each test are (because I suspect they will diverge).

concourse/scripts/tag-release.sh
```diff
@@ -1,17 +1,15 @@
 #!/bin/sh
 
-set -eu -o pipefail
```
dcarley: We'd still want this, wouldn't we?

concourse/scripts/spec/script_testing_helpers.rb
keymon: I don't think so, it should be up to each person who calls the function.

concourse/scripts/tag-release.sh
keymon: But does not work on ubuntu/debian, only on alpine :(

concourse/scripts/spec/script_testing_helpers.rb
dcarley: I disagree. We should make it safe by default and people can use `|| true` or `set +e` if they wish to explicitly opt-out of it.

concourse/scripts/tag-release.sh
dcarley: I don't understand. Do you know why?

Makefile
```diff
@@ -28,15 +28,15 @@ spec:
 		bundle exec rspec
 
 lint_yaml:
-	$(YAMLLINT) -c yamllint.yml .
+	find . -name '*.yml' | grep -v -e '^\./vendor' | xargs $(YAMLLINT) -c yamllint.yml
```
dcarley: Would `-not -path './vendor/*'` do the same thing here without the added complexity of regexes and pipes?

Makefile
```diff
@@ -28,15 +28,15 @@ spec:
...
 
 lint_terraform: dev
 	$(eval export TF_VAR_system_dns_zone_name=$SYSTEM_DNS_ZONE_NAME)
 	$(eval export TF_VAR_apps_dns_zone_name=$APPS_DNS_ZONE_NAME)
 	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t terraform graph > /dev/null
 
 lint_shellcheck:
-	find . -name '*.sh' -print0 | xargs -0 $(SHELLCHECK)
+	find . -name '*.sh' | grep -v -e '^\./vendor' | xargs $(SHELLCHECK)
```
dcarley: Ditto https://github.com/alphagov/paas-cf/pull/202/files#r59742097

Makefile
keymon: Is this a good excuse?

screenshot from 2016-04-14 17 05 56

keymon: fixed

concourse/scripts/spec/configure_aruba.rb
```diff
@@ -0,0 +1,37 @@
...
+
+RSpec.configure do |config|
+  config.before(:all) do
+    # We run scripts with git commands, and as we don't want to mess around
+    # with our own repository, we setup the workdir in a different location.
+    @workdir = Dir.mktmpdir("paas-cf-tmpdir-")
+    # Aruba requires the workdir path to be a relative directory
+    relative_workdir = Pathname.new(@workdir).relative_path_from(Pathname.pwd)
+
+    Aruba.configure do |config|
```
dcarley: [Rubocop's linter](https://github.com/bbatsov/rubocop#lint) notes that:
```
concourse/scripts/spec/configure_aruba.rb:11:25: W: Shadowing outer local variable - config.
    Aruba.configure do |config|
                        ^^^^^^
```

concourse/scripts/spec/configure_aruba.rb
```diff
@@ -0,0 +1,37 @@
...
+      ]
+      if ENV["DEBUG"]
+        config.command_launcher = :debug
+      end
+    end
+  end
+  config.after(:all) do
+    # Ensure we clean up after
+    def rmdir(dir)
+      FileUtils.rmdir(dir) if File.exists? dir
```
dcarley: [Rubocop's linter](https://github.com/bbatsov/rubocop#lint) notes that:
```
concourse/scripts/spec/configure_aruba.rb:25:36: W: File.exists? is deprecated in favor of File.exist?.
      FileUtils.rmdir(dir) if File.exists? dir
                                   ^^^^^^^
```

concourse/scripts/spec/script_testing_helpers.rb
```diff
@@ -0,0 +1,61 @@
+require "pathname"
+require "tempfile"
+
+# Find the directory of the current git repo
+def find_git_repository_path()
+  path = `git rev-parse --show-toplevel`.strip
+  if not File.exists?(path)
```
dcarley: [Rubocop's linter](https://github.com/bbatsov/rubocop#lint) notes that:
```
concourse/scripts/spec/script_testing_helpers.rb:7:15: W: File.exists? is deprecated in favor of File.exist?.
  if not File.exists?(path)
              ^^^^^^^
```

concourse/scripts/spec/configure_aruba.rb
```diff
@@ -0,0 +1,37 @@
...
+    # Aruba requires the workdir path to be a relative directory
+    relative_workdir = Pathname.new(@workdir).relative_path_from(Pathname.pwd)
+
+    Aruba.configure do |aruba_config|
+      aruba_config.working_directory = relative_workdir
+      aruba_config.home_directory = @workdir
+      aruba_config.command_search_paths = [
+        File.join(File.join(__FILE__, "..", "..")),
+      ]
+      if ENV["DEBUG"]
```
saliceti: Enabling DEBUG makes all the tests fail

dcarley: You might have already discussed this in meatspace and maybe it's already been fixed, but it'd be good to include any error messages that you experienced when this happened.

saliceti:
Works OK.
Should we merge commits fe05498, a065015, 6e1d632 to other commits as they don't bring a lot of value seperately?

PR #204, opened 2016-04-14
README: Fix CF link in overview section

dcarley:
## What

This was broken when we combined the BOSH and CloudFoundry pipelines. I
don't think it's worth listing them as separate bullet points when they
point to the same section.

## How to review

View the rendered page and see if it makes sense.

## Who can review

Not @dcarley


mtekel:
links correctly

PR #206, opened 2016-04-15
Update Readme with instructions  on finding the concourse ip

timmow:
**What**

Add instructions to the readme on finding the concourse ip without logging into the console

Read the readme and follow the instructions to try logging into concourse and bosh

I'm not sure if this should be merged or if we should think about moving these
instructions to the makefile, but i'm raising the PR for visibility into the
command I used in case it is useful to others

**Who should review this PR**

Anyone but me

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,16 +263,19 @@ forward some ports to the agent of the new VM.
 Both public and private keys are also uploaded to S3 to be consumed by
 other jobs in the pipelines as resources and/or by us for troubleshooting.
 
-To manually ssh to the *Deployer Concourse*, learn its IP via AWS console and
-download the `id_rsa` file from the s3 state bucket. You will need
-[aws-cli](#aws-cli), to do this:
+To manually ssh to the *Deployer Concourse*, you will need its IP and `id_rsa` key.
+You can get both from command line. You will need [aws-cli](#aws-cli), to do this:
 
 ```
+export CONCOURSE_IP=$(aws ec2 describe-instances \
+--filters 'Name=tag:Name,Values=concourse/0' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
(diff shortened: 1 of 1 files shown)


mtekel:
when I was thinking about using jq once, @dcarley mentioned it's not mentioned in README as a dependency...

dcarley:
> when I was thinking about using jq once, @dcarley mentioned it's not mentioned in README as a dependency...

Good catch. I think we can use the same trick from: https://github.gds/government-paas/aws-account-wide-terraform/pull/34#discussion_r9128
```
➜  paas-cf git:(master) ✗ aws ec2 describe-instances --filters 'Name=tag:Name,Values=concourse/0' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" --query 'Reservations[*].Instances[*].PublicIpAddress' --output text
52.50.65.211
```

timmow:
@dcarley @mtekel Agreed that is much better! Have updated it to use the jmespath expression

README.md
```diff
@@ -267,12 +267,18 @@ To manually ssh to the *Deployer Concourse*, learn its IP via AWS console and
 download the `id_rsa` file from the s3 state bucket. You will need
```
dcarley: The line above this could be amended now to exclude "learn its IP via AWS console".

dcarley:
:+1: 

I think it probably would be better in the `Makefile`. Though, I guess, what we really want is `ssh-deployer` and `ssh-bosh` targets, which is a bit more work. What do you think?

Related: https://www.pivotaltracker.com/story/show/113160603

timmow:
I think it would be better in the `Makefile` ultimately, but the change to the `Readme` seems like a step in the right direction

saliceti:
Tested successfully

PR #207, opened 2016-04-15
[#116983391] Upgrade CF to v233

saliceti:
## What
Story: [Upgrade to CF -N](https://www.pivotaltracker.com/story/show/116983391)

We upgrade CF, diego, stemcell, bosh. This required adjustments to CF configuration and tests properties.

### Prerequisite
This version works with cf version 6.16.1. There are containers ready in docker hub. To use them: 
* open `create-bosh-cloudfoundry.yml`
* replace `docker:///governmentpaas/cf-acceptance-tests` with `docker:///governmentpaas/cf-acceptance-tests#116983391_upgrade_cf_to_v233`
* replace `docker:///governmentpaas/cf-cli` with `docker:///governmentpaas/cf-cli#116983391_upgrade_cf_to_v233`

### Test
Run on top of existing deployment or create a new deployment. Everything should succeed.

### Before merging
Merge PRs on `paas-docker-cloudfoundry-tools` to update `cf-cli` and `cf-acceptance-tests` images.

## Who can review
Anyone but @jimconner @HenryTK or myself.

### DIFF
--- a/manifests/bosh-manifest/deployments/aws/000-bosh-infrastructure.yml
+++ b/manifests/bosh-manifest/deployments/aws/000-bosh-infrastructure.yml
@@ -3,8 +3,8 @@ resource_pools:
 - name: vms
   network: private
   stemcell:
-    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3197
-    sha1: 89a2210b8caf3884855d7db6d48b8863202c7783
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3215
+    sha1: 84c51fed6342d5eb7cd59728c7d691c75b6c1de8
   cloud_properties:
     instance_type: t2.medium
     ephemeral_disk: {size: 40_000, type: gp2}
--- a/manifests/bosh-manifest/reference-bosh-manifest.yml
+++ b/manifests/bosh-manifest/reference-bosh-manifest.yml
@@ -172,8 +172,8 @@ properties:
       -----END RSA PRIVATE KEY-----
 releases:
 - name: bosh
-  sha1: 940956a23b642af3bb24b3cac37c4da746d6f9a9
-  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=253
+  sha1: 6b12652650b87810dcef1be1f6a6d23f1c0c13a7
+  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=255.8
 - name: bosh-aws-cpi
   sha1: a1fe03071e8b9bf1fa97a4022151081bf144c8bc
   url: https://bosh.io/d/github.com/cloudfoundry-incubator/bosh-aws-cpi-release?v=44
--- a/manifests/bosh-manifest/reference-bosh-manifest.yml
+++ b/manifests/bosh-manifest/reference-bosh-manifest.yml
@@ -191,5 +191,5 @@ resource_pools:
   name: vms
   network: private
   stemcell:
-    sha1: 89a2210b8caf3884855d7db6d48b8863202c7783
-    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3197
+    sha1: 84c51fed6342d5eb7cd59728c7d691c75b6c1de8
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3215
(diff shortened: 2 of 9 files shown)


manifests/bosh-manifest/reference-bosh-manifest.yml
```diff
@@ -191,5 +191,5 @@ resource_pools:
   name: vms
   network: private
   stemcell:
-    sha1: 89a2210b8caf3884855d7db6d48b8863202c7783
-    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3197
+    sha1: 84c51fed6342d5eb7cd59728c7d691c75b6c1de8
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3215
```
alext: These changes should be squashed into the commits that made the corresponding changes to the actual manifest.

Having said that, I'm not sure I see much value in the test that compares with the reference manifest. It's the sort of thing that's useful when initially refactoring manifests and sprucing them up, but I'm not sure it has value once that task is complete. We could consider removing this test entirely.

manifests/bosh-manifest/deployments/aws/000-bosh-infrastructure.yml
```diff
@@ -3,8 +3,8 @@ resource_pools:
 - name: vms
   network: private
   stemcell:
-    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3197
-    sha1: 89a2210b8caf3884855d7db6d48b8863202c7783
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3215
+    sha1: 84c51fed6342d5eb7cd59728c7d691c75b6c1de8
```
alext: This bumps the stemcell used for bosh itself, but not for the CF deployment. We should bump both (and possibly also the stemcell used for the deployer-concourse)

manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -316,6 +316,7 @@ properties:
   diego:
     ssl:
       skip_cert_verify: true
+      starting_container_weight: 0.25
```
alext: It would be good to add a reference to the commit message that explains why this change is necessary.

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -27,14 +27,14 @@ meta:
     cloud_controller_ng: {}
 
   api_templates:
+  - name: consul_agent
+    release: (( grab meta.consul_templates.consul_agent.release ))
```
alext: Given the recommendation that this comes first, it's probably worth adding a unit test to ensure that whenever `consul_agent` is present in a job, it's listed first.

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -375,6 +381,8 @@ jobs:
             uaa: {}
       metron_agent:
         zone: z1
+      route_registrar:
+        routes: []
```
alext: Is there any way we can just pull in the `common` package directly instead of having to have this somewhat hacky workaround.

This also sounds like an issue that should be reported upstream. It seems very surprising that `statsd-injector` breaks when `route_registrar` isn't present.

saliceti: As far as I know there is no way to install a package without a job. Unless we create a new fake job in a new bosh-release.
The issue [was reported](https://github.com/cloudfoundry/cf-release/issues/950). Apparently it was fixed in the develop branch, so we can remove the workaround when we upgrade to the right CF version.

alext: Ah, thanks for that link.  Given it's fixed upstream, can we add a `FIXME` comment so that we know to remove this when we upgrade.

manifests/cf-manifest/deployments/aws/000-cf-infrastructure.yml
```diff
@@ -14,8 +14,8 @@ meta:
   default_stemcell:
     name: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
     version: 3197
```
alext: The version needs to change here as well.

saliceti: Done

alext:
I've run into issues while upgrading the etcd cluster. Only the first node is set to serial, which means that nodes 2 and 3 get upgraded at once, resulting in the cluster losing quorum and breaking.  We should update it so that only one node gets upgraded at a time.

alext:
👍  I think this is good to go.  I'll merge tomorrow morning to avoid the cf-cli version change conflicting with the deployment of #213.

PR #210, opened 2016-04-18
[#117356487] Port scan - do not merge

mtekel:
## What

[Check that deployed applications can't see certain things](https://www.pivotaltracker.com/n/projects/1275640/stories/117356487)

We want to ensure that apps running inside CF can't connect to CF components. To verify that, we run port scan from within app container. We check all ports on all CF VMs (including the cell that the app runs on). The app should not be able to connect to anything.

## How to review

Apply. Observe run of the `port-scan` job. It should finish successfully. The job takes about 8 mins.

## Who can review

You! Unless you're me.


mtekel:
There is st. wrong with CF. The scans run either 530s (+-1s very consistent between runs, same for staging and runtime scans), or they get completely stuck and time out. One scan unblocked and finished once I `wsh`d into the container... (after 81 mins). Not sure if the same issue would be present in different CF version...

mtekel:
With the new nmap settings, it seems less likely to get into reduced perf. state.

alext:
I wonder whether this would be better if the orchestration was done by the custom-acceptance-tests - ie have the test suite deploy this nmap app to CF, and collect the results of the test run.

keymon:
Closing this PR. Definitive implementation will be done in other story.

PR #211, opened 2016-04-18
Concourse pipeline linting

Jonty:
## What

Pipecleaner is a small tool I wrote to help spot concourse pipeline errors quickly before we actually run them in concourse. It was written during another story to help with refactoring pipelines and made the whole process substantially faster.

It can check for the following issues:

* Resources being used in a job that have not been defined in the `resources:` block. (Fatal)
* Resources that have been defined in the `resources:` block and are not used in the pipeline. (Warning)
* Resources being used in an `input:` block that have not been `get:`-ted. (Fatal)
* Resources that have been `get:`-ted and are not used in the job. (Warning)
* `output:`s that are not used later in the job. (Warning)

By default it will exit with a nonzero exit code for any Fatal errors, and will exit with a nonzero code for Warnings if you pass the `--fatal-warnings` flag.

If some checks are not desired (e.g. unused outputs) they can be disabled with the `--ignore-types` flag.

Travis is configured to test all concourse pipelines by default, but will only fail the tests on Fatal, not Warning level errors.

## How to review

Run `make lint_concourse`, you should see warnings for minor problems in our pipelines:
```
==  paas-cf/concourse/pipelines/create-bosh-cloudfoundry.yml  ==
* Unused Output: job='smoke-tests', resource='smoke-tests-log'
* Unused Output: job='performance-tests', resource='tests-results'
* Unused Fetch: job='generate-cf-secrets', resource='concourse-tfstate'
* Unused Fetch: job='generate-cf-secrets', resource='vpc-tfstate'

==  paas-cf/concourse/pipelines/destroy-microbosh.yml  ==
* Unused Fetch: job='bosh-init-delete', resource='bosh-tfstate'
```

## Who can review
Someone who isn't @jonty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck check-env-vars
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck lint_concourse check-env-vars
 
 .DEFAULT_GOAL := help
 
@@ -11,7 +11,7 @@ YAMLLINT=yamllint
 check-env-vars:
 	$(if ${DEPLOY_ENV},,$(error Must pass DEPLOY_ENV=<name>))
 
-test: spec lint_yaml lint_terraform lint_shellcheck ## Run linting tests
+test: spec lint_yaml lint_terraform lint_shellcheck lint_concourse ## Run linting tests
 
 spec:
 	cd concourse/scripts &&\
@@ -37,6 +37,9 @@ lint_terraform: dev
 
 lint_shellcheck:
 	find . -name '*.sh' -not -path './vendor/*' | xargs $(SHELLCHECK)
+
+lint_concourse:
+	cd .. && python paas-cf/concourse/scripts/pipecleaner.py paas-cf/concourse/pipelines/*.yml
 
 .PHONY: globals
 globals:
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -169,6 +169,8 @@ jobs:
         image: docker:///governmentpaas/curl-ssl
         inputs:
         - name: paas-cf
+        outputs:
+        - name: ssh-keygen
         params:
           AWS_DEFAULT_REGION: {{aws_region}}
         run:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -38,6 +38,9 @@ lint_terraform: dev
 lint_shellcheck:
 	find . -name '*.sh' -not -path './vendor/*' | xargs $(SHELLCHECK)
 
+lint_concourse:
+	cd .. && python paas-cf/concourse/scripts/pipecleaner.py paas-cf/concourse/pipelines/*.yml
```
alext: Why is the `cd ..` needed here?

Jonty: Where we import external pipeline configuration [from other files](https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-bosh-cloudfoundry.yml#L1145) we refer to the files by their full filepath including the `paas-cf/` component. I _could_ have attempted to determine the root you're in and automatically stripped it, but it seemed cleaner to just switch to the location that concourse will actually run from.

Ugly, but probably easier to understand when it goes wrong.

HenryTK:
If this tool is used in an automated process then we should run it with the `--fatal-warnings` flag, otherwise we would not be forced to clean up code which is sub-standard. Exiting non-zero for warnings seems more like a default behaviour for something called a 'cleaner'.

HenryTK:
The documentation for this script seems to only exist in the PR and commit message. Is there a more visible place for it?

Jonty:
@HenryTK:
1. I have not turned on `--fatal-warnings` as we currently have ~5 warnings in our main manifests and I want to get this merged before fixing them. Also, one of the warnings is non-trivial to fix (`Unused Output: job='smoke-tests', resource='smoke-tests-log'`) as it is due to a shared file between jobs.
2. I was going to write a `README.md` for it, but that would involve moving it to another directory or repo. Suggestions on where it can live are welcome - perhaps I should just add a large docblock to the head of the file?

HenryTK:
@Jonty:

1. Reasonable. You can disregard my comment. It came from a worry that these useful warnings will just become ignored if we only use the script via Travis. We as developers always have the option of running it manually for quality control.
2. A docblock should suffice. README.md may be a bit overkill.

Jonty:
@HenryTK I've added doc to the top of `pipecleaner.py`, does that look ok?

HenryTK:
@Jonty: looking good.

PR #212, opened 2016-04-18
[#117356487] Port scan 2 - do not merge 

combor:
## What

[Check that deployed applications can't see certain things](https://www.pivotaltracker.com/n/projects/1275640/stories/117356487)

This is different than https://github.com/alphagov/paas-cf/pull/210 approach to network scanning.
In this test we are not using `nmap` but our own port scanner written in go by using this library: https://github.com/anvie/port-scanner. This approach brings some pros and cons and we need to decide which one is better.

#### Cons
- nmap is much better as port scanner, it's highly parallel and faster
- this port scanner app scans all host in parallel but does port scanning in serial 
- it's slower as we need ~ 25mins to scan 40 VMs

#### Pros
- much better integration with our tests framework - it runs as a part of custom-acceptance tests
- we've got ( simple ) source code
- we can easily extend it and improve scanning speed ( by parallelising scanning of ports )
- we don't need another custom task to run test - less duplication of efforts

## How to review

Deploy from branch and check if custom-acceptance pass

## Who can review
Not myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1231,6 +1231,7 @@ jobs:
           - get: cf-manifest
             passed: ['cf-deploy']
           - get: bosh-CA
+          - get: concourse-manifest
       - task: generate-test-config
         config:
           platform: linux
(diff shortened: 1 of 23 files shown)


combor:
I just noticed that it fails after I have rebased from master with:
```
FAILED
Server error, status code: 400, error code: 100001, message: The app is invalid: buildpack custom buildpacks are disabled
```

Unfortunately I'm using custom buildpack because of:  https://github.com/anvie/port-scanner/pull/4 . So we need either enable it for testing purposes or merge https://github.com/alphagov/paas-cf/pull/207 which brings Go 1.6

keymon:
Closing this PR. Definitive implementation will be done in other story.

BTW, finally we will likely use nmap.

keymon:
+1 for caching.

PR #213, opened 2016-04-18
[#115869119] Migrate BOSH to S3 / RDS

timmow:
🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 
Do not just merge this - the changes require that we apply this manually to the ci, stage, and prod environments. From the summary in pivotal -

> We have decided we should tear down the cloudfoundry deployment on each environment before applying - this process has been tested by deploying master, running the destroy cloudfoundry pipeline and  applying the changes in the branch.

We have decided to do this because we have the luxury of being able to rebuild these environments, and we are unlikely to need to migrate the bosh state in the future so there is little value in working out how to do this
🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 🔺 

**What**

Story: [Migrate BOSH to S3 / RDS](https://www.pivotaltracker.com/story/show/115869119)

We want to use rds and s3 for bosh so so that we can recreate it if something goes wrong.

This updates the deployment to create the relevant s3 buckets and rds instances, and apply a new instance role to all vms managed by bosh so they are able to access the s3 bucket.

It also moves the secret generation before the bosh terraform run, as we need the secret to be available before the bosh terraform run, and it is better to generate all our secrets in one job rather than having multiple

Checkout this branch

If you have an existing deployment from master - delete the cloud foundry deployment and then apply this. If not building an environment from scratch should work.

Check you can destroy the bosh instance and re run the pipeline and not lose information about the deployment. Don't delete the attached disk or bosh-init will failt and you would have to edit the state file manually.

**Who should review this PR**

* Anyone but me, @jonty, or @saliceti (and probably not @keymon...)


terraform/bosh/rds.tf
```diff
@@ -0,0 +1,61 @@
...
+resource "aws_db_parameter_group" "default" {
+    name = "${var.env}-bosh"
+    family = "postgres9.4"
+    description = "RDS Postgres default parameter group"
+}
+
+resource "aws_db_instance" "bosh" {
+  identifier = "${var.env}-bosh"
+  name = "bosh"
+  allocated_storage = 10
```
jimconner: Utilisation of this DB seems to be around 340MB, so is 10GB allocation perhaps a bit overkill here - we could save a few pennies off our hosting bill with a smaller value.

timmow: Good point, it seems like 5 is the lowest possible value, I will update it

timmow: @jimconner Updated and force pushed, and also fixed a conflict with master

jimconner:
All tested and working superbly - Ready for merging.... just as soon as the Travis checks start passing again.

PR #214, opened 2016-04-18
Enable worldwide access to apps in prod/stage

Jonty:
## What
This PR allows worldwide router access on Staging and Prod so tenants can actually have non-GDS people use them. Staging is exposed so that we have an identical environment to Prod being tested before changes are deployed in Prod.

## How to review
Ugh. Modify your `terraform/dev.tfvars` file to have the same `web_access_cidrs = "0.0.0.0/0"` line that is now in `staging.tfvars` and `prod.tfvars`, then deploy your environment with it.

Push an example application such as [the one](https://github.com/alphagov/paas-cf/tree/master/tests/example-apps/static-app) in our tests.

From a host external to the GDS office `curl` the application and check that it works. If you don't have something to SSH to or a VPN you can use, go to a coffee shop.

## Who can review
No @HenryTK or @jonty.


mtekel:
works

PR #215, opened 2016-04-19
Ensure etcd cluster is updated serially

alext:
## What

Ensure that the etcd cluster is upgraded in a way that maintains the cluster health. This problem was discovered while reviewing #207.

Without this, when upgrading the cluster (eg for a stemcell version
change), the first node will be upgraded, then the other 2 will be
upgraded in parallel, resulting in the cluster losing quorum. This then
requires manual intervention to recover from.

## How to review

Deploy a change that will cause all 3 etcd machines to be re-created (eg a stemcell version bump). Verify that they update cleanly, and that the cluster is healthy afterwards (run `/var/vcap/packages/etcd/etcdctl cluster-health` on one of the nodes).

## Who can review

Anyone but myself.

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -247,6 +247,24 @@ jobs:
           services:
             etcd: {}
 
+  - name: etcd_z2
+    templates: (( grab meta.etcd_templates ))
+    instances: 1
+    persistent_disk: 10024
+    resource_pool: medium_z2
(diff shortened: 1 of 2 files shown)


alext:
The test failure looks to be unrelated to this change. There seem to be problems installing shellcheck in Travis, probably because the version in debian sid that we're installing is no longer compatible with the ubuntu image used in Travis (there are package version conflicts)

mtekel:
Indeed this forces serial updates on etcd.

PR #216, opened 2016-04-19
[#117909253] Add shellcheck binary to the repo.

alext:
## What

Installing shellcheck from debian sid is no longer working, and
resulting in dependency errors (eg https://travis-ci.org/alphagov/paas-cf/builds/124150110).

This adds a shellcheck linux binary to the repo, and updates Travis to
use it. This binary is extracted from the ubuntu wily package version 0.3.7-1.

## How to review

Verify that the Travis build now passes

## Who can review

Anyone but myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,18 @@ env:
 addons:
   apt:
     sources:
-      - debian-sid
       - git-core
     packages:
-      - shellcheck
       - git
 
 before_install:
-  - mkdir ~/bin
-  - export PATH=~/bin:$PATH
+  - |
+    mkdir ~/bin
+    export PATH=~/bin:$PATH
+  - |
+    echo "Fetching shellcheck"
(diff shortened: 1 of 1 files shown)


keymon:
I think it is better to put it as a release as a **temporary solution**, because if not it will be in the git history, increasing the size of the project.

alext:
I've updated this to download from a Github release as suggested.

timmow:
Is it worth making sure [this is cached](https://docs.travis-ci.com/user/caching/#Arbitrary-directories) between builds?

alext:
> Is it worth making sure this is cached between builds?

I don't think so as part of this PR. This is about getting the builds unblocked again, and I don't want to distract from that.  We're already downloading terraform and spruce each time without caching them.

keymon:
Works and LGTM... merging.

PR #217, opened 2016-04-20
Fix bootstrap pipeline.

alext:
## What

#211 updated this pipeline so that it would work with newer versions
of concourse and pass the linter. This fix wasn't complete because the
added outputs created empty directories that were no longer being
populated, meaning that later tasks couldn't find the necessary files.

This updates the pipeline to be closer to our styleguide and have output
names that are distinct from the task names. It is then explicit in
what's added to the output directorties.

## How to review

Run the bootstrap pipeline for a new environment and confirm that it works.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -164,13 +164,13 @@ jobs:
       passed: [init-bucket]
     - get: vpc-terraform-state
 
-    - task: ssh-keygen
+    - task: generate-ssh-keys
       config:
         image: docker:///governmentpaas/curl-ssl
         inputs:
         - name: paas-cf
         outputs:
-        - name: ssh-keygen
+        - name: ssh-keys
         params:
           AWS_DEFAULT_REGION: {{aws_region}}
         run:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-deployer.yml
```diff
@@ -266,13 +267,17 @@ jobs:
             tar xzvf concourse-cert/concourse-cert.tar.gz
             if [ -f concourse.crt ] && [ -f concourse.key ]; then
               echo Certificate and private key already created, nothing to do
-              cp concourse-cert/concourse-cert.tar.gz .
+              cp concourse-cert/concourse-cert.tar.gz generated-concourse-cert
```
jimconner: Is it worth adding a trailing slash after the generated-concourse-cert directory name for clarity? I took me  a few seconds to realise that you weren't copying to a different file name.

jimconner:
I just tried this, and it created the boostrap-deployer OK... but when I tried the create-deployer step, I was met with 
```
cp: can't stat 'ssh-keygen/id_rsa.pub': No such file or directory

vpc-terraform-state

error running command: no matches found for pattern: deploy-vpc/vpc.tfstate
```

concourse/pipelines/create-deployer.yml
alext: Good call, updated.

PR #218, opened 2016-04-20
[#115869119] Remove cf-terraform bosh-ftstate dependency

alext:
## What

The cf and bosh terraform jobs have run in parallel since #213 was
merged. This suggests that the cf-terraform job doesn't depend on the
outputs from bosh-terraform. Reading the code confirms this - there are
no variables in the cf-terraform that match the names of outputs in
bosh-terraform. Additionally the cf-terraform destroy job doesn't rely
on the bosh-ftstate file.

This PR therefore removes the bosh tfstate from the cf-terraform job.

In addition, I've also pinned the bosh-secrets used in the bosh-terraform job to the output of generate-secrets to make this dependency clear

## How to review

Run the create pipeline for a new environment and verify that it works.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -658,11 +658,10 @@ jobs:
           - get: paas-cf
             passed: ['generate-secrets']
           - get: pipeline-trigger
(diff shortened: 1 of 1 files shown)


mtekel:
it does what it says (& not much else)

PR #220, opened 2016-04-22
[#117512347] Service Broker Deployed to Production

combor:
[Service Broker Deployed to Production](https://www.pivotaltracker.com/story/show/117512347)

**Note**

https://github.com/alphagov/paas-aws-broker-boshrelease/pull/1 must be merged first before you can test this PR. 

**What**

As a PaaS Engineer I want an initial version of the Service Broker deployed to Production in order to provide a foundation for further work.

This story doesn't include getting the broker to work (except for the /catalog endpoint), but does include forking the broker and deploying it so it shows up on the marketplace

Checkout this branch and update your pipeline to use it. Deploy the cloudfoundry changes

Login to cloudfoundry and check you can run the marketplace command and see the broker

```
cf marketplace
```

**Who should review this PR**

* Any one but @timmow or @combor

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -90,6 +90,13 @@ resources:
       uri: https://github.com/alphagov/paas-grafana-boshrelease.git
       tag_filter: {{cf_grafana_version}}
 
+  - name: aws-broker-boshrelease
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-aws-broker-boshrelease.git
+      branch: 117512347_rds_service_broker
+      tag_filter: {{cf_aws_broker_version}}
+
   - name: graphite-nozzle
     type: git
     source:
@@ -951,6 +958,7 @@ jobs:
           - get: graphite-statsd-boshrelease
           - get: grafana-boshrelease
           - get: paas-haproxy-release
+          - get: aws-broker-boshrelease
       - task: upload-releases
         config:
           platform: linux
@@ -961,6 +969,7 @@ jobs:
             - name: graphite-statsd-boshrelease
             - name: grafana-boshrelease
             - name: paas-haproxy-release
+            - name: aws-broker-boshrelease
           run:
             path: sh
             args:
@@ -977,6 +986,9 @@ jobs:
 
                 paas-cf/concourse/scripts/bosh_create_and_upload_release.rb paas-haproxy \
                   {{cf-paas-haproxy-release-version}} paas-haproxy-release
+
+                paas-cf/concourse/scripts/bosh_create_and_upload_release.rb aws-broker \
+                {{cf_aws_broker_version}} aws-broker-boshrelease
 
       - task: cf-deploy
         config:
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1055,8 +1055,10 @@ jobs:
               UAA_ENDPOINT=$($VAL_FROM_YAML properties.uaa.url cf-manifest/cf-manifest.yml)
               DOPPLER_ENDPOINT=$($VAL_FROM_YAML properties.loggregator.traffic_controller_url cf-manifest/cf-manifest.yml)
               GRAPHITE_SERVER=$($VAL_FROM_YAML jobs.graphite_z1.networks.cf1.static_ips.0 cf-manifest/cf-manifest.yml)
+              RDS_BROKER_SERVER=$($VAL_FROM_YAML jobs.rds_broker_z1.networks.cf1.static_ips.0 cf-manifest/cf-manifest.yml)
+              RDS_BROKER_PASS=$($VAL_FROM_YAML secrets.rds_broker_admin_password cf-secrets/cf-secrets.yml)
 
-              for var_name in CF_ADMIN CF_PASS FIREHOSE_USER FIREHOSE_PASS API_ENDPOINT UAA_ENDPOINT DOPPLER_ENDPOINT GRAPHITE_SERVER; do
+              for var_name in CF_ADMIN CF_PASS FIREHOSE_USER FIREHOSE_PASS API_ENDPOINT UAA_ENDPOINT DOPPLER_ENDPOINT GRAPHITE_SERVER RDS_BROKER_SERVER RDS_BROKER_PASS; do
                 echo export "${var_name}"=\"$(eval echo \$${var_name})\"
               done > config/config.sh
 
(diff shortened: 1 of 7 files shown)


manifests/cf-manifest/deployments/aws/060-aws-broker-stub.yml
```diff
@@ -0,0 +1,18 @@
...
+    network: cf1
+    stemcell: (( grab meta.stemcell ))
+    env: (( grab meta.default_env ))
+    cloud_properties:
+      instance_type: m3.medium
+      ephemeral_disk:
+        size: 10240
+        type: gp2
+      availability_zone: (( grab meta.zones.z1 ))
+      iam_instance_profile: cf-cloudcontroller
```
alext: This should probably have a dedicated instance profile instead of reusing the cloudcontroller one.

alext:
This is missing acceptance tests. We should add a custom acceptance test that the plan is visible when running `cf marketplace`.

manifests/cf-manifest/scripts/generate-cf-secrets.rb
```diff
@@ -30,6 +30,7 @@
   "loggregator_endpoint_shared_secret" => :simple,
   "consul_encrypt_keys" => :simple_in_array,
   "grafana_admin_password" => :simple,
+  "rds_broker_admin_password" => :simple,
```
alext: This will need to be added to the test fixtures as well (that's why the Travis build is currently failing).

timmow: I fixed that already, I suspect its been rebased away 😢

manifests/cf-manifest/deployments/aws/060-aws-broker-stub.yml
timmow: It should. However, this story wasnt about getting the broker working, to provision rds dbs, it was to get the broker listed in the marketplace - we felt that sorting out the IAM roles / permissions belonged in the story about getting DB provisioning working.

alext: Does this need an instance profile at all then? I think it would be better to not set one than to set this one, which would be a little misleading when it comes to the next story.

manifests/cf-manifest/deployments/060-aws-broker.yml
```diff
@@ -0,0 +1,61 @@
+releases:
+  - name: aws-broker
+    version: v3
```
keymon: Note that this version naming is inconsistent with the other PR: https://github.com/alphagov/paas-cf/pull/221/files#diff-b3a119020efd88e25d5115307a683e4bR32

IMO we should change one of them, regardless other releases are using a different version naming (logsearch/graphite use commit ID)

combor:
I believe all comments are resolved now.

keymon:
As discussing it with @mtekel, he points out that the convention is use the semantic versions for the releases. 

More info:  https://blog.pivotal.io/pivotal-cloud-foundry/products/semantic-yet-flexible-bosh-release-versioning

I changed the HAProxy PR to match that again, and I suggest that this one does the same.

keymon:
I finally managed to get the latest package compiled, but now the broker does not start

```
# monit summary
The Monit daemon 5.2.5 uptime: 23m 

Process 'rds-broker'                Execution failed
System 'system_localhost'           running
```

The error log:
```
# tail /var/vcap/sys/log/rds-broker/rds-broker.stderr.log 
/var/vcap/jobs/rds-broker/bin/rds-broker-ctl: line 16: /var/vcap/packages/rds-broker/bin/rds-broker: No such file or directory
/var/vcap/jobs/rds-broker/bin/rds-broker-ctl: line 16: /var/vcap/packages/rds-broker/bin/rds-broker: No such file or directory
/var/vcap/jobs/rds-broker/bin/rds-broker-ctl: line 16: /var/vcap/packages/rds-broker/bin/rds-broker: No such file or directory

```

The reason is that the binary is called differently:

```
# ls /var/vcap/packages/rds-broker/bin/
paas-rds-broker
```

The reason is that by forking the binary to a new repo called `paas-rds-broker` https://github.com/alphagov/paas-aws-broker-boshrelease/blob/117512347_rds_service_broker/packages/rds-broker/packaging#L10, the ctl script has not being updated: https://github.com/alphagov/paas-aws-broker-boshrelease/blob/117512347_rds_service_broker/jobs/rds-broker/templates/bin/rds-broker-ctl.erb#L16

manifests/cf-manifest/deployments/060-aws-broker.yml
```diff
@@ -0,0 +1,61 @@
...
+                displayName: "AWS RDS Postgres"
+                imageUrl: ""
+                longDescription: "AWS RDS postgres service"
+                providerDisplayName: "Amazon Web Services"
+                documentationUrl: "https://aws.amazon.com/documentation/rds/"
+                supportUrl: "https://forums.aws.amazon.com/forum.jspa?forumID=60"
+              plan_updateable: true
+              plans:
+                - id: "5b8282cf-a669-4ffc-b426-c169a7bbfc71"
+                  name: "5.6-medium"
```
dcarley: Why is it called "5.6"?

tests/acceptance-tests/src/acceptance/rds_broker_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var (
+	brokerName = "rds"
```
dcarley: Should we call this "postgres" so that users are able to easily identify what it is?

manifests/cf-manifest/deployments/060-aws-broker.yml
combor: This a dummy plan that deploys nothing as per acceptance criteria from story description:
> Given A Tenant is logged into the Production platform
> When they run cf marketplace
> Then They can see the Basic Plan Listed

I can change the name to Basic PG plan or something like that.

dcarley: I think it's right to name it after the Postgres version (not MySQL 😉 ) and instance size, because those are the differentiating properties of this plan.

combor: I renamed it to simply `postgres` for now as I don't know what the plan will provide. We need to change this ( and add more plans ) in following stories.

dcarley: It seems confusing that the broker and plan have the same names. Especially as the test is checking for one of those strings. Could you either call it "9.4-medium" (per `rds_properties`) or "basic" for the moment?

combor: OK, renamed.

combor:
OK, so I tested all requested changes and it passes tests in my environment. I hope all will be good this time.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1002,167 +1016,176 @@ jobs:
...
+            run:
+              path: sh
+              args:
+                - -e
+                - -c
+                - |
+                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
+                  . ./config/config.sh
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  if cf service-brokers | grep rds; then exit 0;fi
```
keymon: This check fails, because the broker is called "postgres" not "rds".

I also suggest that you also check the name and the url. something like: `cf service-brokers | grep -e "^postgres[\t ]\+http://${RDS_BROKER_SERVER}"`

PR #221, opened 2016-04-22
[#115931361] Add haproxy in front of gorouter to add HSTS headers (and SSL termination)

keymon:
[#115931361 Ensure correct HSTS headers are being served](https://www.pivotaltracker.com/story/show/115931361)

What?
----

We want to serve [HSTS headers](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) for all HTTPS requests to the apps domains, but it will safeguard existing users from being MITMed over insecure connections and it will improve the user experience when they click on a hostname that doesn't have a protocol.

We want to leave open the option of able to override these headers from the tenant application if they wish.

We decided to use haproxy in front of the gorouter to do the job for the time being, using a custom release as the [current haproxy job from `cf-release`](https://github.com/cloudfoundry/cf-release/tree/master/jobs/haproxy/) does not support adding custom config.

Dependencies
----------------

This should be reviewed with https://github.com/alphagov/paas-haproxy-release/pull/1, which should be merged first and this branch updated to point to the tag.

Context
-------

 * We upload and configure the new haproxy release
 * We inject config in the HAProxy frontends to add the Strict-Transport-Security
   header if it is missing.
 * We add acceptance tests to check the desired behaviour.

Additionally, following the [Boy Scout Rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule) we fix some typo defining the logsearch-shipper job template.

How to test this?
-----------------

Deploy all the platform. All test should pass.

You can check the headers of any app to confirm that the new header is there.

You can also curl any url in the platform:

```
$ curl -Ik https://test.hector.dev.cloudpipelineapps.digital/
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Cf-Routererror: unknown_route
X-Content-Type-Options: nosniff
X-Vcap-Request-Id: 1c3f05f0-72b3-4c35-7498-6c4a8d9544c8
Date: Fri, 22 Apr 2016 17:09:27 GMT
Content-Length: 93
Strict-Transport-Security: max-age=31536000
```

Who?
----

Anyone but @keymon or @HenryTK

### DIFF
--- a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
@@ -29,7 +29,7 @@ releases:
     url: https://bosh.io/d/github.com/cloudfoundry-incubator/etcd-release?v=38
     sha1: c7d303c1b733e77c5c5f3dedfe4dbf75997f30bc
   - name: paas-haproxy
-    version: v1
+    version: 0.0.1
 
 compilation:
   workers: 6
--- a/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/deployments/000-base-cf-deployment.yml
@@ -28,6 +28,8 @@ releases:
     version: 38
     url: https://bosh.io/d/github.com/cloudfoundry-incubator/etcd-release?v=38
     sha1: c7d303c1b733e77c5c5f3dedfe4dbf75997f30bc
+  - name: paas-haproxy
+    version: v1
 
 compilation:
   workers: 6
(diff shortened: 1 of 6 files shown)


manifests/cf-manifest/deployments/000-base-cf-deployment.yml
```diff
@@ -28,6 +28,8 @@ releases:
     version: 38
     url: https://bosh.io/d/github.com/cloudfoundry-incubator/etcd-release?v=38
     sha1: c7d303c1b733e77c5c5f3dedfe4dbf75997f30bc
+  - name: paas-haproxy
+    version: 0.0.1
```
keymon: As https://github.com/alphagov/paas-cf/pull/220/files#r60888205

keymon:
@mtekel I will tag the release and update this branch

mtekel:
works...

PR #222, opened 2016-04-22
[#117704989] Verification of Code for Production

mtekel:
## What

[Verification of Code for Production](https://www.pivotaltracker.com/n/projects/1275640/stories/117704989)

Use git resource from https://github.com/alphagov/paas-git-resource/pull/1 and start verifying commits. This includes modification of concourses (bootstrap & deployer) to use new resource, modification of pipelines, pipeline deploy scripts and adding a list of trusted commiters (which is the 1st signed commit). 

**Note**: You can add your own GPG key (the one you use to sign the commits) to github. It will then display 'Verified' on the commits that match - see the team gpg IDs adding commit in this PR.

## How to review

Deploy or apply against existing environment. Verify functionality by creating signed and unsigned commits and controlling signature verification by editing `SKIP_COMMIT_VERIFICATION` in makefile. 

## Who can review

Not @mtekel or @saliceti


alext:
From the kickoff notes in the story:

> We're not going to enable this in our pipelines at this stage.

I think It's good to review this with the commit verification in place, but it should be disabled before this is merged.  It can be enabled again once https://www.pivotaltracker.com/story/show/117605669 has been played.

mtekel:
Regarding enabling/disabling verification - if you don't specify any GPG IDs, the functionality is disabled. So we can just remove the commit adding team GPG IDs to disable the functionality. Or I don't mind to wait until the signing story is merged...

mtekel:
Rebased, added variable to control signing verification, removed the "hardcoded" resource usage from branch and updated PR description.

dcarley:
> We also enable caching for deployer concourse - which is not a separate PR in order to save one further re-deployment of the deployer (in each environment).

Does this fix [#112582239](https://www.pivotaltracker.com/story/show/112582239)?

It's worth having a separate PR for visibility and testing, but we can be mindful that we should deploy them together to save time.

dcarley:
Could you add a comment to the commit message of a7d9850 to say why we're forking the resource rather than contributing upstream? I had to look back at the story to find out, which will be harder for someone else to do if they want to know in the future.

mtekel:
I'll make a separate PR for caching fix. And it looks like it fixes that story. Unless e.g. upgrade to 1.1 breaks it again. Also, I'm not 100% sure about reasons for forking and I think they should be explained in the resource README instead.

mtekel:
PR for caching: https://github.com/alphagov/paas-cf/pull/224

dcarley:
> Also, I'm not 100% sure about reasons for forking and I think they should be explained in the resource README instead.

Me neither, because I wasn't here, but the kick-off notes in Pivotal say: "we mentioned that running a fork would be satisfactory and it would be reasonable to not submit upstream until we have used it"

A README would be good, but please add it to the commit message regardless - it's low cost and the first place I'd look is `git log -p -G git-resource`.

mtekel:
Updated commit message. I also added 2 commits to readme, describing use of VERIFY_COMMITS and SELF_UPDATE_PIPELINE environment variables.

bleach:
Advance notice that I intend to merge with verification turned on now that [we have played the story to start signing our commits](https://www.pivotaltracker.com/story/show/117605669). Please shout if you think that's a bad idea.

I'll do so once I've finished reviewing.

bleach:
I've tested so far by creating my own branch, deploying a new concourse and then pushing a correctly signed commit to it, which was deployed as expected. Unfortunately the acceptance tests have subsequently failed on both the occasions I've run them in my environment. 

Failures seem to be unrelated and were different each time:

```
• Failure [73.956 seconds]
Security Groups [It] allows external and denies internal traffic during staging based on default staging security rules

• Failure in Spec Teardown (AfterEach) [179.517 seconds]
Service Instance Lifecycle Asynchronous operations when there is an existing service instance when there is an app [AfterEach] when there is an existing binding can unbind a service instance 
/var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry/cf-acceptance-tests/services/service_instance_lifecycle_test.go:433

  Timed out after 45.000s.

  /var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry/cf-acceptance-tests/helpers/app_helpers/app_helpers.go:55
```

I don't want to merge on failing tests and I'm going to be unavailable to do further review or work on this story until Tuesday, so I'm unassigning myself now.

I think what needs to be done next:
- test that pushing an unsigned commit doesn't result in changes being deployed
- test that pushing a commit signed with an untrusted key doesn't result in changes being deployed
- ensure that we see the acceptance tests pass with these changes

bleach:
Since I had the environment/branch already setup I spent 10 minutes confirming that if I push a commit either not signed, or signed by an untrusted key, the resource fails and I can see the message:

`The commit 11340357370fff9af41fb45a4ea01a7cfaafa255 is not signed`

So all that remains is to confirm that the acceptance tests can pass.

concourse/scripts/pipelines-deployer.sh
```diff
@@ -29,9 +29,15 @@ EOF
 
 generate_vars_file > /dev/null # Check for missing vars
 
+gpg_ids="$(awk '{printf "\\n        - %s ", $1}; END {printf "\\n"}' "${SCRIPT_DIR}"/../../.gpg-id)"
```
keymon: Rather than creating a string with new lines and the right amount of spaces, I suggest you use an inline list `[gpg1,gpg2]`. This way you don't need to "know" the exact number of spaces in this script:

For instance:

```diff
--- a/concourse/scripts/pipelines-deployer.sh
+++ b/concourse/scripts/pipelines-deployer.sh
@@ -32,9 +32,9 @@ generate_vars_file > /dev/null # Check for missing vars
 
 generate_manifest_file() {
   if ${COMMIT_VERIFICATION} ; then
-    gpg_ids="$(awk '{printf "\\n        - %s ", $1}; END {printf "\\n"}' < "${SCRIPT_DIR}"/../../.gpg-id)"
+    gpg_ids="[$(cat "${SCRIPT_DIR}/../../.gpg-id" | xargs | tr ' ' ',')]"
   else
-    gpg_ids="null"
+    gpg_ids="[]"
   fi
 
   # This exists because concourse does not support multiline value interpolation by design

```

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -38,6 +38,7 @@ prepare_environment() {
   cf_release_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.cf.version "${cf_manifest_dir}/000-base-cf-deployment.yml")
   cf_graphite_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.graphite.version "${cf_manifest_dir}/055-graphite.yml")
   cf_grafana_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.grafana.version "${cf_manifest_dir}/055-graphite.yml")
+  gpg_ids="$(awk '{printf "\\n        - %s ", $1}; END {printf "\\n"}' < "${SCRIPT_DIR}"/../../.gpg-id)"
```
keymon: Rather than creating a string with new lines and the right amount of spaces, I suggest you use an inline list `[gpg1,gpg2]`. This way you don't need to "know" the exact number of spaces in this script:

For instance:

```diff
--- a/concourse/scripts/pipelines-deployer.sh
+++ b/concourse/scripts/pipelines-deployer.sh
@@ -32,9 +32,9 @@ generate_vars_file > /dev/null # Check for missing vars
 
 generate_manifest_file() {
   if ${COMMIT_VERIFICATION} ; then
-    gpg_ids="$(awk '{printf "\\n        - %s ", $1}; END {printf "\\n"}' < "${SCRIPT_DIR}"/../../.gpg-id)"
+    gpg_ids="[$(cat "${SCRIPT_DIR}/../../.gpg-id" | xargs | tr ' ' ',')]"
   else
-    gpg_ids="null"
+    gpg_ids="[]"
   fi
 
   # This exists because concourse does not support multiline value interpolation by design

```

concourse/scripts/pipelines-deployer.sh
keymon: Same as https://github.com/alphagov/paas-cf/pull/222/commits/37be10a55cb8422c502d0e4a24fec842063ecc9d#r61403722

keymon:
I am not sure about the idea of having a hidden file to drive the verification `.gpg-id`. In the password store they use a hidden file to hide the files from the output. 

But we want to make explicit the existence of this verification and these files.

keymon:
Please see comment https://github.com/alphagov/paas-cf/pull/222/files/166ed4db997dd4d8af2f93c2ee8473707c05b640..37be10a55cb8422c502d0e4a24fec842063ecc9d#r61403722 (hidden by new commits)

alext:
> In case you want to temporarily disable latest commit signature verification,
> e.g. for development purposes where you don't want to sign every temporary commit,
> you can use VERIFY_COMMITS environment variabe set to false.

Can we make this the default for dev deployments.  These are normally deployed from a branch other than master, and therefore will rarely work with commit verification turned on.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -39,6 +40,12 @@ prepare_environment() {
   cf_graphite_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.graphite.version "${cf_manifest_dir}/055-graphite.yml")
   cf_grafana_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.grafana.version "${cf_manifest_dir}/055-graphite.yml")
 
+  if ${COMMIT_VERIFICATION} ; then
```
alext: Will this do the right thin if `VERIFY_COMMITS` is set to `"false"`?  I wonder whether it would be better to invert the sense of this variable - ie name it `SKIP_COMMIT_VERIFICATION`, and then test for empty string here. This would be more consistent with other toggles we have in place.

mtekel: I don't like testing for empty strings, as you could then set SKIP_COMMIT_VERIFICATION=false or SKIP_COMMIT_VERIFICATION=no - and it would actually mean true/yes.

alext: That's true, but also true of all the other variables at the moment.  I think it's better to be consistent.

keymon: I propose we check `[ "${SKIP_COMMIT_VERIFICATION}" == "true" ]`

keymon:
So far LGTM and it works as expected. 

Only pending the minor comments about shell scripting.

mtekel:
Updated.

PR #224, opened 2016-04-26
[#112582239] Enable caching on Deployer Concourse

mtekel:
## What

[Fix caching for Concourse resources](https://www.pivotaltracker.com/n/projects/1275640/stories/112582239)

## How to review

Deploy/apply. When you run pipeline and the resource (e.g. paas-cf) is used for the 2nd time, in the resource logs, you should see `using version of resource found in cache` instead of resource being retrieved again (e.g. `Cloning into '/tmp/build/get'...` in case of paas-cf)

**Note**: if you want to save one concourse re-deploy, then you can merge together with https://github.com/alphagov/paas-cf/pull/222

## Who can review

not @mtekel

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -104,6 +104,8 @@ jobs:
         listen_address: 0.0.0.0:7777
         graph_cleanup_threshold_in_mb: 3072
       groundcrew:
+        baggageclaim:
+          url: "http://127.0.0.1:7788"
         tsa:
           host: 127.0.0.1
         additional_resource_types:


manifests/concourse-manifest/concourse-base.yml
```diff
@@ -99,6 +102,7 @@ jobs:
         forward_host: 127.0.0.1
         atc:
           address: 127.0.0.1:8080
+          old_resource_grace_period: 180m
```
dcarley: This [option is documented as](http://bosh.io/jobs/atc?source=github.com/concourse/concourse&version=1.1.0#p=old_resource_grace_period):

> old_resource_grace_period: 5m 
> How long to cache the result of a get step after a newer version of the resource is found. Use Go duration format (1m = 1 minute).

Is raising this going to have any negative effects? Do it need to be so high?

mtekel: Initially, concourse announced default caching period was 24 hours. Then they realized they set it to 5 mins instead. Hence I'd say both values are (and anything in between is) safe. As this is caching only, it should not affect anything. Resource `check` methods should properly return new versions, which would invalidate old cache. I set this to 3 hours so it can last for 2 pipeline passes (as currently, a single pass including acceptance tests can take ~1h10-20 mins). I can also imagine that caching uses some disk space. I think if we are running out of it, we should increase it instead of decreasing cache period (or disabling caching).

dcarley: Is this the reference for 24 hours?

- Added: https://github.com/concourse/concourse/blob/13b003eb5a252801d6abdaa5037c7f6fedbf4f48/docs/release-notes/v0.64.0.any#L6-L57
- Clarified: https://github.com/concourse/concourse/blob/9bb58ca099aab471521edd819584594778e38a5e/docs/release-notes/v0.65.0.any#L13-L14

v0.69.0 says that the latest version of a resource is always kept:

- Release notes: https://github.com/concourse/concourse/blob/9bb58ca099aab471521edd819584594778e38a5e/docs/release-notes/v0.69.0.any#L50-L53
- Tests: https://github.com/concourse/atc/blob/e10b1d7087b6e451deb9475441f074f21ca601e0/lostandfound/baggage_collector_test.go#L418

In which case, do we benefit much from raising the default? I'm inclined to leave it as-is.

mtekel: I'm a bit confused now. The resources should always provide the latest version in check. That one is cached indefinitely. So why keep previous versions? Maybe it could be useful for debug (hijacking)?

mtekel:
Removed the 3 hour setting as it seems that latest version of resource is really cached forever (in my case it was cached at least for 1 hour, which is how long acceptance tests took).

dcarley:
Thanks for checking. I'm going to re-test after the variable name change.

dcarley:
Works 👍 
```
using version of resource found in cache
```
I'll merge tomorrow when I figure out how to sign.

PR #226, opened 2016-04-26
[#118089721] Split apps and system domain external certs

alext:
**🚨 For review only.  DO NOT MERGE YET.** Please review and indicate in the comments whether it's good to merge.  I'll then merge and deploy myself.

## What

When we came to buy real certs for production and staging, it wasn't
feasible to get a single cert that covered multiple wildcard domains.
We've therefore had to get separate certs for the system and apps
domains.

This updates the pipelines to use separate certs for the system and apps domains. It also updates the dev/ci cert generation to generate separate certs for these.

This PR also updates the metrics ELB to use the system domain cert, instead of a separately generated one.

## How to review

Note: Due to the certs and ELBs being managed in separate terraform runs, it's not possible for this to apply cleanly to an existing environment, so it has to be applied as follows:

When it comes to applying to environments beyond dev, It may be best for me to handle the deployment as I have the context around it.

### dev/ci

* Run the pipeline, the `upload-certs-to-aws` task in the `cf-terraform` job will error
* Log onto the AWS console, and update the following ELBs to use a different cert (it doesn't matter which): `cf-cc`, `cf-doppler`, `cf-loggregator`, `cf-router`, `cf-uaa`.
* Re-run the cf-terraform job, it should run to completion.

### staging/prod

These deployments depend on https://github.gds/government-paas/credentials-high/pull/9

* Run the `manually_upload_certs` make target. It will partially complete, and then error
* Run the pipeline, which should run successfully.
* Re-run the `manually_upload_certs` make task, and it should complete successfully this time.

## Who can review

Anyone but myself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,8 +121,10 @@ manually_upload_certs: ## Manually upload to AWS the SSL certificates for public
 	# check password store and if varables are accesible
 	$(if ${CERT_PASSWORD_STORE_DIR},,$(error Must pass CERT_PASSWORD_STORE_DIR=<path_to_password_store>))
 	$(if $(wildcard ${CERT_PASSWORD_STORE_DIR}),,$(error Password store ${CERT_PASSWORD_STORE_DIR} does not exist))
-	@PASSWORD_STORE_DIR=${CERT_PASSWORD_STORE_DIR} pass certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/router_external.crt > /dev/null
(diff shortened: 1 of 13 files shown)


HenryTK:
@alext: you updated the metrics ELB to use the system domain certificate, should the same be applied to the logsearch ELB?

HenryTK:
Other than the logsearch comment above it is all fine to merge.

alext:
> should the same be applied to the logsearch ELB?

Good spot.  I've updated this to include the logsearch ELB as well.

PR #228, opened 2016-04-27
var_from_yaml: Test CLI and return non-zero exit code on errors

dcarley:
## What

Refactor the existing tests for `val_from_yaml` to use the command line
interface instead of the underlying Ruby object.

This matches how we use the script in our Concourse pipelines, it exercises
more of the code, and allows us to make assertions about the standard output
and exit codes.

.. then ..

Return a non-zero exit code when the script is unable to find the user
provided key in the YAML document. This decreases the chance of another
script silently using an empty value, such as a password.

## How to review

- Look at the code changes and confirm that they make sense.
- Confirm that the Travis CI build completes successfully.

## Who can review

Not @dcarley

### DIFF
--- a/concourse/scripts/spec/val_from_yaml_spec.rb
+++ b/concourse/scripts/spec/val_from_yaml_spec.rb
@@ -23,15 +23,15 @@ val2: b
     EOF
   end
 
-  it "returns nil for non existing properties" do
+  it "exits non-zero with no output for non existing properties" do
     run_with_fixture("x.y.z")
-    expect(last_command_started).to have_exit_status(0)
+    expect(last_command_started).to have_exit_status(1)
     expect(last_command_started.output).to be_empty
   end
 
-  it "returns nil when trasversing simple values" do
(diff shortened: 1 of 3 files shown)


concourse/scripts/spec/val_from_yaml_spec.rb
```diff
@@ -1,52 +1,73 @@
...
+    run_with_fixture("foo.bar")
+    expect(last_command_started).to have_exit_status(0)
+    expect(last_command_started.output).to eq <<-EOF
+---
+val1: a
+val2: b
+    EOF
   end
+
   it "returns nil for non existing properties" do
```
alext: I wonder whether this test description should be updated to reflect the commandline behaviour (exits non-zero with no output)

dcarley: Yep, good point. Will do.

dcarley: Rebased into the last commit.

PR #229, opened 2016-04-27
[#117706115] Automated Organisation and Org Manager Creation

keymon:
[#117706115 Automated Organisation and Org Manager Creation](https://www.pivotaltracker.com/story/show/117706115)

What
----

We want to provide a automated script to create organisations, org managers
and email the password.

Context
-------

We added ./scripts/create-tenant.sh script

./scripts/create-tenant.sh will create a user and organisation in the CF
API where you are currently logged in and send an email to the user
if the password changes.

The script will leverage the cf client to create the organisation, space
and user, and use AWS cli with SES to send emails.

How to test
-----------

Prerequisites:

 1. You must have installed the latest AWS client and CF client.
 2. You must have your AWS credentials exported and login into CF with an admin user.
 3. Your email account must be verified as a account to receive emails from SES.
    Later in prod we will remove the account from the SES sandbox.
    You can do it with:
    ```
aws ses verify-email-identity --email-address hector.rivas.gandara@digital.cabinet-office.gov.uk --region eu-west-1
```

Then run the script to create your account. You will get an email with the credentials

```
./scripts/create-tenant.sh -o orgname -e "hector.rivas.gandara@digital.cabinet-office.gov.uk"
```

You can test to recreate it:

```
./scripts/create-tenant.sh -o orgname -e "hector.rivas.gandara@digital.cabinet-office.gov.uk" -r
```

Who?
----

Anyone but @keymon or @jimconner

### DIFF
--- a/scripts/create-tenant.sh
+++ b/scripts/create-tenant.sh
@@ -1,0 +1,229 @@
+#!/bin/bash
+set -euo pipefail
+SCRIPT=$0
+
+###########################################################################
+# Defaults
+DEFAULT_SPACE=sandbox
+FROM_ADDRESS='the-multi-cloud-paas-team@digital.cabinet-office.gov.uk'
+# shellcheck disable=SC2016
+SUBJECT='Welcome to the Government PaaS'
+# shellcheck disable=SC2016,SC1078
+MESSAGE='Hello,
+
+Your account has been created for the  Government PaaS service.
+
+Your organisation is \"${ORG}\" and your login and password:
+
+ - login: ${EMAIL}
+ - password: ${PASSWORD}
+
+To begin using the PaaS you will need to install the cf command line client
+and execute the following commands.
+
+cf target https://api.cloud.service.gov.uk
+cf login
+
+Further documentation on the cf command line client is available from
+http://docs.cloudfoundry.org/cf-cli/install-go-cli.html
+
+
+Regards,
+Government PaaS team.
+'
+
+
+###########################################################################
+usage() {
+  cat <<EOF
+Usage:
+
+  $SCRIPT [-r] -e <email> -o <orgname>
+
+$SCRIPT will create a user and organisation in the CF service where you
+are currently logged in and send an email to the user if the password changes.
+
+Nothing will change if the organisation or the user already exists
+(unless the -r flag is used, which recreates the user with a new password).
+This way you can add a user to multiple organisations by running the script
+multiple times.
+
+Requirements:
+
+ * You must login with the cf client with an "admin" user.
+ * You must have a functional aws client with credentials configured.
+
+Where:
+  -r           Delete/recreate the user. The user will be recreated
+               and the password reset.
+
+  -e <email>   User email to add and configure as organization and
+               space manager. If the user is created or recreated,
+               they will receive a mail with the new password.
+
+  -o <orgname> Organisation to create and add the user to. If the
+               organisation already exists the script will carry on.
+
+EOF
+  exit 1
+}
+
+load_colors() {
+  if [ -t 1 ] ; then
+    export ESC_SEQ="\x1b["
+    export COL_RESET="${ESC_SEQ}39;49;00m"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/create-tenant.sh [hunk 1]
```diff
@@ -0,0 +1,231 @@
...
+  fi
+
+  if ! aws ses get-send-quota >/dev/null 2>&1; then
+    abort "You must have AWS cli installed and configured with valid credentials. Test it with: aws ses get-send-quota"
+  fi
+
+}
+
+generate_password() {
+  PASSWORD=$(tr -dc _A-Z-a-z-0-9 < /dev/urandom | head -c32 || true)
```
alext: This errors for me.  I get `tr: Illegal byte sequence`. This is on OS X.

scripts/create-tenant.sh
```diff
@@ -0,0 +1,231 @@
...
+
+  if [ -z "${ORG:-}" ]; then
+    abort_usage "Org must be defined"
+  fi
+
+  if ! aws ses get-send-quota >/dev/null 2>&1; then
+    abort "You must have AWS cli installed and configured with valid credentials. Test it with: aws ses get-send-quota"
+  fi
+
+}
```
alext: It would be good to check that `cf` has a valid login to the API in this step.

scripts/create-tenant.sh
alext: Instead of using `tr`, it's probably easier to use openssl here (eg `openssl rand -hex 16` or `openssl rand -base64 21 | tr '/+' '-_'`)

scripts/create-tenant.sh
```diff
@@ -0,0 +1,231 @@
...
+get_subject() {
+  eval "echo ${SUBJECT}" | \
+    sed 's/\([^\/]\)"/\1\\"/g'
+}
+
+# Expand variables from message body, escaping new lines and quotes
+get_body() {
+  eval "echo \"${MESSAGE}\"" | \
+    sed ':a;N;$!ba;s/\n/\\n/g'| \
+    sed 's/\([^\/]\)"/\1\\"/g'
```
alext: These don't do the right thing with BSD sed (as found on OS X).

The email sending results in an error:
```
Error parsing parameter '--message': Invalid JSON: Invalid control character at: line 9 column 26 (char 170)
JSON received:
    {
      "Subject": {
        "Data": "Welcome to the Government PaaS",
        "Charset": "utf8"
      },
      "Body": {
        "Text": {
          "Data": "Hello,

Your account has been created for the  Government PaaS service.

Your organisation is \"alext-test\" and your login and password:

 - login: me@example.com
 - password: super_secret

To begin using the PaaS you will need to install the cf command line client
and execute the following commands.

cf target https://api.cloud.service.gov.uk
cf login

Further documentation on the cf command line client is available from
http://docs.cloudfoundry.org/cf-cli/install-go-cli.html


Regards,
Government PaaS team.",
          "Charset": "utf8"
        }
      }
    }
```

alext:
Given all these portability issues, I wonder whether this script would be better written in something like ruby, where we would avoid some of these issues, and it could be unit tested etc.

keymon:
> Given all these portability issues, I wonder whether this script would be better written in something like ruby, where we would avoid some of these issues, and it could be unit tested etc.

As commented in person: Yeah, I understand, but I wanted to keep the simplest version with the less possible dependencies. This is also a temporary solution to create users, so spending too much time implementing this might be overkill.

I will fixed these issues and tested on macosx, should work now.

scripts/create-tenant.sh
keymon: It will fail anyway a report a reasonable error. That is why we avoid that.

scripts/create-tenant.sh
keymon: Fixed. I added the `LC_CTYPE=C` variable to avoid tr get confused.

I did not use openssl to avoid another dependency.

PR #232, opened 2016-04-28
[#118552679] bosh renderer can read properties under jobs

saliceti:
## What
Story: [Make Concourse 1.1 work with bosh-init](https://www.pivotaltracker.com/story/show/118552679)

bosh renderer was only considering global properties. But some
manifests (like our concourse manifest) don't provide global
properties but use properties under the jobs.

This now considers both, with the job ones taking priority.

## How to review
Test with the concourse manifest for example.

* retrieve the concourse manifest from S3
* get https://github.com/alphagov/paas-concourse and checkout our current version (v0.74.0)
* This should render the template:

```
$./render.rb [...]/paas-concourse/jobs/atc/templates/atc_ctl.erb [...]/paas-concourse/jobs/atc/spec concourse-manifest.yml concourse
```
* run the tests with `make test`

## Who can review
Anyone but @HenryTK or myself

### DIFF
--- a/tests/bosh-template-renderer/render.rb
+++ b/tests/bosh-template-renderer/render.rb
@@ -2,10 +2,10 @@
 require 'yaml'
 require File.join(File.dirname(__FILE__), 'render_lib.rb')
 
-template_path, spec_path, manifest_path = ARGV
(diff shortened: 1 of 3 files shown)


keymon:
LGTM

PR #234, opened 2016-04-29
Prune base_job_templates from cf manifest

mtekel:
## What

base_job_templates meta is used to define templates for diego jobs.
After rendering, these templates are present in the diego jobs. The
base_job_templates tree (which is on top level with properties,
compilation, releases etc.) has no purpose anymore and is essentially
"dead yaml". Prune it like we prune other scaffolding.

## How to review

Apply against existing environment. Check that there are no changes in the deploy job.

## Who can review

not @mtekel

### DIFF
--- a/manifests/cf-manifest/deployments/040-cf-properties.yml
+++ b/manifests/cf-manifest/deployments/040-cf-properties.yml
@@ -1,15 +1,3 @@
-meta:
-  environment: ~
-  default_quota_definitions:
-    default:
-       memory_limit: 2048
-       total_services: 10
-       non_basic_services_allowed: false
-       total_routes: 1000
-  secrets:
-    consul_ca_cert: (( grab secrets.bosh_ca_cert ))
-    bbs_ca_cert: (( grab secrets.bosh_ca_cert ))
-
 properties:
   app_ssh: ~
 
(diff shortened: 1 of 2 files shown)


alext:
I wonder whether instead of having another top-level item that's being pruned we should rename this item to something like `meta.diego_templates`

mtekel:
That can be done as well (see the 2nd commit), but it's much bigger change (in terms of diff/lines changed).

dcarley:
> it's much bigger change (in terms of diff/lines changed)

Omitting whitespace changes by adding `?w=1` to the pull request URL or using `git (show|diff) -b` can make it a bit easier to review.

manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -448,85 +436,96 @@ properties:
...
+  default_quota_definitions:
+    default:
+       memory_limit: 2048
+       total_services: 10
+       non_basic_services_allowed: false
+       total_routes: 1000
+  secrets:
+    consul_ca_cert: (( grab secrets.bosh_ca_cert ))
+    bbs_ca_cert: (( grab secrets.bosh_ca_cert ))
+  base_job_templates:
```
alext: I think this should be renamed to `diego_templates` to make it clear what it's scope is.

alext:
> That can be done as well (see the 2nd commit), but it's much bigger change (in terms of diff/lines changed).

I think the result is better for it though.  We should squash these 2 commits before merging though.

mtekel:
updated

PR #235, opened 2016-04-29
Add CONCOURSE_IP to show environment script

mtekel:
## What

In the recent days I found myself needing to get concourse IP way too often. I have realized that other team members might have to do the same. Enhance environment script to show CONCOURSE_IP as well.

## How to review

run `make <environment> showenv` and check that the output now also contains CONCOURSE_IP

## Who can review

not @mtekel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,9 @@ pipelines: ## Upload pipelines to Concourse
 .PHONY: showenv
 showenv: ## Display environment information
 	$(eval export TARGET_CONCOURSE=deployer)
+	@echo CONCOURSE_IP=$$(aws ec2 describe-instances \
+		--filters 'Name=tag:Name,Values=concourse/0' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
+		--query 'Reservations[].Instances[].PublicIpAddress' --output text)
 	@concourse/scripts/environment.sh
 
 .PHONY: manually_upload_certs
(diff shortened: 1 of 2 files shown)


dcarley:
👍  for this.

Can you replace the `aws s3` command that's in the README with `showenv`?

https://github.com/alphagov/paas-cf#ssh-to-deployer-concourse-and-microbosh

alext:
This does mean that the `aws ec2` command will be called for most of the targets in the `Makefile`. This includes the self-update-pipeliens concourse job.  I wonder whether there's a way to restrict this to only apply to the `showenv` target...

mtekel:
I can add it to makefile directly instead...

dcarley:
Yeah, makes sense.

mtekel:
Updated. Initially I didn't want to put it there because of Makefile bloat. I also updated Readme, but left the original script there as it is a convenient copy-paste, whereas using make showenv requires additional step where you copy-paste outputs on the commandline.

dcarley:
I don't want to have it duplicated in the README. It's a long command and we're unlikely to remember to update both places if we ever change the arguments.

I think you can change the README to use `eval $(make dev showenv)` or maybe even `eval $(make dev showenv | grep CONCOURSE_IP=)`

mtekel:
updated

dcarley:
Followed the README and it worked 👍

PR #237, opened 2016-04-29
[#118089721] CI does not use manually uploaded certs.

alext:
## What

CI is using self-generated certs, so it should not be set to skip the
upload of these generated certs.

## How to review

Verify that the change looks sensible.

## Who can review

Anyone but myself.


keymon:
LGTM

PR #239, opened 2016-04-29
[#116230113] Revert "Workaround terraform codecommit issue"

dcarley:
**NOTICE: alphagov/paas-docker-cloudfoundry-tools#49 needs to be merged first**

## What

This reverts commit 15367aeb79f73e5c0b3cc83d81f9787d1ee1ea40.

This has been fixed upstream in Terraform 0.6.15 which we're upgrading to in
the PR: alphagov/paas-docker-cloudfoundry-tools#49

The output file is passed between two tasks, rather than into an s3
resource, so we don't need to do any other cleanup.

## How to review

You need to run the `create-deployer` pipeline twice:

1. first without the codecommit repo (or branch) already there
1. then with the codecommit repo (or branch) already there

I did manage to test this on an existing environment, but it was a lot of hassle because:

- it seems we don't have permissions to delete codecommit repos using our ordinary credentials so you have to hijack an `awscli` container on the Bootstrap Concourse
- Terraform gets upset if you delete the resource it was managing so you have to remove it from the statefile
- it might have been easier to delete the branch from a container on the Deployer, but I'd already got so far..

Frankly it's probably easier if you create a new environment with a different `DEPLOY_ENV`:

1. Checkout this branch: `git checkout feature/116230113-revert_codecommit_workaround`
1. Create a Bootstrap: `BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev bootstrap`
1. Run the `create-deployer` pipeline once and confirm that it completes.
1. Run the `create-deployer` pipeline again  and confirm that it completes.
1. Run the `destroy-deployer` pipeline.
1. Destroy the Boostrap: `make dev bootstrap-destroy`

## Who can review

Not @dcarley

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -280,41 +280,6 @@ jobs:
         params:
           file: generated-concourse-cert/concourse-cert.tar.gz
 
-    # Temporary task to retrieve the codecommit default branch if it exists
-    - task: get-codecommit-default-branch
-      config:
-        image: docker:///governmentpaas/awscli
-        inputs:
-        - name: paas-cf
-        outputs:
-        - name: codecommit-default-branch
-        params:
-          DEPLOY_ENV: {{deploy_env}}
-        run:
-          path: sh
-          args:
-          - -c
-          - |
-            aws codecommit get-repository \
-              --region us-east-1 \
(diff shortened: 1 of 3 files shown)


alext:
This works cleanly now. Tested 2 runs with a new deployment.

PR #240, opened 2016-05-03
[#115870117] Use FQDN for BOSH

combor:
## What

This PR changes the way how bosh-agents are connecting to microBOSH instance. Currently we are using hardcoded IP and we want to use FQDN instead. FQDN is based on `${SYSTEM_DNS_ZONE_NAME}` and points to local IP of BOSH instance. The IP depends on AZ we want to deploy BOSH to, currently AZ is hardcoded to `eu-west-1a` and following story should provide a way of specifying BOSH AZ by environmental variable.  IP to AZ mapping is done by terraform. 

## How to review

- destroy your CF and microBOSH deployment
- run create-bosh-cloudfoundry pipeline
- login to bosh by executing `make dev bosh-cli`
- all VMs should be in `running` state

## Who can review

Not @combor, @keymon, @HenryTK


terraform/bosh/outputs.tf
```diff
@@ -15,7 +15,19 @@ output "region" {
...
 output "bosh_subnet_id" {
-  value = "${element(split(",", var.infra_subnet_ids), 0)}"
+  value = "${element(split(",", var.infra_subnet_ids), lookup(var.zone_index, var.bosh_az))}"
+}
+
+output "bosh_subnet_cidr" {
+  value = "${lookup(var.infra_cidrs, concat("zone", lookup(var.zone_index, var.bosh_az)))}"
+}
+
+output "default_gw" {
```
alext: I think this needs a more specific name to make it clear that it's the default gateway for the Bosh director.

alext:
The "Fix BOSH reference manifest" commit should be squashed into the commit that made the corresponding changes to the manifest. I think that's better than having separate commits to fix the tests.

manifests/bosh-manifest/deployments/010-bosh-template.yml
```diff
@@ -81,7 +85,7 @@ jobs:
     ntp: (( grab meta.ntp ))
 
     agent:
-      mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" meta.bosh_service_ip ":4222" ))
```
alext: We should also remove `meta.bosh_service_ip` from the manifest if it's no longer used.

manifests/bosh-manifest/deployments/010-bosh-template.yml
```diff
@@ -9,7 +9,11 @@ meta:
     recursor: 8.8.8.8
 
   default_agent:
-    mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" meta.bosh_service_ip ":4222" ))
+    mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" terraform_outputs.bosh_fqdn ":4222" ))
+
+  director_instances:
```
saliceti: Is this used anywhere?

terraform/bosh/outputs.tf
saliceti: `bosh_default_gw` seems reasonable

manifests/bosh-manifest/deployments/010-bosh-template.yml
combor: It was used but I replaced it with `terraform_outputs.bosh_fqdn` as we want to switch to hostname.

terraform/bosh/outputs.tf
combor: This is done.

manifests/bosh-manifest/deployments/010-bosh-template.yml
combor: You mean director_instances? I'm not sure. It might not be used but...frankly speaking, this manifest needs a proper cleaning in the future.  I wanted to do a boy scout rule but that would ended up doing serious changes and I though it's not in the scope of this spike/story.

combor:
>The "Fix BOSH reference manifest" commit...

Done.

manifests/bosh-manifest/deployments/aws/090-bosh-aws-stub.yml
```diff
@@ -21,7 +21,6 @@ meta:
 
   bosh_private_ip: (( grab terraform_outputs.microbosh_static_private_ip ))
```
saliceti: We can probably remove `bosh_private_ip` and `bosh_public_ip` and use terraform_outputs directly when reserving the IPs.
`bosh_public_ip` is also used in `default_dns` but I don't think we use it.

saliceti:
It fails destroying bosh at task cleanup-orphaned-disks:

```
./paas-cf/concourse/scripts/bosh_login.sh: line 6: 2: parameter not set
```

manifests/bosh-manifest/deployments/aws/090-bosh-aws-stub.yml
combor: Yes we can, but I'm not sure if it's a part of this story to clean up bosh manifest. I noted this in our pipeline improvements document.

combor:
> It fails destroying bosh at task cleanup-orphaned-disks:

Fixed. Probably id got broken during a rebase.

PR #244, opened 2016-05-04
[#117705909] Create ephemeral user for use duing tests

jimconner:
## What

In order to prevent potential credential leakage from the smoke tests, acceptance tests and performance tests, we wish to create a temporary user account for each of the tests to use whilst they run. The user should be destoyed at the end of the tests (regardless of whether they succeed or fail.)

We have added a cf-uaac container to dockerhub (already merged) which allows the `uaac` command to run - this is needed for creating 'admin' users as the `cf` cli cannot create admin user accounts (which are needed for the tests to run). 

The tasks for the tests have been modified to create an admin user account with a random suffix and a randomly generated password - this is removed after the test run using an `ensure` step.

## How to review

Push the pipelines from this branch and run through a deployment.
You will notice that there are extra tasks added to each of the test steps for creating and deleting a temporary admin user account for each test. 
You can verify that the users have been successfully removed using the `uaac users` command after the tests have completed.

## Who can review
Anyone except @mtekel or myself.


dcarley:
I had no luck testing this today because of problems with my environment. Going to try again tomorrow.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1211,11 +1212,23 @@ jobs:
           - get: cf-manifest
             passed: ['cf-deploy']
           - get: bosh-CA
+          - get: cf-secrets
+            passed: ['cf-deploy']
+
+      - task: create-temp-user
+        file: paas-cf/concourse/tasks/create_admin.yml
```
dcarley: Do the smoke tests need a user with admin privileges? Can we create a less privileged user?

(same applies to normal smoke tests and failure testing)

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1300,13 +1334,23 @@ jobs:
...
```
dcarley: Do the custom acceptance tests need a user with admin privileges? Can we create a less privileged user?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1379,13 +1427,23 @@ jobs:
           - get: cf-manifest
             passed: ['smoke-tests', 'acceptance-tests', 'custom-acceptance-tests']
           - get: bosh-CA
+          - get: cf-secrets
+            passed: ['cf-deploy']
+
+      - task: create-temp-user
+        file: paas-cf/concourse/tasks/create_admin.yml
```
dcarley: Do the performance tests need a user with admin privileges? Can we create a less privileged user?

concourse/tasks/create_admin.yml
```diff
@@ -0,0 +1,37 @@
...
+  path: sh
+  args:
+    - -e
+    - -c
+    - |
+      [ -z "${PREFIX}" ] && echo "You need to specify \$PREFIX" && exit 1
+
+      SUFFIX=$(tr -cd '[:alpha:]0-9' < /dev/urandom | head -c10)
+      PASSWORD=$(tr -cd '[:alpha:]0-9' < /dev/urandom | head -c32 )
+      NAME=${PREFIX}-${SUFFIX}
```
dcarley: Can you add  `echo "Adding user ${NAME}"` so that we know username, regardless of whether the add succeeds or fails?

concourse/tasks/create_admin.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+      UAA_ADMIN_CLIENT_PASS=$($VAL_FROM_YAML secrets.uaa_admin_client_secret cf-secrets/cf-secrets.yml)
+      UAA_ENDPOINT=$($VAL_FROM_YAML properties.uaa.url cf-manifest/cf-manifest.yml)
+
+      uaac target "${UAA_ENDPOINT}" --skip-ssl-validation
```
dcarley: Can we add `./paas-cf/concourse/scripts/import_bosh_ca.sh` before this and remove `--skip-ssl-validation` like df2f0803f0f3a573d99d36e229110855da4398cd?

concourse/tasks/delete_admin.yml
```diff
@@ -0,0 +1,25 @@
...
+    - -e
+    - -c
+    - |
+      NAME=$(cat admin-creds/username)
+
+      VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+      UAA_ADMIN_CLIENT_PASS=$($VAL_FROM_YAML secrets.uaa_admin_client_secret cf-secrets/cf-secrets.yml)
+      UAA_ENDPOINT=$($VAL_FROM_YAML properties.uaa.url cf-manifest/cf-manifest.yml)
+
+      uaac target "${UAA_ENDPOINT}" --skip-ssl-validation
```
dcarley: Can we add `./paas-cf/concourse/scripts/import_bosh_ca.sh` before this and remove `--skip-ssl-validation` like df2f0803f0f3a573d99d36e229110855da4398cd?

concourse/tasks/delete_admin.yml
```diff
@@ -0,0 +1,25 @@
...
+      uaac token client get admin -s "${UAA_ADMIN_CLIENT_PASS}"
+      
+      uaac user delete "${NAME}"
+      echo "Removed user ${NAME}"
```
dcarley: I don't think we need to say that it was removed at the end, because `uaac` outputs the following on success:
```
user account successfully deleted
```

But it would be really useful to put `echo "Removing user ${NAME}"` at the top of the script so that we know username regardless of whether the delete succeeds or fails.

concourse/pipelines/create-bosh-cloudfoundry.yml
mtekel: Previously they were using the admin user. The upgrade here is that the new admin user they use is temporary. We did not explore for each type of tests which kind of user do they need (which exact access levels does that user need). We considered it out of scope of this story. We can of course do that, but it will take quite some more effort.

dcarley:
Tests pass 👍 

Some of the commits could be squashed together, for example:
- the tasks to create (2471b3f298db5b935fb248fc32bdf62c802e49d8) and delete (9937193f53e01a3a51faa2f210c140c029df86a1) the user
- creating (7b38a5dfa7f8d7071e2fbcccc3e24dc90edfdaa6) and using (7b38a5dfa7f8d7071e2fbcccc3e24dc90edfdaa6) the user for smoke tests

Could you also add a description of "why" we're creating temporary users to the commit messages?

concourse/tasks/create_admin.yml
mtekel: We considered the generated user name to be a part of secrets, so explicitly didn't want to print it. But we can add that.

dcarley: You're right - I even wrote in the kick-off notes that it was out of scope. I'll make sure that it's captured by a separate story.

concourse/tasks/delete_admin.yml
mtekel: We only wanted to print user name after it was removed, as we considered the generated user name to be part of the login/user secret. We can change that...

mtekel:
Updated

combor:
This fails with:
```
Adding bosh-CA to root certificates
tar: can't open 'bosh-CA/bosh-CA.tar.gz': No such file or directory
```
In create-temp-user task.

We need to re-open this.

dcarley:
Doh, I'd forgotten it was rebased for code in addition to commit messages. I'll test and submit a fix.

mtekel:
i'm just fixing it...

PR #246, opened 2016-05-05
[#118843481] Allow different profiles of Cell Instances per environment

HenryTK:
## What

[#118843481 - Allow different profiles of Cell Instances per environment](https://www.pivotaltracker.com/story/show/118843481)

We want to define higher number of instances of cells in the production environment only. Environment-specific manifest files have been generated. The Makefile determines what file name is passed to the script which renders the Concourse config file. A Concourse task then includes this file name in the arguments passed to Spruce. 

## How to review

* Edit `paas-cf/manifests/cf-manifest/deployments/env-specific/cf-dev.yml` to define the number of `cell_z*` instances you want to deploy. 
* Check it deploys correct number of instances.
* Run `rspec` from this directory: `paas-cf/manifests/cf-manifest`.

## Who can review

Anyone but @mtekel @dcarley or me.


manifests/cf-manifest/spec/support/manifest_helpers.rb
```diff
@@ -14,15 +14,15 @@ def terraform_fixture(key)
...
-      [
+  def load_default_manifest(environment = "dev")
+    arg_list = [
         File.expand_path("../../../build_manifest.sh", __FILE__),
         File.expand_path("../../fixtures/terraform/*.yml", __FILE__),
         File.expand_path("../../fixtures/cf-secrets.yml", __FILE__),
         File.expand_path("../../fixtures/cf-ssl-certificates.yml", __FILE__),
-      ].join(' ')
-    )
+        File.expand_path("../../fixtures/cf-#{environment}.yml", __FILE__),
```
dcarley: From meatspace conversation with @mtekel: we're going to change this to test against the real files rather than fixtures.

manifests/cf-manifest/spec/fixtures/cf-dev.yml
```diff
@@ -0,0 +1,6 @@
+---
+meta:
+  cell_z1:
+    instances: 1
```
dcarley: From meatspace conversation with @mtekel: we're going to move the attributes for dev/CI/staging to a "defaults" file, which will reduce duplication and simplify the possible combinations of tests that we need to write in the future.

mtekel:
Updated, ready for review.

keymon:
LGTM, not really happy with the `eval` thing, but I understand that is required due the limitations of concourse.

mtekel:
eval is a bit "evil" but at least it works. We tried putting quotes around the whole thing (that is, re-define the variable to contain full path) but it didn't help.

dcarley:
It might help to add a comment with a link to the upstream issue so that we know we can possibly remove it in future:
```
# FIXME: Remove eval after https://github.com/concourse/concourse/issues/360
```

(I know, I'm normally against code comments, but it might help in this case)

mtekel:
I had a discussion with @HenryTK yesterday - exactly about this. We had a comment there and decided to remove it because you prefer to have information in commit messages...

mtekel:
Added the comment.

PR #247, opened 2016-05-06
[#116309951] Setup ProxyProtocol between ELB and HAProxy to support  X-Forwarded headers

keymon:
[#116309951 X-Forwarded headers available to applications](https://www.pivotaltracker.com/n/projects/1275640/stories/116309951)

What
----

As a tenant deploying an application to the PaaS, I want an X-Forwarded-Proto and X-Forwarded-For headers to be available to my application so that the application can determine whether it is being served externally over HTTPS.

In order to implement this, we need to get the client origin IP from the ELB in front of HAProxy, for that we want to use [ProxyProtocols](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) in [ELB](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html).

Prerequisites
-------------

**Note** This PR depends in this other one: https://github.com/alphagov/paas-haproxy-release/pull/2 and they should be tested together. After that PR is merge, you must remove the latest commit in this PR.

Context
-------

In this PR:

 * We use a new version of haproxy release too support proxy protocol and other features related to X-Forwarded headers.
 * We disable the HAProxy SSL endpoint and enable the HTTP+ProxyProtocol one
 * We change the ELB to send the traffic to the new endpoint [with ProxyProtocol enabled.](https://www.terraform.io/docs/providers/aws/r/proxy_protocol_policy.html)
 * We update the related custom acceptance tests for X-Forwarded headers.

How to test this
----------------

In order to test this.

 1. Deploy the environment as usual.
 2. You can check the ELB config and confirm that now sends the traffic to 81 (N.B. Proxy protocol config cannot be check from the console, but you can use this command: `aws elb describe-load-balancers --load-balancer-name hector-cf-router`. More info [here](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html))
 3. The tests must pass, specially the custom ones

If the test is successful, remember merge the PR https://github.com/alphagov/paas-haproxy-release/pull/2

Who?
---

Anyone but @keymon or @combor

### DIFF
--- a/terraform/cloudfoundry/router_elb.tf
+++ b/terraform/cloudfoundry/router_elb.tf
@@ -11,17 +11,22 @@ resource "aws_elb" "cf_router" {
   ]
 
   health_check {
-    target = "TCP:443"
+    target = "TCP:81"
     interval = "${var.health_check_interval}"
     timeout = "${var.health_check_timeout}"
     healthy_threshold = "${var.health_check_healthy}"
     unhealthy_threshold = "${var.health_check_unhealthy}"
   }
   listener {
-    instance_port = 443
-    instance_protocol = "ssl"
+    instance_port = 81
+    instance_protocol = "tcp"
     lb_port = 443
     lb_protocol = "ssl"
     ssl_certificate_id = "${var.apps_domain_cert_arn}"
   }
+}
+
+resource "aws_proxy_protocol_policy" "http_haproxy" {
+  load_balancer = "${aws_elb.cf_router.name}"
+  instance_ports = ["81"]
 }
--- a/tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
+++ b/tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
@@ -14,26 +14,23 @@ import (
 	"strings"
 )
 
-type Headers struct {
-	X_Forwarded_For   []string `json:"X-Forwarded-For"`
-	X_Forwarded_Proto []string `json:"X-Forwarded-Proto"`
-}
-
-var _ = PDescribe("X-Forwarded headers", func() {
-	var headers Headers
+var _ = Describe("X-Forwarded headers", func() {
+	var headers struct {
+		X_Forwarded_For []string `json:"X-Forwarded-For"`
+	}
 
 	var _ = BeforeEach(func() {
 		appName := generator.PrefixedRandomName("CATS-APP-")
 		Expect(cf.Cf(
 			"push", appName,
 			"-b", config.GoBuildpackName,
-			"-p", "../../example_apps/print_request_headers",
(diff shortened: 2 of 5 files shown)


tests/acceptance-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -14,44 +14,50 @@ import (
...
 
-	It("should indicate that the user connection was encrypted", func() {
-		Expect(headers.X_Forwarded_Proto).Should(HaveLen(1))
-		Expect(headers.X_Forwarded_Proto[0]).To(Equal("https"))
+	It("should append to existing X-Forwarded-For header", func() {
+		Expect(headers.X_Forwarded_For).Should(HaveLen(1))
+
+		remoteIP := strings.Split(headers.X_Forwarded_For[0], ",")[0]
+		parsedRemoteIP := net.ParseIP(remoteIP)
+
```
mtekel: I wonder why there are differences in parsing IP from x-forwarded for in this and previous test, e.g.:
remoteIP := strings.Split(headers.X_Forwarded_For[0], ",")[1] vs [0]   (is that supposed to be st like 'last' vs 'first'?)

parsedRemoteIP := net.ParseIP(remoteIP) vs net.ParseIP(strings.TrimSpace(remoteIP)) - why different handling of the remoteIP ?

combor: Now, we always send X-Forwarded-For header via curl request. The detected IP is in different place as pre-set IP should be preserved. To be more specific: if you append to the X-Forwarded-For  header it will look like that: 
```
X-Forwarded-For: 1.2.3.4, 4.3.2.1, 6.6.6.6
```
where:
- 1.2.3.4 is preserved [0]
- 4.3.2.1 is detected by ELB [1]
- 6.6.6.6 is added by gorouter [2]

Note the spaces after commas - TrimSpace removes space from the string as when you split it by comma you are gonna get " 4.3.2.1" which is not parsable IP.

terraform/cloudfoundry/router_elb.tf
```diff
@@ -11,17 +11,22 @@ resource "aws_elb" "cf_router" {
...
-    instance_protocol = "ssl"
+    instance_port = 81
+    instance_protocol = "tcp"
     lb_port = 443
     lb_protocol = "ssl"
     ssl_certificate_id = "${var.apps_domain_cert_arn}"
   }
 }
+
+resource "aws_proxy_protocol_policy" "https" {
```
mtekel: should this be called `http`?

keymon: I will rename that

PR #248, opened 2016-05-06
[#118836855] Load CIDRs of tenants from file per environment and Whitelist Bit Zesty IP address

keymon:
[#118836855 Whitelist Bit Zesty IP address](https://www.pivotaltracker.com/story/show/118836855)

What
----

As we currently restrict access to Aviation House for internet facing system components, we need to  allow access to Bit Zesty before we implement a more secure mechanism to access the platform.

We want to implement it in a way that will:
 * Allow us to easily add more tenants
 * Allow different IPs in prod than dev/staging (3rd party tenants will access prod)

Context
-------

To implement this, we allow define in file tenant CIDRs with access to API+web by using terraform interpolation.

Terraform will read from a file defined by the variable ${var.tenant_cidrs_file}, one CIDR per line and allowing comments.

Prod will be the environment access for the actual tenants, so for dev and staging it will load the file
terraform/tenant-cidrs-default.txt, and prod will load terraform/tenant-cidrs-prod.txt as defined in  erraform/prod.tfvars.

We move the office CIDRs to these files too, but keep the variable office_cirds for more specific and operator related security groups.

The variable ${var.web_access_cidrs} is now empty, but will be updated in prod to allow access from the world.

How to test it
--------------

 1. Deploy the environment and confirm that the office IPs are still being properly added to the SG for web and api access.
 2. Optionally add other IP to `terraform/tenant-cidrs-default.txt`
 3. Optionally override the variable `$var.tenant_cidrs_file}` to point to prod and see that it adds the Bit Zesty IP to the SG.

Who?
----

Anyone but @keymon or @combor


terraform/cloudfoundry/router_security_groups.tf
```diff
@@ -17,7 +17,7 @@ resource "aws_security_group" "web" {
     to_port   = 443
     protocol  = "tcp"
     cidr_blocks = [
-      "${split(",", var.web_access_cidrs)}",
+      "${compact(split("\n", replace(file("${path.root}/../${var.tenant_cidrs_file}"), "/ *#.*/", "")))}",
       "${var.concourse_elastic_ip}/32",
```
mtekel: We probably want to maintain ability to define different access levels to API vs WEB. E.g. in prod open web to all, but API only to tenants...

keymon: sorry, I shouldn't have deleted that :-m

terraform/globals.tf
```diff
@@ -100,5 +100,11 @@ variable "microbosh_static_private_ip" {
 /* Note: This is overridden in prod.tfvars and staging.tfvars to allow world access */
 variable "web_access_cidrs" {
   description = "CSV of CIDR addresses for which we allow web access"
-  default     = "80.194.77.90/32,80.194.77.100/32,85.133.67.244/32"
+  default     = ""
+}
+
+/* Note: This is overridden in prod.tfvars to allow access only in prod*/
```
mtekel: 'to allow access only in prod' ? Also, web_access_cirds now acts as extra IPs to enable - but is in a different format/way to the new method of this PR

keymon: Yes, eventually `web_access_cirds` will be all the world: 0.0.0.0 in the case of prod. 

But tenants will still have access to the API.

terraform/cloudfoundry/cf_api_security_groups.tf
```diff
@@ -17,7 +17,7 @@ resource "aws_security_group" "cf_api_elb" {
     to_port   = 443
     protocol  = "tcp"
     cidr_blocks = [
-      "${split(",", var.office_cidrs)}",
```
mtekel: I think we should keep office cirds here. Have office cirds defined in one place only (this variable) and the tennant_cidrs would only contain additional IPs to enable

terraform/globals.tf
combor: I changed this comment so it's more clear now.

terraform/cloudfoundry/router_security_groups.tf
```diff
@@ -18,6 +18,7 @@ resource "aws_security_group" "web" {
     protocol  = "tcp"
     cidr_blocks = [
       "${split(",", var.web_access_cidrs)}",
```
mtekel: I think we should add office cirds here, as we always want to have access to our deployment. We should drop web access cidrs and only use the new way of adding IPs, but rename the file to e.g. additional_web_access_cidrs. For prod that would include 0.0.0.0. Or we could keep additional web access cidrs and only define additional cirds that way. It however removes ability to add comments. (but you can still comment in tfvars file).

keymon: Sorry, I don't really understand what you mean.

PR #249, opened 2016-05-06
Make merge_pr visible in make help

mtekel:
## What

Add description to merge_pr target so that it becomes visible in make
help (when you do just `make` or `make help`).

## How to review

type `make` or `make help`. Check that it displays merge_pr target with the description added in this PR.

## Who can review

not @mtekel


keymon:
LGTM, will merge as the tests in travis finish

PR #252, opened 2016-05-09
[#118836855] Separate whitelisted CIDRs per role and Whitelist Bit Zesty IP address

keymon:
[#118836855 Whitelist Bit Zesty IP address](https://www.pivotaltracker.com/story/show/118836855)

What
----

As we currently restrict access to Aviation House for internet facing system components, we need to  allow access for Bit Zesty before we implement a more secure mechanism to access the platform.

We want to implement it in a way that will allow different IPs in prod than dev/staging (3rd party tenants will access prod)

Notes
-----

This is a different implementation than #248

In this alternate PR we try to:
 * Keep the minimum changes in terraform
 * Do not repeat IPs
 * All CIDRs are defined in the same way, a variable with a list of IPs.

Context
-------

Split endpoint access cidrs by role type

In order to simplify the access rules and definition of CIDRS,
we split the endpoint cidrs list definitions for the security groups
in three types of roles:

 * admin_cidrs: Access to everything, from operation endpoints to CF API, SSH proxy and web access to apps.
 * tenant_cidrs: Access to CF API, SSH proxy and web access to apps.
 * web_access_cidrs: Web access to apps.

We make use of the function [`compact()` in terraform to](https://www.terraform.io/docs/configuration/interpolation.html#compact_list_) to allow have empty list of IPs in all cases.

These variables can be overwritten in the terraform/*.tfvars for each environment. For instance in `terraform/prod.tfvars` we currently override:

 * `access_cidrs = "0.0.0.0/0"`
 * `tenant_cidrs = "52.19.165.187/32"`

How to test it
--------------

 1. Deploy the environment and confirm that the office IPs are still being properly added to the SG for web and api access.
 2. Optionally add other IP to `tenant_cidrs` and confirm that they are added to the SG for API, SSH and web access.

Who?
----

Anyone but @keymon or @combor


terraform/globals.tf
```diff
@@ -97,8 +91,21 @@ variable "microbosh_static_private_ip" {
...
-  description = "CSV of CIDR addresses for which we allow web access"
+
+/* Operators will mainly be from the office. See https://sites.google.com/a/digital.cabinet-office.gov.uk/gds-internal-it/news/aviationhouse-sourceipaddresses for details. */
+variable "admin_cidrs" {
+  description = "CSV of CIDR addresses with access to operator/admin endpoints"
   default     = "80.194.77.90/32,80.194.77.100/32,85.133.67.244/32"
 }
+
+/* Note: This is overridden in prod.tfvars to allow specific tenant access */
+variable "tenant_cidrs" {
```
dcarley: Should this be called `api_access_cidrs` to match the function of the endpoint like `web_access_cidrs`?

keymon: So, do we do it by endpoint? 

Now is per role (see description)

keymon: If we do it per endpoint we will have to repeat the IPs.

dcarley: 👌

saliceti:
Tested successfully, with and without the web access and tenant cidrs.

PR #253, opened 2016-05-09
Disable verbose output for platform tests

dcarley:
## What

There are three related changes:

- Disable verbose mode for our own tests suites so that the output is easier to read
- Ensure that the Vegeta report is still written to the output
- Fix a typo in the `run_tests.sh` script

More detail can be found in the individual commit messages.

## How to review

1. Deploy the pipelines from this branch
1. Run the `create-bosh-cloudfoundry` pipeline
1. Confirm that all of the test jobs pass
1. Confirm that the custom acceptance and performance tests don't output verbose information
1. Confirm that the Vegeta report is still visible in the performance test job

## Who can review

Not @dcarley. @combor should have a look, because he has some context about the Vegeta output.

### DIFF
--- a/tests/performance-tests/src/performance/load_test.go
+++ b/tests/performance-tests/src/performance/load_test.go
@@ -73,7 +73,7 @@ var _ = Describe("Load performance", func() {
 		It("has a response latency within our threshold", func() {
 			metrics, latency := loadTest(appName, loadTestRate, loadTestDuration, false)
 			generateJsonReport(metrics, "load-test-no-keep-alive.json")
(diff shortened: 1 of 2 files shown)


keymon:
I was going to propose that we remove the verbose execution in the script of acceptance tests, but that is defined upstream :(

https://github.com/cloudfoundry/cf-release/blob/master/jobs/acceptance-tests/templates/run.erb#L3

keymon:
Again, -2 points because is a PR without story....

But works and LGTM.

PR #254, opened 2016-05-09
Fix bosh_login invocation - missing bosh FQDN in failure testing pipeline

combor:
## What

https://github.com/alphagov/paas-cf/pull/240 missed required bosh_login.sh change in failure testing pipeline. Right now this script is using two parameters: bosh FQDN and secrets file and failure pipeline still uses only one. This PR corrects this.

## How to review

Run failure testing pipeline from this branch and check if it passes. 

## Who can review

Not @combor


mtekel:
works fine, merging

PR #255, opened 2016-05-09
[#118837127] Add trade tariff quota

keymon:
[#118837127 Create trade-tariff organisation and org manager with quota](https://www.pivotaltracker.com/story/show/118837127)

## What

Create the specific quota for Trade Tariff for 3x environments. Their current deployment is 15G, so we will set the quota to 60G.

We will create the quota named `trade_tariff` and in the future we will create more generic quotas (`big_mem`, `medium`...) as soon as we define our quota model.

We also want to make it easy to add define more quotas in the future.

## Context

The quotas are set during deployment, but we have a task in concourse to update them directly from the manifest.yml in consequent deployments. In this PR we:

 * Add the new quota for trade tariff
 * Add logic to allow define and update multiple quotas,

## How to test

 1. Deploy the environment
 2. Run `cf quotas`, it should report this quota:

```
trade_tariff                            60G                  unlimited               1000     10                  disallowed  
```

## Who can review

Anyone but @colin or @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1063,9 +1063,9 @@ jobs:
                   echo export "${var_name}"=\"$(eval echo \$${var_name})\"
                 done > config/config.sh
 
-                CF_MANIFEST=cf-manifest/cf-manifest.yml $(pwd)/paas-cf/concourse/scripts/extract_quota_settings.rb > config/quota_settings.sh
+                CF_MANIFEST=cf-manifest/cf-manifest.yml $(pwd)/paas-cf/concourse/scripts/extract_quota_settings.rb config
       - aggregate:
-        - task: update-default-quota
+        - task: update-quotas
           config:
             platform: linux
             inputs:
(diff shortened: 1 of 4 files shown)


concourse/scripts/extract_quota_settings.rb
```diff
@@ -2,11 +2,15 @@
 require 'json'
 require 'yaml'
 
+output_dir=ARGV[0]
```
dcarley: Can we add some error handling to prevent it being run without a directory argument? At the moment it tries to write to the root directory:
```
➜  paas-cf git:(118837127_trade_tariff_quota) ✗ CF_MANIFEST=tests/cf-manifest.yml concourse/scripts/extract_quota_settings.rb
concourse/scripts/extract_quota_settings.rb:15:in `write': Permission denied @ rb_sysopen - /default_quota.sh (Errno::EACCES)
        from concourse/scripts/extract_quota_settings.rb:15:in `block in <main>'
        from concourse/scripts/extract_quota_settings.rb:10:in `each'
        from concourse/scripts/extract_quota_settings.rb:10:in `<main>'
```

`ARGV.fetch(0)` would work, but the feedback isn't very friendly. `raise "…" unless File.directory?` might be OK.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1085,14 +1085,18 @@ jobs:
...
-                  . ./config/quota_settings.sh
-                  ${QUOTA_non_basic_services_allowed} || prefix="dis"
-
-                  cf update-quota "${QUOTA_DEFAULT}" \
-                    -m "${QUOTA_memory_limit}"M \
-                    -r "${QUOTA_total_routes}" \
-                    -s "${QUOTA_total_services}" \
-                    --${prefix}allow-paid-service-plans
+                  for quota_file in config/*_quota.sh; do
+                    (
```
dcarley: 👍  for the subshell.

The syntax is quite subtle and easy to overlook though. What do you think about refactoring the contents of the subshell out to a script and calling it with `env -i` to have the same effect?

I'm not sure what the behaviour would be if one of the environment variables was empty because someone omitted or typoed a key in the quota hash. Would it fail or set an unlimited quota? It'd be good to `set -u` to catch that (although L1090 would need refactoring or excluding).

dcarley:
+5 karma for adding tests.

dcarley:
Could you add a bit more detail to the commit message of ebbc00a5e19f06b3bfe1f2758ca1ef252c852fd9 to say how we got the 15G number. Such as a link to the graphs we used and an explanation that it's the peak usage for the 3 applications in the GOV.UK staging environment. Otherwise we might be similarly confused in the future as we were by the 150G number.

concourse/pipelines/create-bosh-cloudfoundry.yml
keymon: +1 for setting `set -u`

But I am not sure about exporting the subshell to script. that means create a script for just a pair of lines that are called from a for. If we do so we are breaking the flow in the loop, and it will it makes it more difficult to understand how everything works together.

If we are worried about the risk of overlooking the subshell, we can add a comment.

dcarley:
We agreed not to split it out to a separate script.

Confirmed that it updates an incorrect quota too.

PR #256, opened 2016-05-09
[#117705909] Make really sure temp. admin gets removed

mtekel:
## What

Wrap all the tasks including temp. admin creation into a `do` block
to which we assign ensure step which removes the temp. admin. Previously
only last task had ensure step and if there were any issues in any tasks
before the last one (including the admin creation task), the admin user
would not be deleted.

Note that this is not displayed correctly in concourse, see https://github.com/concourse/concourse/issues/434

## How to review

Apply. You can try canceling jobs as soon as create-temp-user task starts. Ensure task should run even when the job is cancelled by user.

## Who can review
not @mtekel

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1219,17 +1219,19 @@ jobs:
           - get: cf-secrets
             passed: ['cf-deploy']
 
-      - task: create-temp-user
-        file: paas-cf/concourse/tasks/create_admin.yml
-        config:
-          params:
-            PREFIX: smoketest-user
+      - do:
+        - task: create-temp-user
+          file: paas-cf/concourse/tasks/create_admin.yml
+          config:
+            params:
+              PREFIX: smoketest-user
 
-      - task: smoke-tests-config
-        file: paas-cf/concourse/tasks/smoke-tests-config.yml
+        - task: smoke-tests-config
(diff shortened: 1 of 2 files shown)


mtekel:
Regarding failed tests: @Jonty is updating his pipecleaner to support `do` block.

dcarley:
Blocked by #257.

keymon:
I merged #257 and rebased this repo with that

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1642,38 +1650,40 @@ jobs:
...
-              - name: smoke-tests-log
-            run:
-              path: sh
-              args:
-              - -e
-              - -c
-              - |
-                paas-cf/concourse/scripts/smoke_tests_email.sh \
-                  ${DEPLOY_ENV} ${SYSTEM_DNS_ZONE_NAME} ${ALERT_EMAIL_ADDRESS}
+              PREFIX: continuous-smoketest-user
```
keymon: As commented in person, this won't work in the newer versions of concourse (>=1.1) because it complains:

```
DEPRECATION WARNING: Specifying both `file` and `config.params` in a task step is deprecated, use params on task step directly
``` 

I think we will have to remove the `config` key. 

But that is a issue that should be fix in the story about upgrading concourse.

keymon:
Works as announced! merging...

PR #257, opened 2016-05-09
[#117705909] Fix concourse `do` blocks in pipecleaner lint

Jonty:
117705909-ensure-temp-admin-deletion introduces `do` blocks into our pipelines, which causes ordering issues in pipecleaners dependency resolution and spurious fatal errors to be reported.

Until this fix is merged branches with `do` blocks cannot be merged. https://travis-ci.org/alphagov/paas-cf/builds/128887561

### DIFF
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -62,15 +62,23 @@ class Pipecleaner(object):
 
             while plan:
                 item = plan.pop(0)
-                # Flatten aggregate blocks as we don't care
-                if 'aggregate' in item:
-                    plan = item['aggregate'] + plan
-                    continue
+                discovered_blocks = []
 
-                # Flatten blocks we don't care about
+                # Flatten array blocks as we don't care
+                for block_type in ('aggregate', 'do'):
+                    if block_type in item:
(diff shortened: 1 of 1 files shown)


dcarley:
Would be good to have tests for this, sometime 😄

keymon:
The PR works as expected.

But there is a small bug in this script, but I don't think it is related to this PR so it is fine.

The bug is that we don't check that one item has 2x different task types. For instance:

```
resources:
- name: concourse
  type: git2
  source:
    uri: https://github.com/concourse/git-resource.git
    branch: master

jobs:
- name: test
  plan:
  - get: concourse
    params: {submodules: none}
    do:
    - get: concourse
  - get: concourse
    put: concourse
  - put: concourse
```

The two first "tasks" are wrong, because we mix `do` and `get` or `put` and `get`. 

Concourse fails with this error, interestingly in the server side:

```
error: Unexpected Response
Status: 400 Bad Request
Body:
{"errors":["invalid jobs:\n\tjobs.test.plan[0] has multiple actions specified (do, get)\n\tjobs.test.plan[1] has multiple actions specified (get, put)\n"]}
```

keymon:
I will merge as it fixes the issue described and will unblock #256

PR #258, opened 2016-05-10
[#118837191] Create space users

combor:
## What

As a Paas Team Member in order create users for an organisation I want to run a simple process to create a new user(s)

## How to review

Run the script and check if it creates unprivileged user.

Prerequisites:

 1. You must have installed the latest AWS client and CF client.
 1. You must have your AWS credentials exported and login into CF with an admin user.
 1. You have to export aws region `export AWS_DEFAULT_REGION=eu-west-1`
 1. Your email account must be verified as a account to receive emails from SES.
    You can do it with:
    ```
aws ses verify-email-identity --email-address piotr-komborski@digital.cabinet-office.gov.uk --region eu-west-1
```

Then run the script to create your account. You will get an email with the credentials

```
./scripts/create-tenant.sh -m -o orgname -e "piotr-komborski@digital.cabinet-office.gov.uk"
```
 Now check if you are not an Organisation manager

## Who can review

Not @timmow or @combor

### CODE REVIEW

File: scripts/create-user.sh [hunk 1]
```diff
@@ -143,9 +145,11 @@ create_user() {
 }
 
 set_user_roles() {
+  if [[ "${ORG_MANAGER}" == "true" ]]; then
   cf set-org-role "${EMAIL}" "${ORG}" OrgManager
```
dcarley: Could you indent the lines within this new `if` block.

combor: Done.

keymon:
LGTM, but just one comment:

There is a flaw here. If the org-manager decides remove sandbox and add, for instance: dev, staging and prod, whenever we create a new developer user we will add a space called "sandbox" and add the user to that space.

Does not it make sense to make the script:
 - If the org does not exist, create it.
 - If the space does not exist and there is no other space, create it
 - add the new user to all the spaces.

This way the user will always be added to all the spaces, and if they deleted sandbox, it won't be created.

dcarley:
@combor @timmow should non-org-managers be added to a space at all? Isn't that up to the org-manager to do?

combor:
We discussed this with @dcarley and @Rory80Hz and decided that it's OK to assign users to the sandbox space by default. It will change in the future with invitation UI.

keymon:
merging in that case

PR #259, opened 2016-05-10
[#118272895] - Upgrade CF, stemcells and BOSH

bleach:
## What

[Story](https://www.pivotaltracker.com/projects/1275640/stories/118272895)

We upgraded:

- ~~BOSH from 255.8 to 256.2~~ Cancelled because of [this issue](https://github.com/cloudfoundry/bosh/issues/1263).  This will be done separately.
- CF from 233 to 235 and its recommended versions of diego, garden, etcd
- ~~Stemcells from 3215 to 3232.2 to get the latest security fixes~~ Cancelled because of time out issues. This will be done separately.

Changes:

- [graphite BOSH release
  changes](https://github.com/alphagov/paas-graphite-statsd-boshrelease/pull/5)
  were needed to work with the new BOSH
- cf-acceptance, smoke tests and app that our custom acceptance tests use to
  print headers all switched to go 1.6
- ~~the new BOSH doesn't use redis~~
- We turned off the V3 suite in cf-acceptance-tests as they break when presented with HTTPS endpoints and we don't currently use the [not yet stable v3 API](http://v3-apidocs.cloudfoundry.org/version/release-candidate/)

## How to review

- Switch your environment to the branch and run the pipeline.
- Review the changes
- First merge the [PR against graphite BOSH release](https://github.com/alphagov/paas-graphite-statsd-boshrelease/pull/5) and create a new tag.
- Amend the manifests in this branch to use the new tag in the graphite BOSH  release.
- Merge this PR.

## Who can review

Anyone but @saliceti @bleach @jimconner

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1333,7 +1333,7 @@ jobs:
                   mkdir -p /var/vcap/jobs/acceptance-tests/bin/ /var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry
                   ln -snf $(pwd)/test-config/config.json /var/vcap/jobs/acceptance-tests/bin/config.json
                   ln -snf $(pwd)/test-config/run /var/vcap/jobs/acceptance-tests/bin/run
-                  ln -snf /usr/local/go /var/vcap/packages/golang1.4
+                  ln -snf /usr/local/go /var/vcap/packages/golang1.6
                   sed -i 's/bits=@"%s"/bits=@%s/' cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/helpers/v3_helpers/v3.go
                   ln -snf $(pwd)/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests /var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry/
                   /var/vcap/jobs/acceptance-tests/bin/run
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1334,7 +1334,6 @@ jobs:
                   ln -snf $(pwd)/test-config/config.json /var/vcap/jobs/acceptance-tests/bin/config.json
                   ln -snf $(pwd)/test-config/run /var/vcap/jobs/acceptance-tests/bin/run
                   ln -snf /usr/local/go /var/vcap/packages/golang1.6
-                  sed -i 's/bits=@"%s"/bits=@%s/' cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/helpers/v3_helpers/v3.go
                   ln -snf $(pwd)/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests /var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry/
                   /var/vcap/jobs/acceptance-tests/bin/run
 
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -83,6 +83,7 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-graphite-statsd-boshrelease.git
       tag_filter: {{cf_graphite_version}}
+      branch: gds_master
 
   - name: grafana-boshrelease
     type: git
--- a/concourse/tasks/smoke-tests-run.yml
+++ b/concourse/tasks/smoke-tests-run.yml
@@ -19,7 +19,7 @@ run:
       mkdir -p /var/vcap/jobs/smoke-tests/bin/ /var/vcap/packages/smoke-tests/src/github.com/cloudfoundry
       ln -snf $(pwd)/test-config/config.json /var/vcap/jobs/smoke-tests/bin/config.json
       ln -snf $(pwd)/test-config/run /var/vcap/jobs/smoke-tests/bin/run
-      ln -snf /usr/local/go /var/vcap/packages/golang1.4
+      ln -snf /usr/local/go /var/vcap/packages/golang1.6
       ln -snf $(pwd)/cf-release/src/smoke-tests /var/vcap/packages/smoke-tests/src/github.com/cloudfoundry/cf-smoke-tests
       set -o pipefail
       git --git-dir paas-cf/.git log -1 > smoke-tests-log/last-commit.log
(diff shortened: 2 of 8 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -83,6 +83,7 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-graphite-statsd-boshrelease.git
       tag_filter: {{cf_graphite_version}}
+      branch: 118272895_bosh20
```
mtekel: The new PR (https://github.com/alphagov/paas-graphite-statsd-boshrelease/pull/5) is against `gds_master` branch. Nothing wrong with that, but after that PR is merged, this commit should stay in, specify `gds_master` branch and update commit message.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1323,7 +1323,7 @@ jobs:
                   mkdir -p /var/vcap/jobs/acceptance-tests/bin/ /var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry
                   ln -snf $(pwd)/test-config/config.json /var/vcap/jobs/acceptance-tests/bin/config.json
                   ln -snf $(pwd)/test-config/run /var/vcap/jobs/acceptance-tests/bin/run
-                  ln -snf /usr/local/go /var/vcap/packages/golang1.4
+                  ln -snf /usr/local/go /var/vcap/packages/golang1.6
                   sed -i 's/bits=@"%s"/bits=@%s/' cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/helpers/v3_helpers/v3.go
```
mtekel: The commit message should probably contain that this is for acceptance-tests and why we need to bump the version. This commit could be probably combined with the commit that's bumping the go version for smoke tests, as I think the reasons for bumping the version could be the same (...as this is same shared dependency for both tests).

mtekel:
We should remove `max_retries: 8` from 
```
manifests/bosh-manifest/reference-bosh-manifest.yml:11:      max_retries: 8
manifests/bosh-manifest/reference-bosh-manifest.yml:55:      max_retries: 8
manifests/bosh-manifest/deployments/aws/090-bosh-aws-stub.yml:13:      max_retries: 8
manifests/bosh-manifest/deployments/aws/090-bosh-aws-stub.yml:20:      max_retries: 8
```
as this is default since v50 of the AWS CPI & we're upgrading to v52 in this PR. No need to override default with the same value...

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: Good point.

Would you mind reviewing alphagov/paas-graphite-statsd-boshrelease#5, as you're testing the release with the new version of CF?

Do we need to specify `branch`if we're pinning the commit with `tag_filter` too?

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: Incidentally, we can remove L1327 (a revert of 4865dc54ab9199779bad4819f26861c31c4a5b3f), because it's now fixed upstream:
```
➜  cf-release git:(master) git submodule-contains src/github.com/cloudfoundry/cf-acceptance-tests cfacc2d8c7c04c3e540622a53d747ac6ed2ac332
d1cdf1b59d08ca30ee39a593e2512231db3e949e
➜  cf-release git:(master) git tag --contains d1cdf1b59d08ca30ee39a593e2512231db3e949e
v232
v233
v234
v235
v236
```

It might be worth doing that now so that we don't have cruft for tests that we're not running.

mtekel:
As it turns out, `bosh_admin_password` is the correct password for HM:
```
diff --git a/manifests/bosh-manifest/deployments/010-bosh-template.yml b/manifests/bosh-manifest/deployments/010-bosh-template.yml
index be902c5..89c709b 100644
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -68,7 +68,7 @@ jobs:
     hm:
       director_account:
         user: admin
-        password: (( grab secrets.bosh_hm_director_password ))
+        password: (( grab secrets.bosh_admin_password ))
       resurrector_enabled: true
 
     ntp: (( grab meta.ntp ))
```

concourse/pipelines/create-bosh-cloudfoundry.yml
bleach: > Do we need to specify `branch` if we're pinning the commit with `tag_filter` too?

Yes. @saliceti investigated and the git resource looks at the `branch`, only returning tags that match `tag_filter`.

bleach: If the tag points to a commit in the default branch then it can be omitted, so once alphagov/paas-graphite-statsd-boshrelease#5 is merged I think it could be omitted.

dcarley: > If the tag points to a commit in the default branch then it can be omitted, so once alphagov/paas-graphite-statsd-boshrelease#5 is merged I think it could be omitted.

Possibly not then, because the commit would be on `gds_master` rather than `master`. In which case I think Michal's original suggestion is correct.

dcarley:
NB: Disabling the v3 tests has been pulled out to #262 because that test suite is also causing another problem. We're going to amend the commit to include details of the new problem, so this PR will need rebasing afterwards.

mtekel:
I was able to deploy for the 2nd time. After VMs are created, rest of deployment is fine. All tests pass. Also, @combor was able to deploy this without issues.  To conclude, there are few outstanding things on this PR:

* update the e4c303aac2db1879c8e818b618a1e4ea549b6d87 commit and message to point to correct branch and contain correct description
* revert 4865dc54ab9199779bad4819f26861c31c4a5b3f
* squash d2ffbc64abd6ef4aead7717e4151d3dffe3896d1 and 048f0ddf8862e85e1e4703af3bd30c9d720a2696 and update commit description with why go version had to be bumped
* remove the `max_retry:8` setting as this is now default
* fix and remove `bosh_hm_director_password`
* remove c3b3c41c4e52c7776e1d8d2dc92db11ed086701e as it was already merged in https://github.com/alphagov/paas-cf/pull/262

I have summarized encountered BOSH issues in https://github.com/cloudfoundry/bosh/issues/1263.
After above things are fixed, someone should try to deploy this again and see if they encounter the BOSH issues as well.

saliceti:
Regarding the CPR upgrade, this will not be done here but in [this PR](https://github.com/alphagov/paas-cf/pull/260).

mtekel:
All comments addressed. The only outstanding issue is BOSH behaviour when deleting or creating VMs. If during this phase there is any hiccup in worker <-> NATS connection, the operation will get stuck and BOSH will end up in weird state (see https://github.com/cloudfoundry/bosh/issues/1263)

That's why I think someone else should also test this (mainly repeated deployment and deletion of VMs by deploying and deleting the deployment).

PR #260, opened 2016-05-10
[#117512371] Creation and Deletion of Basic RDS Service Instance

alext:
## What

Building on the work in #220, this adds all the necessary config to allow the service broker to create/destroy RDS instances, and to bind them to apps. It includes an acceptance test that tests this lifecycle.

This includes an upgrade to the bosh-aws-cpi so that we can apply security-groups at a resource_pool level.

## How to review

Deploy from this branch, and verify that the custom acceptance tests pass.

## Before Merging

* [x] Merge alphagov/paas-docker-cloudfoundry-tools#52
* [x] Remove the "FIXME: Use bosh-init container from branch." commit from this branch
* [x] Merge alphagov/paas-docker-cloudfoundry-tools#53 and remove the "FIXME: Use cf-cli from branch" commit from this branch.

## Who can review

Anyone but myself, @timmow, @jimconner and @mtekel.


HenryTK:
* The acceptance tests do not cover the unbinding of a service. 
* After the service is bound and the application is started the response status code is asserted on. Does this guarantee the application can successfully retrieve data from the database?
* Regarding use of docker image tag `docker:///governmentpaas/bosh-init#bosh_init_cpi_52` in the pipelines. It has been added to the create-bosh-cloudfoundry pipeline for review, but has it been tested with the create-deployer pipeline where this image is also used? [7d23bc60791b2ae58443f8bd86a87ca2c7a894a3](https://github.com/alphagov/paas-cf/pull/260/commits/7d23bc60791b2ae58443f8bd86a87ca2c7a894a3).

tests/acceptance-tests/src/acceptance/rds_broker_test.go
```diff
@@ -1,22 +1,101 @@
...
+
+			Expect(cf.Cf("bind-service", appName, dbInstanceName).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
+
+			Expect(cf.Cf("start", appName).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+
+			// Hit app endpoint to verify functionality.
+			fmt.Fprintln(GinkgoWriter, "Sending request to DB Healthcheck app")
+			resp, err := httpClient.Get(helpers.AppRootUri(appName))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode).To(Equal(200))
```
HenryTK: Is the db-healthcheck application written in such a way that a 200 status code definitely means the database is up an running? Is there any output to the page we should be asserting on?

HenryTK:
I get the following failure in custom acceptance tests:

```
• Failure [560.104 seconds]
RDS broker
/tmp/build/f541ec31/paas-cf/tests/src/acceptance/rds_broker_test.go:95
  creating a database instance
  /tmp/build/f541ec31/paas-cf/tests/src/acceptance/rds_broker_test.go:94
    can be used from an app [It]
    /tmp/build/f541ec31/paas-cf/tests/src/acceptance/rds_broker_test.go:93

    Got 500 response from healthcheck app. Response body:
    dial tcp 10.0.54.199:5432: getsockopt: connection refused
    
    
    Expected
        <int>: 500
    to equal
        <int>: 200

    /tmp/build/f541ec31/paas-cf/tests/src/acceptance/rds_broker_test.go:92
```

timmow:
> After the service is bound and the application is started the response status code is asserted on. Does this guarantee the application can successfully connect to the database?

The app should catch all errors and respond with a 5xx error, the only path that leads to a 200 is on a successful connection and query of the database, so I think a 200 response code should mean database communication is successful

mtekel:
There was issue with applying the new security groups. We were not sure if it was accidental or persistent. It looks like it is persistent and is reason why you get 500 error instead of 200 OK. The solution is to add & apply the RDS security group in the same way we do `graphite-nozzle` SG.

terraform/cloudfoundry/variables.tf
```diff
@@ -25,19 +25,18 @@ variable "router_cidrs" {
...
+    zone2 = "10.0.54.0/24"
+  }
 }
 
-variable "router_cidr_all" {
-  description = "CIDR for all router subnets"
-  default     = "10.0.48.0/20"
+variable "aws_backing_service_cidr_all" {
+  description = "CIDR for all aws_backing_service subnets"
+  default     = "10.0.52.0/20"
```
HenryTK: Having 20 mask bits seems to create a subnet much larger than needed. Should this be changed to 22 or 23?

mtekel:
I will add a new way to update security groups, done in a similar way we handle quotas. The issue is very similar: https://github.com/cloudfoundry/capi-release/issues/9

I will also look into adding unbind test.

mtekel:
Added explicit SG configuration, rebased on master to solve merge conflicts. I have analysed unbind behaviour and found some issues. Thus spinning of addition of acceptance test (+investigation & fix of found issues) into separate story: https://www.pivotaltracker.com/projects/1275640/stories/119530189

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1064,7 +1064,44 @@ jobs:
                 done > config/config.sh
 
                 CF_MANIFEST=cf-manifest/cf-manifest.yml $(pwd)/paas-cf/concourse/scripts/extract_quota_settings.rb config
+
+                cd config
+                CF_MANIFEST=../cf-manifest/cf-manifest.yml $(pwd)/../paas-cf/concourse/scripts/extract_security_groups.rb
+
       - aggregate:
+        - task: update-default-security-groups
```
HenryTK: The `update-default-security-groups` task does not cover the situation where we would want to remove a default security group (not implying said situation will ever occur).

HenryTK:
There are no tests for `concourse/scripts/extract_security_groups.rb`.

HenryTK:
Performance test job is currently broken. Unsure if it has been caused by this story (given the recent pipeline failures that have stopped things progressing passed acceptance tests).

``` 
Adding bosh-CA to root certificates
Updating certificates in /etc/ssl/certs... 1 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d....done.
Running tests
# performance_test
./load_test.go:23: undefined: vegeta in vegeta.Metrics
./load_test.go:25: undefined: vegeta in vegeta.NewStaticTargeter
./load_test.go:25: undefined: vegeta in vegeta.Target
./load_test.go:29: undefined: vegeta in vegeta.Metrics
./load_test.go:32: undefined: vegeta in vegeta.NewAttacker
./load_test.go:32: undefined: vegeta in vegeta.KeepAlive
./load_test.go:40: undefined: vegeta in vegeta.Metrics
./load_test.go:41: undefined: vegeta in vegeta.NewJSONReporter
./load_test.go:75: undefined: vegeta in vegeta.NewTextReporter
./load_test.go:84: undefined: vegeta
./load_test.go:84: too many errors
FAIL	performance [build failed]
```

mtekel:
Indeed I was able to reproduce the issue.

alext:
That's my bad.  I've removed in import line from that test by mistake.  I'll fix that.

tests/src/performance/load_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gexec"
-	"github.com/tsenart/vegeta/lib"
```
dcarley: Looks like this is the cause of the error @HenryTK and @mtekel were discussing.

alext: Indeed. I think this was `goimports` removing it because it couldn't resolve it locally on my machine. I've put it back again now.

alext:
> There are no tests for concourse/scripts/extract_security_groups.rb.

I've changed this so that there's a single script that extracts the definitions from the manifests and applies them to CF. This also has unit tests.

alext:
I think all comments have been addressed. This is ready for review again.

timmow:
I've merged https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/53, so the FIXME commit can be removed

alext:
> I've merged alphagov/paas-docker-cloudfoundry-tools#53, so the FIXME commit can be removed

Done.

PR #262, opened 2016-05-11
[#119177161] Turn off V3 API tests, which don't currently work

jimconner:
## What

The acceptance-tests suite for v233 includes a test for the v3 api which fetches a go buildpack from github. Within the last 48 hours, the go buildpack has been changed in github in a manner which appears to make it incompatible with the version of the tests we're running. Later versions of the acceptance tests have been swtiched to use a ruby buildpack instead of the go buildpack - We need to disable the v3 acceptance tests (which are disabled already in the PR for upgrading to CF 235) to unblock the pipelines.

## How to review

If you have an existing environment deployed from our current master branch, you will find that the acceptance tests are already failing. Deploying from this branch should fix that problem.

## Who can review

Anyone on the team excecpt for myself or @saliceti .

### DIFF
--- a/manifests/cf-manifest/deployments/900-cf-tests.yml
+++ b/manifests/cf-manifest/deployments/900-cf-tests.yml
@@ -10,7 +10,7 @@ properties:
     client_secret: (( grab secrets.uaa_clients_gorouter_secret ))
     skip_diego_unsupported_tests: true
     include_tasks: false
-    include_v3: true
+    include_v3: false
     include_security_groups: true
     include_routing: true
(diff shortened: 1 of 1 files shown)


dcarley:
Waiting for tests to run..

combor:
https://github.com/alphagov/paas-cf/pull/259 also disables this here: https://github.com/alphagov/paas-cf/pull/259/files#diff-657ec4d8555241f88c182201d568500cR13

combor:
Uh, I just noticed that this is probably cherry picked commit :)

dcarley:
v3 was skipped. Tests passed.

PR #263, opened 2016-05-11
[#119268815] Cross Zone Logging for CF

combor:
## What

This PR removes hardcoded zone name from metron zone property and replaces
it with empty string. Metron doppler selection is based on preferred zone
prefix that matches zone property value with doppler instance name. If
there is no match, metron forwards logs to randomly picked doppler despite
the zone it is deployed to.

This will also unblock [#119269637](https://www.pivotaltracker.com/story/show/119269637) as zone won't be hardcoded any more.

More details here:
https://github.com/cloudfoundry/loggregator/blob/a4675b0821d40743ff03424456da9b5ce8de6598/src/doppler/dopplerservice/finder.go#L193

## How to review

Deploy from this branch and do `bosh ssh` to any doppler instance.

Execute as root:
```
tcpdump -i eth0 port 3457 | awk '{print $3}' | grep -v ${doppler_zone}
```
and check if you can see hostnames from other zones.

## Who can review
Not @timmow or @combor

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -219,7 +219,7 @@ jobs:
         agent:
           mode: server
       metron_agent:
-        zone: z1
+        zone: ""
 
   - name: consul_z2
     templates: (( grab meta.consul_templates ))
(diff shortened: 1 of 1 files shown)


Jonty:
Tested and working!

PR #264, opened 2016-05-12
[#118836855] Update BitZesty IP for API access to prod

dcarley:
## What

The original address provided was incorrect. They have now provided this
updated address.

## How to review

There's not much we can test here. Confirm that the IP address is the same as the last address commented by @Rory80Hz in the story: https://www.pivotaltracker.com/story/show/118836855

## Who can review

Not @dcarley


Rory80Hz:
Looks good, checks passed, IP matches the new one from BitZesty.

PR #269, opened 2016-05-18
[#115870117] BOSH failover to other AZ

saliceti:
## What
Story: [Process for recreating BOSH on AZ Failure](https://www.pivotaltracker.com/story/show/119090295)

We want to be able to recreate BOSH in another AZ in case of AZ failure and test it can still see the VMs as running.

## How to review

* Apply on existing pipeline
* Kill BOSH via AWS console
* Upload pipeline with:

```
BOSH_AZ=eu-west-1b make dev pipelines
```

* This should recreate BOSH and test all VMs are in running state. You can check this by doing:

```
make dev bosh-cli
```

or check the output of bosh-tests job.

## ⚠️  Before merging ⚠️ 

As the BOSH state file name is changed in this PR, when this is applied to CI, Staging, Prod bosh-init would try to recreate BOSH and fail because of IP conflict. To solve this rename BOSH state file in S3 to `bosh-manifest-state-eu-west-1a.json` in each environment.

## Who can review
Anyone but @combor or myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,3 +376,6 @@ pass the `-m` option to the script
 To run the above script, you need to verify your email address with [AWS SES](https://aws.amazon.com/ses/)
 To do this, run `aws ses verify-email-identity --email-address youremail@example.com --region eu-west-1`
 and click on the link in the resulting email
+
+## BOSH failover
+Visit [BOSH failover page](https://github.com/alphagov/paas-cf/doc/bosh_failover.md)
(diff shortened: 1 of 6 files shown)


jimconner:
All works exactly as intended. I can't find anything to fault. The new bosh-tests works great and spotted that I had a failing VM after my first deployment (before recreating bosh in a new AZ). All works great, so... Merging!!

PR #270, opened 2016-05-18
Fix bosh_login.sh when password starts with '-'

keymon:
What
----

If the random generated password for the bosh admin starts with '-', the
`bosh_login.sh` won't work because it recognises the password with an
option.

Context
-------
To avoid that we add '--' to tell bosh that the following parameters are not options.
    
Additionally, if the string starts with '-', ruby yaml will quote the string in the cf-secrets.yml with `"`. We need to remote these quotes before passing the password as argument.

How to test
-----------

Check that works by running rerunning the pipeline with cf-deploy, which
executes the `bosh_login.sh`.

I don't think it is necessary test when the password has a `-`

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/scripts/bosh_login.sh
+++ b/concourse/scripts/bosh_login.sh
@@ -5,6 +5,6 @@ set -eu
 bosh_host=$1
 bosh_secrets_file=$2
 
-bosh_password=$(awk '/bosh_admin_password/ { print $2 }' "${bosh_secrets_file}")
-bosh -t "${bosh_host}" login admin "${bosh_password}"
+bosh_password=$(awk '/bosh_admin_password/ { print $2 }' "${bosh_secrets_file}" | tr -d '"')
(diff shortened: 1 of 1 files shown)


keymon:
The issue is that there are quotes in the password from the yaml. I can remove them.

dcarley:
`bosh_login` seems to always be used from `governmentpaas/bosh-cli` which is based on Ruby, so could we use `concourse/scripts/val_from_yaml.rb` for proper YAML parsing instead?

dcarley:
@keymon As discussed on Slack, I've added a commit to use `val_from_yaml` and tested it using the create pipeline. Could you ACK it? If you're happy then I'll merge.

dcarley:
Oops, fixed shellcheck error.

keymon:
I did misunderstood you in slack, I though you wanted to move the login logic inside of `bosh_create_and_upload_release` script.

This is good for me :+1:

PR #271, opened 2016-05-19
[#119803023] Create paas_demo quota

saliceti:
## What
Story: [Increase Quota for paas-demo org](https://www.pivotaltracker.com/story/show/119803023)

Create new paas_demo quota in code so we can aassign it to the paas-demo org.

## How to review
* Apply on new or existing environment
* `cf quotas` should show the new quota

## Who can review
Anyone but me


mtekel:
This indeed creates new quota on the deployment, merging...

PR #273, opened 2016-05-19
[#119090295] TF_VAR_bosh_az variable missing in microbosh-destroy pipeline

combor:
## What

It seems that we missed the update for microbosh-destroy pipeline during implementation of deployment to multiple AZs for bosh. This PR fixes this.

## How to review

Run microbosh-destroy and make sure if passes. Especially terraform-destroy task shouldn't throw:

```
Errors:

  * 1 error(s) occurred:

* Required variable not set: bosh_az
```

## Who can review

not @combor


jimconner:
Tested and I can confirm that the microbosh is now cleanly destroyed without any errors from terraform... Merging now.

PR #274, opened 2016-05-19
[#119903539] Cleanup temp. admin on failure

mtekel:
## What

Store new user credentials in output immediately after user is created, so that if any of the group additions fail, the ensure task would have access to the username in the outputs.

## How to review

Apply. Run your pipeline enough times till you get failure from UAA:
```
error response:
{
  "error_description": "Incorrect result size: expected 1, actual 0",
  "error": "scim",
  "message": "Incorrect result size: expected 1, actual 0"
}
```
Check that the delete temp user task ran successfully (previously this would fail).

## Who can review

not @mtekel


saliceti:
Tested by stopping the smoke test job when the permissions are being added. The user is successfully removed.

PR #275, opened 2016-05-19
[#118484173] RDS broker cleanup

alext:
## What

This includes a number of changes to clean up the RDS broker and deployment:

* Removed support for AWS Aurora (alphagov/paas-rds-broker#5)
* Removed some unused jobs from the bosh release (alphagov/paas-aws-broker-boshrelease#6)
* Clarify the broker name. 

On the last point - the broker had been given the same name as the service it provided, which had led to some confusion (especially if we were to enable the mysql services). See the commit for more details

## How to review

Deploy from this branch and verify that everything works (the custom-acceptance tests pass etc).

## Who can review

Anyone but @HenryTK or myself.


jimconner:
Works nicely for me - confirmed that it renames the broker and deploys correctly. Merging...

PR #276, opened 2016-05-19
Clear environment before running PR specs.

alext:
## What

Without this the tests will fail if you have `GITHUB_API_TOKEN` set in
your environment as the test expectations won't match the requests that
are actually made.

## How to review

Ensure that the Travis build passes.
Set `GITHUB_API_TOKEN` to something in your local environment, and verify that the tests still pass.

## Who can review

Anyone but myself.


mtekel:
works...

PR #277, opened 2016-05-19
[#119818913] Security Groups on resource pools instead of networks 

combor:
## What

Right now, security group configuration work on the network property. To have two jobs in the same network but with different security groups, you'd have to define two networks. We want to move security groups definitions from subnets to resource pools to reduce number of networks, reduce manifest complexity and prepare for BOSH AZ feature implementation.

## How to review

Deploy from branch and check if acceptance-tests pass.

## Who can review

Not @HenryTK or @combor


manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -763,7 +763,7 @@ properties:
   support_address: "http://support.cloudfoundry.com"
   description: null
   ssl:
-    skip_cert_verify: true
+    skip_cert_verify: false
```
dcarley: Good catch. Looks like this fixes a story we'd forgotten about: https://www.pivotaltracker.com/story/show/111678348

Could you pull it out into a separate PR for visibility and tag it with that story ID? Will it cause merge conflicts with this PR?

combor: I've created https://github.com/alphagov/paas-cf/pull/282. There shouldn't be any merge conflicts.

jimconner:
Works as expeted. Looks good. Merging.

dcarley:
Goodbye AWS stub. Goodbye me taking 5 attempts every time I want to see a VM definition 🎉

PR #278, opened 2016-05-20
WIP: [#119516287] Add tests for network MTU

dcarley:
## Notice

Please don't merge straight away. I'd like to get a few opinions from the team about the code and whether we think:

- this is something worth testing for future regressions
- is it possible to automate this in the pipeline?
  - would we build and transfer the binary? how?
  - would we use an errand on a job that has Golang available?
  - how would we choose the VM it runs on?
- is this worth merging before we have it automated in the pipeline?

## What

Here's an overview. Please see the commits for more details:

### Ginkgo suite for "run_from_vm"

Add a new test suite that will need to be run from a BOSH VM rather than a
Concourse or Diego container.

### tests/run_from_vm: Requests of different sizes

We experienced a bug in CF on AWS whereby TPS was unable to query Doppler
with an `Authorization` header that contained a username of a specific
length.

This test reproduces the same behaviour by making HTTP or HTTPS requests
with increasingly large packet sizes near to the standard network MTU of
1500 bytes. It should prevent the same from happening again in the future
without us needing to debug the internal CF calls.

### tests/run_from_vm: Check MTU of host

Test that the suite is being run from a host with a network interface
configured with an MTU of 9000~, because that is needed to exercise the bug
found by the other test in this suite.

## How to review

1. You'll need Go, Godep, and Ginkgo installed locally.
1. Run the following to build the test binary:

    ```
➜  paas-cf git:(feature/119516287-mtu_test) ✗ export GOPATH=${GOPATH}:$(pwd)/tests
➜  paas-cf git:(feature/119516287-mtu_test) ✗ cd tests/src/run_from_vm
➜  run_from_vm git:(feature/119516287-mtu_test) ✗ godep restore
➜  run_from_vm git:(feature/119516287-mtu_test) ✗ ginkgo build
Compiling run_from_vm...
    compiled run_from_vm.test
```
1. Transfer the binary to a VM such as `colocated_z1`. You might be to be creative about how you do this - I `scp`ed it to a personal VM and `wget`ed it back, but you could use SSH tunnels or S3.
1. Run the binary.
1. Confirm the host "MTU test" passes and the "request size" test fails.

## Who can review

Not @dcarley

### DIFF
--- a/tests/src/run_from_vm/Godeps/Godeps.json
+++ b/tests/src/run_from_vm/Godeps/Godeps.json
@@ -1,0 +1,141 @@
+{
+	"ImportPath": "run_from_vm",
+	"GoVersion": "go1.6",
+	"Deps": [
+		{
+			"ImportPath": "github.com/onsi/ginkgo",
+			"Comment": "v1.2.0-42-g07d85e6",
+			"Rev": "07d85e6b10c4289c7d612f9b13f45ba36f66d55b"
+		},
+		{
+			"ImportPath": "github.com/onsi/ginkgo/config",
+			"Comment": "v1.2.0-42-g07d85e6",
+			"Rev": "07d85e6b10c4289c7d612f9b13f45ba36f66d55b"
+		},
+		{
+			"ImportPath": "github.com/onsi/ginkgo/internal/codelocation",
+			"Comment": "v1.2.0-42-g07d85e6",
+			"Rev": "07d85e6b10c4289c7d612f9b13f45ba36f66d55b"
+		},
+		{
+			"ImportPath": "github.com/onsi/ginkgo/internal/containernode",
+			"Comment": "v1.2.0-42-g07d85e6",
+			"Rev": "07d85e6b10c4289c7d612f9b13f45ba36f66d55b"
+		},
+		{
+			"ImportPath": "github.com/onsi/ginkgo/internal/failer",
+			"Comment": "v1.2.0-42-g07d85e6",
+			"Rev": "07d85e6b10c4289c7d612f9b13f45ba36f66d55b"
+		},
+		{
+			"ImportPath": "github.com/onsi/ginkgo/internal/leafnodes",
+			"Comment": "v1.2.0-42-g07d85e6",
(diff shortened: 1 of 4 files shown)


jimconner:
Copying binaries onto the VMs feels a bit dirty. I wonder if this might be a good use case for the binary-buildpack - we could then run the MTU tests from within a garden container (assuming that the MTU of the container matches the MTU of it's parent VM - I don't have a deployed environment to check that on at the moment)

dcarley:
> we could then run the MTU tests from within a garden container (assuming that the MTU of the container matches the MTU of it's parent VM - I don't have a deployed environment to check that on at the moment)

It doesn't, unfortunately. The virtual interface gets `1500`.

dcarley:
NB: commit 88b5d1a would need to be omitted now that we've reduced the MTU of all hosts to 1500 in #301. Though I'm not 100% sure whether there's value in running this from hosts that don't have jumbo frames - whether that's ever likely to provoke a similar bug.

keymon:
Although the tests are OK and nicely crafted, I doubt about the value of testing a such specific issue in a specific IaaS that is about to be fixed. 

I think it is better to add a simpler test or monitoring that it checks that the solution (or workaround) is in place, in this case that the VMs will have the MTU configured properly to 1500. It would have associated the description.

Using bosh cli we could run something like: `bosh vms | grep running | awk '{print $2}' | xargs -n 1 -I{} bosh ssh {} "ifconfig | grep encap:Ethernet -A 4  | sed -n 's/.*MTU:\([0-9]*\).*/\1/p'"`

dcarley:
> I think it is better to add a simpler test or monitoring that it checks that the solution (or workaround) is in place, in this case that the VMs will have the MTU configured properly to 1500. It would have associated the description.

I'm hoping that the workaround won't be in place long enough for that to matter as much.

dcarley:
Closing after a discussion with Henry and Piotr. I still think in principle that this would be worth testing, but it's not worth the complexity and time required to test it - which would involve transferring binaries or creating errand releases.

PR #279, opened 2016-05-20
[#119090295] Separate bosh cli and bosh vms tests

saliceti:
## What
Story: [Process for recreating BOSH on AZ Failure](https://www.pivotaltracker.com/story/show/119090295)
There should be 2 different tests instead of one because they serve different purposes:
* making sure that make <env> bosh-cli works for humans
* making sure that VM agents are connected to BOSH after failover

## How to review
* Apply to new or existing environment
* There should be 2 different bosh tests now
* You can kill a CF VM to make bosh vms fail

## Who can review
Anyone but me

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1452,7 +1452,9 @@ jobs:
           - get: paas-cf
             passed: ['cf-deploy']
           - get: concourse-manifest
-      - task: test-bosh-vms-via-bosh-cli
+          - get: cf-manifest
+          - get: bosh-secrets
+      - task: test-bosh-cli
         config:
           platform: linux
           image: docker:///ruby#2.2-slim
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1472,6 +1472,26 @@ jobs:
...
+          - name: concourse-manifest
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
+              VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+              CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
+              export CONCOURSE_ATC_PASSWORD
```
dcarley: Can this task use the `bosh-cli` container and `bosh_login` script instead? It'll be much faster and simpler than looping back through Concourse.

dcarley:
> There should be 2 different tests instead of one because they serve different purposes:
> 
> making sure that make bosh-cli works for humans
> making sure that VM agents are connected to BOSH after failover

Could you add this to the commit message? 👼

saliceti:
@dcarley: Done

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1472,7 +1474,27 @@ jobs:
               CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
               export CONCOURSE_ATC_PASSWORD
               echo Looking for non running VMs
-              ./paas-cf/concourse/scripts/bosh-cli.sh bosh vms | tee vms.txt
+              ./paas-cf/concourse/scripts/bosh-cli.sh bosh status
+      - task: test-bosh-vms
+        config:
+          platform: linux
+          image: docker:///governmentpaas/bosh-cli
+          params:
```
dcarley: These params shouldn't be needed for this task now that it doesn't use bosh-cli.

dcarley:
Thanks 👍

PR #281, opened 2016-05-20
val_from_yaml: STDERR msg when unable to find key

dcarley:
## What

Write an error message to STDERR when `val_from_yaml` is unable to find a
key in the YAML document. This will make it easier to diagnose typos when
running from within the pipeline. Previously the non-zero exit code would
make the task fail without any output.

For example, when called as we do in the pipeline:

    ➜  paas-cf git:(master) ✗ TEST=$(./concourse/scripts/val_from_yaml.rb foo manifests/concourse-manifest/concourse-base.yml)
    Unable to find key: foo

## How to review

Confirm that the tests look correct and that they pass in Travis. The changes don't affect the happy path of the code, so it's probably not necessary for you to test the whole thing in a pipeline.

## Who can review

Not @dcarley

### DIFF
--- a/concourse/scripts/spec/val_from_yaml_spec.rb
+++ b/concourse/scripts/spec/val_from_yaml_spec.rb
@@ -26,13 +26,19 @@ val2: b
   it "exits non-zero with no output for non existing properties" do
     run_with_fixture("x.y.z")
     expect(last_command_started).to have_exit_status(1)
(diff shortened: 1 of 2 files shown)


jimconner:
Tests look correct to me, and they pass in travis. All looks good, so merging.

PR #282, opened 2016-05-20
[#111678348] SSL cert verify for internal Diego Communication

combor:
## What

As we are verifying SSL certificates everywhere there is no reason to skip verification for internal Diego components.

## How to review

Deploy from this branch and check if tests pass. 

## Who can review

NOT @combor

### DIFF
--- a/manifests/cf-manifest/deployments/030-cf-jobs.yml
+++ b/manifests/cf-manifest/deployments/030-cf-jobs.yml
@@ -763,7 +763,7 @@ properties:
   support_address: "http://support.cloudfoundry.com"
   description: null
   ssl:
-    skip_cert_verify: true
+    skip_cert_verify: false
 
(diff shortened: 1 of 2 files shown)


dcarley:
Looks like this has pulled in some unrelated commits from Alex. Does it need rebasing against master?

combor:
Fixed.

dcarley:
Looks like this is used by the jobs that currently live on our colocated VM:

- https://github.com/cloudfoundry/capi-release/search?utf8=%E2%9C%93&q=skip_cert_verify&type=Code

Confirmed that they now use `-skipCertVerify=false` (or variations thereof) 👍 

Acceptance tests failed on something that seems unrelated though. I'm re-running them.

dcarley:
Tests passed this time. I'm convinced that it was unrelated.

PR #285, opened 2016-05-23
[#113528323] destroy-microbosh: Abort if any deployments exist

dcarley:
## What

It's possible to run the `destroy-microbosh` pipeline before the
`destroy-cloudfoundry` pipeline. When this happens, all of the Cloud Foundry
VMs and their related EBS volumes are left "orphaned". You have to cleanup
manually using the AWS console, which is tiresome and prone to error.

Prevent this from happening by aborting the `destroy-microbosh` pipeline
before it calls `bosh-init delete` if there are any existing deployments.
The script lists the deployments that you need to delete.

To test this I've used a library called `mimic` to mock the minimal API
responses from BOSH and overridden `BOSH_CONFIG` to point `bosh_cli` at that
fake BOSH.

I looked at using `webmock` as we've used elsewhere, but it only works for
HTTP calls within the same Ruby interpreter and I want to test the script in
a different process using Aruba so that we can assert on the exit code and
STDERR. I looked at the library `webmock-server`, which allows you to
configure a daemon using normal webmock stubs, but it looked more
complicated and could have affected other tests in this suite.

## How to review

1. Checkout the branch and deploy:

  ```
git checkout feature/113528323-dont_destroy_if_deployments
```
1. Run the `create-bosh-cloudfoundry` pipeline to get a CF environment
1. Run the `destroy-microbosh` pipeline and see it fail with a message
1. If you wish, run `destroy-cloudfoundry` and then `destroy-microbosh`, see it pass.

## Who can review

Not @dcarley

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,4 @@ gem 'rspec', '~> 3.3'
 gem 'bosh_cli'
 gem 'webmock', "~> 1.24"
 gem 'rubocop'
+gem 'mimic'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -66,11 +66,17 @@ GEM
     highline (1.6.21)
     httpclient (2.7.1)
     jmespath (1.1.3)
+    json (1.8.3)
     json_pure (1.8.3)
     little-plugger (1.1.4)
     logging (1.8.2)
       little-plugger (>= 1.1.3)
       multi_json (>= 1.8.4)
+    mimic (0.4.3)
+      json
+      plist
+      rack
+      sinatra
     minitar (0.5.4)
     multi_json (1.11.2)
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: concourse/scripts/spec/bosh_pre_destroy.rb [hunk 1]
```diff
@@ -0,0 +1,52 @@
+require 'yaml'
```
alext: This file should be named ending `_spec.rb` otherwise it won't be picked up and run by default.

dcarley: Whoops, fixed and rebased.

alext:
I think this is ready to merge as soon as Github recovers enough to allow this to happen.

PR #283, opened 2016-05-20
[#118458297] Postgres single AZ plan

alext:
## What

This adds the first real service plan for postgres. This is a single AZ plan (not highly available). Other plans will follow in subsequent stories.

This additionally makes the following changes:

* Removes the temporary test plan that was in place solely to validate the broker.
* Update the trade-tariff quota to allow non-free plans
* Refactor the quota-setting task to me more robust, and have the logic in a single place.

## How to review

Deploy from this branch.  Verify that the new plan is available, and that the description matches what's specified in the spreadsheet linked from [the pivotal story](https://www.pivotaltracker.com/story/show/118458297).
Verify that the custom acceptance tests pass.

## Who can review

Anyone but myself.

PR #288, opened 2016-05-24
Remove unnecessary test setup in quota setting test.

alext:
## What

This was previously fixed as part of https://github.com/alphagov/paas-cf/pull/283, but due to Github having
issues at the time, a stale branch got merged which didn't include this
update.

## How to review

Read conversation on https://github.com/alphagov/paas-cf/pull/283 and verify that this change is good.

## Who can review

Anyone but myself.


saliceti:
Tests still work.
Thanks!

PR #289, opened 2016-05-24
[#118838943] Individual CF Production Accounts for Paas Team Members

timmow:
**What**

As a Paas Engineer in order to be able to track changes made to production I want team members to have their own CF Admin Accounts.

* Check out the branch
* Edit the makefile and remove the following line from the `dev` section

```
$(eval export DISABLE_USER_CREATION=true)
```

* Edit `config/admin_users.yml` and remove everyone except yourself
* Verify your email with SES in the dev account if you have not done this already - run

```
aws ses verify-email-identity --email-address <email> --region eu-west-1
```

* Click on the link in the generated email

* Run the create bosh cloudfoundry pipeline - you should be emailed some gpg encrypted credentials - check you can login with these.

It is possible to generate the email again by deleting the admin user using [cf-uaac](http://docs.cloudfoundry.org/adminguide/uaa-user-management.html) and re runing the pipeline

```
uaac token client get admin -s <uaac creds from state bucket>
uaac user delete <email-to-delete>
```

**Who should review this PR**

* Anyone but me, @jonty, @keymon


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1144,6 +1144,34 @@ jobs:
 
                   cf push
 
+        - task: sync-admin-users
+          config:
+            platform: linux
+            image: docker:///ruby#2.2
```
saliceti: Could we use [the cf-uaac container image](https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/cf-uaac)?
This would be faster, and its version of uaac is more recent.

Gemfile.lock
```diff
@@ -43,9 +45,18 @@ GEM
     builder (3.2.2)
     cf-uaa-lib (3.2.5)
       multi_json
+    cf-uaac (3.1.6)
```
saliceti: Could we use the most recent gem? The latest one is 3.2.0: https://rubygems.org/gems/cf-uaac

scripts/lib/mail_credentials_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    Dir.mktmpdir("gpg-mail") do |tmpdir|
+      GPGME::Engine.home_dir = tmpdir
+      GPGME::Key.import(response.body)
+      GPGME::Crypto.encrypt(message, :recipients => key, :always_trust => true)
+    end
+  end
+
+  def self.send_email(opts)
+    ses = Aws::SES::Client.new(:region => "eu-west-1")
```
saliceti: Can we change to a constant declared at the top (like DEFAULT_SOURCE_ADDRESS)

concourse/pipelines/create-bosh-cloudfoundry.yml
Jonty: @saliceti Sadly not, it doesn't include GPG.

Gemfile.lock
Jonty: @saliceti Seems reasonable, not sure why @keymon used 3.1.6.

scripts/lib/mail_credentials_helper.rb
Jonty: @saliceti Seems reasonable.

scripts/lib/mail_credentials_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+
+A CF admin user has been created for you with the following details:
+
+ - API url: #{api_url}
+ - Login: #{user[:username]}
+ - Password: Attached as a GPG encrypted file with key ID #{user[:gpg_key]}
+
+You can login and change your password executing:
+
+  gpg -d < paas_admin_password.gpg
```
saliceti: This would be clearer as a separate step before, such as:
Retrieve your password executing: gpg -d paas_admin_password.gpg

Also it looks like "<" is not necessary (at least on my mac)

scripts/lib/mail_credentials_helper.rb
```diff
@@ -0,0 +1,88 @@
...
+  cf login -a #{api_url} -u #{user[:username]}
+  cf passwd
+
+Regards,
+The Government PaaS Team.
+
+... We are building a World Class Goverment PaaS based on OpenSource... that's the best!!!
```
saliceti: I'd love to keep that, but...

scripts/sync-admin-users.rb
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'yaml'
+require 'net/https'
+
+require File.expand_path("../lib/mail_credentials_helper", __FILE__)
+require File.expand_path("../lib/uaa_sync_admin_users", __FILE__)
+
+def get_uaa_target(api_url)
```
saliceti: This looks convoluted as [we already know the UAA endpoint](https://github.com/alphagov/paas-cf/blob/53086d5f5e01204ed423ba711b91b38ce1e7174f/terraform/cloudfoundry/dns-records.tf#L9-L15)
Could we expose it as terraform output and read it as env variable?

scripts/sync-admin-users.rb
```diff
@@ -0,0 +1,66 @@
...
+    {
+      username: u["username"] || u["email"],
+      email: u["email"] || raise("User #{i} defined in file #{filename} is missing email"),
+      gpg_key: u["gpg_key"] || raise("User #{i} defined in file #{filename} is missing gpg_key"),
+    }
+  }
+end
+
+api_url = ARGV[0] || raise("You must pass API endpoint as first argument")
+users_filename = ARGV[1] || raise("You must pass API endpoint as second argument")
```
saliceti: cut/paste mistake?

scripts/lib/mail_credentials_helper.rb
Jonty: I've made this clearer.

The "<" means the input is read from STDIN rather than actually decrypting the file, which means it doesn't leave a decrypted credentials file on your hard drive.

scripts/lib/mail_credentials_helper.rb
Jonty: ...but it's awesome, so don't remove it?! 👍

scripts/lib/uaa_sync_admin_users.rb
```diff
@@ -0,0 +1,198 @@
...
+      else
+        log_level = @options[:log_level] || :info
+      end
+      @logger = Logger.new($stdout)
+      @logger.level = Logger::Severity.const_get(log_level.to_s.upcase)
+    end
+    @logger
+  end
+
+  def token=(token)
```
saliceti: Is this used anywhere?

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: It would be preferable not to run `bundle install` within the pipeline. I think it'd also be good to make the dependency more explicit. What do either of you think about adding it (and a test for the binary) to the `cf-uaac` container?

scripts/sync-admin-users.rb
Jonty: @saliceti I agree, however I don't think shipping with this is going to cause a huge problem and would rather not make such a fairly large change now.

scripts/sync-admin-users.rb
Jonty: @saliceti Well spotted, fixed.

scripts/lib/uaa_sync_admin_users.rb
Jonty: @saliceti Yes: https://github.com/alphagov/paas-cf/pull/289/files#diff-ad029f10343380c2cdce21020f848f7aR46

concourse/pipelines/create-bosh-cloudfoundry.yml
saliceti: This would add other gems like aws-sdk and mail that shouldn't be in uaac container.

As discussed, the current solution is ok for now and we may revisit it in another story.

scripts/lib/mail_credentials_helper.rb
Jonty: @saliceti I was wrong, this doesn't do in-place encryption. Fixing it now!

concourse/pipelines/create-bosh-cloudfoundry.yml
Jonty: @dcarley @saliceti I think we probably _should_ revisit it in another story, possibly one about pipeline speedup improvements.

scripts/spec/uaa_sync_admin_users_spec.rb
```diff
@@ -0,0 +1,303 @@
...
+        expect(@created_users[0]).to include(:username => "new_user")
+        expect(@deleted_users.length).to be(1)
+        expect(@deleted_users[0]).to include(:username => "removed_user")
+      end
+    end
+  end
+
+end
+
+def user_uuid(username)
```
alext: It would be good for these helper methods to be scoped inside the `describe`. This will prevent them from polluting the global namespace and potentially clashing with other tests.

scripts/lib/uaa_sync_admin_users.rb
Jonty: @saliceti Yes, that's a setter method, so this is using it: https://github.com/alphagov/paas-cf/blob/6fc39c62f282c3d13f822790bc6b590ae9028d6a/scripts/lib/uaa_sync_admin_users.rb#L110

scripts/spec/uaa_sync_admin_users_spec.rb
Jonty: @alext Done, just for you. There's no duck emoji, so have this pigeon instead: 🐦

scripts/lib/uaa_sync_admin_users.rb
saliceti: token is an instance variable. `self.token =` can be replaced by `@token =` to access it internally instead of via the setter.
The tests pass with this change and without the setter method.

saliceti:
The tests emit debug logs as info and it pollutes the output:
```
Randomized with seed 17699
..I, [2016-05-25T17:26:17.781492 #31967]  INFO -- : Creating user new_user
I, [2016-05-25T17:26:17.794869 #31967]  INFO -- : Adding user new_user to group cloud_controller.admin
I, [2016-05-25T17:26:17.794958 #31967]  INFO -- : Updating group cloud_controller.admin
I, [2016-05-25T17:26:17.801063 #31967]  INFO -- : Adding user new_user to group uaa.admin
I, [2016-05-25T17:26:17.801113 #31967]  INFO -- : Updating group uaa.admin
```
Could you change them to debug?
Ex: `self.get_logger.debug("Creating user #{user[:username]}")`

Jonty:
@saliceti Have raised the log level for that class to :warn only in tests, so in production it'll log any changes it makes but it won't make the tests noisy.

scripts/lib/uaa_sync_admin_users.rb
Jonty: @saliceti removed.

saliceti:
Thanks!

scripts/lib/mail_credentials_helper.rb
keymon: Ohhh, it was cool! :(

PR #290, opened 2016-05-25
[#119818859] Manifest cleanup (part 1)

alext:
## What

In order to make the transition to Bosh 2.0 features (to use the build-in zones support) easier, this refactors the cf-manifest to simplify the layout. I've removed the `aws/` subdirectory and merged all these properties into the top-level files. I've also moved some things around to make them more consistent.

This is the first part of the changes to the manifests. I'm raising this as a separate PR because the changes so far should be a no-op as far as the generated manifest is concerned, and are therefore easier to review separately.

## How to review

Code review of the changes, including the commit structure/messages.

These changes should all be a no-op in the generated manifest.  It should therefore be sufficient to compare the generated manifests before and after these changes.

* Generate a manifest from `master` using the test fixtures
```
cd manifests/cf-manifest
./build_manifest.sh spec/fixtures/terraform/*.yml spec/fixtures/cf-s* deployments/env-specific/cf-default.yml > /tmp/cf-manifest-before.yml
```
* Generate a manifest from this branch in the same way
```
git checkout origin/flatten_manifests_119818859
./build_manifest.sh spec/fixtures/terraform/*.yml spec/fixtures/cf-s* deployments/env-specific/cf-default.yml > /tmp/cf-manifest-after.yml
```
* Compare the 2 manifests
```
diff -u /tmp/cf-manifest-before.yml /tmp/cf-manifest-after.yml
```
* Check there are no significant differences (the only difference should be in the ordering of the `disk_pools` array.

## Who can review

Anyone but myself.


combor:
LGTM as this is no-op and I can't see and difference in diff ... merging.

PR #291, opened 2016-05-25
[#119818859] Manifest cleanup (part 2)

alext:
## What

In order to make the upcoming transition to cloud-config and the Bosh zones features easier we want to remove things from the manifests that aren't used. This therefore makes the following changes:

* Remove NFS job definitions and related properties
* Removed unused ha-proxy and Postgres job definitions
* Remove route_registrar FIXMEs
* Remove unused nginx release
* Remove some unused resource_pools.

See the individual commits for more details.

Note: this doesn't remove the unused Diego job definitions as we're likely to use them instead of the colocated VMs at some point.

## How to review

Deploy this branch. Due to the way Bosh handles jobs being removed from VMs, it's probably necessary to test both a clean deployment, and an update to an existing deployment.

## Who can review

Anyone but myself.


combor:
All tests have passed. Merging.

PR #292, opened 2016-05-26
Update bosh_cli gem to 1.3232.0

saliceti:
## What
The previous version was causing an error related to Bosh::Cli::Config.
We now pin the version to a known working version.

## How to review
* Run `bundle`
* Run the tests via `make test`
* The bosh_pre_destroy test should pass
* If it fails with an error related to Bosh::Cli::Config, you may have to remove the old bosh_cli gem, or recreate the gemset entirely.

## Who can review
Anyone but me

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gem 'aruba'
 gem 'rspec', '~> 3.3'
-gem 'bosh_cli'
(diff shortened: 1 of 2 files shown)


dcarley:
From conversation in Slack: we're going to combine this into #289 because we can't test it without adding `cf-uaac` which is done in that other PR.

PR #295, opened 2016-05-26
[#117512517] Check postgres instances deployed via broker are available

HenryTK:
## What

[#117512517 - Check postgres instances deployed via broker are available](https://www.pivotaltracker.com/story/show/117512517)

We want to know whether any changes to our environment affect the service instances and their binding to applications. For this we have deployed a health check application which serves a static endpoint, which replaces the existing health check application, and a database endpoint.

Pingdom health checks can now be set up to monitor the staging and production environments. This should result in the existing health check to be updated, and a new health check created for the database endpoint.

## How to review

* Deploy using `117512517_db_check_app`. 
	* Make sure `cf-deploy` passes.
	* Check The following end points in a browser:
	* `https://healthcheck.${DEPLOY_ENV}.dev.cloudpipelineapps.digital`
	* `https://healthcheck.${DEPLOY_ENV}.dev.cloudpipelineapps.digital/db`
* Set up the Pingdom checks for CI, staging, and production, using the instructions in the README. You can review the contents of the README in the process.
* Log into the Pingdom account and make sure there are checks for both end points for CI, staging and production. Instructions are in the README, which can also be reviewed for clarity.

## Who

Anyone but @henrytk or @combor


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1235,6 +1209,42 @@ jobs:
...
+                echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                cf create-org testers
+                cf set-quota testers test_apps
+                cf create-space healthcheck -o testers
+                cf target -o testers -s healthcheck
+
+                cd paas-cf/tests/example-apps/healthcheck
+                cf push --no-start
+
+                cf create-service postgres M-dedicated-9.5 healthcheck-db
```
alext: This needs to be written in an idempotent way (ie it needs to check if the service already exists before attempting to create it)

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1235,6 +1209,42 @@ jobs:
...
+                while ! cf service healthcheck-db | grep -q 'Status: create succeeded'; do
+                  echo "Waiting for creation of service to complete..."
+                  sleep 30
+                done
+                cf bind-service healthcheck healthcheck-db
```
alext: Similarly, this needs to check whether the binding exists before attempting to create it.

alext:
We need to add something to the destroy pipeline to ensure that the db instance gets destroyed before the CF install is deleted.

concourse/pipelines/create-bosh-cloudfoundry.yml
HenryTK: Running the `cf create-service` and `cf bind-service` commands when the services and bindings already exists does not cause a problem. It prints a friendly message saying it does not need to do anything and then exits 0.

alext:
> Remove the existing healthcheck-app from staging and production using the CF CLI.

Can we do this in the pipeline automatically rather than having to have this be a manual step. This is a much more reliable way of doing it.  We can leave it in with a FIXME comment so that we know to remove it later (eg https://github.com/alphagov/paas-cf/commit/8639c6a)

HenryTK:
> Can we do this in the pipeline automatically rather than having to have this be a manual step. This is a much more reliable way of doing it. We can leave it in with a FIXME comment so that we know to remove it later (eg 8639c6a)
We decided it would be less work to just do it as a one-off manual task. Do you think the automated approach would be quicker?

dcarley:
> We decided it would be less work to just do it as a one-off manual task. Do you think the automated approach would be quicker?

It would be preferable. It'll reduce the effort required by the person merging the PR, reduce the chance of it not being applied to an environment (e.g. CI wasn't listed), and increase visibility of how the change made it into our persistent/shared environments.

combor:
> reduce the chance of it not being applied to an environment

At the other hand it increases the chance of not being de-applied, and having leftovers like [this story](https://www.pivotaltracker.com/story/show/118085963) that we actually implemented right now, not realising that binaries were not updated and pingdom checks were left using http: https://github.com/alphagov/paas-cf/pull/295/files#diff-5ae1cbe3159d21d9c29141f9c68cc3bdR19

HenryTK:
@dcarley:
> reduce the chance of it not being applied to an environment (e.g. CI wasn't listed)

I thought we decided not to have checks against CI?

dcarley:
> At the other hand it increases the chance of not being de-applied

Piotr and I talked about this offline. It's alright for the "de-apply" code to be there for a while, since it's effectively a no-op. If we mark it with a `FIXME` comment then it'll be easier to identify and remove. Piotr also suggested requesting a new story in the story summary to remove the FIXME afterwards.

> I thought we decided not to have checks against CI?

Pingdom checks aren't configured to query the environment, but the app is still deployed.

concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -29,6 +50,16 @@ jobs:
...
+      - get: cf-secrets
+      - get: cf-manifest
+      - get: bosh-CA
+
+      - task: get-cf-cli-config
+        file: paas-cf/concourse/tasks/get-cf-cli-config.yml
+
+      - task: remove-healthcheck-app
+        file: paas-cf/concourse/tasks/remove-healthcheck-app.yml
+
```
keymon: Why is the healthcheck-app deleted during the autodelete?

concourse/pipelines/destroy-cloudfoundry.yml
```diff
@@ -103,7 +117,17 @@ jobs:
...
+          - get: bosh-CA
           - get: paas-cf
+          - get: cf-secrets
+          - get: cf-manifest
+
+      - task: get-cf-cli-config
+        file: paas-cf/concourse/tasks/get-cf-cli-config.yml
+
+      - task: remove-healthcheck-app
+        file: paas-cf/concourse/tasks/remove-healthcheck-app.yml
```
keymon: As we are deleting CF, the application will get deleted with all the platform, so there is no need to delete it during the destroy.

concourse/pipelines/autodelete-cloudfoundry.yml
combor: This not only removes an app itself but also removes RDS instance.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1258,6 +1232,44 @@ jobs:
 
                   cf push
 
+      - task: deploy-healthcheck
```
keymon: For the sake of consistency and to  make it more clear, shall we move this to a dedicated separated file in tasks? This way we can have a `paas-cf/concourse/tasks/remove-healthcheck-app.yml` and a `paas-cf/concourse/tasks/push-healthcheck-app.yml`

concourse/pipelines/autodelete-cloudfoundry.yml
dcarley: Would it be more explicit and faster to only delete the service? Is that possible?

combor: As I remember to delete a service you need to unbind an app. During unbinding the app is restarted so it takes time. Deletion should be quicker. I renamed the task name to `remove-healthcheck-db`.

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: This one is only in `create-bosh-cloudfoundry.yml` so there is no code duplication. I'm not sure about moving it to separate task as we prefer to have tasks inline if possible ( and reasonable ).

keymon:
Everything LGTM, I am about to merge

PR #296, opened 2016-05-27
[#118838943] Fix user sync scripts directory

saliceti:
## What
Fix the sync users directory issue from https://github.com/alphagov/paas-cf/pull/289

## How to review
Same test as https://github.com/alphagov/paas-cf/pull/289

## Who can review
Anyone but me


alext:
Code change looks good to me. The commit message could use some more detail though. At the least it should detail the current issue that this is fixing.

jimconner:
Looks good. Worked in my dev environment OK. Merging.

PR #301, opened 2016-05-27
[#120368533] Set MTU via bosh addon

saliceti:
## What
Story: [Reduce MTU of VMs to 1500](https://www.pivotaltracker.com/story/show/120368533)

Add a bosh addon job to set the MTU on the VMs.

This relies on PR https://github.com/alphagov/paas-os-conf-release/pull/2 to be reviewed at the same time. Once the paas-os-conf-release is merged, the [TEMP] commits here must be updated.

## How to review
* Deploy
* Check the MTU is 1500 where it was 9001 originally on api_z1 for example

```
$ ifconfig |grep MTU
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
```
## Who can review
Anyone but @jimconner or me

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -97,6 +97,12 @@ resources:
       uri: https://github.com/alphagov/paas-aws-broker-boshrelease.git
       tag_filter: {{cf_aws_broker_version}}
 
+  - name: os-conf-boshrelease
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-os-conf-release.git
+      tag_filter: {{cf_os_conf_version}}
+
   - name: graphite-nozzle
     type: git
     source:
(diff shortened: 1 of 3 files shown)


dcarley:
Could you add some detail to the commit messages about why we're doing this and what circumstance it will be safe to revert this in future?

It might seem like repetition, because we've been talking about MTU a lot and have a lot of context, but it could come in useful later if we have new people on the team or AWS don't fix the bug in the expected timescale.

combor:
I have verified that it works and we don't have MTU problem anymore by running:
```
export ELB="piotr-dev-cf-loggregator-2007548262.eu-west-1.elb.amazonaws.com"
curl -k -m 20 -v  https://$ELB:443/apps/3bf24a38-cfdb-4e55-ad38-83e1ad7823f7/containermetrics -H 'Host: doppler.cloud.service.gov.uk:443' -H 'Accept: ' -H 'User-Agent: Go-http-client/1.1'   -H "Authorization: bearer $(yes = | head -1227 | tr -d "\n")" -H 'Accept-Encoding: gzip' ; echo
```

PR #268, opened 2016-05-16
[#118830743] Command to check our github forks against upstream

Jonty:
## What

We want to keep an eye on the repositories we have forked as they quickly diverge from upstream, this tool allows us to do that in an automated fashion.

## How to review

This requires the python `pygithub` module, which can be installed using `pip install pygithub`.

Running `make find_diverged_forks GITHUB_TOKEN=<personal access token>` will output a list of forks on github belonging to the PaaS team that have diverged from the upstream repository, it will also output any open pull requests belonging to us on those forks.

While the script will run without a personal access token it will quickly run into rate limits due to the size of alphagov and the size of the repositories we have forked, so the makefile makes it mandatory.

You can get a personal access token here:
https://github.com/settings/tokens

The output should look similar to this:
````
$ make find_diverged_forks GITHUB_TOKEN=<REDACTED>
./scripts/find_diverged_forks.py alphagov --prefix=paas --extra-repo=cf-release --extra-repo=graphite-nozzle --github-token=82e9ff45189854f161b4c6c7d8dee8f1d4953058

alphagov/paas-terraform:
 * Upstream ahead by 2791 commits: https://github.com/alphagov/paas-terraform/compare/alphagov:master...hashicorp:master

alphagov/cf-release:
 * Upstream ahead by 1309 commits: https://github.com/alphagov/cf-release/compare/alphagov:master...cloudfoundry:master

alphagov/graphite-nozzle:
 * Upstream ahead by 4 commits: https://github.com/alphagov/graphite-nozzle/compare/alphagov:master...CloudCredo:master

alphagov/paas-graphite-statsd-boshrelease:
 * Open pull request: https://github.com/CloudCredo/graphite-statsd-boshrelease/pull/8

alphagov/paas-s3-resource:
 * Upstream ahead by 26 commits: https://github.com/alphagov/paas-s3-resource/compare/alphagov:master...concourse:master

alphagov/paas-semver-resource:
 * Upstream ahead by 26 commits: https://github.com/alphagov/paas-semver-resource/compare/alphagov:master...concourse:master

alphagov/paas-bosh-aws-cpi-release:
 * Upstream ahead by 141 commits: https://github.com/alphagov/paas-bosh-aws-cpi-release/compare/alphagov:master...cloudfoundry-incubator:master

alphagov/paas-docs-bosh:
 * Upstream ahead by 159 commits: https://github.com/alphagov/paas-docs-bosh/compare/alphagov:master...cloudfoundry:master

[...]

PR #303, opened 2016-06-01
[#118830743] Don't crash when master is not the default branch

Jonty:
Related: [#118830743 SPIKE: Investigate if forked code requires updating](https://www.pivotaltracker.com/n/projects/1275640/stories/118830743)

## What

In #268 we added a script `find_diverged_forks.py` to check our forked repositories to their upstream

This script, in cases where the default branch (of either our fork or the origin) is not `master`, will crash as the comparison is invalid. we must use the current default branch to allow tracking of branch renames and reconfigurations.

This also includes some slightly improved formatting, and a requirements.txt addition to simplify installation.

## How to review

1) Get a personal access token from https://github.com/settings/tokens
2) Run `./scripts/find_diverged_forks.py alphagov --github-token=<YOUR_TOKEN>` - this will check _every_ GDS fork
3) Wait for a really long time
4) Observe that it doesn't crash

## Who can review
Not @jonty

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 awscli>=1.10.16
 yamllint>=1.0.0
+PyGithub>=1.26.0
--- a/scripts/find_diverged_forks.py
+++ b/scripts/find_diverged_forks.py
@@ -2,6 +2,9 @@
 import argparse
 from github import Github
 
+BOLD = '\033[1m'
+ENDC = '\033[0m'
+
 class ForkChecker(object):
     def __init__(self, github_token):
         self.github_token = github_token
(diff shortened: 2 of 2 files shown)


keymon:
The description of the PR does not give enough context. 

What crashes?  is that a tool? what is the tool for? When was it added and for what?

keymon:
I also suggest change the commit title to something like: `find_diverged_forks.py: not crash if branch != master`

keymon:
For reference:
 - `find_diverged_forks.py` was added in #268 
 - The related story is [#118830743](https://www.pivotaltracker.com/n/projects/1275640/stories/118830743)

keymon:
I noticed that the egg `github` is not in the `requirements.txt` so in my case fails:

```
Traceback (most recent call last):
  File "./scripts/find_diverged_forks.py", line 3, in <module>
    from github import Github
ImportError: No module named github

```

can you add that dependency in this PR?

keymon:
Works as expected. @Jonty do you mind add that to `requirements.txt` and change the commit title to add more context. Once that is done I will merge.

Jonty:
@keymon All issues fixed.

PR #304, opened 2016-06-01
[#118466261] Force SSL in AWS RDS broker provisioned instances

keymon:
[#118466261 SPIKE: Ensuring applications use TLS to connect to postgres](https://www.pivotaltracker.com/story/show/118466261)

What
----

We want to ensure that applications only use TLS to talk to postgreSQL.

We want to be sure that the [instances provisioned only allow TLS connections](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#d0e59747) and test that the applications cannot connect in plain text.

Prerequisites
-------------

This PR must be merged and applied to all the environments:

https://github.gds/government-paas/aws-account-wide-terraform/pull/44

Context
-------

In this PR:

 * We create a new RDS Parameters Group for postgres provisioned instances for the RDS broker using terraform.
 * Add the RDS CA certificates to the rootfs image for the warden containers, so they can verify the SSL certificates from AWS.
 * Add a test to check that the apps cannot connect without using SSL.

How to test
-----------

Deploy as usual, all tests should pass.

Who?
----

Anyone but @keymon or @combor


terraform/cloudfoundry/rds_broker.tf
```diff
@@ -37,3 +36,21 @@ resource "aws_security_group" "rds_broker_dbs" {
...
   }
 }
+
+resource "aws_db_parameter_group" "rds_broker_postgres95" {
+  name = "rdsbroker-postgres95-${var.env}"
+  family = "postgres9.5"
+  description = "RDS Broker Postgres 9.5 parameter group"
+
+  parameter {
+    apply_method = "pending-reboot"
```
timmow: It might be useful to have some rationale in the commit message about why the `apply_method` is different for both parameters (although I have no context of db parameter groups so may be missing something)

timmow:
All LGTM except the one comment about the commit message, assuming acceptance tests pass I'm happy to merge

manifests/cf-manifest/deployments/040-cf-properties.yml
```diff
@@ -615,3 +615,290 @@ properties:
         client_key: (( grab secrets.bbs_client_key ))
         require_ssl: true
       traffic_controller_url: (( grab properties.loggregator.traffic_controller_url ))
+    rootfs_cflinuxfs2:
+      trusted_certs: |
+        -----BEGIN CERTIFICATE-----
```
dcarley: Do you think it would be easier to maintain this, especially if we add other CAs in the future, if it was a separate property/file that could be named (e.g. `aws_rds_certificates`)?

keymon:
@timmow I updated the commit. Actually both needed to be `pending-reboot`

I also split the properties for the SSL CA certs to a separate file.

PR #306, opened 2016-06-02
[#120446955] Remove Jim from admin users, and commit signers.

alext:
## What

Jim has left the team, so this removes him from the list of commit signers, and admin accounts.

## How to review

Verify that the correct key has been removed from `.gpg-id`. Verify that Jim has indeed been removed from `admin_accounts.yml`

## Who can review

Anyone but myself.


keymon:
LGTM

PR #309, opened 2016-06-03
[#118838943] Fix typo in admin sync logging

dcarley:
#### Use Hash#fetch in all admin sync scripts

Make all hash lookups where we're getting (not setting) a value use `#fetch`
instead of `#[]` so that an exception is raised if we mis-type the key name
or the object isn't populated as expected. Previously a `nil` object would
be returned and the scripts could fail in a silent or unclear way.

This change highlights one such case, whereby we're using a key name with
the wrong capitalisation, which I'll fix in a following commit. We don't
explicitly test the log function where it occurs, but it now makes the tests
fail because we exercise the code path and get an exception:

     Failure/Error: get_logger.info("Deleting admin user #{user_info.fetch("userName")} which is not in the list.")

     KeyError:
       key not found: "userName"

I haven't changed the `self.ua.all_pages` calls which return the first item
because it seems that it's expected for them to return `nil` if there are no
results.

#### Fix typo in admin sync logging

When `sync-admin-users` was first run in the staging and prod environments
it deleted a lot of old test users but the logs didn't indicate what their
usernames were:

    I, [2016-05-31T09:26:48.723823 #15203]  INFO -- : Deleting admin user  which is not in the list.
    I, [2016-05-31T09:26:48.840621 #15203]  INFO -- : Deleting admin user  which is not in the list.
    I, [2016-05-31T09:26:48.986202 #15203]  INFO -- : Deleting admin user  which is not in the list.
    I, [2016-05-31T09:26:49.097718 #15203]  INFO -- : Deleting admin user  which is not in the list.
    I, [2016-05-31T09:26:49.219233 #15203]  INFO -- : Deleting admin user  which is not in the list.
    …

This is because it was using a hash key with the incorrect capitalisation;
`userName` instead of `username`. This resulted in a `nil` object which is
cast to an empty string.

This makes the tests pass after the preceding commit to use `Hash#fetch`
made them fail and should prevent this from happening again in the future.

## How to review

Sorry, this is a bit involved..

To test deleting users you'll need to create one:

1. Use a Bundler setup with UAAC:

    ```
cd scripts/
```
1. Install the dependencies:

    ```
bundle install
```
1. Target your environment:

    ```
bundle exec uaac target https://uaa.${DEPLOY_ENV}.dev.cloudpipeline.digital
```
1. Login using the password from the state bucket:

    ```
bundle exec uaac token client get admin -s $(aws s3 cp s3://${DEPLOY_ENV}-state/cf-secrets.yml - | awk '/uaa_admin_client_secret/ {print $2}')
```
1. Add a user:

    ```
bundle exec uaac user add this-should-be-deleted -p "$(pwgen 16 1)" --emails ignored
```
1. Add them to the admin group:

    ```
bundle exec uaac member add uaa.admin this-should-be-deleted
```

To run this on your development environment you'll need to make some further changes:

1. Edit `config/admin_users.yml` and remove all users but yourself.
2. Apply the following patch to delete unmanaged users sooner:

    ```
diff --git a/scripts/lib/uaa_sync_admin_users.rb b/scripts/lib/uaa_sync_admin_users.rb
index f871701..9675837 100644
--- a/scripts/lib/uaa_sync_admin_users.rb
+++ b/scripts/lib/uaa_sync_admin_users.rb
@@ -21,7 +21,7 @@ class UaaSyncAdminUsers
     "doppler.firehose"
   ]

-  HOURS_TO_KEEP_TEST_USERS = 2
+  MINS_TO_KEEP_TEST_USERS = 2

   # Initialise the class with to the given UAA target with the given
   # credentials.
@@ -179,7 +179,7 @@ class UaaSyncAdminUsers
         raise "User #{user_id} not found"
       end

-      if Time.parse(user_info.fetch("meta").fetch("created")) < (Time.now - HOURS_TO_KEEP_TEST_USERS * 60 * 60)
+      if Time.parse(user_info.fetch("meta").fetch("created")) < (Time.now - MINS_TO_KEEP_TEST_USERS * 60)
         get_logger.info("Deleting admin user #{user_info.fetch("username")} which is not in the list.")
         ua.delete(:user, user_id)
         deleted_users << { username: user_info.fetch("username"), email: user_info.fetch("emails").fetch(0).fetch("value") }
```
1. Wait for 2 minutes to have passed since you added the temporary user.
1. Commit and push those changes to a temporary branch e.g. `wip/<yourname>`
1. Deploy that branch and enable account creation (normally disabled in dev):

    ```
NEW_ACCOUNT_EMAIL_ADDRESS=the-multi-cloud-paas-team@digital.cabinet-office.gov.uk BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines
```
1. Run `create-bosh-cloudfoundry` pipeline and check in the `cf-deploy` job that it has:
  1. Created yourself a user.
  1. Deleted the temporary user.

## Who can review

Not @dcarley

### DIFF
--- a/scripts/lib/mail_credentials_helper.rb
+++ b/scripts/lib/mail_credentials_helper.rb
@@ -28,8 +28,8 @@ module EmailCredentialsHelper
     region = opts.fetch(:region, DEFAULT_REGION)
     ses = Aws::SES::Client.new(:region => region)
     mail = Mail.new do
-      from    opts.fetch(:source_address)
-      to      opts.fetch(:email)
+      from    opts.fetch(:from)
+      to      opts.fetch(:to)
       subject opts.fetch(:subject)
 
       text_part do
--- a/scripts/lib/mail_credentials_helper.rb
+++ b/scripts/lib/mail_credentials_helper.rb
@@ -28,21 +28,21 @@ module EmailCredentialsHelper
     region = opts.fetch(:region, DEFAULT_REGION)
     ses = Aws::SES::Client.new(:region => region)
     mail = Mail.new do
-      from    opts[:source_address]
-      to      opts[:email]
-      subject opts[:subject]
+      from    opts.fetch(:source_address)
+      to      opts.fetch(:email)
+      subject opts.fetch(:subject)
 
       text_part do
-        body opts[:message]
+        body opts.fetch(:message)
       end
     end
-    opts[:attachments].each { | name, value |
+    opts.fetch(:attachments, {}).each { | name, value |
       mail.attachments[name] = value
     }
 
     ses.send_raw_email({
-      source: opts[:from],
-      destinations: [opts[:to]],
+      source: opts.fetch(:from),
+      destinations: [opts.fetch(:to)],
       raw_message: {
         data: mail.to_s,
       },
@@ -50,11 +50,11 @@ module EmailCredentialsHelper
   end
 
   def self.send_admin_credentials(api_url, user, source_address)
-    attachment_source = encrypt_message_to(user[:password], user[:gpg_key])
+    attachment_source = encrypt_message_to(user.fetch(:password), user.fetch(:gpg_key))
 
     send_email(
       :from           => source_address,
-      :to             => user[:email],
+      :to             => user.fetch(:email),
       :subject        => "PaaS admin account credentials",
       :message        => %Q{
 Hello,
@@ -62,8 +62,8 @@ Hello,
 A CF admin user has been created for you with the following details:
 
  - API url: #{api_url}
- - Login: #{user[:username]}
- - Password: Attached as a GPG encrypted file with key ID #{user[:gpg_key]}
+ - Login: #{user.fetch(:username)}
+ - Password: Attached as a GPG encrypted file with key ID #{user.fetch(:gpg_key)}
 
 You can decrypt the password attachment using:
 
@@ -71,7 +71,7 @@ You can decrypt the password attachment using:
 
 You can login and change your password by executing:
 
-  cf login -a #{api_url} -u #{user[:username]}
+  cf login -a #{api_url} -u #{user.fetch(:username)}
   cf passwd
 
 Regards,
--- a/scripts/lib/uaa_sync_admin_users.rb
+++ b/scripts/lib/uaa_sync_admin_users.rb
@@ -180,7 +180,7 @@ class UaaSyncAdminUsers
       end
 
       if Time.parse(user_info.fetch("meta").fetch("created")) < (Time.now - HOURS_TO_KEEP_TEST_USERS * 60 * 60)
-        get_logger.info("Deleting admin user #{user_info.fetch("userName")} which is not in the list.")
+        get_logger.info("Deleting admin user #{user_info.fetch("username")} which is not in the list.")
         ua.delete(:user, user_id)
         deleted_users << { username: user_info.fetch("username"), email: user_info.fetch("emails").fetch(0).fetch("value") }
       else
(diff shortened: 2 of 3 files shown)


alext:
I get an error when I run this:
```
Sending credentials to new user alex.tomlins@digital.cabinet-office.gov.uk
bundler: failed to load command: sync-admin-users.rb (sync-admin-users.rb)
KeyError: key not found: :source_address
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:31:in `fetch'
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:31:in `block in send_email'
  /usr/local/bundle/gems/mail-2.6.4/lib/mail/message.rb:133:in `instance_eval'
  /usr/local/bundle/gems/mail-2.6.4/lib/mail/message.rb:133:in `initialize'
  /usr/local/bundle/gems/mail-2.6.4/lib/mail/mail.rb:51:in `new'
  /usr/local/bundle/gems/mail-2.6.4/lib/mail/mail.rb:51:in `new'
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:30:in `send_email'
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:55:in `send_admin_credentials'
  sync-admin-users.rb:56:in `block in <top (required)>'
  sync-admin-users.rb:54:in `each'
  sync-admin-users.rb:54:in `<top (required)>'
```

I think this is because the keys don't match in the method call chain there.  `send_admin_credentials` [calls `send_email`](https://github.com/alphagov/paas-cf/blob/c7434aafe994b6f11ab226d2c410f73eb54cf84e/scripts/lib/mail_credentials_helper.rb#L55-L59) with `:from` and `:to` options, but `send_email` then [looks for](https://github.com/alphagov/paas-cf/blob/c7434aafe994b6f11ab226d2c410f73eb54cf84e/scripts/lib/mail_credentials_helper.rb#L30-L33) `:source_address`, and `:email`.

dcarley:
Ah, thanks. I think I failed to re-test the account creation after making some further changes. I'll investigate..

dcarley:
OK.. fixed, tested, and pushed.

I didn't get the email because my recipient address isn't verified, but I think that's good enough:
```
Sending credentials to new user dan.carley@digital.cabinet-office.gov.uk
bundler: failed to load command: sync-admin-users.rb (sync-admin-users.rb)
Aws::SES::Errors::MessageRejected: Email address is not verified. The following identities failed the check in region EU-WEST-1: dan.carley@digital.cabinet-office.gov.uk
  /usr/local/bundle/gems/aws-sdk-core-2.3.8/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
  /usr/local/bundle/gems/aws-sdk-core-2.3.8/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'
  /usr/local/bundle/gems/aws-sdk-core-2.3.8/lib/seahorse/client/plugins/response_target.rb:21:in `call'
  /usr/local/bundle/gems/aws-sdk-core-2.3.8/lib/seahorse/client/request.rb:70:in `send_request'
  /usr/local/bundle/gems/aws-sdk-core-2.3.8/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:43:in `send_email'
  /tmp/build/fc57e4e1/paas-cf/scripts/lib/mail_credentials_helper.rb:55:in `send_admin_credentials'
  sync-admin-users.rb:56:in `block in <top (required)>'
  sync-admin-users.rb:54:in `each'
  sync-admin-users.rb:54:in `<top (required)>'
```

dcarley:
@alext bump!

PR #311, opened 2016-06-06
[#120846679] Feature flag destroy pipelines off in non-dev envs

dcarley:
## Notice

Before merging we need to capture logs from the `destroy-cloudfoundry` pipeline in prod. I'm not able to do this at the moment, so if somebody else could..

## What

Add a new `ENABLE_DESTROY` feature flag to control whether the
`destroy-cloudfoundry` and `destroy-microbosh` pipelines are setup on the
Deployer Concourse. This is enabled on development environments and disabled
by default on all other environments such as CI, staging, and production.
Instructions for enabling the destroy pipelines in these environments, if
necessary, is added to the non-dev deployments documentation.

This will prevent the destroy pipelines from being run by accident in
persistent environments. We've seen this happen once and it's easy to
do when you have multiple Concourse tabs open to diagnose a pipeline failure
in another environment.

I've not included a warning like `ENABLE_AUTODELETE` does because it will be
quite noisy and the effect isn't as surprising as autodelete; which contains
a pipeline that runs automatically. I've used a glob to capture both destroy
pipelines in the case statement so that we don't have to match their exact
names. These pipelines aren't reinstated by `self-update-pipeline` because
that passing in a single pipeline name.

We have talked about adding a Concourse feature to make the background a
different colour for different environments or adding some text to the
pipeline screen. Those things might still be desirable, but we should leave
this in place for extra safety.

## How to review

1. Checkout the branch: `git fetch && git checkout 120846679-feature_flag_destroy_pipelines`
1. Update the pipelines on your dev environment: `make dev pipelines`
1. Confirm that the destroy pipelines are still on the Deployer Concourse.
1. Temporarily delete the `ENABLE_DESTROY=true` line from the `Makefile` to simulate a non-dev environment.
1. Update the pipelines again: `make dev pipelines`
1. Confirm that the destroy pipelines are *not* on the Deployer Concourse.
1. Run the `create-bosh-cloudfoundry` and confirm that the `init` job doesn't reinstate them.

## Who can review

Not @dcarley


doc/non_dev_deployments.md
```diff
@@ -102,3 +102,12 @@ AWS key to your own, it's necessary to export the concourse password:
...
 ```
+
+## Destroying an existing environment
+
+The pipelines to destroy Cloud Foundry and MicroBOSH are not setup
+automatically on non-development environments in order to prevent accidental
+deletion. They can be setup if you do need them by running:
+```
+ENABLE_DESTROY=true make <env> pipelines
+```
```
alext: This doesn't quite work. This adds the pipeline back in, but then when you run the destroy pipeline, the `self-update-pipeline` step deletes the pipeline while it's running.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -104,12 +104,22 @@ update_pipeline() {
...
 
   case $pipeline_name in
-    create-bosh-cloudfoundry|destroy-microbosh|destroy-cloudfoundry|failure-testing)
+    create-bosh-cloudfoundry|failure-testing)
       bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
         "${env}" "${pipeline_name}" \
         <(generate_manifest_file) \
         <(generate_vars_file)
     ;;
+      destroy-*)
```
alext: The indentation is off on this line.

doc/non_dev_deployments.md
dcarley: Good catch. I've added the environment variables into the `params` of those jobs so that it's persisted.

Do you want me to rewrite the review instructions to include this?

concourse/scripts/pipelines-bosh-cloudfoundry.sh
dcarley: Fixed

alext:
I'll apply this to CI/staging/prod environments.

alext:
> I'll apply this to CI/staging/prod environments.

Done.

dcarley:
Thanks @alext 👍

PR #310, opened 2016-06-03
[#112234445] Use correct argument for aws_route_table gateway

dcarley:
## What

Change `terraform/cloudfoundry/` to use the `nat_gateway_id` argument
instead of `gateway_id` because we're passing the ID of a NAT Gateway rather
than an Internet Gateway.

When creating or updating a `route_table` resource the AWS API will let you
use either argument interchangeably for either type of gateway. However,
when you read it back, it modifies it to the right argument. This causes
Terraform to think that the resource needs updating on every run.

I've started a discussion with upstream in hashicorp/terraform#6551 about
how this could be handled in a more explicit way.

In the meantime this change will stop Terraform from modifying our route
tables on every run of the `cf-terraform` job:

    aws_route_table.internet.1: Modifying...
      route.28127694.cidr_block:                  "0.0.0.0/0" => ""
      route.28127694.gateway_id:                  "" => ""
      route.28127694.instance_id:                 "" => ""
      route.28127694.nat_gateway_id:              "nat-0435bb1217614301c" => ""
      route.28127694.network_interface_id:        "" => ""
      route.28127694.vpc_peering_connection_id:   "" => ""
      route.3849577180.cidr_block:                "" => "0.0.0.0/0"
      route.3849577180.gateway_id:                "" => "nat-0435bb1217614301c"
      route.3849577180.instance_id:               "" => ""
      route.3849577180.nat_gateway_id:            "" => ""
      route.3849577180.network_interface_id:      "" => ""
      route.3849577180.vpc_peering_connection_id: "" => ""
    aws_route_table.internet.0: Modifying...
      route.1177416365.cidr_block:                "" => "0.0.0.0/0"
      route.1177416365.gateway_id:                "" => "nat-08218b6154223426a"
      route.1177416365.instance_id:               "" => ""
      route.1177416365.nat_gateway_id:            "" => ""
      route.1177416365.network_interface_id:      "" => ""
      route.1177416365.vpc_peering_connection_id: "" => ""
      route.2849130247.cidr_block:                "0.0.0.0/0" => ""
      route.2849130247.gateway_id:                "" => ""
      route.2849130247.instance_id:               "" => ""
      route.2849130247.nat_gateway_id:            "nat-08218b6154223426a" => ""
      route.2849130247.network_interface_id:      "" => ""
      route.2849130247.vpc_peering_connection_id: "" => ""
    aws_route_table.internet.2: Modifying...
      route.2265194627.cidr_block:                "0.0.0.0/0" => ""
[...]

PR #313, opened 2016-06-06
[#120938265] Change GPG key ID for Hector Rivas Gandara pt2

dcarley:
## What

I forgot to sign the merge commit for #312. So I've reverted the change and reapplied.

## How to review

1. Verify that this PR makes no changes.
1. Merge in the correct way.

## Who can review

Not @dcarley. It's probably OK for Hector to review since I OKed the original change.


dcarley:
No need to do this now that #310 has been merged in the meantime.

PR #314, opened 2016-06-07
WIP: [#118479861] Configure prefix master password generation with a seed

keymon:
What
----

In a previous version of rds-broker, if you know the service GUID for a database instance, you can  determine the master password. This should not be possible as this password is supposed to be secret.

In https://github.com/alphagov/paas-rds-broker/pull/7 we added a configuration variable master_password_seed which is added to the service GUID before hashing. This means that it's no longer possible to determine the master password from the instance_id alone.

This PR we point to the new version of aws-broker-boshrelease which supports this.

Dependency
----------

We must merge first:
 - https://github.com/alphagov/paas-rds-broker/pull/7
 - https://github.com/alphagov/paas-aws-broker-boshrelease/pull/7

But you can test this PR now before merging.

After merge that PR, you must tag https://github.com/alphagov/paas-aws-broker-boshrelease to 0.0.4 and remove the latest commit of this PR.

Considerations
-------------------

As we changed the method to generate the master password, we need to update it for the existing DBs. We solved by using the script `scripts/update-rds-broker-passwords.sh` that runs after CF deploy. 

We must revert the latest commit once it has been deployed in all the environments.

How to test
-----------

Deploy CF and run the custom acceptance tests.

You can also check that the exiting DBs have their master password updated, by binding an existing instance (e.g. `healthcheck_db`) to other app:

```
cd tests/example-apps/static-app
cf org testers && cf space healthcheck
cf push my-static-app
cf bind-service my-static-app healthcheck-db # It should work

cf delete my-static-app
```

Who?
----

Anyone but @keymon or @alext

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1246,6 +1246,27 @@ jobs:
 
                   cf push
 
+      # FIXME: Delete this once https://github.com/alphagov/paas-cf/pull/314 is
+      # deployed in prod
+      - task: temp-update-rds-broker-master-password
+        config:
+          platform: linux
+          image: docker:///governmentpaas/awscli
+          inputs:
+            - name: paas-cf
+            - name: cf-secrets
+          params:
+            AWS_DEFAULT_REGION: {{aws_region}}
+            DEPLOY_ENV: {{deploy_env}}
+          run:
+            path: sh
+            args:
+              - -e
+              - -u
+              - -c
+              - |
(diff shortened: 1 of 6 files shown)


dcarley:
Are there any special considerations for deploying this to existing environments that have service instances created?

keymon:
> Are there any special considerations for deploying this to existing environments that have service instances created?

We are working on that. We have created a temporary script to update the RDS instances.

keymon:
we decided not implement this in paas-cf until the implementation of ration of passwords is in place #118841005

PR #315, opened 2016-06-07
[#119269637] Move to cloud config

mtekel:
## What

[Move to cloud config](https://www.pivotaltracker.com/n/projects/1275640/stories/119269637). This PR includes:
* new manifest directory structure
* new, unified way to render manifests
* cleanup of unnecessary properties and settings
* various minor fixes to make move easier
* move of cf-manifest to cloud config + manifest
* update pipeline to facilitate cloud config rendering and stemcell upload (as this is now not done by BOSH anymore)
* further simplify manifest where possible
* adjust tests: 
  -  remove tests for things that are no longer in cf-manifest
  -  add tests for cloud-config
  -  add cloud-config and cf-manifest integration tests
* ensure persistent data and job names are migrated
* ensure interrupt-less future deploys

# NOTE: THERE'S NO WAY BACK AFTER CLOUD CONFIG
if you want to go back to resource_pools and manifest only, you need to delete your BOSH (inc. database). Don't forget to run cleanup [repeatedly](https://github.com/cloudfoundry/bosh/issues/1294)
## How to review

Do at least these 3 intependent tests:

* Review commits and code
* Build from scratch. All tests should pass.
* Apply against existing deployment (built from master). Upgrade/migration takes about 1 hour, there is some downtime. Verify that persistent data for graphite and elasticsearch was migrated correctly. You should be able to see historical data in the deployment after migration. All tests should pass.

## Who can review

Not @mtekel or @HenryTK or @saliceti


manifests/cf-manifest/cloud-config/000-base-cloud-config.yml
```diff
@@ -0,0 +1,21 @@
...
+- name: z1
+  cloud_properties: {availability_zone: eu-west-1a}
+- name: z2
+  cloud_properties: {availability_zone: eu-west-1b}
+- name: z3
+  cloud_properties: {availability_zone: eu-west-1c}
+
+compilation:
+  vm_type: compilation
+  az: z1
```
combor: Do we need to hardcode this ?

manifests/cf-manifest/deployments/030-cf-jobs.yml
```diff
@@ -231,6 +231,8 @@ meta:
 disk_pools:
   - name: database_disks
     disk_size: 1024
+    cloud_properties:
+      type: gp2
```
keymon: This probably can be merged with b60a5596b2daba079857f4c0d9a430eb6cb9a1ea

manifests/cf-manifest/cloud-config/010-cf-networks.yml
```diff
@@ -1,7 +1,6 @@
 ---
-
 networks:
-- name: cf1
+- name: cf
   subnets:
     - range: 10.0.16.0/24
       reserved:
```
combor: We, don't want to reserve 10.0.16.1 ? And the dynamic range goes to 10.0.16.255 ( which is broadcast ) not .254

manifests/cf-manifest/cloud-config/000-base-cloud-config.yml
mtekel: yes, you HAVE TO specify az

keymon: So, you can only compile in one AZ? Well, I guess it is not a huge risk, one can always change that.

mtekel: see https://bosh.io/docs/cloud-config.html#compilation

manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
```diff
@@ -0,0 +1,125 @@
...
+  default_env:
+    bosh:
+      password: (( grab secrets.vcap_password ))
+
+vm_types:
+  - name: small
+    network: cf
+    env: (( grab meta.default_env ))
+    cloud_properties:
+      instance_type: m3.medium
```
combor: Do we really want to have `m3.medium` machine for `small` type?

manifests/cf-manifest/cloud-config/010-cf-networks.yml
mtekel: Even more so, dynamic range overlaps with reserved...

manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
mtekel: `m3.medium` is the smallest of the new instances. Normally T2 instances would be sufficient for this, but you only get 100 T2 instance starts per day per account (not sure if this is even per master account) which start with full CPU credits. As we auto-delete our environments, just the casual morning re-deployment of the environments exhausts this. When T2 instances start without CPU credits, they are _very_ slow.

mtekel: Btw. new stemcells are HVM and from previous gen instances, only CC2, CR1, HI1, and HS1 support it. All of which are quite huge instances.

manifests/cf-manifest/cloud-config/010-cf-networks.yml
mtekel: Btw. this has been there before, in this PR we only rewrite BOSH semantics, so I'll fix this in a new commit (or potentially a new PR - as this change takes a while to test).

manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
combor: Shall we than remove this block and move all references to `small` type to medium? It's a little bit confusing now.

manifests/cf-manifest/cloud-config/010-cf-networks.yml
combor: New commit sounds good. It's a perfect opportunity now to fix such small things.

manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
mtekel: On the 2nd look, we use only 7 smalls in the current deployment, potentially removing 2 of them in the future. We could switch to T2.small, which should still have 20% of CPU performance with no CPU credits - which looks like it's enough for the type of workload small instances do in our deployment. Added to improvements spreadsheet.

mtekel: Indeed our small are the same as mediums, but I think it's good to keep different references, as that indicates how much resources the given job really needs. We could also completely review and optimise this, however due to nature of T2 we probably can't switch too many VMs to it.

mtekel:
Squashed commit (see @keymon comment above), made zone name be picked up from TF outputs instead of hardcoded.

manifests/cf-manifest/spec/networks_spec.rb
```diff
@@ -1,28 +0,0 @@
...
-
-  CF_NETWORK_NAMES.each do |net_name|
-    describe "#{net_name} network" do
-      let(:network) { networks.find {|n| n["name"] == net_name } }
-
-      it "should have a single subnet" do
-        expect(network["subnets"].length).to eq(1)
-      end
-
-      it "should set the correct subnet ID" do
```
combor: I think we have lost this test after move to cloud-config.

manifests/cf-manifest/spec/resource_pools_spec.rb
```diff
@@ -1,49 +0,0 @@
...
-        let(:pool) { resource_pools.find {|p| p["name"] == "#{base_name}_#{zone_suffix}" } }
-
-        it "should specify the correct AWS AZ" do
-          expect(pool["cloud_properties"]["availability_zone"]).to eq(terraform_fixture(ZONE_KEYS[zone_suffix]))
-        end
-      end
-    end
-  end
-
-  describe "the compilation pool" do
```
combor: This test seems to be gone in new CC world.

manifests/cf-manifest/spec/networks_spec.rb
combor: No, sorry, it has been moved here: https://github.com/alphagov/paas-cf/blob/cc-move-2/manifests/cf-manifest/spec/cloud-config/networks_spec.rb#L19

manifests/cf-manifest/spec/resource_pools_spec.rb
saliceti: Thanks. I've added new specific tests for compilation.

combor:
`cf-deploy` job failed with version miss-match for stemcell. Upload-stemcell task pushed `3232.6` to bosh:
```
  Started update stemcell > Uploading stemcell bosh-aws-xen-hvm-ubuntu-trusty-go_agent/3232.6 to the cloud. Done (00:00:07)
```
even if version `3232.4` was properly extracted from manifest and passed to the concourse pipeline. Then the deployment failed with:
```
Director task 13
  Started preparing deployment > Preparing deployment. Failed: Stemcell `bosh-aws-xen-hvm-ubuntu-trusty-go_agent/3232.4' doesn't exist (00:00:02)

Error 50003: Stemcell `bosh-aws-xen-hvm-ubuntu-trusty-go_agent/3232.4' doesn't exist
```

keymon:
Reading the new schema, we could move the `cloud_properties` to the new [`vm_extensions`](https://bosh.io/docs/cloud-config.html#vm-extensions).

With that we can define our ELBs and the security groups or AWS roles as "VM Extensions" which can be [specified per job](https://bosh.io/docs/manifest-v2.html#instance-groups). If we do so, we can get rid of several `vm_types` that only exist to be able to custom those specific properties.

But I am OK to do this in a different story (if we do, we should create a story).

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -913,56 +928,90 @@ jobs:
...
+                - -e
+                - -c
+                - |
+                  echo "Generating cloud config..."
 
-                  echo "Converting ${cn} certs to YAML"
-                  ../paas-cf/concourse/scripts/file-to-yaml.sh secrets ${yaml_key_name}_key ${cn}.key > ${cn}_key.yml
-                  ../paas-cf/concourse/scripts/file-to-yaml.sh secrets ${yaml_key_name}_cert ${cn}.crt > ${cn}_cert.yml
-                done
+                  # FIXME: Remove eval after https://github.com/concourse/concourse/issues/360
```
keymon: mmm are you sure this is the same issue? You are not expanding a concourse parameter, you are using an environment variable.

dcarley:
> Reading the new schema, we could move the cloud_properties to the new vm_extensions.
> 
> With that we can define our ELBs and the security groups or AWS roles as "VM Extensions" which can be specified per job. If we do so, we can get rid of several vm_types that only exist to be able to custom those specific properties.
> 
> But I am OK to do this in a different story (if we do, we should create a story).

Sounds good, but definitely a separate story. This is already huge 😸

concourse/pipelines/create-bosh-cloudfoundry.yml
mtekel: In order to get rid of the quotes added by concourse, we need to do the additional eval...

combor:
`bosh-io-stemcell-resource` check script has a bug:

```
{
  "source": {
    "name": "bosh-aws-xen-hvm-ubuntu-trusty-go_agent"
  },
  "version": { "ref": "3232.4" }
}
```

```
cat x.json | ./check
[
  {
    "version": "3232.6"
  }
]
```

combor:
Ok, that's not a bug. It seems that we misread documentation ( but it's confusing anyway ):

>  otherwise check returns all versions from the version specified on.

concourse/pipelines/create-bosh-cloudfoundry.yml
mtekel: Good find, we don't need to eval there...

mtekel:
Will remove the stemcell resource & download stemcell every time. The micro stemcell file size is about 25KB. We can fix/extend stemcell resource in some other story, to let specify version of stemcell to get...

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -988,40 +1032,76 @@ jobs:
...
+            inputs:
+              - name: bosh-secrets
+              - name: paas-cf
+            run:
+              path: sh
+              args:
+                - -e
+                - -c
+                - |
+                  wget https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v={{stemcell_version}} -O stemcell.tgz
```
combor: That's minor but it would be better to get stemcell name from manifest instead hardcoding it here.
But that can be another story so we can figure out how to get all the required stemcells if we will decide to use multiple different stemcells/versions for jobs.

mtekel: Speaking of which, I can probably val_from_yaml it directly instead of passing the version as a parameter to fly...

combor: True, you don't have to pass it to the resource any more.

dcarley:
I'm deploying this to my existing dev environment. Have setup Pingdom checks with a frequency of 1min (the min, it seems) to get an idea of downtime.

combor:
My deployment from scratch succeeded and all checks passed. I'll destroy it now and re-apply on the deployment from master.

dcarley:
Deploy took 1h1m. Pingdom report 404s from the healthcheck app for 4m at the beginning of the deployment. I wasn't able to tell (from smoke tests or logsearch) how API operations were affected during the deploy, but I did see it update all of the jobs one-by-one except `consul`. Acceptance tests passed. So 👍  from me.

@combor is starting a test now.

combor:
Test reported 15mins downtime for apps but we have decided to give it a go so ... merging.

PR #318, opened 2016-06-10
[#121291013] Add copy recipients to user create script

combor:
## What

We need to create CF users for people without access to email accounts so we need
to CC all the emails to one common address so they can get their login details.

## How to review
```
scripts/create-user.sh -r -e your.name@digital.cabinet-office.gov.uk -c cc.name@digital.cabinet-office.gov.uk -o GDS
```

## Who can review

not @combor

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -31,7 +31,7 @@ usage() {
   cat <<EOF
 Usage:
 
-  $SCRIPT [-r] [-m] -e <email> -o <orgname>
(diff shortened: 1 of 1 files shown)


dcarley:
@combor @Rory80Hz Having second thoughts after kick-off. I'm not sure that we'd use this functionality normally, because we wouldn't want to know the passwords of tenants. Should we use a single generic account for the user research session?

Rory80Hz:
@combor @dcarley agree that this isn't something we want in future. Let's not merge this in...

PR #321, opened 2016-06-16
[#119272203] Test application availability during deployment of CF

keymon:
[#119272203 Test Upgrade CF without application downtime](https://www.pivotaltracker.com/projects/1275640/stories/119272203)

## What

We want to test in every deployment that there is no downtime in the application.

## Context

We decided to implement this logic:

 1. Add a test that constantly queries the `healthcheck app` (e.g. every 100ms)
 2. The test will be triggered just before any deployment (terraform, bosh, CF deploy...)
 3. The test will monitor concourse to check if `cf-deploy` finished successfully.
     If it runs 
 4. The test will stop if:
     - `cf-deploy` has finished
     - timeout after 2h
     - the rate of errors after 15 seconds is > 50%
 5. The test will be successful if `cf-deploy` finished in the given time and no errors were detected.
 6. We add some logic to wait for the test to start running by checking the logs of `healthcheck app`  

## Issues of this PR

 * We assume healthcheck app is deployed and accessible. If not the test will fail.
 * We assume CF is already deployed, if not the the test will fail.
 * If this is the first deployment and CF is not deployed:
    * pre-deploy will wait for 5m for the test to run, but pas green after.
    * `availability-test` will fail and the rest tests will not be triggered.
    * Re-running `availability-test` after that will succeed immediately, because `cf-deploy` is finished.

## How to review

Review the code.

Run the pipeline with:
 * A fresh deployment (`availability-test` will fail)
 * A deployment on a already running environment.
 * A deployment changing the stemcell version (Note: optional)

## Who can review
 
Anyone but @keymon or @combor


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -309,9 +309,9 @@ jobs:
     plan:
       - get: pipeline-trigger
         trigger: true
-        passed: [pipeline-lock]
+        passed: ['pipeline-lock']
       - get: paas-cf
```
mtekel: Less is sometimes more. I'd normalise on format that doesn't add quotes around job names... (we don't have jobs that have space in name).

keymon: The majority of cases is with quotes ¯\_(ツ)_/¯

mtekel: That sounds like opportunity for improvement.... :)

tests/src/availability/healthcheck_availability_during_deployment_test.go
```diff
@@ -0,0 +1,118 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	availabilityTestRate              = 10
+	availabilityTestMaxDuration       = 2 * time.Hour
+	availabilityTestMaxLatency        = 300 * time.Millisecond
+	availabilityTestMaxErrors         = 0
+	availabilitySuccessyRateThreshold = 99.9
```
mtekel: availability _Success_ RateThreshold ?

tests/src/availability/healthcheck_availability_during_deployment_test.go
```diff
@@ -108,7 +104,7 @@ var _ = Describe("Availability test", func() {
 			attacker.Stop()
 			metrics.Close()
 
-			Expect(len(metrics.Success*100)).To(
+			Expect(metrics.Success*100).To(
```
mtekel: squash this in to previous commit?

tests/src/availability/concourse_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+func getConcoursePassword() string {
+	return getConfigFromEnvironment("CONCOURSE_ATC_PASSWORD")
+}
+func getSkipSSLValidation() bool {
+	return getConfigFromEnvironment("SKIP_SSL_VALIDATION") == "true"
+}
+
+func filterAndSortBuildsByJobName(builds []atc.Build, jobName string) []atc.Build {
+	var filteredBuilds []atc.Build
```
mtekel: just a nitpick, but it should be more like filterByNameAndSortByTime, or more simply SortByTime(builds, jobname) - jobname parameter indicates that you want to sort this job by time

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1750,10 +1789,10 @@ jobs:
     plan:
       - aggregate:
           - get: pipeline-trigger
-            passed: ['performance-tests']
+            passed: ['performance-tests', 'smoke-tests', 'acceptance-tests', 'custom-acceptance-tests', 'bosh-tests']
             trigger: true
```
mtekel: this adds a lot of lines to concourse. I am not sure this is necessary, as previous job of performance tests depends on all these, so they have to be passed if perf-test is passed....

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -483,6 +486,59 @@ jobs:
...
+                  PIPELINE_TRIGGER_VERSION=$(cat pipeline-trigger/number)
+
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS} > /dev/null
+                  cf target -o testers -s healthcheck > /dev/null
+
+                  echo "Waiting two minutes for availability test to start:"
+                  for i in $(seq 24); do
+                    if cf logs healthcheck --recent | grep -q "availability-test=${PIPELINE_TRIGGER_VERSION}"; then
+                      echo "Request detected"
+                      break
```
mtekel: how about exit 0 here and exit 1 after done? This way pipeline can proceed only if it is monitored...

tests/src/availability/healthcheck_availability_during_deployment_test.go
combor: Oh, typos everywhere...

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: Yes, true. I'll remove this.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1440,14 +1479,14 @@ jobs:
     plan:
       - aggregate:
           - get: pipeline-trigger
-            passed: [ 'cf-deploy']
+            passed: ['post-deploy','availability-tests']
```
mtekel: I guess post-deploy should depend on `avail tests` and then if this depends on post deploy, then it doesn't need to depend on `avail tests` again...

combor: No, if `post-deploy` depends on `avail tests` than during first run `avail-test` will fail and `post deploy` wont be run so health-check app will not be deployed.

mtekel: But then if post-deploy runs in parallel with avail tests, it will take the healthcheck app briefly down and thus interfere with avail. tests...

combor: It's doesn't run in parallel with tests. post-deploy has `passed: ['cf-deploy']` dependency and `avail-tests` are stopped when `cf-deploy` has finished.

mtekel: This is not 100% guaranteed and there is small race condition/window (depending on how often _eventually_ checks the function results). Nevertheless, why would we _not_ want to run smoke/acceptance/other tests if the availability tests have failed? I wouldn't change the previous behaviour and still run them (simply after post-deploy, without depending on `avail tests`).

combor: Yes, there is small race condition here. The check frequency is every 5 sec. If `post-deploy` can deploy healthcheck app within 5 seconds than avail-test will fail. 

 > why would we not want to run smoke/acceptance/other 

Because we already have one test failure so it doesn't make sense to run others tests?

Why would we like to run more tests if we have failure already? `tag-release` shouldn't be run anyway.  We should fix the issue and re-run entire pipeline. If you want to run `acceptance-tests` after `availability-tests` failure run them manually.

mtekel: I want to run the tests to help me tell _what_ the failure was. I see no harm in running them... It is quite possible that even though there _was_ some downtime during the deploy, now the environment is perfectly healthy again - which is a thing good to know.

combor: `avail-test` will tell you if app is reachable again. It's not failing immediately but measuring error rate and providing statistics at the end. I see no point in continuing running the pipeline after failure. If you want to run other tests, run them manually.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -477,21 +486,116 @@ jobs:
...
+            BRANCH: {{branch_name}}
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                ./paas-cf/concourse/scripts/import_bosh_ca.sh
+
+                export CONCOURSE_ATC_PASSWORD=$(awk '/basic_auth_password/ { print $2 }' concourse-manifest/concourse-manifest.yml)
```
dcarley: Can you use `var_from_yaml.rb` here instead of `awk`?

It'll be safer and avoid the problems described in 9b732a66980010a8e4ee468d09d93af6514259e6 and cdcab37f2fe21d4b6e5229fda89de321c39c9b04.

tests/src/availability/concourse_test.go
```diff
@@ -0,0 +1,122 @@
+package availability_test
```
dcarley: I was a bit confused by this file at first because I:
- half expected a `_test.go` file to contain another suite of tests rather than helpers (based on our other tests, rather than knowledge of Go)
- wasn't sure which of the methods were local to this file or use by the other test file

So I think it'd be clearer to make this a separate package and export the one function (I think it's only one?) as `DeploymentHasFinishedInConcourse`

tests/src/availability/concourse_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Expect(resourceVersionID).NotTo(Equal(0), "Resource: %s with version: %s should exist in Concourse", resourceName, resourceVersion)
+
+	builds, _, err := client.BuildsWithVersionAsInput(pipelineName, resourceName, resourceVersionID)
+
+	return filterBuildsByNameAndSortByTime(builds, jobName)
+}
+
+func deploymentHasFinishedInConcourse() bool {
+	client := newConcourseClient(getConcourseAtcUrl(), getConcourseUserName(), getConcoursePassword())
+	builds := buildsWithVersion(client, pipelineName, resourceName, getResourceVersion())
```
dcarley: Unlike the `get*()` methods on the line above, `getResourceVersion()` is defined in the other file.

Could we pass this as an argument to `deploymentHasFinishedInConcourse()` instead?

tests/src/availability/healthcheck_availability_during_deployment_test.go
```diff
@@ -0,0 +1,114 @@
...
+					return true
+				}
+				if errorRateThreshold(metrics, 15*time.Second, 0.5) {
+					return true
+				}
+				return false
+			}
+		})
+
+		It("does not get reuest success rate less than 99.9%", func() {
```
dcarley: Can the test description use `availabilitySuccessRateThreshold` in case we change one value and forget to the update the other?

tests/src/availability/healthcheck_availability_during_deployment_test.go
```diff
@@ -0,0 +1,114 @@
...
+	Context("when runs (until the deployment is finished or error rate > 50%)", func() {
+		var attacker *vegeta.Attacker
+		var resultChannel <-chan *vegeta.Result
+
+		BeforeEach(func() {
+			stopAttackCriteria = func() bool {
+				if deploymentHasFinishedInConcourse() {
+					return true
+				}
+				if errorRateThreshold(metrics, 15*time.Second, 0.5) {
```
dcarley: Could the `time.Duration` be a named constant?

At first I thought it was a rolling window that we were checking error rates within.

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: `governmentpaas/cf-acceptance-tests` doesn't have ruby. I can do the same thing as in https://github.com/alphagov/paas-cf/commit/cdcab37f2fe21d4b6e5229fda89de321c39c9b04. Is it ok ?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -477,21 +486,116 @@ jobs:
...
+            run:
+              path: sh
+              args:
+                - -e
+                - -u
+                - -c
+                - |
+                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
+
+                  CF_PASS=$(awk '/uaa_admin_password/ {print $2}' cf-secrets/cf-secrets.yml)
```
dcarley: Like https://github.com/alphagov/paas-cf/pull/321#discussion_r68070427 can we use `val_from_yaml.rb` here where we're using a container that has Ruby available?

(sorry, missed this yesterday)

dcarley:
There are some code comments to address or talk about. Plus some follow-up stories that we talked about and I'll writeup in Pivotal. Otherwise this looks good to go. I ran it several times yesterday while upgrading the stemcell, downgrading the stemcell and upgrading CF from #324 - each time the failure rate (about 3 reqs) was under the threshold. I know the implementation has been slightly unsatisfying, but this is really valuable functionality, so thank you.

concourse/pipelines/create-bosh-cloudfoundry.yml
dcarley: That's a shame. I think it'll be alright without the `| tr` thing though, because we're consuming the environment variable in Go, which is probably safer than passing it as a shell argument in that other commit.

tests/src/availability/concourse_test.go
combor: I moved it to separate helpers package now and re-arranged which functions it contain. I've moved some from init_test.go as well.

tests/src/availability/concourse_test.go
combor: Now `getResourceVersion()` is in helpers package and is exported.

tests/src/availability/healthcheck_availability_during_deployment_test.go
combor: Done.

tests/src/availability/healthcheck_availability_during_deployment_test.go
combor: Done.

concourse/pipelines/create-bosh-cloudfoundry.yml
combor: Now it's using `tr` to ensure that there are no quotes inside the string.

dcarley:
Let's go!

PR #322, opened 2016-06-17
[#120376951] Workaround UAA race condition for user membership

mtekel:
## What
Use uaac curl to work around uaa group add concurrency issue (see https://github.com/cloudfoundry/uaa/issues/372).

Use the suggested atomic endpoint instead of using `uaac` member add functionality.

## How to review

Deploy. Run pipeline several times and observe that starting all tests concurrently doesn't cause concurrent user addition to fail.

## Who can review
not @mtekel

### DIFF
--- a/concourse/tasks/create_admin.yml
+++ b/concourse/tasks/create_admin.yml
@@ -18,7 +18,7 @@ run:
       ./paas-cf/concourse/scripts/import_bosh_ca.sh
 
       SUFFIX=$(tr -cd '[:alpha:]0-9' < /dev/urandom | head -c10)
-      PASSWORD=$(tr -cd '[:alpha:]0-9' < /dev/urandom | head -c32 )
+      PASSWORD=$(tr -cd '[:alpha:]0-9' < /dev/urandom | head -c32)
       NAME=${PREFIX}-${SUFFIX}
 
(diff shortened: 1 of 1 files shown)


concourse/tasks/create_admin.yml
```diff
@@ -29,11 +29,22 @@ run:
...
-      uaac member add uaa.admin "${NAME}"
-      uaac member add scim.read "${NAME}"
-      uaac member add scim.write "${NAME}"
-      uaac member add doppler.firehose "${NAME}"
+      set +e
+      add_member(){
+        GID=$(uaac group get $1 -a id | awk '{print $2}')
+        uaac curl -k "/Groups/${GID}/members" -XPOST -H'Content-Type: application/json' \
+                  -d'{"origin":"uaa","type":"USER","value":"'${2}'"}' | tee result | grep -q '201 Created'
+        [ $? != 0 ] && cat result && exit 1
```
saliceti: Could you try simplifying by merging the 2 lines with `grep ... || cat ...` ?
So you could remove `set +e`

saliceti: I've tested with @mtekel and it's not that simple.
Let's leave like this.

saliceti:
Tested successfully by running 2 loops of this scripts 2 concourse containers.
In 15 min there was no error.
With the previous script, we would get errors before 30s.

PR #323, opened 2016-06-17
[#121133283] Fix RDS postgres broker user permission handling

alext:
## What

This fixes issues with how permissions were being setup for Postgres instances.

Due to how Postgres' permissions model works, we need to re-use the same user for all binds that need access to make schema-level changes (at present, that's all binds). To achieve this, we now need to persist the password to a database on the RDS instance so that it can be returned from bind calls. This therefore adds an encryption key param so that these passwords can be encrypted in the DB.

## How to review

Deploy from this branch, and verify that the new custom acceptance tests pass.

## Before merging this:

The following has to be done before this can be merged:

* [x] Merge the [paas-rds-broker PR](https://github.com/alphagov/paas-rds-broker/pull/12)
* [x] Update the submodule in the [paas-aws-broker-boshrelease PR](https://github.com/alphagov/paas-aws-broker-boshrelease/pull/9) to point to the updated master branch.
* [x] Merge the [paas-aws-broker-boshrelease PR](https://github.com/alphagov/paas-aws-broker-boshrelease/pull/9).
* [x] Tag paas-aws-broker-boshrelease with `0.0.5`
* [x] Remove the commit "WIP: deploy rds-broker from branch" from this PR.

## Who can review

Anyone but @HenryTK, @Jonty, @timmow or myself.


tests/example-apps/healthcheck/db_permissions.go
```diff
@@ -0,0 +1,94 @@
...
+	dbURL, err := url.Parse(os.Getenv("DATABASE_URL"))
+	if err != nil {
+		return err
+	}
+	db, err := sql.Open("postgres", dbURL.String())
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
```
mtekel: I wonder if we can DRY this DB connection above.

manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         aws_secret_access_key: ""
         aws_region: "eu-west-1"
         password: (( grab secrets.rds_broker_admin_password ))
+        rds_broker_state_encryption_key: (( grab secrets.rds_broker_state_encryption_key ))
```
alext: Doh, I'll fix it now...


Done.

mtekel:
Works. I have also tried the rails app & re-binding and it is fine. Will merge if there are no further comments...

PR #324, opened 2016-06-21
[#120591859] Upgrade cloudfoundry to 237

alext:
## What

This upgrades Cloudfoundry to release 237. It also upgrades the related components to the recommended versions for this release:

* diego to 0.1472.0
* etcd to 49

The diego release includes some changes to how the releases are structured:

* the rootfses have been split into a separate release
* the cc-bridge jobs have been extracted into a separate release `capi-release` which is now included as part of the CF release.

See the corresponding commits for more details.

## How to review

This should be tested by deploying both a clean environment, and an upgrade of an existing environment.

## Who can review

Anyone but myself or @Jonty.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -107,8 +107,8 @@ resources:
   - name: graphite-nozzle
     type: git
     source:
-      uri: https://github.com/alphagov/graphite-nozzle
-      branch: 106191208-deaggregated_names
+      uri: https://github.com/alphagov/paas-graphite-nozzle
+      branch: upgrade_go
 
   - name: vpc-tfstate
     type: s3-iam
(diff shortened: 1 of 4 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -107,8 +107,8 @@ resources:
   - name: graphite-nozzle
     type: git
     source:
-      uri: https://github.com/alphagov/graphite-nozzle
-      branch: 106191208-deaggregated_names
+      uri: https://github.com/alphagov/paas-graphite-nozzle
+      branch: upgrade_go
```
mtekel: shall we have some sort of "gds-master" that we merge against? Are we even interested to push changes upstream? What happened to this nozzle after CC got merged into pivotal? Is it officially supported? Or scrapped?

alext: Maybe, but whatever we do on this repo, we need to ensure that this manifest is pinned to a specific version. I don't think we want to be in a position where commits to that repo cause an upgrade here without us having a PR here first.

combor:
All acceptance tests passed when deploying on fresh env.

mtekel:
passing for me as well, merging

PR #327, opened 2016-06-23
[#122117381] Use gp2 storage for Bosh RDS instance.

alext:
## What

This has better performance than the default (magnetic) storage that
it's currently provisioned with.

This modification will not apply immediately; it will apply during the
configured maintenance window. During this time there will be a few
minutes of downtime. In my dev environment, bosh was unavailable
for 3 minutes while the RDS instance restarter. After that, all was well.

## How to review

* Deploy from this branch.
* Using the RDS console to change the maintenance window of your bosh RDS instance to happen in a few minutes time (note: the window is specified in UTC).
* Watch while the change is applied during the maintenance window. Verify that everything is working correctly after the change has completely applied.

Run another deploy to set the maintenance window back to the correct value.

## Who can review

Anyone but myself.

### DIFF
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -45,7 +45,7 @@ resource "aws_db_instance" "bosh" {
   password = "${var.secrets_bosh_postgres_password}"
   db_subnet_group_name = "${aws_db_subnet_group.bosh_rds.name}"
   parameter_group_name = "${aws_db_parameter_group.default.id}"
-
+  maintenance_window = "Thu:03:00-Thu:04:00"
   multi_az = "${var.bosh_db_multi_az}"
   backup_retention_period = "${var.bosh_db_backup_retention_period}"
   final_snapshot_identifier = "${var.env}-bosh-rds-final-snapshot"
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -38,6 +38,7 @@ resource "aws_db_instance" "bosh" {
   identifier = "${var.env}-bosh"
   name = "bosh"
   allocated_storage = 5
+  storage_type = "gp2"
   engine = "postgres"
   engine_version = "9.4.5"
   instance_class = "db.t2.medium"


dcarley:
> note the current maintenance window (so you can set it back afterwards)

Should we manage this with Terraform to make it explicit and prevent any manual mistakes?

We do for the CF RDS:

- https://github.com/alphagov/paas-cf/blob/89b7f485dfde4cd769e42bab9cd6962ce869591b/terraform/cloudfoundry/cf_rds.tf#L56

alext:
> Should we manage this with Terraform to make it explicit and prevent any manual mistakes?

The manual thing in this case is purely to speed up reviewing, otherwise it could take up to a week for the change to apply. Setting it in terraform wouldn't help with this.

By default RDS selects a random 30 min window within a range that's appropriate for the region the instance is in (for eu-west-1 it's 22:00-06:00 UTC) - http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBMaintenance.html

We could pin it in terraform, but I'm not sure what that would gain us in this instance, as the default behaviour is reasonable.

dcarley:
> The manual thing in this case is purely to speed up reviewing, otherwise it could take up to a week for the change to apply. Setting it in terraform wouldn't help with this.

To clarify - I mean the need to note the original time, modify it using the console, and reset it again afterwards. If it was managed in code then it could be tested with a temporary commit on a wip branch and then removed afterwards.

alext:
I've updated this, and the instructions for reviewing.

dcarley:
Sorry, taking longer to test than I expected. I had problems with the instructions because changing the maint window also seemed to undo the pending storage change. So I used a WIP commit on a branch to change the window instead.

When I applied it took 16mins and rebooted twice because it also did "Applying modification to database instance class". Is that expected?

alext:
> When I applied it took 16mins and rebooted twice because it also did "Applying modification to database instance class". Is that expected?

I think so. I think my instance did 2 reboots during the window - one to change the instance class (which is what actually changes the storage type), and one to update the storage size. I'm not sure why it needed to update the size given that hadn't changed, but I assume that's necessary with the change of storage type.

dcarley:
I saw the same. I don't think the extended downtime will cause any issue at that time of night. Merging..

PR #329, opened 2016-06-23
[#120654489] Start from specific job

mtekel:
## What

[Run specific Pipeline Jobs / Tasks](https://www.pivotaltracker.com/n/projects/1275640/stories/120654489)

Add a makefile target to run pipeline from specific job, ensuring that job is able to pick up latest changes in paas-cf. We do this by adjusting the pipeline by removing 'passed' for paas-cf in the given job.

## How to review

Run `JOB=myjob make dev run`. Observe that job having it's paas-cf "unbound" from previous jobs and being triggered.

## Who can review

not @mtekel or @combor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,3 +150,8 @@ merge_pr: ## Merge a PR. Must specify number in a PR=<number> form.
 find_diverged_forks: ## Check all github forks belonging to paas to see if they've diverged upstream
 	$(if ${GITHUB_TOKEN},,$(error Must pass GITHUB_TOKEN=<personal github token>))
 	./scripts/find_diverged_forks.py alphagov --prefix=paas --extra-repo=cf-release --extra-repo=graphite-nozzle --github-token=${GITHUB_TOKEN}
+
+.PHONY: run_job
+run_job: check-env-vars ##  Unbind paas-cf of $JOB in create-bosh-cloudfoundry pipeline and then trigger it
+	$(if ${JOB},,$(error Must pass JOB=<name>))
+	./concourse/scripts/run_job.sh ${JOB}
--- a/concourse/scripts/run_job.sh
+++ b/concourse/scripts/run_job.sh
@@ -1,0 +1,21 @@
+#!/bin/bash -eu
+set -o pipefail
+
+JOB=$1
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+PIPE='create-bosh-cloudfoundry'
+TEMP="/tmp/concourse-run-job.${RANDOM}"
+trap 'rm -f $TEMP' ERR
+
+export TARGET_CONCOURSE=deployer
+# shellcheck disable=SC2091
+$("${SCRIPT_DIR}/environment.sh")
+"${SCRIPT_DIR}/fly_sync_and_login.sh"
+FLY="$FLY_CMD -t ${FLY_TARGET}"
+
+# 2 steps prevent fly reading empty input in case step 1 fails
+${FLY} get-pipeline -p ${PIPE} | "${SCRIPT_DIR}"/unbind_job.rb "${JOB}" >$TEMP
+${FLY} set-pipeline -p ${PIPE} -c=$TEMP -n
+
+curl -k -u "${CONCOURSE_ATC_USER}:${CONCOURSE_ATC_PASSWORD}" "${CONCOURSE_URL}/api/v1/pipelines/${PIPE}/jobs/${JOB}/builds" -d ''
+rm -f $TEMP
--- a/concourse/scripts/unbind_job.rb
+++ b/concourse/scripts/unbind_job.rb
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+require 'yaml'
+pipe = YAML.load(STDIN)
+
+def remove_passed(obj)
+  case obj
+    when Array
+      obj.each {|v| remove_passed(v)}
+    when Hash
+      if obj.has_key?('get') and obj['get']=='paas-cf' then
+        obj.delete('passed')
+      else
+        obj.each {|k,v| remove_passed(v) if k == 'do' or k == 'aggregate'}
+      end
+  end
+end
+
+abort "Unable to parse YAML hash from the input" if pipe.class != Hash
+abort "Can't find job definitions in the input"  if pipe['jobs'].nil?
+abort "Jobs definition not an array"             if pipe['jobs'].class != Array
+
+myJob=pipe['jobs'].find{|j| j['name']==ARGV[0]}
+abort "Job " + ARGV[0] + " not found in the pipeline" if myJob.nil?
+
+remove_passed(myJob['plan'])
+puts YAML.dump(pipe)
--- a/README.md
+++ b/README.md
@@ -381,4 +381,4 @@ To do this, run `aws ses verify-email-identity --email-address youremail@example
 and click on the link in the resulting email
 
 ## BOSH failover
-Visit [BOSH failover page](https://github.com/alphagov/paas-cf/doc/bosh_failover.md)
+Visit [BOSH failover page](https://github.com/alphagov/paas-cf/blob/master/doc/bosh_failover.md)
--- a/README.md
+++ b/README.md
@@ -217,6 +217,21 @@ can use SELF_UPDATE_PIPELINE environment variable, set to false (true is default
 
 See [doc/non_dev_deployments.md](doc/non_dev_deployments.md).
 
+## Optionally run specific job in the create-bosh-cloudfoundry pipeline
+
+`create-bosh-cloudfoundry` is our main pipeline. When we are making changes or
+adding new features to our deployment we many times wish to test only the
+specific changes we have just made. To do that, it's many times enough to run
+only the job that is applying the change. In order to do that, you can use
+`run_job` makefile target. Specify the job name you want to execute by setting
+the `JOB` variable. You also have to specify your environment type, e.g.
+`JOB=performance-tests make dev run_job`.
+
+This will not only tigger the job, but before that it will modify the pipeline
+to remove `passed` dependencies for `paas-cf` in the specified job. This means
+that your job will pick the latest changes to `paas-cf` directly, without the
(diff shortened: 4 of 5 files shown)


Makefile
```diff
@@ -150,3 +150,8 @@ merge_pr: ## Merge a PR. Must specify number in a PR=<number> form.
 find_diverged_forks: ## Check all github forks belonging to paas to see if they've diverged upstream
 	$(if ${GITHUB_TOKEN},,$(error Must pass GITHUB_TOKEN=<personal github token>))
 	./scripts/find_diverged_forks.py alphagov --prefix=paas --extra-repo=cf-release --extra-repo=graphite-nozzle --github-token=${GITHUB_TOKEN}
+
+.PHONY: run
+run: check-env-vars ## Unbind paas-cf of and run specified $JOB in create-bosh-cloudfoundry pipeline
```
dcarley: Can we think of a more specific name for this?

concourse/scripts/run_job.sh
```diff
@@ -0,0 +1,22 @@
...
+TEMP="/tmp/concourse-run-job.${RANDOM}"
+trap "rm -f $TEMP" ERR
+
+export TARGET_CONCOURSE=deployer
+# shellcheck disable=SC2091
+$("${SCRIPT_DIR}/environment.sh")
+"${SCRIPT_DIR}/fly_sync_and_login.sh"
+FLY="$FLY_CMD -t ${FLY_TARGET}"
+
+# 2 steps prevent fly reading empty input in case step 1 fails
```
dcarley: Could you describe the kind of circumstances the first step would fail? Are they returning the correct exit codes? Does `pipefail` not prevent it from proceeding?

It would be preferable not to use a temporary file, if possible.

concourse/scripts/run_job.sh
```diff
@@ -0,0 +1,22 @@
...
+${FLY} get-pipeline -p ${PIPE} | "${SCRIPT_DIR}"/unbind_job.rb "${JOB}" >$TEMP
+cat $TEMP | ${FLY} set-pipeline -p ${PIPE} -c=/dev/stdin -n
+
+# shellcheck disable=SC2086
+curl -k https://${CONCOURSE_ATC_USER}:${CONCOURSE_ATC_PASSWORD}@${CONCOURSE_URL:8}/api/v1/pipelines/${PIPE}/jobs/${JOB}/builds -d ''
```
dcarley: Could you split the authentication out to a separate argument so that we don't need to do string manipulation:
```sh
curl -k -u "${CONCOURSE_ATC_USER}:${CONCOURSE_ATC_PASSWORD}" "${CONCOURSE_URL}/api/v1/pipelines/${PIPE}/jobs/${JOB}/builds" -d ''
```

concourse/scripts/run_job.sh
```diff
@@ -0,0 +1,22 @@
...
+# shellcheck disable=SC2091
+$("${SCRIPT_DIR}/environment.sh")
+"${SCRIPT_DIR}/fly_sync_and_login.sh"
+FLY="$FLY_CMD -t ${FLY_TARGET}"
+
+# 2 steps prevent fly reading empty input in case step 1 fails
+${FLY} get-pipeline -p ${PIPE} | "${SCRIPT_DIR}"/unbind_job.rb "${JOB}" >$TEMP
+cat $TEMP | ${FLY} set-pipeline -p ${PIPE} -c=/dev/stdin -n
+
+# shellcheck disable=SC2086
```
dcarley: If we quote the arguments like the comment below then we don't need to disable shellcheck.

concourse/scripts/unbind_job.rb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env ruby
+require 'yaml'
+pipe = YAML.load(STDIN)
+
+def clean_passed(obj)
```
dcarley: Can we call this `remove_passed` or `delete_passed` to give a better indication of what it's doing?

concourse/scripts/unbind_job.rb
```diff
@@ -0,0 +1,26 @@
...
+  case obj
+    when Array
+      obj.each {|v| clean_passed(v)}
+    when Hash
+      if obj.has_key?('get') and obj['get']=='paas-cf' then
+        obj.delete('passed')
+      else
+        obj.each {|k,v| if k == 'do' or k == 'aggregate' then clean_passed(v) end}
```
dcarley: The `then`/`end` can be omitted to make this more readable:
```ruby
clean_passed(v) if k == 'do' or k == 'aggregate'
```

dcarley:
Works as described and separation of files looks good 👍 

Could you explain in the commit messages why this functionality is useful to us? I think it's especially important to capture as we had trouble defining it in kick-off. Should we also add it to the README?

We need tests for `unbind_job.rb`. There isn't much code, but it uses recursion and a specific YAML structure, which is going to make it hard for somebody else to work on it with confidence that they haven't broken existing functionality.

The tests can use a basic pipeline, with other `get` steps, and check that `passed` is removed from `paas-cf`. We probably need to assert on the entire output rather than specific keys in order to make sure that we get the correct structure out. The following code paths should be covered with separate `context`/`it` blocks to describe the specific case:

- non-zero exit code and STDERR message when STDIN is empty and closed
- non-zero exit code and STDERR message when STDIN is not YAML
- non-zero exit code and STDERR message when job name cannot be found in STDIN
- a plan that uses `get` in `aggregate`
- a plan that uses `get` in `do`
- a plan that uses `get` outside of `aggregate` or `do`

dcarley:
PS: I'm not clear why the `tag-release` tests failed. The right version of `git` got installed in https://travis-ci.org/alphagov/paas-cf/builds/139802136 but not in https://travis-ci.org/alphagov/paas-cf/builds/139807032

Makefile
mtekel: I thought about "run-job", but then, having to specify 'JOB=' makes it obvious that you want to run a job. I am happy for suggestions what to rename this into...

alext: I think `run_job` is fine. This follows the same pattern as the `merge_pr` target. I don't think the repetition with `JOB=` is a problem.

alext: As an aside: We've not been very good at being consistent on `-` vs `_` in existing jobs. Most seem to use `_`, but a few use `-`.

concourse/scripts/run_job.sh
mtekel: There are several issues:
1. I initially did pipefail and <get manifest> | remove_passed | upload manifest. The problem is that this starts all 3 processes in parallel. If 1st process fails (e.g. due to network connectivity). Or 2nd process breaks (because of e.g. bug in code or broken yaml on input), the last process will receive empty input - and will upload it. Effectively deleting the pipeline.
2. Then I thought about using named pipes instead of a temp file. However named pipes are blocking. Which means I'd have to do (a) <get manifest> | remove_passed >pipe & (meaning, let it run in the background). And then (b) <upload manifest> <pipe. Which would either get stuck forever if there is no output from (a), or read empty/broken input in case (a) fails (this setup thus makes set -e/pipefail ineffective for capturing/preventing errors).
3. So in the end I chose the temp file (even though I don't like it so much). Because I can do 1st step where I render the mainfest, but if that fails, set -e/pipefail will prevent 2nd step. I have also thought that the pipeline could contain sensitive data (e.g. passwords), but we render pipeline on the disk anyway before uploading, so that data is on disk anyaway.

Nevertheless, I am happy for suggestions how to solve this in a better way. I don't want to trigger specific pipeline rendering scripts again (& then remove passed & upload), as:
* that would make the solution very pipeline specific
* would overwrite any changes user did by hand to the pipeline

concourse/scripts/run_job.sh
mtekel: The problem is then that if you quote "user":"password" - you get another shellcheck error that says that the quotes around : don't do nothing. I'll change this as I didn't realize you can pass username/pass separately to curl.

concourse/scripts/run_job.sh
dcarley: Thanks for explaining. I'll have a play and see if I can suggest anything. If we can't find another solution then it would be good to capture that rationale in the commit message, in case somebody attempts to refactor it in the future.

Makefile
dcarley: Agree, the repetition is a little weird, but it will make the output of `make help` clearer. Lets rename to `run_job`.

concourse/scripts/run_job.sh
dcarley: I think the whole argument should be quoted. The command in the other comment passed with shellcheck 0.3.7

concourse/scripts/run_job.sh
dcarley: Yep, right you are. I hadn't realised that about chained pipes.

I'm not sure how important it is that we don't use the rendering scripts to prevent overwriting manual changes. Don't people normally make those changes in their local repo and upload them with `make dev pipelines` anyway?

I realise that it's a large change now, but I suspect it would also simplify the implementation if `pipelines-bosh-cloudfoundry.sh` took an additional environment variable that piped the manifest through `unbind_job.rb`. What do you think?

mtekel: Let me think about some more general mechanism. All pipeline create scripts in the end use `deploy-pipeline.sh` to upload it. If we do the `passed` removal there, it would work for any pipeline. Triggering the job can also happen here.

mtekel:
Added tests, updated commit messages.

concourse/scripts/run_job.sh
dcarley: We concluded that this wasn't possible because it's not possible to parse the pipeline with the `{{ }}` variable placeholders present.

dcarley:
Thanks!

I'm un-assigning my because I'm not going to have a chance to re-review this today and I'm on leave Monday.

keymon:
Tested and LGTM. Tests cover all cases and all comments were resolved.

Minor question: Shall we fail if the `$JOB` does not exist in the pipeline? E.g. the user makes a typo in the name.

keymon:
LGTM, merging

PR #330, opened 2016-06-24
[#122187235] Fix delete-deployment when testers org doesn't exist.

alext:
## What

At present the remove-healthcheck-db task in the destroy-cloudfoundry
and autodelete-cloudfoundry pipelines will error if the CF api is
available, but the testers org doesn't exist. This can happen if the
cf-deployment succeeds, but one of the post-deploy tasks fails for some
reason.

This updates the task to check for the exsitence of the testers org
before attempting to delete the healthcheck app and db. If the org isn't
present, the app and db can't exist, so it's safe to just skip the task.

## How to review

* Use the cf cli to delete the healthcheck app, db and then delete the whole testers org.
* Run the `destroy-cloudfoundry` pipeline (you may want to pause the terraform destroy job if you don't want the dstroy to go this far), and observe it fail.
* Update concourse to use this branch, and re-run. It should now succeed.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/tasks/remove-healthcheck-db.yml
+++ b/concourse/tasks/remove-healthcheck-db.yml
@@ -13,11 +13,17 @@ run:
     - |
       ./paas-cf/concourse/scripts/import_bosh_ca.sh
       . ./config/config.sh
-      if curl -I -f $API_ENDPOINT/info; then
-        echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
-        cf target -o testers -s healthcheck
-        cf delete healthcheck -f -r
-        if cf services | grep -q healthcheck-db; then
-          cf delete-service healthcheck-db -f
-        fi
+      if ! curl -I -f $API_ENDPOINT/info; then
+        echo "CF API unavailable. Skipping..."
+        exit 0
+      fi
+      echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+      if ! cf org testers > /dev/null; then
+        echo "Org 'testers' not found. Skipping..."
+        exit 0
+      fi
+      cf target -o testers -s healthcheck
+      cf delete healthcheck -f -r
+      if cf services | grep -q healthcheck-db; then
+        cf delete-service healthcheck-db -f
       fi


concourse/tasks/remove-healthcheck-db.yml
```diff
@@ -13,11 +13,17 @@ run:
...
-        cf delete healthcheck -f -r
-        if cf services | grep -q healthcheck-db; then
-          cf delete-service healthcheck-db -f
-        fi
+      if ! curl -I -f $API_ENDPOINT/info; then
+        echo "CF API unavailable. Skipping..."
+        exit 0
+      fi
+      echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+      if ! cf org testers > /dev/null; then
```
combor: I'm not sure if it's ok to assume that if org doesn't exist that db instance is not present. I know that right now we create it after org creation but if for any reason we will change it int the future we'll have to remember to update it in two places. Also creation of service is not dependant on existence of an org. Can we do sth like this instead:

alext: Services are scoped to the org, so `cf services` will only show services for the org you're currently targetting.

Also, given this, I don't see how it's possible for the healthcheck db to exist when the org doesn't.

combor: Ok, that's right. At the other hand I checked if we can just delete the org but it starts deleting the service instance and fails with "An operation for service instance test is in progress" and doesn't delete the app.  So I have just small improvement suggestion and we should be good to go.

Can we replace
```
if ! curl -I -f $API_ENDPOINT/info; then
```

with 

```
if echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS} ; then
```
and remove the login command below so we'll have api check and login in one line.

alext: That would work, but I think it's better as it is now for a couple of reasons:

* I had to think for a bit about how the pipe in the command would interact with the conditional, which means other reading this are likely to have to as well. I think it's clearer what's going on at the moment.

* It conflates login failure with the API being unavailable. We want to skip this if the API isn't there (because the deployment doesn't exist), but not if authentication fails for another reason (eg wrong password etc).

combor: Well, I see it differently. API might not be available for other reasons ( networking issue, broken deployment as this is dev ) and we should continue deleting the deployment as this is ultimate task of this pipeline. It's only there as we want to reduce costs by not keeping VMs during night and that's the current behaviour of
```
      if ! curl -I -f $API_ENDPOINT/info; then
        echo "CF API unavailable. Skipping..."
        exit 0
      fi
```
To continue if API isn't there.

Apart from that there is an issue with using `curl -f`.
```
       -f, --fail
              (HTTP)  Fail  silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In
              normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This
              flag will prevent curl from outputting that and return error 22.

              This  method  is  not  fail-safe  and  there  are occasions where non-successful response codes will slip through, especially when authentication is
              involved (response codes 401 and 407).
```

alext: I think this has wandered well outside the scope of this PR. It's focussed in allowing this job to continue when the org doesn't exist. At this point I think we should capture these issues to be addressed separately.

combor: Alex, you are proposing a change that touches that line `if ! curl -I -f $API_ENDPOINT/info; then` I don't see how it not is related.

alext: I haven't changed that check, I've just restructured the conditionals to avoid deep nesting.

This PR is about ensuring that the deployment can be deleted when the testers org doesn't exist. I don't think this is the time to dig into how we check for the existence of the API itself.

dcarley: I'm unsure about this too..

It's seem that part of the problem is that we use this task in two different contexts:

- `autodelete-cloudfoundry` which is unattended and it's acceptable for the task to fail silently because the damage is limited
- `destroy-cloudfoundry` which is attended and it's not acceptable for this to fail silently because the instance will be orphaned - I think it's right that this pipeline expects a full environment to be created

Would this also be addressed by [#120765225: disable healthcheck DB in dev](https://www.pivotaltracker.com/story/show/120765225) because we'd no longer need to include this task in `autodelete-cloudfoundry`?

alext: > I'm unsure about this too..

Which part?  As I see it there are 2 things being discussed here:

1. The additional check for the testers org.
2. Refactoring the existing check for the existence of the API.

I think 2 is out of scope for this PR, which was intend to be a small change to fix a problem that had been encountered.

dcarley: Well, both. I was getting a bit confused.

@combor's point about silent failure if we change the org name is correct. I think it's probably OK to make this change for the moment, but raises the importance of https://www.pivotaltracker.com/story/show/120002281

combor: Ok, so I'll merge it now.

mtekel:
re @dcarley 
>destroy-cloudfoundry which is attended and it's not acceptable for this to fail silently because the instance will be orphaned - I think it's right that this pipeline expects a full environment to be created

I'd say destroy-cloudfoundry job should be robust, to enable us to cleanly remove CF which is in various stages of broken, including failing deployment. It should then _not_ expect full environment to be created.

dcarley:
> I'd say destroy-cloudfoundry job should be robust, to enable us to cleanly remove CF which is in various stages of broken, including failing deployment. It should then not expect full environment to be created.

Not if it comes at the cost of complicated logic and silent failures. It could be better to re-run the create pipeline to resolve the problems. There are plenty of nuances though, so it might be better to discuss in-person.

PR #331, opened 2016-06-24
[#122188819] Fix capture of bosh-init state on error.

alext:
## What

Because the updated state file was only copied into the output directory
after the bosh-init run, if bosh-init errored for any reason, this would
not happen, and the ensure block had no file available to upload.

This refactors the task to use the output dir as the working dir for
bosh-init meaning that the updated state file is always available to the
ensure block.

## How to review

Given an environment with an existing bosh VM:

* Rename the existing bosh-manifest-state file in your S3 bucket.
* Run the pipeline from this branch.
* Observe bosh-init error when trying to create a new bosh vm due to an IP conflict.
* Verify that the updated state file is uploaded to S3
* Restore the original state file.
* Run the pipeline again, and verify that it succeeds.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -769,7 +769,7 @@ jobs:
             - name: bosh-init-state
             - name: ssh-private-key
           outputs:
-            - name: updated-bosh-init-state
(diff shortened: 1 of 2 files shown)


HenryTK:
All worked perfectly

PR #332, opened 2016-06-24
[#121843167] Ensure Collectd is configured on all VMs

HenryTK:
## What

https://www.pivotaltracker.com/projects/1275640/stories/121843167

A Collectd release has been forked from the CloudFoundry Community respositories and applied as an addon in Bosh, making it available on all VMs managed by Bosh. Collectd is configured to send metrics directly to Graphite. Logsearch-shipper has been removed as most of the logs were duplicates of what is already sent by Metron. See commit message for full details.

## How to review

* Deploy using the branch `collectd_121843167`.
* Check Collectd data is sent to Graphite (get Graphite URL using `make dev showenv`).

## Who can review

Anyone but @Jonty or me

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -106,6 +106,12 @@ resources:
       uri: https://github.com/alphagov/paas-aws-broker-boshrelease.git
       tag_filter: {{cf_aws_broker_version}}
 
+  - name: collectd-boshrelease
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-collectd-boshrelease.git
+      tag_filter: {{cf_collectd_version}}
+
   - name: os-conf-boshrelease
     type: git
     source:
@@ -1129,6 +1135,7 @@ jobs:
           - get: grafana-boshrelease
           - get: paas-haproxy-release
           - get: aws-broker-boshrelease
+          - get: collectd-boshrelease
           - get: os-conf-boshrelease
           - get: cf-secrets
             passed: ['generate-cf-config']
@@ -1147,6 +1154,7 @@ jobs:
               - name: grafana-boshrelease
               - name: paas-haproxy-release
               - name: aws-broker-boshrelease
+              - name: collectd-boshrelease
               - name: os-conf-boshrelease
             run:
               path: sh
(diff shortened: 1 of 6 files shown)


combor:
> Check Collectd data is sent to Graphite (get Graphite URL using make dev showenv).

What specifically should I check?

HenryTK:
@combor: 

> What specifically should I check?

Sorry, should have been more specific. The data in question is defined in [the runtime config](https://github.com/alphagov/paas-cf/blob/collectd_121843167/manifests/cf-manifest/manifest/runtime/runtime.yml#L19-L59). Under `collectd` on the left hand side of the Graphite UI there should be an entry for each VM, and sub folders for all of the configurations in the runtime config.

combor:
Is it possible to change the frequency of how data is being send to graphite? As I see from graphs it's every minute ( I couldn't find such setting in the code ) which might be not enough for us. And the graphs look not entirely readable because of that. 

image

dcarley:
I peeked at Graphite in @combor's environment..

The nodes are keyed by BOSH job name and index which is 👍 

And snap ^ we should fix the polling/storage frequencies because we won't be able to rely on the data if they're incorrect.

mtekel:
Default interval should be 10 seconds. I don't see it changed in the release. It could be possible that something on the way (e.g. statsD?) is doing something to metrics, like merging multiple values into 1 minute averages and then submitting them to graphite. It is also possible that metric resolution in graphite is set to 1 minute.

Regarding graph "niceness" - just change it to connected line and you'll get a nice graph...

dcarley:
The BOSH job has a default `Interval` of `60s`:

- https://github.com/alphagov/paas-collectd-boshrelease/blob/258bfb5563eca15a58603364bb1954d4514b276d/jobs/collectd/spec#L15-L17

We should use collectd's default of `10s` because it will give us better visibility. We may need to check that Carbon is configured to store at that max resolution.

We shouldn't workaround it in the UI because it's an indication of a bigger problem with the accuracy of our metrics.

mtekel:
> I don't see it changed in the release
I have used git search for 'interval', which still doesn't find anything...

HenryTK:
I have overridden the default interval. It is now set to 10 seconds.

combor:
I'm not sure why but some metrics are not being send to Graphite. For instance in my deployment it's TX packets for eth0 on api_worker_0.

image

combor:
Ok, so I got some TX graphs after a while. Ignore the previous comment.

combor:
Graphs are now much more readable after changing data interval and it looks good to me. Merging.

PR #333, opened 2016-06-24
Update resource pinning to simplify pipeline.

alext:
## What

This reduces the number of loops/bypasses in the pipeline and makes it
easier to see what's flowing where. It generally makes the input pinning
more consistent for several jobs - they're all pinned to the same
previous job if possible.

* Pin cf-terraform inputs to pre-deploy
* Pin post-deploy inputs to cf-deploy
* Pin cf-secrets and cf-manifest for smoke/acceptance tests to
  post-deploy.
* Pin cf-secrets in performange-tests to smoke/acceptance tests

### Early part of pipeline:

Before:
image
After:
image

### Later part of pipeline:

Before:
image

After:
image

## How to review

Check changes look sensible, and the pipeline dependencies are still correct.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -810,11 +810,11 @@ jobs:
           - get: cf-tfstate
           - get: cf-certs-tfstate
           - get: bosh-CA
-            passed: ['generate-secrets']
+            passed: ['pre-deploy']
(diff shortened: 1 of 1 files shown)


dcarley:
I haven't looked at the diff in detail, but this looks sensible and extra karma for the before/after images 👍

mtekel:
Rebased on master. All works still well.

PR #334, opened 2016-06-27
[#122117381] Set Bosh RDS backup window.

alext:
## What

When this is not specified, RDS will pick a random window that's
appropriate for the region. There's therefore a chance that this will
conflict with the maintenance window that we're specifying here.

Specifying the backup window avoids the possibility of these
conflicting.

## How to review

Deploy from this branch, and verify that it applies cleanly.

## Who can review

Anyone but myself.


keymon:
The docs say:

>  Note that if the backup requires more time than allotted to the backup window, the backup will continue to completion. 

Although I think terraform will deploy OK, I wonder if there is any potential issue of having the maintenance window just after backup. I cannot find any reference about that, only that they cannot overlap.

alext:
That's a fair point.  I'll move it an hour earlier to be sure.

keymon:
LGTM

PR #341, opened 2016-06-29
[#121933113] Fix gorouter issue for zero downtime deploy

combor:
## What
Story: https://www.pivotaltracker.com/story/show/121933113

It's a part of zero downtime deployment epic. 

This PR implements custom healthcheck for haproxy to enable connection draining to make sure we don't server any errors during cf-deploy.

Also we split routing-release from main cf-release and use our fork to deliver gorouter fix for early exit when more than one signal is received. More details about that fix in routing-release PR: https://github.com/alphagov/paas-routing-release/pull/1

As a result of above listed changes we increase the availability test success rate threshold to 100% as we expect true zero downtime deployment now  

## How to review

- [x] Review: https://github.com/alphagov/paas-haproxy-release/pull/3
- [x] Review: https://github.com/alphagov/paas-routing-release/pull/1
- [x] Run pipeline from `121933113_routers` branch, it will force full router update as it contains  https://github.com/alphagov/paas-cf/commit/9292aae6d40b45aa9e2703b5eb297264ab8904fd. Later you will remove it. Check if availability test passes.
- [x] Merge: https://github.com/alphagov/paas-haproxy-release/pull/3
- [x] Remove https://github.com/alphagov/paas-cf/blob/2937b4c8d0ff8612256503607c4393ac1f2de145/concourse/pipelines/create-bosh-cloudfoundry.yml#L89, put a new tag on `paas-haproxy` and update https://github.com/alphagov/paas-cf/blob/2937b4c8d0ff8612256503607c4393ac1f2de145/manifests/cf-manifest/manifest/000-base-cf-deployment.yml#L29
- [x] Remove https://github.com/alphagov/paas-cf/commit/2c2e606cd1c3641e47440edb7b258cc1a254a249 , put new tag on `routing-release` and update https://github.com/alphagov/paas-cf/blob/758e0b5eef87228e640f6a33fb79c69d1bb251f1/manifests/cf-manifest/manifest/000-base-cf-deployment.yml#L31
- [x] Remove https://github.com/alphagov/paas-cf/commit/9292aae6d40b45aa9e2703b5eb297264ab8904fd
- [x] Merge this PR

## Who can review

Not @saliceti , @timmow or @combor

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -148,6 +148,7 @@ properties:
     extra_headers_to_log:
     debug_addr:
     enable_routing_api:
+    drain_wait: 15
 
   cc:
     jobs:
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -131,6 +131,7 @@ properties:
     additional_frontend_config: |
       capture response header strict-transport-security len 128
       http-response add-header Strict-Transport-Security max-age=31536000 unless { capture.res.hdr(0) -m found }
+    enable_healthcheck_frontend: true
 
   router:
     enable_ssl: false
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -119,6 +119,13 @@ resources:
       tag_filter: {{cf_os_conf_version}}
       branch: gds_master
 
+  - name: routing-release
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-routing-release.git
+      tag_filter: {{cf_routing_version}}
+      branch: cf_237
+
   - name: graphite-nozzle
     type: git
     source:
@@ -1141,6 +1148,7 @@ jobs:
             passed: ['generate-cf-config']
           - get: bosh-CA
           - get: graphite-nozzle
+          - get: routing-release
 
       - aggregate:
         - task: upload-releases
@@ -1156,6 +1164,7 @@ jobs:
               - name: aws-broker-boshrelease
               - name: collectd-boshrelease
               - name: os-conf-boshrelease
+              - name: routing-release
             run:
               path: sh
               args:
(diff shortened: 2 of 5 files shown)


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -131,6 +131,7 @@ properties:
     additional_frontend_config: |
       capture response header strict-transport-security len 128
       http-response add-header Strict-Transport-Security max-age=31536000 unless { capture.res.hdr(0) -m found }
+    user_agent_frontend: true
```
dcarley: > This will enable to do better healtcheck from the router ELB.

Can you explain in the commit message why this is better and what it's doing?

keymon: I commented something similar it in https://github.com/alphagov/paas-haproxy-release/pull/3#issuecomment-229394248

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -148,6 +148,7 @@ properties:
     extra_headers_to_log:
     debug_addr:
     enable_routing_api:
+    drain_wait: 15
```
keymon: Can you explain better in the commit message why this happens:

> Set gorouter drain_wait to 15s 
>
> With this setting the gorouter will keep serving to new connections for 15 seconds after getting the drain signal, but respond 500 error when the HTTPMonitor user agent is used. 
>
> As the ELBs  healthcheck runs every 5 seconds and fails after 2 failures, and it will send the custom header (currently by using a custom frontent in haproxy). 
>
> This way we ensure that we don't drop any incoming connection before the ELB marks the gorouter down.

terraform/cloudfoundry/router_elb.tf
```diff
@@ -11,7 +11,7 @@ resource "aws_elb" "cf_router" {
   ]
 
   health_check {
-    target = "TCP:81"
+    target = "HTTP:82/"
```
keymon: Again, explain that port 82 is different

keymon:
Hehe, this PR completely breaks the service while deploying. 

The reason is that in the PR the healthcheck of the ELB is updated to point to port 82 in the job `cf-terraform` with the new healthcheck for HAProxy, but that frontend is not deployed as `cf-deploy` did not run!

So we must split this in 2 different PR if we don't want downtime: one to add the new healthcheck frontend in haproxy, and other one to change the ELB healthcheck in terraform :)

combor:
> The reason is that in the PR the healthcheck of the ELB is updated to point to port 82 
I have removed https://github.com/alphagov/paas-cf/commit/2242315ce5ee5a6f992cbd2cd67d07283118156a from this PR and I'll create separate one to apply later.

combor:
I believe all comments have been addressed.

keymon:
I updated the branch as https://github.com/alphagov/paas-routing-release/pull/1 has been merged

keymon:
Everything good. I removed a trailing whitespace that it was getting on my nerves :)

PR #344, opened 2016-07-01
[#122183539] Upgrade Grafana to the latest tagged version.

combor:
## What

As we are going to create a set of dashboards we want them to be compatible with recent versions of Grafana and avoid upgrade pain in the future.

## How to review

Deploy from this branch. Go to Grafana - you can check what the URL is with `make dev showenv` and confirm that the version is 3.0.3. You can find it on the bottom of the login screen. 

## Who can review

Not @combor or @saliceti

### DIFF
--- a/manifests/cf-manifest/manifest/040-graphite.yml
+++ b/manifests/cf-manifest/manifest/040-graphite.yml
@@ -15,7 +15,7 @@ releases:
 - name: graphite
   version: commit-ccc206f3ba21fe5ba4f3e617bb7dca3e66a652ad
 - name: grafana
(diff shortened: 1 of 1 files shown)


dcarley:
> Deploy from this branch. Go to Grafana - you can check what the URL is with make dev showenv and confirm that the version is 3.0.3. You can find it on the bottom of the login screen.

Do graphs of Graphite metrics still work? 😉

keymon:
It should work. I will review this too, it is small :)

combor:
@dcarley It works:

image

keymon:
LGTM and works for me too.

keymon:
rebased

dcarley:
Thanks both 👍

PR #346, opened 2016-07-01
[#115404231] Remove collector

mtekel:
## What

[Remove collector](https://www.pivotaltracker.com/projects/1275640/stories/115404231)

In our current deployment, collector is not configured to send metrics anywhere (even though it might seem so from logs, but historian has no targets configured). Also remove etcd_metrics_server, as that was sending etcd metrics to NATS for them to be picked up by collector.

## How to review

Deploy to new/existing env. In existing env observe collector being deleted and etcd upgated. Check that there is no metrics job on etcd servers.

## Who can review

not @mtekel

### DIFF
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -103,14 +103,6 @@ meta:
   - name: haproxy
     release: paas-haproxy
 
-  stats_templates:
-  - name: consul_agent
-    release: (( grab meta.consul_templates.consul_agent.release ))
-  - name: collector
-    release: (( grab meta.release.name ))
-  - name: metron_agent
-    release: (( grab meta.release.name ))
-
   uaa_templates:
   - name: consul_agent
     release: (( grab meta.consul_templates.consul_agent.release ))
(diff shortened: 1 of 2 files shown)


dcarley:
> Deploy to new/existing env. In existing env observe collector being deleted and etcd upgated. Check that there is no metrics job on etcd servers.

Can you describe how the reviewer can verify that these components weren't configured/used?

I think it'd also be useful to reference this post on cf-dev which describes the intention to remove collector from cf-release and the linked spreadsheet which gives specifics about why it's no longer required:

- https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/message/MBAR3MQK6B6B23S7SEGOUIIAW6FDQNUB/

mtekel:
To check collector configuration:
* check the manifest, properties for `collector:` should be empty
* on collector server, check /var/vcap/jobs/collector/config/config.yml . Compare against https://github.com/cloudfoundry/cf-release/blob/v237/jobs/collector/templates/config.yml.erb
* you can check historian code in the deployed collector: var/vcap/packages/collector/lib/collector/historian.rb (you can also note that it supports cfMetrics backend, but which can't be configured as it's missing in spec and config.erb template)

I'll update the commit messages.

keymon:
I confirm that what @mtekel says is right, stats are not configured.

The code removes that vm and the job from etcd.

keymon:
LGTM, letting tests to run and I will merge.

keymon:
rebased with master

dcarley:
👍 👍

PR #347, opened 2016-07-04
[#124496741] Alert on temp. admin removal failure

mtekel:
## What

[Alert on temp. admin removal failure](https://www.pivotaltracker.com/projects/1275640/stories/124496741)

We have noticed that we are having a lot of temporary admin user removal
failures. We would like to receive notification when this happens, as
this indicates that there are possible issues with UAA and also leaves
an admin user behind. This can impact UAA (as we are leaking an user)
and security (as there would be many admin users that are not supposed
to exist).

## How to review

Deploy against your existing environments. Enable continuous smoketests by specifying `ALERT_EMAIL_ADDRESS`. e.g.
`SELF_UPDATE_PIPELINE=false ALERT_EMAIL_ADDRESS=my@email.address make dev pipelines`. Stop UAA by using bosh ssh and `monit stop uaa` on both servers. Then
* if your continuous smoketests are running, they might fail & send you smoketests failed email and then attempt temp. admin removal
* if smoketests are not running, you may trigger them by hand or wait until they get triggered by the timer resource. Creating temp user would fail and then removal will be attempted.
* removal will fail (as UAA is down) and notification email would be sent.

Recover your environment by starting UAA again on both servers. Smoketests should start passing again.

On an environment that is not deployed - it is possible that smoketests will start running (and failing) straight away. In that case it is enough to verify receival of the alert email.

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2012,3 +2012,23 @@ jobs:
         ensure:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
+          on_failure:
+            task: alert
+            config:
+              platform: linux
+              image: docker:///governmentpaas/awscli
+              params:
+                AWS_DEFAULT_REGION: {{aws_region}}
+                DEPLOY_ENV: {{deploy_env}}
+                SYSTEM_DNS_ZONE_NAME: {{system_dns_zone_name}}
+                ALERT_EMAIL_ADDRESS: {{ALERT_EMAIL_ADDRESS}}
+              inputs:
+                - name: paas-cf
+              run:
+                path: sh
+                args:
+                - -e
+                - -c
+                - |
+                  paas-cf/concourse/scripts/failed_admin_removal_email.sh \
+                    ${DEPLOY_ENV} ${SYSTEM_DNS_ZONE_NAME} ${ALERT_EMAIL_ADDRESS}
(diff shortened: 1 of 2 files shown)


bleach:
I appreciate the intent behind this, but I think work on this should be deferred until we have a state tracking monitoring system in place, because email is not a great user interface to be able to tell if a problem is still present.

From an email I can't tell if the user was already deleted eg. by another run of the job or by someone tidying up manually.

Also, a style note - IMO alerts should have explicit calls to action: "first step to resolving this is to do X" (with a URL if possibe)

mtekel:
First step to fixing a problem is knowing about it. The 'alert' itself is not really meant to say anything about state of a problem, it's just to notify us. Like we want to be notified in case of smoketest failure. 

Next run of the job creates new user, so that temp. user that failed removal (or maybe partially succeeded?) will not be removed automatically. This is a problem that requires user interaction to fix, hence notify user.

Regarding contents of mail - it's hard to include link to job, since build number is currently only present in the resource containers (see https://github.com/concourse/concourse/issues/494).

dcarley:
Summarising in-person conversation with @mtekel, we're not going to proceed with this because:

- the temporary users are eventually cleaned up by [`sync-admin-users`](https://github.com/alphagov/paas-cf/blob/master/scripts/sync-admin-users.rb)
- we have some coverage of UAA failures from our other tests (acceptance, smoke, etc.)
- we're not terribly happy with emails alerts from the pipeline (not just specific to this PR - not enough detail, requires people to check inbox, no indication of state, no correlation between failures, duplication of code)
- we're doing work now on system and component health monitoring which will improve those things
- we should make sure that we have kick-offs on all stories including tech debt while on support

PR #350, opened 2016-07-07
[#123862321] Upgrade to CloudFoundry 238

alext:
## What

In order to get the fix for CVE-2016-4468 (UAA SQL Injection). This upgrades to Cloudfoundry release 238.

This includes a release of Diego that supports the Bosh 2.0 zones feature, so I've also combined the cell_z*N* and colocated_z*N* jobs.

While upgrading, we ran into issues where existing deployed apps would stop working. This turned out to be because our diego jobs weren't ordered correctly so that the cells were upgraded before the database migrations had been applied.

## How to test

This should be tested twice, one as a clean deploy, and once upgrading from current master:

* test that a new deploy from this branch works correctly.
* Given an existing deployment running current master, test that deploying from this branch upgrades CF without downtime.

## Who can review

Anyone but myself or @bleach.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -124,7 +124,7 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-routing-release.git
       tag_filter: {{cf_routing_version}}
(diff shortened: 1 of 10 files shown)


keymon:
LGTM :+1:

PR #351, opened 2016-07-07
[#122183539] Visibility of important platform metrics

HenryTK:
## Who

[#122183539 - Visibility of important platform metrics](https://www.pivotaltracker.com/story/show/122183539)

We want to be able to deploy Grafana with dashboard configuration files (JSON files) supplied via the manifest. This pull request configures Grafana to use Graphite as a datasource, updates the pipeline to generate YAML containing the minified JSON, and adds the dashboard JSON file.

## How to review

**Note:** review after [pull request 58 on paas-docker-cloudfoundry-tools](https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/58) has been merged and [this FIXME commit](https://github.com/alphagov/paas-cf/pull/351/commits/2287e3ad5cf794e8f4c3ae21faf08109084f00ee) on this repository has been removed.

* Complete the review in [pull request 1 on paas-grafana-boshrelease](https://github.com/alphagov/paas-grafana-boshrelease/pull/1). **This must be done first**.
* Once the pull request above is merged we can point to the new Grafana bosh release. I am happy to do that, but for reference the steps are:
    * removing [this FIXME commit](https://github.com/alphagov/paas-cf/pull/351/commits/bcd8ff44a8dcb2f7e05db87b30af5933de29e05e)
    * tagging the merge commit. The tag name tends to follow [this convention](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/manifest/040-graphite.yml#L18)
    * updating the manifest so `releases.grafana.version` is the same as the new tag name
    * update the pipeline to use the new release via tag filter
* Re-deploy using the newly pinned resource.
* Log into Grafana using `make dev showenv` to get the URL. 
* Check the User Impact dashboard contains the metrics as defined in Pivotal Tracker.
* Consider whether the scales and times used are appropriate. Your development environment may not have a lot of metrics, so you can get creative generating some.

## Who

Anyone but @saliceti @combor or me

### DIFF
--- a/manifests/cf-manifest/grafana/user-impact.json
+++ b/manifests/cf-manifest/grafana/user-impact.json
@@ -1,0 +1,699 @@
+{
+  "id": 4,
+  "title": "User Impact",
+  "originalTitle": "User Impact",
+  "tags": [],
+  "style": "dark",
+  "timezone": "utc",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": true,
+  "rows": [
+    {
+      "collapse": false,
+      "editable": true,
+      "height": "150px",
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "graphite",
+          "editable": true,
+          "error": false,
+          "fill": 1,
+          "grid": {
+            "threshold1": null,
+            "threshold1Color": "rgba(216, 200, 27, 0.27)",
+            "threshold2": null,
+            "threshold2Color": "rgba(234, 112, 112, 0.22)",
+            "thresholdLine": false
+          },
+          "height": "250",
+          "id": 3,
+          "isNew": true,
+          "legend": {
+            "alignAsTable": false,
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 2,
+          "links": [],
+          "nullPointMode": "connected",
+          "percentage": false,
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "hide": false,
+              "refId": "A",
+              "target": "alias(scale(sumSeries(stats.counters.cfstats.router.*.ops.total_requests.rate), 60), 'Apps requests per minute')",
+              "textEditor": false
+            }
+          ],
+          "timeFrom": null,
+          "timeShift": null,
+          "title": "",
+          "tooltip": {
+            "msResolution": false,
+            "shared": true,
+            "value_type": "cumulative"
+          },
+          "type": "graph",
+          "xaxis": {
+            "show": true
+          },
+          "yaxes": [
+            {
+              "format": "short",
+              "label": "",
+              "logBase": 1,
+              "max": null,
+              "min": 0,
+              "show": true
+            },
+            {
+              "format": "short",
+              "label": "",
+              "logBase": 1,
+              "max": null,
+              "min": 0,
+              "show": true
+            }
+          ]
+        },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "graphite",
+          "editable": true,
+          "error": false,
+          "fill": 1,
+          "grid": {
+            "threshold1": null,
+            "threshold1Color": "rgba(216, 200, 27, 0.27)",
+            "threshold2": null,
+            "threshold2Color": "rgba(234, 112, 112, 0.22)"
+          },
+          "height": "250",
+          "id": 6,
+          "isNew": true,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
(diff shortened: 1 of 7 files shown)


alext:
It would be good to pin the Y axes of these graphs to 0.

Otherwise looks good.

HenryTK:
@alext 
I have:
* rebased against master, now that the upgrade to CF v238 has been merged.
* Removed both FIXME commits
* Tagged the merge commit in the [paas-grafana-boshrelease repository](https://github.com/alphagov/paas-grafana-boshrelease)
* Updated the `releases.grafana.version` value in the manifest to reflect the new tag.
* Set the minimum Y-axis values to 0.

**Note:** I never needed to update the pipeline, because removing the merge commit made that update for me.

You can now deploy from `visibility-of-metrics-122183539-minify` again and review.

manifests/cf-manifest/grafana/user-impact.json
```diff
@@ -0,0 +1,699 @@
...
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "hide": false,
+              "refId": "A",
+              "target": "alias(maxSeries(stats.gauges.cfstats.colocated_z*.*.ops.bbs.LRPsDesired), 'Desired apps')"
```
alext: This is no longer correct now that it's been rebased. The colocated_z* bosh jobs were merged into a single colocated job as part of the CF upgrade.

manifests/cf-manifest/grafana/user-impact.json
```diff
@@ -0,0 +1,699 @@
...
+            },
+            {
+              "refId": "B",
+              "target": "alias(maxSeries(stats.gauges.cfstats.colocated_z*.*.ops.bbs.LRPsRunning), 'Running apps')"
```
alext: Same here with the colocated_z* jobs.

manifests/cf-manifest/grafana/user-impact.json
```diff
@@ -0,0 +1,699 @@
...
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "alias(summarize(derivative(stats.gauges.cfstats.api.0.ops.cc.requests.completed), '1min', 'sum', false), 'API requests per minute')"
```
alext: I think the `0` in this target should be a `*`. As it stands, I'm not getting any metrics for this graph. I think this also needs a `sumSeries` before the derivative to combine the requests from the 2 api instances.

manifests/cf-manifest/grafana/user-impact.json
```diff
@@ -0,0 +1,699 @@
...
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "alias(sumSeries(stats.timers.cfstats.router.*.http.responsetimes.http.mean_90), 'Router response time')"
```
alext: This is taking the sum of 2 means, which doesn't really make sense. I think we should be averaging them instead.

manifests/cf-manifest/grafana/user-impact.json
```diff
@@ -0,0 +1,699 @@
...
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 6,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "refId": "A",
+              "target": "alias(sumSeries(stats.gauges.cfstats.api.*.ops.cc.http_status.4XX), 'API HTTP 4xx')"
```
alext: This should probably apply the same formula as the "API requests per minute" graph (ie take the derivative, and summarise per minute).

alext: I'm slightly puzzled as to why these metrics are being reported as a gauge and not a counter in the first place, but that's out of the scope of this story.

HenryTK:
@alext: I have made the changes in the comments for the diff of `c7ea017ad42af4d2ade3e6b3ef7389b3fa6c23f6`. The changes were squashed into that commit.

mtekel:
After discussing with @saliceti , I think we should rename `Router Xxx per minute` to `App Xxx per minute`. If the app is missing, router returns 404, but that's not an app metric (as app is missing) and is not added into the metric that we display...

HenryTK:
@mtekel: The change has been made and squashed into `d790681`.

mtekel:
There's one more thing - I think it would be good to test that the dashboards don't break our manifest. In some future versions grafana could change format or include weird characters (e.g. for color specification, or some unicode for labels), which when included into the final manifest could break yaml.

In this case we don't need to use any additional fixtures, but can use dashboards directly, as they are present in the repository at the test time...

mtekel:
It also looks like `cc-generic` doesn't involve all tasks in CC job queue. I would change the metric displayed to `total` instead, which seems to be more accurate - and less delayed.

HenryTK:
@mtekel: test helpers updated and changed `cc-generic` to `total`.

mtekel:
Nice. I've tried the tests by creating a bogus dashboard that would not output from script correctly, breaking the tests...

PR #353, opened 2016-07-12
[#121928523] Fix and disable cf ssh support

alext:
## What

This fixes the cf SSH support, and then disables it.

## How to review

Step1: verify that the ssh support is fixed:

* Deploy the branch up to and including commit "Generate and use a ssh host key for the ssh-proxy"
* Verify that `cf ssh` works and gives you a shell inside a container.

Step2: verify that ssh is disabled

* Deploy the complete branch.
* Verify that `cf ssh` no longer works
* Verify that the custom acceptance tests pass

## Who can review

Anyone but myself or @HenryTK


saliceti:
Tested with and without SSH, acceptance test ok, clean commits. Merging!

PR #354, opened 2016-07-13
Add reminder to change password

benh-gds:
## What

Change to the welcome email text, reminding user to change password.

This addresses user concern about receiving emails in plain text, found in user research.

See [Include Change Password Instructions in Welcome Mail][1] for
details. There is no point putting the actual command to change
password in the welcome email, because that won't work until user is
logged in, but adding this reminder makes clear that the password that was
included in the email can be changed by the tenant.

Also style fixes to email copy.

## How to review

Check the email copy makes sense. Check the email is still sent correctly when a the script runs.

## Who can review

Anyone but BH.

[1]: https://www.pivotaltracker.com/story/show/125537351

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -11,15 +11,17 @@ SUBJECT='Welcome to the Government PaaS'
 # shellcheck disable=SC2016,SC1078
 MESSAGE='Hello,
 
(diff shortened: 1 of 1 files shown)


mtekel:
This works. Mail still sends correctly. Thanks for the fixes.

PR #355, opened 2016-07-14
[#115884129] Switch to separate diego job distribution

alext:
## What

We want to switch to using separate VMs for the diego components, as recommended by upstream. From http://cf-dev.70369.x6.nabble.com/cf-dev-How-to-minimize-VMs-used-in-cf-diego-td4216.html:

> Also, as a warning, the Diego team doesn't regularly test the colocated-VM deployment configuration, so it's possible that it may accidentally break in the future (or even that it's broken now!). So I wouldn't recommend its use in a production deployment.

To avoid downtime, we're doing this as a 2 phase approach. This first step, adds the new VMs, but leaves the collocated VMs in place for now. This means that we'll have 4 instances of each component for the moment. Once this has been deployed to prod, we'll do a follow-up PR to remove the collocated VMs.

This PR also cleans up a couple of `FIXME`s, and inlines the job templates (see relevant commits for details)

## How to review

Deploy from master, then deploy from this branch. Verify that everything still works paying particular attention to the availability tests.

## Who can review

Anyone but @HenryTK, @benhyland or myself.

### DIFF
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -370,10 +370,6 @@ jobs:
         zone: ""
     update:
       serial: true
-    # FIXME: remove once deployed to prod.
-    migrated_from:
-      - {name: colocated_z1, az: z1}
-      - {name: colocated_z2, az: z2}
 
   - name: database
     azs: [z1, z2]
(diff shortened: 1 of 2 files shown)


keymon:
Good for me, deploys fine by upgrading and without downtime.

PR #357, opened 2016-07-15
Add .idea directory to .gitignore

benhyland:
## What

Add `.idea/` to `.gitignore`.

Prefer adding to the project rather than the global gitignore,
since other non-PaaS projects may choose to commit IDE config files.

## How to review

No special advice.

## Who can review

Anyone except @benhyland

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/*.tfstate
 vagrant/.vagrant/
 bin/fly*
+.idea/


keymon:
Fine with me, but I will close this one and open it again. the reason is that I think this will break the automated script that we have for merge PRs.

keymon:
https://github.com/alphagov/paas-cf/pull/358

mtekel:
@keymon merge script works also for this kind of PRs. See #354

keymon:
I see, sorry I assume it wouldn't.

Now is done, only more work for me :)

PR #361, opened 2016-07-18
[#122189093] Add collectd to bosh and concourse

mtekel:
## What

[Configure Collectd on Concourse and Bosh VMs](https://www.pivotaltracker.com/n/projects/1275640/stories/122189093)

## How to review

Deploy against existing or new environments. 
* Start with bootstrap concourse to update your deployer concourse. 
* Then watch deployer update bosh. 
* After deployment, observe concourse and bosh metrics in graphite under `collectd.<bosh/concourse>.collectd_0.<metrics>`. 
* Once [the pull request on paas-collectd-boshrelease](https://github.com/alphagov/paas-collectd-boshrelease/pull/2) has been merged, @mtekel or @HenryTK will upload the release to github and the FIXME commit can be removed.
* remove release 0.4 from your bosh and deploy again to test that the new release works (the URL is correct, hash matches etc.)
* after depoying to prod, sync master of https://github.com/alphagov/paas-collectd-boshrelease with upstream (= rollback of one PR merge we did on to master), as we are deploying from tarballs now and our main branch is now gds_master

## Who can review

not @mtekel or @HenryTK


manifests/bosh-manifest/reference-bosh-manifest.yml
```diff
@@ -151,6 +153,50 @@ networks:
...
+      <Plugin vmem>
+        Verbose false
+      </Plugin>
+
+      LoadPlugin "write_graphite"
+      <Plugin "write_graphite">
+       <Node "myNode">
+         #FIXME: hard coded static IP to be removed during #121602315
+         # This address comes from 040-graphite.yml:31 and has been copied
+         # as `grab` cannot interpolate in a multi-line string.
```
alext: It would be good to clarify that `040-graphite.yml` is from the CF manifest and not from this manifest. It's probably worth updating the comment in that file to mention that that IP is now also replicated in this manifest.

This is also another reason we can't use `grab` here (in addition to this being a multi-line string).

manifests/bosh-manifest/reference-bosh-manifest.yml
```diff
@@ -160,10 +206,14 @@ properties:
...
       0000000000000000000000000000000000000000000000000000000000000000
       -----END RSA PRIVATE KEY-----
+
 releases:
 - name: bosh
   sha1: 6b12652650b87810dcef1be1f6a6d23f1c0c13a7
   url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=255.8
+- name: collectd
+  sha1: 5e5c196f9c95fe12eb17bba0f40fdaecaddd6424
+  url: https://github.com/alphagov/paas-collectd-boshrelease/releases/download/0.4/collectd-0.4.tgz
```
alext: We're using version 0.3 [in the cf manifest](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/manifest/runtime/runtime.yml#L5). Should we use the same version here to be consistent?

HenryTK: @alext - there is a [corresponding pull request](https://github.com/alphagov/paas-collectd-boshrelease/pull/1) in the paas-collectd-boshrelease repository for upgrading to 0.4

manifests/bosh-manifest/reference-bosh-manifest.yml
HenryTK: @alext - good spot, these messages have now been updated.

manifests/bosh-manifest/deployments/040-collectd.yml
```diff
@@ -0,0 +1,56 @@
+---
+releases:
+- name: collectd
+  sha1: 5e5c196f9c95fe12eb17bba0f40fdaecaddd6424
+  url: https://github.com/alphagov/paas-collectd-boshrelease/releases/download/0.4/collectd-0.4.tgz
```
dcarley: Could you explain in the two commit messages where this URL is used:

- why we're using a fork, so that we know when we can go back to using upstream
- the steps to create and upload the release, so that someone can update it again in future

mtekel: The initial reason why we used fork was that we could add custom tags and pin our deployment to these tags.
Now, because bosh-init can only consume tarballs (you can't upload a release to bosh-init like you can to BOSH), I had to create a dev-release tarbal (using standard bosh-cli way `bosh create release --with-tarball --force --version 0.4`, more documentation about bosh releases: https://bosh.io/docs/create-release.html#dev-release-release). I have created the tarballs from the same code as is tagged. That is, 0.4 release is created from code that's tagged as 0.4.

It seemed to me that a good place to put the release tarball was in github release. This practice is also used in some CF repositories (e.g. diego). We have a spike to analyse our situation with releases completely, but we did not want to block this story on it (#115142265 - see comments in the story). In the future, we can easily move the release location to elsewhere and we can change the way to build the releases.

Regarding changing the commit messages:
* I think it should be instead described in the repo why we have a fork of it (e.g. in readme). It seems a bit odd to have to search through commits in other repos hoping for explanation.
* steps to create and upload release are a standard thing, documented on BOSH side or when you print help of bosh-cli. If we have a special workflow (e.g. where to put the releases), they should be described in ADR or team manual or some other generally accessible place, instead of random commit messages.
* both information above are not really related to the code in the commit

manifests/bosh-manifest/deployments/040-collectd.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+- name: bosh
+  templates:
+  - {name: collectd, release: collectd}
+
+properties:
+  collectd:
+    interval: 10
+    config: |
```
dcarley: Are the three copies of this config (Deployer, BOSH, BOSH managed VMs) all the same? Should we include it from a shared file to prevent duplication and ensure that they all benefit from changes (such as additional plugins) at the same time?

manifests/bosh-manifest/deployments/040-collectd.yml
dcarley: > The initial reason why we used fork was that we could add custom tags and pin our deployment to these tags.
> Now, because bosh-init can only consume tarballs (you can't upload a release to bosh-init like you can to BOSH), I had to create a dev-release tarbal (using standard bosh-cli way bosh create release --with-tarball --force --version 0.4, more documentation about bosh releases: https://bosh.io/docs/create-release.html#dev-release-release). I have created the tarballs from the same code as is tagged. That is, 0.4 release is created from code that's tagged as 0.4.
> 
> It seemed to me that a good place to put the release tarball was in github release. This practice is also used in some CF repositories (e.g. diego). We have a spike to analyse our situation with releases completely, but we did not want to block this story on it (#115142265 - see comments in the story). In the future, we can easily move the release location to elsewhere and we can change the way to build the releases.

That's good information. Please put it in the commits 👍 

> I think it should be instead described in the repo why we have a fork of it (e.g. in readme). It seems a bit odd to have to search through commits in other repos hoping for explanation.

It could work either way. I think that putting it in the commit message that uses it will be easier to maintain though, because the reasons for use may change over time and we're unlikely to keep the README up-to-date. In this particular case, our previous requirement to fork for stability and our new requirement to use it with bosh-init has meant that we've had to do something different, so it's relevant to that commit.

I don't think it's hard to find. Chances are you'd want to see where the release was used before you changed it, so you could use `git log -p -G <repo>` to find occurrences in our single repo.

> steps to create and upload release are a standard thing, documented on BOSH side or when you print help of bosh-cli. If we have a special workflow (e.g. where to put the releases), they should be described in ADR or team manual or some other generally accessible place, instead of random commit messages.

Nothing is standard if it's not automated. I hope it will be in the future after that SPIKE, but in the meantime we have no history of how it go there. What is standard and obvious to you may not be to the next person that has to work on it.

An example of this was the Terraform provider that we had to compile and vendor in 6f1014eaddc4bd32c01c5b595f0e7563c9a1842a - the next person that worked on it asked how they were created, even though they were familiar with Go, but was able to find all the information they needed from the commit.

manifests/bosh-manifest/deployments/040-collectd.yml
```diff
@@ -0,0 +1,56 @@
...
+        Verbose false
+      </Plugin>
+
+      LoadPlugin "write_graphite"
+      <Plugin "write_graphite">
+       <Node "myNode">
+         # Host is the static IP of Graphite, which is defined in the
+         # CloudFoundry manifest. The CloudFoundry manifest is rendered
+         # separately, and therefore this value cannot be templated in.
+         Host "10.0.16.20"
```
dcarley: I can't see anything in the Pivotal notes, but was using a DNS hostname or ELB to replace the hardcoded IPs discussed in kick-off? Not suggesting you do it immediately or within this story, but worth considering now.

(might also be alleviated by my comment about de-duping the config file)

manifests/bosh-manifest/deployments/040-collectd.yml
HenryTK: > I think it should be instead described in the repo why we have a fork of it (e.g. in readme). It seems a bit odd to have to search through commits in other repos hoping for explanation.

If we did that and there was an intention to merge changes upstream we will have to use some sort of branching to keep those README edits out of any pull requests we raise.

manifests/bosh-manifest/deployments/040-collectd.yml
mtekel: Right now they are the same, except for comments. I was looking how to make it the same (meaning, included from single place) for all of them but it is non-trivial. In cf-manifest, this is a runtime addon. We'd have to parse it out from that file to include it into bosh - in a different location. Concourse manifest is not even composed from multiple files, so that would require another way again. Altogether this looks like a big piece of work that we might not want to do for this story. Second question is if we really want to have all the settings the same. E.g. if we need to deploy a collectd fix for concourse (0.4), there's no need to re-deploy bosh as well and do rolling update of whole CF deployment....

manifests/bosh-manifest/deployments/040-collectd.yml
mtekel: Introducing ELB/DNS is a whole new thing. We might have to do it for graphite HA. For now, fixed IP is fine. A comment was added to graphite part in cf-manifest to ensure IP doesn't change when we switch back to IPs generated by spruce.

In kick-off we decided to do minimal viable implementation.

manifests/bosh-manifest/deployments/040-collectd.yml
alext: > Concourse manifest is not even composed from multiple files

Concourse is actually still run through spruce (it's the single file here merged with the terraform outputs etc.).

It should be possible to put all this config in a single file, and then add it to the list of files pulled in by spruce in the various pipeline tasks.

manifests/bosh-manifest/deployments/040-collectd.yml
mtekel: I'll add description to the release. That seems the best place to describe what and why.

manifests/bosh-manifest/deployments/040-collectd.yml
mtekel: That opens a second question - do we want to do this? Feels like we should have a 2nd kickoff for this story...

manifests/bosh-manifest/deployments/040-collectd.yml
dcarley: Please add it to the commit messages, too.

manifests/bosh-manifest/deployments/040-collectd.yml
dcarley: 👌 

It's not ideal that we have it hardcoded and duplicated in 6 places now. If we're not able to de-dupe the configs (above), then we should note it as follow-up work in the story summary.

manifests/bosh-manifest/deployments/040-collectd.yml
dcarley: Discussed - not currently possible because they have different prefixes to workaround another issue and we can't interpolate into multiline strings. We should record it as follow-up work.

keymon:
The PR seems fine, but I am not happy with the collectd release PR: https://github.com/alphagov/paas-collectd-boshrelease/pull/1#issuecomment-233667311

dcarley:
Other PR is being addressed now.

@HenryTK also found a way to make it possible to consolidate the duplicate configs by using `collectd.hostname_prefix` so they're going to give it a try 😄

HenryTK:
The collectd properties used across CloudFoundry, Bosh, and Concourse have now been unified into the same file, meaning a complete refactoring of the original commits for this pull request.

dcarley:
> The collectd properties used across CloudFoundry, Bosh, and Concourse have now been unified into the same file, meaning a complete refactoring of the original commits for this pull request.

Thanks 👏 

Looks like the "How to review" instructions need updating to mention the FIXME commit.

mtekel:
This is still wip, we need to do some more updates (e.g. unify deployment from tarball).

mtekel:
Ready for review again.

keymon:
Do you mind rename `manifests/cf-manifest/runtime-config/runtime-config.yml ` to `manifests/cf-manifest/runtime-config/runtime-config-base.yml` or so? 

It is a little bit confusing that you use spruce to merge that file and collectd config but the resulting file has the same name.

HenryTK:
`runtime-config.yml` has now been renamed to `runtime-config-base.yml` to distinguish it from the final rendered file.

keymon:
Everything seems good. Doing final checks of last changes but so far so good.

mtekel:
For some reason, travis test results are missing for this PR. I wonder if it is mergable...

alext:
> For some reason, travis test results are missing for this PR. I wonder if it is mergable

Travis seems to be being quite slow in picking up builds at the moment. It is currently possible to merge PRs before Travis has picked them up. Whether that's desirable behaviour or not is a separate question...

PR #362, opened 2016-07-18
[#115884129] Fix LRP metrics

saliceti:
## What
The LRP metrics come from BBS but BBS has moved from 'colocated' to
'database' VM. The metric name contains the VM name so it must be
updated.

## How to review
Deploy and load the metrics dashboard, it should show the LRP metrics.

## Who can review
Anyone but myself

### DIFF
--- a/manifests/cf-manifest/grafana/user-impact.json
+++ b/manifests/cf-manifest/grafana/user-impact.json
@@ -541,11 +541,11 @@
             {
               "hide": false,
               "refId": "A",
-              "target": "alias(maxSeries(stats.gauges.cfstats.colocated.*.ops.bbs.LRPsDesired), 'Desired apps')"
+              "target": "alias(maxSeries(stats.gauges.cfstats.database.*.ops.bbs.LRPsDesired), 'Desired apps')"
             },
             {
               "refId": "B",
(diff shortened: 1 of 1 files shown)


dcarley:
Subject should reference a Pivotal story. Is https://www.pivotaltracker.com/story/show/122183539 the right one? (nm: Alex fixed before merge)

I'm interested in how you noticed that this was broken. Did Grafana display a red error message box? It seems likely that we'll break these again in future without noticing until the change reaches production (where we look at the dashboard more regularly). Not sure there's anything we can do about that especially for metrics that we're unlikely to alert on in future.

alext:
> I'm interested in how you noticed that this was broken. Did Grafana display a red error message box? It seems likely that we'll break these again in future without noticing until the change reaches production (where we look at the dashboard more regularly). Not sure there's anything we can do about that especially for metrics that we're unlikely to alert on in future.

Actually, the dashboard was showing stale data. Upon reload, it showed an empty graph. This is presumably because the metrics for the colocated VMs still exist in graphite, but are no longer receiving any data.

I was thinking about whether it would be possible to test these dashboards in any way, but couldn't come up with anything workable.

PR #364, opened 2016-07-19
[#121043905] Set kibana timezone to UTC

combor:
## What

Story: [Set logsearch timezone to UTC](https://www.pivotaltracker.com/story/show/121043905)

Change kibana settings to display log timestamps in UTC instead of the browser's local timezone (e.g. BST). This will make it easier to correlate with other logs and incident reports.

## How to review

Apply on top of master branch and:

- go to Kibana ( you can find the URL from `showenv` make target )
- go to settings
- go to advance settings 
- check if `dateFormat:tz` is set to UTC
- search logs in Kibana and select absolute time to be one hour ahead - you should see no logs there  

## Who can review

not @benhyland or @combor

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1514,6 +1514,31 @@ jobs:
 
                   cf push
 
+        - task: update-kibana-timezone
+          config:
+            platform: linux
+            image: docker:///ruby#2.2-slim
+            params:
+              SYSTEM_DNS_ZONE_NAME: {{system_dns_zone_name}}
+              DEPLOY_ENV: {{deploy_env}}
+            inputs:
+            - name: paas-cf
+            - name: cf-tfstate
+            run:
+              path: sh
+              args:
+              - -e
+              - -c
+              - |
+                VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+                SCRIPTS_PATH="./paas-cf/concourse/scripts"
+                TF_VARS_SCRIPT="extract_tf_vars_from_terraform_state.rb"
+                $SCRIPTS_PATH/$TF_VARS_SCRIPT < cf-tfstate/cf.tfstate | grep logsearch_elastic_master_elb_dns_name > ./logsearch_var.sh
+                . ./logsearch_var.sh
+                export ES_HOST="${TF_VAR_logsearch_elastic_master_elb_dns_name}"
+                export ES_PORT="9200"
+                $(pwd)/paas-cf/concourse/scripts/kibana_set_utc.rb
+
   - name: post-deploy
     plan:
       - aggregate:
@@ -1527,6 +1552,7 @@ jobs:
           - get: cf-secrets
             passed: ['cf-deploy']
           - get: bosh-CA
+          - get: cf-tfstate
 
       - task: retrieve-config
         config:
--- a/concourse/scripts/kibana_set_utc.rb
+++ b/concourse/scripts/kibana_set_utc.rb
@@ -1,0 +1,76 @@
+#! /usr/bin/env ruby
+
+require 'net/http'
+require 'json'
+
+def config_uri(es_url)
+  config_url = "#{es_url}/.kibana/config/4.3.1"
+  URI(config_url)
+end
+
+def index_settings_uri(es_url)
+  index_settings_url = "#{es_url}/.kibana"
+  URI(index_settings_url)
+end
+
+def update_es_index(uri, document)
+  send_request(
+    uri: uri,
+    http_method: :Put,
+    body: document,
+    allowed_response_codes: ['200', '201']
+  )
+end
+
+def send_request(uri:, http_method:, allowed_response_codes:, body: nil)
+  raise "Document is not a hash: #{body}" unless body == nil || body.is_a?(Hash)
+  response = nil
+  Net::HTTP.new(uri.host, uri.port).start do |http|
+    method = Net::HTTP.const_get(http_method)
+    headers = body ? { 'Content-Type' => 'application/json' } : {}
+    req = method.new(uri.path, headers)
+    req.body = body.to_json if body
+    response = http.request(req)
+    unless allowed_response_codes.include?(response.code)
+      raise "Unexpected response code: #{response.code}\n#{response.body}"
+    end
+  end
+  response
+end
+
+def need_to_create_index(response)
+  !response["found"]
+end
+
+def need_to_set_timezone(response)
+  !response["_source"] || response["_source"]["dateFormat:tz"] != "UTC"
+end
+
+def set_utc_config
+  es_host = ENV.fetch("ES_HOST")
+  es_port = ENV.fetch("ES_PORT")
+
+  es_url = "http://#{es_host}:#{es_port}"
+
+  config_uri = config_uri(es_url)
+  response = send_request(uri: config_uri, http_method: :Get, allowed_response_codes: ['200', '404'])
+  response_json = JSON.parse(response.body)
+
+  if need_to_create_index(response_json)
+    index_settings = {
+      settings: {
+        index: {
+          number_of_shards: 1,
+          number_of_replicas: 1,
+        }
+      }
+    }
+    update_es_index(index_settings_uri(es_url), index_settings)
+  end
+
+  if need_to_set_timezone(response_json)
+    update_es_index(config_uri(es_url), { "dateFormat:tz" => "UTC" })
+  end
+end
+
+set_utc_config
--- a/concourse/scripts/spec/kibana_set_utc_spec.rb
+++ b/concourse/scripts/spec/kibana_set_utc_spec.rb
@@ -1,0 +1,116 @@
+require 'mimic'
+
+RSpec.describe "kibana_set_utc.rb", :type => :aruba do
+
+  ES_HOST = "127.0.0.1"
+  ES_PORT = "9200"
+  KIBANA_CONFIG_PATH = "/.kibana/config/4.3.1"
+  KIBANA_INDEX_PATH = "/.kibana"
+
+  Thread.abort_on_exception = true
+
+  def run_set_utc_script
+    run("./kibana_set_utc.rb")
+  end
+
+  before :all do
+    @elasticsearch = Mimic.mimic(:hostname => ES_HOST, :port => ES_PORT)
+    set_environment_variable 'ES_HOST', ES_HOST
+    set_environment_variable 'ES_PORT', ES_PORT
+  end
+
+  after :each do
+    Mimic.reset_all!
+  end
+
+  after :all do
+    Mimic.cleanup!
+  end
+
+  it "creates index and adds utc config if no index exists" do
+
+    @elasticsearch.instance_variable_get(:@app).not_found do
+      # override the not_found handler provided by mimic,
+      # so that we don't emit an empty body instead of the json we want with the 404
+      [404, {}, nil]
+    end
+
+    @elasticsearch.get(KIBANA_CONFIG_PATH).returning('{ "found": false }', 404)
+    @elasticsearch.put(KIBANA_INDEX_PATH).returning('{}', 201)
+    @elasticsearch.put(KIBANA_CONFIG_PATH).returning('{}', 201)
+
+    run_set_utc_script
+
+    expect(last_command_started).to have_exit_status(0)
+    expect(@elasticsearch.received_requests.size).to be(3)
+    expect(@elasticsearch.received_requests).to contain_request('GET', KIBANA_CONFIG_PATH)
+    expect(@elasticsearch.received_requests).to contain_request('PUT', KIBANA_INDEX_PATH)
+    expect(@elasticsearch.received_requests).to contain_request('PUT', KIBANA_CONFIG_PATH)
+  end
+
+  it "adds utc config if index exists but config is not present" do
+    @elasticsearch.get(KIBANA_CONFIG_PATH).returning('{ "found": true }', 200)
+    @elasticsearch.put(KIBANA_CONFIG_PATH).returning('{}', 200)
+
+    run_set_utc_script
+
+    expect(last_command_started).to have_exit_status(0)
+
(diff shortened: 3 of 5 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1525,6 +1526,30 @@ jobs:
...
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
+              VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+              CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
+              export CONCOURSE_ATC_PASSWORD
+              ES_HOST=$(./paas-cf/concourse/scripts/bosh-cli.sh bosh vms 2>&1 | awk -F'|' '$2~/elasticsearch_master/ {print $6}' | head -1)
```
dcarley: I was originally going to say that I don't think we should use `bosh-cli.sh` to loop back through Concourse for tasks like this because it's complex and will make it harder for us to change Concourse authentication in future.

Then I realised that it might be simpler still to use the hostname of the ELB, which points to all of the ES masters and has a basic healthcheck to remove dead nodes:

https://github.com/alphagov/paas-cf/blob/999fd66836efa1fbeaabafa670289e2d475fd4e4/terraform/cloudfoundry/outputs.tf#L81-L83

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
+#! /usr/bin/env ruby
+
+require 'net/http'
+require 'json'
+
+def ingest_to_es(document, uri)
```
dcarley: Could we rename this to something like `update_es_index` or `upsert_es_index`?

I found `ingest` confusing because it already has a meaning in the logsearch release and our manifests.

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+  raise "Document is not a hash" unless document.is_a? Hash
+  Net::HTTP.new(uri.host, uri.port).start do |http|
+    req = Net::HTTP::Put.new(uri.path, 'Content-Type' => 'application/json')
+    req.body = document.to_json
+    http.request(req)
+  end
+end
+
+def date_format_not_utc(response_json)
```
dcarley: This method isn't used. Can it be removed?

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+end
+
+es_host = ENV.fetch("ES_HOST").delete(' ')
+es_port = ENV.fetch("ES_PORT").delete(' ')
+
+es_url = "http://#{es_host}:#{es_port}"
+date_format_url = "#{es_url}/.kibana/config/4.3.1"
+date_format_uri = URI(date_format_url)
```
dcarley: Could you change these variables to `config_url` and `config_uri`?

The existing name sounds more like a verb than a noun and doesn't describe what the endpoint is (which returns more than just `dateFormat`).

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+
+response = JSON.parse(Net::HTTP.get(date_format_uri))
+
+if ! response["found"] || response["_source"]["dateFormat:tz"] != "UTC"
```
dcarley: This silently fails to set the timezone if there's a problem connecting to ES or the index doesn't exist. I think the script should return a non-zero exit code when these happen, so we'd be aware of the failure in the pipeline.

Are there any cases where we would want it to proceed on failure? Is the index only created the first time you log in to Kibana? This might be related to how we deal with the version number comment above.

We should also have some basic Aruba tests for this script, which test it from the outside and assert these kind of behaviours. The ES endpoint could probably be mocked using Mimic like we have for `bosh_pre_destroy`.

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+  index_settings_url = es_url + '/.kibana'
+  index_settings_uri = URI(index_settings_url)
+  index_settings = {
+    "settings" =>
+    { "index" => {
+      "number_of_shards" => 1,
+      "number_of_replicas" => 1,
```
dcarley: Could you describe why we set these?

Should the URL/URI be outside the block like the other one?

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+end
+
+def date_format_not_utc(response_json)
+end
+
+es_host = ENV.fetch("ES_HOST").delete(' ')
+es_port = ENV.fetch("ES_PORT").delete(' ')
+
+es_url = "http://#{es_host}:#{es_port}"
+date_format_url = "#{es_url}/.kibana/config/4.3.1"
```
dcarley: Hardcoding the Kibana version number here seems quite fragile, because I'm not confident that we'd remember to change it when Kibana is upgraded. Any thoughts on how we can make it more resilient?

Related to the comment below about testing exit codes, I think this would also fail silently and we wouldn't know that the timezone hadn't been reconfigured.

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,43 @@
...
+
+require 'net/http'
+require 'json'
+
+def ingest_to_es(document, uri)
+  raise "Document is not a hash" unless document.is_a? Hash
+  Net::HTTP.new(uri.host, uri.port).start do |http|
+    req = Net::HTTP::Put.new(uri.path, 'Content-Type' => 'application/json')
+    req.body = document.to_json
+    http.request(req)
```
dcarley: I don't think this will raise an exception for an unexpected response. What do we expect, status code of <300?

Related to exit code testing comment below.

dcarley:
Could you add a brief description to the commit message about why we're changing the timezone? Most of the team have context at the moment, because it's annoyed us recently, but people may not in future.

Is it also worth noting in the commit message that this could be refactored when Kibana 5.x is released and a link to elastic/kibana#6849?

concourse/scripts/kibana_set_utc.rb
combor: It's possible that `.kibana` index doesn't exist when we push `dateFormat` setting to Elasticsearch. In that case new index is created with default settings - 5 primary shards and one replica per shard. That's way too many of them ( each shard is lucene process ). Kibana UI creates one primary shard and one replica so we replicate this behaviour here. I will update commit message to explain that.

concourse/scripts/kibana_set_utc.rb
combor: > This silently fails to set the timezone if there's a problem connecting to ES or the index doesn't exist. 

`response = JSON.parse(Net::HTTP.get(date_format_uri))` will fail non-silently if we can't connect to ES. If index is not found, which is perfectly fine, ES will return 404 code and json document in the response body containing `"found": false`, in such case script will create `.kibana` index with `dateFormat:tz` setting.

> Are there any cases where we would want it to proceed on failure?

In my opinion there are no such cases. If we can't update ( or create ) index we shouldn't continue. 

>Is the index only created the first time you log in to Kibana?

Yes, but it can be updated later to store dashboards etc. If it already exists it is (re)used.

> We should also have some basic Aruba tests for this script

I'll look into this.

concourse/scripts/kibana_set_utc.rb
combor: > Should the URL/URI be outside the block like the other one?

I thought that we can avoid assigning unnecessary variables if condition is not met.

concourse/scripts/kibana_set_utc.rb
combor: > Hardcoding the Kibana version number here seems quite fragile

True, I thought about it as well but I couldn't find any place where Kibana version is exposed. I can't even get it from ES as it's possible that index doesn't exists. If anyone has an idea how to fix this than I'll be glad to remove hardcoded version.

concourse/scripts/kibana_set_utc.rb
dcarley: It might also be clearer to break it out to two separate conditions:
```ruby
if ! response["found"]
  ingest_to_es(index_settings, index_settings_uri)
end

if response["_source"]["dateFormat:tz"] != "UTC"
  ingest_to_es(timezone_settings, date_format_uri)
end
```

HenryTK:
Now that [pull request 374](https://github.com/alphagov/paas-cf/pull/374) has been merged is there an intention to update this branch prior to review?

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1525,6 +1526,31 @@ jobs:
                   echo export "${var_name}"=\"$(eval echo \$${var_name})\"
                 done > config/config.sh
 
+      - task: update-kibana-timezone
```
alext: Is there a reason for this to be in this post-deploy job as opposed to in the main cf-deploy job with most of the other post-deploy tasks. The advantage of doing that is that it can be run in parallel with them.

AIUI, this job was only introduced for the healthcheck app because we're not currently deploying it in a zero-downtime manner, and it therefore would have broken the availability tests if it was included in the main cf-deploy job. With that in mind, this job probably doesn't have the best name.

concourse/scripts/spec/kibana_set_utc_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    run_set_utc_script
+
+    expect(last_command_started).to have_exit_status(1)
+
+    expect(@elasticsearch.received_requests.size).to be(1)
+    expect(@elasticsearch.received_requests).to contain_request('GET', KIBANA_CONFIG_PATH)
+  end
+
+  RSpec::Matchers.define :contain_request do |expected_method, expected_path|
```
alext: Within a describe block, you can just do:
```ruby
matcher :contain_request do ...
```
This also scopes the matcher to the describe block.  I'm not sure whether the same happens when using `RSpec::Matchers.define`.

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,76 @@
...
+  !response["found"]
+end
+
+def need_to_set_timezone(response)
+  !response["_source"] || response["_source"]["dateFormat:tz"] != "UTC"
+end
+
+def set_utc_config
+  es_host = ENV.fetch("ES_HOST").delete(' ')
+  es_port = ENV.fetch("ES_PORT").delete(' ')
```
alext: You could use `strip` here instead of `delete(' ')` - http://apidock.com/ruby/v1_9_3_392/String/strip

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,76 @@
...
+def send_request(uri:, http_method:, allowed_response_codes:, body: nil)
+  raise "Document is not a hash: #{body}" unless body == nil || body.is_a?(Hash)
+  response = nil
+  Net::HTTP.new(uri.host, uri.port).start do |http|
+    method = Net::HTTP.const_get(http_method)
+    headers = body ? { 'Content-Type' => 'application/json' } : {}
+    req = method.new(uri.path, headers)
+    req.body = body.to_json if body
+    response = http.request(req)
+    if allowed_response_codes.all? { |code| code != response.code }
```
alext: This might read better as:
```ruby
unless allowed_response_codes.include?(response.code)
```

concourse/scripts/kibana_set_utc.rb
benhyland: `strip` only removes leading and trailing whitespace, while we need to remove internal spaces.

concourse/scripts/kibana_set_utc.rb
```diff
@@ -0,0 +1,76 @@
...
+    allowed_response_codes: ['200', '201']
+  )
+end
+
+def send_request(uri:, http_method:, allowed_response_codes:, body: nil)
+  raise "Document is not a hash: #{body}" unless body == nil || body.is_a?(Hash)
+  response = nil
+  Net::HTTP.new(uri.host, uri.port).start do |http|
+    method = Net::HTTP.const_get(http_method)
+    headers = body ? { 'Content-Type' => 'application/json' } : {}
```
HenryTK: This means if body has been set (not `Nil`) then overwrite it with `{ 'Content-Type' => 'application/json' }`, which is confusing. Should it be the other way around?

concourse/scripts/kibana_set_utc.rb
alext: I'm not clear why we would we expect these to contain internal whitespace. If there is a case like this, we should at least have a test to cover it.

concourse/scripts/kibana_set_utc.rb
benhyland: The intention here is to set the headers - we only send a body in json format, so send the appropriate header if the body was given, or send no headers if it wasn't

HenryTK:
This branch needs updating against master to stop it regressing the collectd version on all CloudFoundry VMs.

alext:
The commit "Test for script to set Kibana default timezone" should probably be squashed into the commit that creates that script.

HenryTK:
In the absence of @combor and @benhyland I have squashed the commits containing the script and the tests for the script and updated the branch.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1514,6 +1514,31 @@ jobs:
 
                   cf push
 
+        - task: update-kibana-timezone
+          config:
+            platform: linux
+            image: docker:///ruby#2.2-slim
```
mtekel: note that after https://github.com/alphagov/paas-cf/pull/381 is merged, this should be changed to image_resource

HenryTK: @mtekel Thanks. This PR will be merged first. @keymon knows to do the rebasing of #381

PR #366, opened 2016-07-20
[#115884129] Remove colocated vm type.

alext:
## What

In #359 we removed the colocated VMs, but omitted to remove this vm
type definition from the cloud-config.

## How to review

Verify that the Travis tests pass (there's a unit-test that verifies that no jobs reference a vm_type that doesn't exist - https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/spec/integration/integration_spec.rb#L43)

## Who can review

Anyone but myself.


alext:
Good spot.  I've added a commit to clean that up.

dcarley:
> Good spot. I've added a commit to clean that up.

Could you add a summary to the commit message of the conversation about we didn't replace it with something else: we didn't know whether to replace it with `database` or `brain` (which was more stateful and important), we aren't running these regularly at the moment, and we need to extend the list of components tested anyway.

I've added a note in https://www.pivotaltracker.com/story/show/115237899

PR #367, opened 2016-07-20
[#126398487] Set RDS Broker backup retention period to 7 days

keymon:
[#126398487 Enable Backups for Postgres instances](https://www.pivotaltracker.com/n/projects/1275640/stories/126398487)

## What

We want to ensure that RDS instances created by the broker have backups enabled and retained for 7 days.

We also want to change the existing provisioned instances to this default value.

## Dependencies

- [x]  All tenants must be contacted about the downtime when enabling the retention period for preexisting instances.
- [ ] We must allow deployer to update the instances by applying https://github.gds/government-paas/aws-account-wide-terraform/pull/51 

## How to review

 1. Deploy the environment with `master`
 2. In AWS console, change the "Maintenance window" of the `healthcheck-db` (search for a `rdsbroker-*` RDS instance in your VPC) to something in the next 20-30 minutes or so. Be aware it is UTC time, so 1h less than London. Check the "Apply immediately" checkbox. 
 3. Deploy this branch. The RDS broker instances should be updated and the post-deploy job should update the existing instances.

Check that new instances have the right retention:

 1. Try to create a new RDS broker instance:

     ```
DEPLOY_ENV=...
CF_ADMIN_PWD=$(aws s3 cp s3://${DEPLOY_ENV}-state/cf-secrets.yml  - | grep uaa_admin_password | awk '{print $2}')
cf login --skip-ssl-validation -a https://api.${DEPLOY_ENV}.dev.cloudpipeline.digital -u admin -p ${CF_ADMIN_PWD}
cf target -o testers -s healthcheck
cf create-service postgres  M-dedicated-9.5  test-retention-db
```
 2. Wait for the new instance to be created, go to the AWS console and check that it has retention policy 7 days. 

Check that the existing instances are updated: 

After the maintenance window has passed, check that the existing instance for `healthcheck-db` has the right retention period. You can see that has been restarted in the logs.
    
## Who can review

Anyone but @keymon or @timmow

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-awscli>=1.10.16
+awscli>=1.10.38
 yamllint>=1.0.0
 PyGithub>=1.26.0
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1547,10 +1547,31 @@ jobs:
                 CF_ADMIN=admin
                 CF_PASS=$($VAL_FROM_YAML secrets.uaa_admin_password cf-secrets/cf-secrets.yml)
                 API_ENDPOINT=$($VAL_FROM_YAML properties.cc.srv_api_uri cf-manifest/cf-manifest.yml)
+                RDS_BROKER_BACKUP_RETENTION_PERIOD=$($VAL_FROM_YAML jobs.rds_broker.properties.rds-broker.catalog.services.0.plans.0.rds_properties.backup_retention_period cf-manifest/cf-manifest.yml)
 
-                for var_name in CF_ADMIN CF_PASS API_ENDPOINT; do
+                for var_name in CF_ADMIN CF_PASS API_ENDPOINT RDS_BROKER_BACKUP_RETENTION_PERIOD; do
                   echo export "${var_name}"=\"$(eval echo \$${var_name})\"
                 done > config/config.sh
+
+      # FIXME: Remove this task once the retention is updated in all running instances
+      - task: update-rds-broker-backup-retention-period
+        config:
+          platform: linux
+          image: docker:///governmentpaas/awscli
+          params:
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by HenryTK** [comment]
Regarding updating the retention period (commit `d6b45e828a7c2af4eac53ee0c3177b78dc0ece44`): can we guarantee future layout/Sprucing of the manifest won't change the order of the list of plans?

Also, if you are to assume all retention periods are going to be the same, we may as well provide a test asserting that assumption. Possibly two tests:

1. Assert the first plan in the list has a retention period set.
2. Assert all retention periods are the same.

**Comment by keymon** [comment]
@HenryTK Good suggestion, I will add it in a while.

File: requirements.txt [hunk 1]
```diff
@@ -1,3 +1,3 @@
-awscli>=1.10.16
+awscli>=1.10.48
```
alext: Where is this file used? I'd expect us to have to update one of our containers if we need a newer version of this.

keymon: It is not used by any of the containers or automation, it is only used by ourselves for local development.

The awscli container has a good version... On the other hand, it would make sense to use the same version than awscli containers :-m

keymon:
@HenryTK I added the test

keymon:
@alext updated the commit message to explain that we use requirements.txt for local dev, and set the version to the one used in the container.

keymon:
@HenryTK ready to be merged.

Notes:
 * I changed the order or the commits
 * We contacted the tenants, they are OK with it.
 * We confirmed with @dcarley that it is OK to not add tests on the update retention period script, as it is one-off.
 * We should remove the task to update the existing instances once it is deployed in prod.

alext:
I reckon the "Test retention period for RDS broker pg plans" commit could be squashed into the previous commit. IMO, changes to tests should be made along with the corresponding changes to the code, not as a separate commit.

keymon:
Yes, I guess you are right. I will squash them.

PR #368, opened 2016-07-21
[#126686741] Create some standard quotas

alext:
## What

Now that we're creating more orgs we want some standard quota definitions that can be applied to to these orgs.

This PR also removes the `paas_demo` quota as this is identical to the new `small` quota. The quota setting script doesn't delete quotas not present in the manifest (it's create/update only), so this won't actually remove the quota from the deployment. Once this is deployed, the quota for the paas_demo org can be updated to use, and then I'll add a temporary task to clean up this quota.

## How to review

Deploy against an existing environment from this branch. Verify that the new quotas are created. Verify that the existing `paas_demo` quota isn't removed.

## Who can review

Anyone but myself.

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -261,6 +261,21 @@ properties:
         total_services: 10
         non_basic_services_allowed: false
         total_routes: 1000
+      small:
+        memory_limit: 10240
+        total_services: 10
+        non_basic_services_allowed: true
+        total_routes: 1000
+      medium:
+        memory_limit: 61440
+        total_services: 10
+        non_basic_services_allowed: true
+        total_routes: 1000
+      large:
+        memory_limit: 102400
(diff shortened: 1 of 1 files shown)


HenryTK:
Do we have a reason for why Trade Tariff is on a non-standard quota?

alext:
> Do we have a reason for why Trade Tariff is on a non-standard quota?

Historical reasons.  I'll ask @Rory80Hz about that when he gets back.

alext:
I've updated this after a conversation with @Rory80Hz. I've updated the medium definition to be 60Gig, and removed the trade-tariff one.

HenryTK:
@alext:

> Once this is deployed, the quota for the paas_demo org can be updated to use, and then I'll add a temporary task to clean up this quota.

Does exactly the same apply for the Trade Tariff quota?

alext:
> Does exactly the same apply for the Trade Tariff quota?

Yes it does.  I'll take care of that one as well.

PR #370, opened 2016-07-21
[#121564477] Update CF stemcell to 3232.13

saliceti:
## What
Story: [Upgrade stemcells](https://www.pivotaltracker.com/story/show/121564477)

This includes security fixes and we should apply them.
The chosen version matches our BOSH version. BOSH and the stemcell are in the same repository and stemcell updates are applied to different BOSH versions in separate branches.

Separate PRs will be made for BOSH and Concourse.

## How to review
* Deploy to an existing environment
* Check that all tests pass

## Note
I had a short failure in the availability test once but I tried again and it didn't happen. It might happen again, in that case it would be useful to understand at which stage of the upgrade.

## Who can review
Anyone but myself

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -33,7 +33,7 @@ releases:
 stemcells:
   - alias: default
     name: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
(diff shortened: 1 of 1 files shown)


alext:
Closing this because we're getting 404's during the consul re-election on the upgrade. We don't think this is related to this upgrade, but we still don't want to proceed with this until we've fixed this issue.

PR #380, opened 2016-07-22
[#126686741] Remove unneeded org-specific quotas.

alext:
[#126686741 Create digitalmarketplace Org and Users](https://www.pivotaltracker.com/n/projects/1275640/stories/126686741)

## What

Now that we have some standard quotas in place (#368), these org-specific quotas are no
longer needed. This cleans them up from existing environments.

I've manually verified that these quotas are not currently in use by anything on production.

Once this has run everywhere, we'll need a follow-up PR to revert this.

## How to review

Run the deploy against an existing environment that has the `paas_demo` and/or `trade_tariff` quotas defined. Verify that the quotas get removed. Verify that a second run of the cf-deploy job doesn't error.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1395,6 +1395,9 @@ jobs:
 
                   ./paas-cf/concourse/scripts/set_quotas_from_manifest.rb cf-manifest/cf-manifest.yml
 
+                  # FIXME Remove these once they've run on prod.
+                  cf delete-quota -f trade_tariff
+                  cf delete-quota -f paas_demo
 
         - task: sync-admin-users
           config:


dcarley:
Can you explain (in the commit message) what steps have been taking to migrate existing orgs from those old quotas? Are there any caveats, like being able to burst beyond their original quota size if they have apps started under the old quotas and new apps under the new quotas?

alext:
@dcarley I've updated the commit message to explain what's been done to migrate orgs away from these quotas. I don't believe there are any caveats as quotas are applied at an org level, not an app level. Additionally, the standard quotas have identical parameters to these org-specific ones, so there's no change in available resources to these orgs.

keymon:
You did not include link to the story. I have updated the description it to include it.

keymon:
How did you migrate the exiting orgs? Manually? 

If yes, maybe it would be better to write some script to do the migration automatically, and then delete the quotas?

alext:
> How did you migrate the exiting orgs? Manually?

> If yes, maybe it would be better to write some script to do the migration automatically, and then delete the quotas?

I did do it manually (commands detailed in the commit message). It was only 3 orgs, and they weren't all using the same quota, so it didn't seem worth it to script this. Secondly, as we are managing orgs manually at the moment, it seemed odd to automate this one part of their management.

keymon:
We tend to do everything always so automated! 

Anyway, fine for me :)

PR #381, opened 2016-07-22
[#117991263] upgrade concourse to 1.5.1

mtekel:
## What

[Upgrade Deployer Concourse](https://www.pivotaltracker.com/n/projects/1275640/stories/117991263)

## How to review

Run bootstrap pipeline & watch it upgrade your concourse. Run your new concourse to confirm all still works fine. A good test is to run pipeline against existing environment and observe if anything is being changed.

## Merging
Before you merge check if any other PRs have been merged since this PR was opened. If so, check if any changes to pipelines use/introduce `image:` definition of container. If so, rebase on master and convert this to `image_resource` like the rest of container specifications in the pipelines. Note that bootstrap create/destroy and auto-delete pipelines should still use `image:` as they are meant for older bootstrap concourse.

This PR depends on https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/60. Review that as a part of reviewing this PR. Remove 922fec7 after https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/60 is merged.

## Who can review

not @combor @mtekel or @benhyland

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -65,6 +65,22 @@ groups:
     jobs:
       - continuous-smoke-tests
 
+resource_types:
+- name: git-gpg
+  type: docker-image
+  source:
+    repository: governmentpaas/git-resource
+
+- name: s3-iam
+  type: docker-image
+  source:
+    repository: governmentpaas/s3-resource
+
+- name: semver-iam
+  type: docker-image
+  source:
+    repository: governmentpaas/semver-resource
+
 resources:
   - name: pipeline-trigger
     type: semver-iam
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -305,7 +305,10 @@ jobs:
       - task: init-pipeline-pool
         config:
           platform: linux
-          image: docker:///governmentpaas/git-ssh
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/git-ssh
           inputs:
             - name: paas-cf
             - name: git-ssh-private-key
@@ -346,7 +349,10 @@ jobs:
       - task: self-update-pipeline
         config:
           platform: linux
-          image: docker:///governmentpaas/self-update-pipelines
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/self-update-pipelines
           inputs:
             - name: paas-cf
             - name: concourse-manifest
@@ -363,7 +369,10 @@ jobs:
       - task: bootstrap-s3-state
         config:
           platform: linux
-          image: docker:///governmentpaas/curl-ssl
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/curl-ssl
           inputs:
             - name: paas-cf
           run:
@@ -400,7 +409,10 @@ jobs:
       - task: generate-bosh-CA
         config:
           platform: linux
-          image: docker:///governmentpaas/certstrap
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/certstrap
           inputs:
             - name: paas-cf
             - name: bosh-CA
@@ -433,7 +445,11 @@ jobs:
       - task: generate-bosh-secrets
         config:
           platform: linux
-          image: docker:///ruby#2.2-slim
+          image_resource:
+            type: docker-image
+            source:
+              repository: ruby
+              tag: 2.2-slim
           inputs:
             - name: paas-cf
             - name: bosh-secrets
@@ -458,7 +474,10 @@ jobs:
       - task: generate-cf-certs
         config:
           platform: linux
-          image: docker:///governmentpaas/certstrap
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/certstrap
           inputs:
             - name: bosh-CA
             - name: paas-cf
@@ -494,7 +513,11 @@ jobs:
       - task: generate-cf-secrets
         config:
           platform: linux
-          image: docker:///ruby#2.2-slim
+          image_resource:
+            type: docker-image
+            source:
+              repository: ruby
+              tag: 2.2-slim
           inputs:
             - name: paas-cf
             - name: cf-secrets
@@ -541,7 +564,10 @@ jobs:
             params:
               SYSTEM_DNS_ZONE_NAME: {{system_dns_zone_name}}
               CF_ADMIN: admin
-            image: docker:///governmentpaas/cf-cli
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
             run:
               path: sh
               args:
@@ -582,7 +608,10 @@ jobs:
       - task: run-tests
         config:
           platform: linux
-          image: docker:///governmentpaas/cf-acceptance-tests
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/cf-acceptance-tests
           inputs:
             - name: paas-cf
             - name: pipeline-trigger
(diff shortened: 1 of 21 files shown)


jimconner:
It is not possible for outsiders to view the 'What' on this story as we don't all have access to the private tracker used.

mtekel:
That's a pity. It links to an epic story of concourse upgrade where we attempted to upgrade to version 1.0 at first, finding out we couldn't. We made a PR to fix the issues for bosh-init, which were merged to 1.3 concourse, which unfortunately had issue with too many containers. Then we tried 1.4.0 and 1.4.1, which still didn't fix the issue. Finally 1.5.1 seems to work OK. Except we now need to extend fly hijack to be able to specify container type to hijack, so that we don't get multiple container choices when doing bosh-cli...

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1814,14 +1969,17 @@ jobs:
             - -c
             - |
               VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
-              CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.atc.basic_auth_password concourse-manifest/concourse-manifest.yml)
+              CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.properties.basic_auth_password concourse-manifest/concourse-manifest.yml)
               export CONCOURSE_ATC_PASSWORD
               echo Looking for non running VMs
-              ./paas-cf/concourse/scripts/bosh-cli.sh bosh status
+              echo 2 | ./paas-cf/concourse/scripts/bosh-cli.sh bosh status
```
dcarley: Can we specify `fly intercept -c2 …` in `bosh-cli.sh` instead of using `echo` here and documenting in the README?

mtekel: `error: invalid argument for flag `-c, --check' (expected flaghelpers.ResourceFlag): argument format should be <pipeline>/<resource>`

dcarley: Does using `<pipeline>/<resource>` work? Using a name would be more clear and reliable than using a number.

dcarley: If that doesn't work, then it might be `-s2` that we can use.

mtekel: But I don't want to hijack resource, but a job. run-bosh-cli is not a name of resource...

mtekel: -s2 doesn't work. '2' is not a name of step. I only get these options in the fly:
```
[hijack command options]
      -j, --job=PIPELINE/JOB        Name of a job to hijack
      -c, --check=PIPELINE/CHECK    Name of a resource's checking container to hijack
      -b, --build=                  Build number within the job, or global build ID
      -s, --step=                   Name of step to hijack (e.g. build, unit, resource name)
      -a, --attempt=                Attempt number of step to hijack. Can be specified multiple times for nested retries
```
I wish I did not have to do this but I did not find a way to tell fly that I want to hijack a 'task' type container instead of 'check':
```
1: build #20, step: run-bosh-cli, type: check
2: build #20, step: run-bosh-cli, type: task
```
These 2 have same pipeline name, job name and step name. You can't currently specify any further details in fly.

keymon:
Deploys good for me... lets see how it deploys :)

mtekel:
It seems like t2.medium is not fast enough for this version/setup of concourse. I left mine running over the weekend doing nothing. Initially it was regaining credits, but then it stopped next day. Current load is 0.4, which is exactly how much credits that instance gets. Looking at the server, there's not much going on. It is possible that if e.g. there is some issue/change with external dependencies (dockerhub, github), resource checks can consume more CPU. I suggest we play [Change deployer-concourse to use a non-burstable instance type.](https://www.pivotaltracker.com/n/projects/1275640/stories/126234443) story. c4.large looks like a good replacement or m4.large if we need 8G of RAM (for e.g. disk cache).

dcarley:
@mtekel thanks for checking. Is the new version usable on the existing instance type? Will we be able to upgrade first and then change the instance type afterwards, in separate stories?

@bleach any concerns about the cost? Urmi won't be about to ask.

mtekel:
Yes we'll be able to do 2 changes, but it means 2 changes (someone has to manually run bootstrap pipeline in all environments for each of the stories). 

Regarding costs: c4.large is 2x more expensive than t2.medium. m4.large is 2.5x times more expensive. For our ~ 12 environments (dev + ci + prod + stage), the difference per year was about $5k (we checked this with @benhyland on Friday).

alext:
For reference, a t2.medium costs $0.056/hr and a c4.large costs $0.119/hr

keymon:
Can we explain why we cannot use `link()` in 4f236663c15844f8ab18e2f8614897145211b6a7

Something like:

> `bosh_init` does not support [link services as bosh does](https://bosh.io/docs/links.html). Concourse release now relies on links to configure services between jobs, although it can be disabled specifying all the properties. This can break in the future, as concourse does not recommend use `bosh_init` to deploy it.

keymon:
I would squash b0c38ae1566c38e7d7fb7689de747f2a7accb6c4 and23375c703a27daf7cc94c5272dfbd7047c67cd14, and also explain that we need to keep `image` in the job definition until https://github.com/concourse/fly/issues/98 is fixed.

keymon:
The PR seems fine, concourse deploys and the pipeline runs successfully.

We need to clarify if we want to upgrade the instance type or not.

bleach:
I think the cost of upgrading to a c4.large is justified if the current instance size causes our concourses to effectively and the larger size fixes that problem without us spending too much engineer time on it. It seems like this is the case from the comments above.

If this problem is introduced or made much worse by the upgrade then I would opt to fix it during the upgrade; if it does cause problems it will hamper our ability to get anything delivered to production.

bleach:
tl;dr; if the choice is between between paying $5k/ year to upgrade all our concourses or having them grind to a halt every so often, we should pay the $5k

mtekel:
Given the situation (and the fact that it keeps breaking), I'd say changing instance type is not much of a choice. If we care about the costs, we can possibly save more elsewhere (e.g. different VM types in cloud config for dev).

I have updated the comment messages, squashed commits and PR description. 
922fec7 should be removed after https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/60 is merged.

dcarley:
@mtekel thanks for finding a solution to the bosh-cli usage and let's go ahead with the instance type change 👍

mtekel:
half of the credit for the solution goes to @keymon

keymon:
LGTM, merging!

HenryTK:
Now that #364 has been merged there is a reference to `image` used instead of `image_resource`, in the [`update-kibana-timezone` task](https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-bosh-cloudfoundry.yml#L1520), so this branch needs updating before merging.

keymon:
> Now that #364 has been merged there is a reference to image used instead of image_resource, in the update-kibana-timezone task, so this branch needs updating before merging.  

Rebased and updated

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1402,7 +1527,11 @@ jobs:
         - task: sync-admin-users
           config:
             platform: linux
-            image: docker:///ruby#2.2
+            image_resource:
+              type: docker-image
+              source:
+                repository: ruby
+                tag: 2.2-slim
```
alext: This changed the container from `2.2` to `2.2-slim`. This has led to errors running the sync-admin-users script as it can no longer `bundle install` because the slim container doesn't include a compiler.

I'll do a PR to fix this.

PR #386, opened 2016-07-27
[#120829637] Remove some dependencies from scripts/Gemfile

dcarley:
## What

`paas-cf/scripts/Gemfile` was forked from `paas-cf/Gemfile` in d271b43
because at the time `bosh_cli` and `cf-uaac` couldn't be loaded at the same
time due to cloudfoundry/cf-uaac#24. This has now been released but changing
it will require more changes to dependencies and testing.

Some of the dependencies that were copied are not needed by `scripts/`, at
either test-time in Travis or run-time in the pipeline. Removing them will
mean less dependencies to manage and a small speed improvement to how fast
the Travis tests and `sync-admin-users` task take to run.

NB: The referenced story has not been started. I noticed this was slowing me down slightly when working on something else.

## How to review

Foremost:

1. Check that the Travis build still passes.

I have tested that `sync-admin-users` still works. If you also want to confirm it:

1. Remove all users but yourself from `config/admin_users.yml`
1. Commit that change to a branch and push it.
1. Deploy the pipeline with user creation enabled:

  ```
NEW_ACCOUNT_EMAIL_ADDRESS=<your_email> BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipeline
````
1. Run `cf-deploy`:
  ```
make dev run_job JOB=cf-deploy
```

1. Confirm that `sync-admin-users` logs that it's creating users, though it may fail sending email because of SES sandbox:

  ```
Adding bosh-CA to root certificates
Updating certificates in /etc/ssl/certs... 1 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d....done.
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Installing addressable 2.4.0
Installing json_pure 1.8.3
Installing multi_json 1.12.1
Installing cookiejar 0.3.0
Installing eventmachine 1.0.9.1 with native extensions
Installing http_parser.rb 0.6.0 with native extensions
Installing highline 1.6.21
Installing rack 1.5.5
Installing safe_yaml 1.0.4
Installing diff-lcs 1.2.5
Installing mini_portile2 2.1.0
Installing hashdiff 0.3.0
Installing mime-types-data 3.2016.0521
Installing rspec-support 3.4.1
Using bundler 1.12.5
Installing launchy 2.4.3
Installing jmespath 1.2.4
Installing cf-uaa-lib 3.4.0
Installing em-socksify 0.3.1
Installing crack 0.4.3
[...]

PR #388, opened 2016-07-27
[#127073275] Replace tests that use Mimic with Ginkgo/Gomega

dcarley:
#### Create new Ginkgo suite for concourse/scripts

This is the first step at replacing some (and later all) of our tests for
scripts under `concourse/scripts/` to use Ginkgo and Gomega instead of Rspec
and Aruba+Mimic. We're not terribly happy with the current libraries
because:

Aruba magically hooks into Rspec and is quite complex. The way we're using
it through Rspec (instead of Cucumber, which I'd rather avoid) doesn't seem
to be that commonly used, which is reflected in the unclear API and
documentation. Several people in the team have struggled to write new tests
for scripts because of these issues.

Mimic seems the best choice for creating a real HTTP server with stub
responses that is accessed by a different process in Ruby. However we've run
into problems with it not listening on the correct interface, not being able
to stub 404 bodies, not being able to assert on the contents of requests,
and not being able to query the address it's bound to. We have raised PRs
for some of these issues, but it's unclear whether the project is
maintained.

Alex pointed out that Ginkgo and Gomega have some very good libraries for
executing binaries (gexec) and creating HTTP servers (ghttp). Using them
feels far less "magic" than Aruba and Mimic. I think it's a good fit for us
because we also have other tests in Ginkgo, so we're familiar with the
tooling, and using a separate language from the scripts (which are mostly
shell and Ruby) will help enforce blackbox testing (which is most important
for these scripts where we care about the input/output/exit code).

The only oddity is that we'll need to call the scripts with `bundle exec`
because we're no longer running them from within a Bundler session.

I'm focussing on converting the scripts that use Mimic at the moment because
the listen bug is preventing us from running them locally under OS X. If
this approach proves successful then we should write all new tests that
would have used Aruba in Ginkgo instead and slowly transition existing tests
over. I have no intention of changing other Rspec tests which don't use
Aruba. So the focus is primarily on `concourse/scripts/`.

Travis only lets you specify a single language for a build. With most of our
other unit tests still being written in Ruby we should keep that as our
primary language for Travis. Go is available in that container, but it's an
older version of 1.4 which might not be supported by Ginkgo in future (they
dropped 1.3). So I'm installing a later version using the `gimme` tool which
Travis use in Go builds. We also need to install Ginkgo and other
dependencies.

#### Convert bosh_pre_destroy tests to Ginkgo

For the reasons described in the preceding commit which adds a test suite
for the `concourse/scripts/` directory.

The structure and assertions have been copied pretty much like-for-like
which will make it easier to compare whether we are testing the same things.

#### Convert kibana_set_utc tests to Ginkgo

For the reasons described in the preceding commit which adds a test suite
for the `concourse/scripts/` directory.

The structure and assertions have been copied pretty much like-for-like
which will make it easier to compare whether we are testing the same things.
Though please note the following:

- I've split the "if <thing>" and "it <thing>" descriptions out to separate
  `Context` and `It` blocks so that I have somewhere to put the
  `BeforeEach`.
- `Context` blocks could be DRYed at a later date.
- The assertions for `ReceivedRequests()` aren't as valuable now that we're
  asserting the contents of requests and ghttp will assert that undefined
  stubs aren't hit.
- I've added assertions for STDOUT and STDERR content which we didn't have
  before and I believe is important to tell how the script is working.

## How to review

1. Compare the new tests with the previous tests and confirm that they cover the same functionality.
1. Confirm that the Travis build passes.
1. Run the tests locally using `make test` or `make spec`.

I'm also interested in feedback from more than one person about:

1. Are the tests any harder or easier to understand?
1. Are the tests any harder to run locally?

## Who can review

Not @dcarley


alext:
This approach looks good to me. I've run into a couple of issues with the implementation though:

* both tests fail with errors like `bundler: command not found: bosh_pre_destroy.rb`. This can be fixed by replacing the command with `./bosh_pre_destroy.rb`. This feels more correct to me anyway because the scripts directory won't be on the PATH. It seems that newer versions of bundler may have started implicitly looking in the current directory.

* I've intermittently had the following error:
  ```
BoshPreDestroy
/Users/alex/paas/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:94
  no deployments
  /Users/alex/paas/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:74
    should return a zero exit code and no output [It]
    /Users/alex/paas/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:73

    Timed out after 1.001s.

    /Users/alex/paas/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:70
  ```
  I reckon the default timeout (1s) is too short here.

* I get the following error from the `kibana_set_utc` test.
  ```
KibanaSetUtc
/Users/alex/paas/paas-cf/concourse/scripts/kibana_set_utc_test.go:210
  elastic search is not available
  /Users/alex/paas/paas-cf/concourse/scripts/kibana_set_utc_test.go:159
    reports an error [It]
    /Users/alex/paas/paas-cf/concourse/scripts/kibana_set_utc_test.go:158

    Got stuck at:

    Waiting for:
        bundler: failed to load command
  ```
  Again, I reckon this is a string that's only output with newer versions of bundler.

For reference, I'm running bundler 1.11.2.

dcarley:
Looking into the problem described above..

dcarley:
I've rebased with the following changes:
- added a path to all execs
- `kibana_set_utc.rb` is no longer run through `bundle exec`. I think you're right, that there's no need and very little danger if we're not using third-party gems. It speeds up the execution a bit and no longer pollutes STDOUT/STDERR. I've changed the STDOUT/STDERR assertions to account for this. My only concern is how will other people adding new tests know which example to copy?
- `bosh_pre_destroy.rb` now has an execution timeout of 2sec. I can't reproduce the intermittent failure anymore with this change. It does seem these tests are slower though, probably because of Bundler.
- reverted the commit which adds extra debugging to `bosh_pre_destroy.rb` for the problems we were having under Rspec in Travis.

alext:
:+1: Looks good to me now.  I'll leave this for a second reviewer to merge.

dcarley:
PS: I didn't link to the individual Mimic PRs because I didn't want this PR, where we no longer use Mimic, to be linked and affect the likelihood of them being merged.

bleach:
I've not reviewed the coverage, but I have run the tests locally and they all pass on Ubuntu 14.04.

saliceti:
I've hit the timeout issue:

```
------------------------------
• Failure [2.008 seconds]
BoshPreDestroy
/Users/colin/Documents/Boulot/gds/git/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:99
  no deployments
  /Users/colin/Documents/Boulot/gds/git/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:79
    should return a zero exit code and no output [It]
    /Users/colin/Documents/Boulot/gds/git/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:78

    Timed out after 2.000s.

    /Users/colin/Documents/Boulot/gds/git/paas-cf/concourse/scripts/bosh_pre_destroy_test.go:75
------------------------------
•

Summarizing 1 Failure:

[Fail] BoshPreDestroy no deployments [It] should return a zero exit code and no output 
```

The second time it worked fine. Can we increase it more?

dcarley:
> The second time it worked fine. Can we increase it more?

Do you think 3s would be enough?

Out of interest, are you able to reproduce it with `ginkgo -untilItFails`?

saliceti:
`ginkgo -untilItFails` didn't make it fail even after 90 attempts (and 90 funny messages, thanks ginkgo!).
I have deleted my gemset, bundle installed again and ran the test with 3s timeout. It succeeded and took just above 2s: `2.099s`. The second time it took only `0.950s`.
I reckon 3s is fine.

saliceti:
Can we remove aruba from scripts/Gemfile? It doesn't look like it's used there.

dcarley:
Rebased with 3secs.

> Can we remove aruba from scripts/Gemfile? It doesn't look like it's used there.

Do you mean..

- `paas-cf/scripts/Gemfile`? It was removed in #386
- `paas-cf/concourse/scripts/Gemfile`? It's still used by the other tests in `concourse/scripts/spec/` at the moment

saliceti:
OK it's already in master, thanks.
Merging.

PR #390, opened 2016-07-27
[#117991263] Increase disk space&reduce cleanup threshold on deployer Concourse

combor:
## What

We found out that current disk size of Concourse ephemeral disk in combination with `graph_cleanup_threshold_in_mb` property can lead to disk space problems and failure of Concourse pipeline. 

New disk size set to 64GB and `graph_cleanup_threshold_in_mb` reduced to 32GB should allow Concourse to clean space proactively leaving enough headroom for new downloads of containers. 

## How to review

Use bootstrap concourse to provision Deployer Concourse and run create-bosh-cloudfoundry pipeline at least twice to check whether disk space problem will not appear.

## Who can review

Not @alext or @combor

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -28,7 +28,7 @@ resource_pools:
       elbs:
       - (( grab terraform_outputs.concourse_elb_name ))
       ephemeral_disk:
-        size: 50240
+        size: 65536
         type: gp2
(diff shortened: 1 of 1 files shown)


alext:
There's good detail in the "What" for this PR that would be good to include in the commit message.

alext:
We're going to revisit this in more detail, as we've found some other details that need fixing.

PR #393, opened 2016-07-28
[#120100877] Set up pingdom email alerts

saliceti:
## What
Story: [Email Alerts from Pingdom](https://www.pivotaltracker.com/story/show/120100877)

We want to be able to set up Pingdom checks that alert us via email.
The resolution is changed to be alerted more quickly in case of downtime.
The pingdom configuration and build is now documented extensively.

## How to review
* Install the custom provider using the instructions from the [pingdom doc](https://github.com/alphagov/paas-cf/blob/7f0ef19d6437d7594548705bf4a2f5c3c5ae88f2/doc/pingdom.md). Use branches in https://github.com/alphagov/paas-terraform-provider-pingdom/pull/3 and https://github.com/alphagov/paas-go-pingdom/pull/2. Run the make command.
* Connect to pingdom and observe the checks are created and configured correctly
* Delete the checks

## Who can review
Anyone but @HenryTK or myself

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,7 @@ manually_upload_certs: ## Manually upload to AWS the SSL certificates for public
 .PHONY: pingdom
 pingdom: ## Use custom Terraform provider to set up Pingdom check
 	$(eval export PASSWORD_STORE_DIR?=~/.paas-pass)
+	$(eval export PINGDOM_CONTACT_IDS=11089310)
 	@terraform/scripts/set-up-pingdom.sh
 
 merge_pr: ## Merge a PR. Must specify number in a PR=<number> form.
--- a/doc/pingdom.md
+++ b/doc/pingdom.md
@@ -1,0 +1,69 @@
+# Pingdom checks
+
+## Setting up the checks
+
+The checks are defined in terraform configuration file `terraform/pingdom/pingdom.tf`.
+
+The IDs of the contacts to be notified are stored as a comma-delimited string in variable `$PINGDOM_CONTACT_IDS` in the Makefile.
+
+### Requirements
+
+* [Terraform](https://www.terraform.io/downloads.html) must be installed. We currently support version `0.6.*`
+* Make sure you have access to the PaaS credential store, this is required for Pingdom credentials.
+* Load the AWS credentials for the environment you are setting up checks for. These are required as the Terraform state file for the Pingdom checks is stored in an S3 bucket.
+
+### Usage
+Run `make <ENV> pingdom` to set up the Pingdom checks.
+
+
+## Build from source
+
+#### Requirements
+You must have a [golang environment](https://golang.org/doc/install) configured. We currently test with Go 1.6.
+ 
+#### Download
+Download the provider and its dependencies:
+
+```
+go get github.com/russellcardullo/terraform-provider-pingdom
+```
+
+#### Custom provider
+For bleeding-edge features, it may be necessary to build the provider from our fork.
+
+```
+# go-pingdom library
+cd $GOPATH/src/github.com/russellcardullo/go-pingdom
+git remote add alphagov https://github.com/russellcardullo/go-pingdom.git
+
+# terraform-provider-pingdom
+cd $GOPATH/src/github.com/russellcardullo/terraform-provider-pingdom
+git remote add alphagov https://github.com/russellcardullo/terraform-provider-pingdom.git
+```
+
+#### Terraform version
+You may need to ensure your Terraform version is compatible with the terraform library used to compile the provider. You can run `terraform -v` to get your version of Terraform and find the corresponding git tag for this version in `$GOPATH/src/github.com/hashicorp/terraform`. Use the tag to checkout that version of the terraform library prior to installing the provider.
+
+#### Install
+Run `go install github.com/russellcardullo/terraform-provider-pingdom`. This will build and install the binary in `$GOPATH/bin`. Make sure `$GOPATH/bin` is in your `$PATH`.
+
+Add the content of `terraform/providers/terraformrc` to `$HOME/.terraformrc`.
+
+The binary should now be installed and Terraform knows where to find it.
+
+## Publishing the custom provider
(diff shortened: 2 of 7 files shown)


doc/pingdom.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Run
+Run `make <ENV> pingdom` to set up the Pingdom checks.
+
+## Use the binary
+Binaries for Linux and MacOS are available in our [paas-terraform-provider-pingdom repository](https://github.com/alphagov/paas-terraform-provider-pingdom/releases).
+
+Download and copy to a directory in your `$PATH`.
+
+## Build from source
```
dcarley: Could you make it a bit clearer that you would only want to do the following steps if you were developing changes to the provider or library? We want to guide people towards the pre-built binaries in normal circumstances, otherwise there's a chance that they'll waste time or use an incompatible version.

doc/pingdom.md
```diff
@@ -0,0 +1,76 @@
...
+#### Configure
+Configure the IDs of the contacts to be notified as a comma-delimited string in variable PINGDOM_CONTACT_IDS in the Makefile.
+
+#### Run
+Run `make <ENV> pingdom` to set up the Pingdom checks.
+
+## Use the binary
+Binaries for Linux and MacOS are available in our [paas-terraform-provider-pingdom repository](https://github.com/alphagov/paas-terraform-provider-pingdom/releases).
+
+Download and copy to a directory in your `$PATH`.
```
dcarley: Do you also need to modify `~/.terraformrc`? It's only mentioned under the "build from source" section.

doc/pingdom.md
```diff
@@ -0,0 +1,76 @@
...
+### Usage
+
+#### Configure
+Configure the IDs of the contacts to be notified as a comma-delimited string in variable PINGDOM_CONTACT_IDS in the Makefile.
+
+#### Run
+Run `make <ENV> pingdom` to set up the Pingdom checks.
+
+## Use the binary
+Binaries for Linux and MacOS are available in our [paas-terraform-provider-pingdom repository](https://github.com/alphagov/paas-terraform-provider-pingdom/releases).
```
dcarley: Would it be preferable for the script to download these for you automatically? There's quite a lot of documentation here and I feel like it might confuse someone that wants to deploy a change, rather than develop changes.

doc/pingdom.md
keymon: +1

doc/pingdom.md
keymon: The problem with download automatically is the `.terraformrc` file. I wouldn't like this script to change files in my `$HOME`.

But one solution can be download the binary to a tmp dir, copy the `.terraformrc` there and change the `$PATH` and `$HOME` in the script:

BTW, the binaries should be suffixed with the output of ` $(uname -s)-$(uname -a)`

``` 
mkdir /tmp/terraform-pingdom
wget https://github.com/alphagov/paas-terraform-provider-pingdom/releases/download/0.2.1/terraform-provider-pingdom- $(uname -s)-$(uname -a) -O  /tmp/terraform-pingdom/terraform-provider-pingdom
cp terraform/providers/.terraformrc /tmp/terraform-pingdom/

PATH=$PATH:/tmp/terraform-pingdom HOME=/tmp/terraform-pingdom terraform  destroy  -var pingdom_user=$(paas-pass pingdom.com/username) -var pingdom_password=$(paas-pass pingdom.com/password) -var pingdom_api_key=$(paas-pass pingdom.com/api_key) -var pingdom_account_email=$(paas-pass pingdom.com/account_email) -var apps_dns_zone_name=example.com -var env=hector
```

Makefile
```diff
@@ -141,6 +141,7 @@ manually_upload_certs: ## Manually upload to AWS the SSL certificates for public
 .PHONY: pingdom
 pingdom: ## Use custom Terraform provider to set up Pingdom check
 	$(eval export PASSWORD_STORE_DIR?=~/.paas-pass)
+	$(eval export PINGDOM_CONTACT_IDS=11089310)
```
keymon: Can we say in the commit message which contact is this? or add the link https://my.pingdom.com/contacts#contact=11089310

doc/pingdom.md
dcarley: I think the previous approach, of warning that you have to create `~/.terraformrc` manually, is OK.

Or your suggestion might work. Not sure if there are any other consequences of overriding `$HOME`?

dcarley:
Could you split the `resolution` and `sendnotificationwhendown` changes to a separate commit? They're important enough to highlight independently from setting the email contact.

The last three commits could be squashed into the first commit which changes the provider. There isn't enough detail in them as they are and they are closely related to the first change - if somebody looked at them in isolation they might wonder why the changes were made.

keymon:
The code works fine, but I vote for automate the download of the binaries. Also fix the comments above

doc/pingdom.md
keymon: My suggestion works, I tested it. overriding `$HOME` for **only terraform** seems OK for these .tf files. Other providers might have problems, but currently is OK.

terraform/scripts/set-up-pingdom.sh
```diff
@@ -9,26 +12,31 @@ PINGDOM_PASSWORD=$(pass pingdom.com/password)
...
-terraform apply -state="${STATEFILE}" \
+# Configure Terraform remote state
+terraform remote config \
+    -backend=s3 \
+    -backend-config="bucket=${DEPLOY_ENV}-state" \
+    -backend-config="key=${STATEFILE}" \
+    -backend-config="region=${AWS_DEFAULT_REGION}"
+
+# Run Terraform Pingdom Provider. We change $HOME so Terraform can find terraformrc
+HOME=/tmp/terraform-pingdom \
```
keymon: I think you have to also add PATH=$PATH:/tmp/terraform-pingdom

saliceti: You don't need to as I've set the absolute path in terraformrc.

keymon:
It works fine, at least in linux :)

PR #394, opened 2016-07-29
[#127256345] Fix issues with concourse running out of disk space

alext:
## What

We've been running into issues where the concourse machine will grind to a halt. Investigation lead to baggageclaim running into 'no space left on device' errors form it's loopback mounted volumes filesystem. This was created as a sparse file set to the full size of the `/var/vcap/data` filesystem. This causes problems because this filesystem is also used by other things (bosh packages and jobs, garden's graph and depot, logs etc). When this volume starts to get full, this can lead to errors when the btrfs driver gets a `no space left on device` error when writing to the sparse image file. This in turn causes a kernel error, and the loopback filesystem gets remounted read-only.

This PR does a few things to address this:

* Increase the ephemeral disk size to give more headroom
* Pull in a forked concourse release that includes https://github.com/alphagov/paas-concourse/pull/2 to make the sparse image not use 100% of the filesystem.

This also updates the instructions for SSH access to the concourse VM to make it a little easier.

## How to review

Code review of this and https://github.com/alphagov/paas-concourse/pull/2

Run the bootstrap pipeline from this branch.

Verify that the deployer concourse is working by running the pipeline a couple of times.

SSH onto the deployer concourse (using instructions in the README), and verify that the `/var/vcap/data/baggageclaim/volumes.img` file is less than the size of the filesystem.

## Before merging

Before this is merged, the [concourse PR](https://github.com/alphagov/paas-concourse/pull/2) will need to be merged, and the resulting boshrelease generated and uploaded to a github release. The last commit in this PR will then need to be updated to point at that release.

## Who can review

Anyone but @mtekel or myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,11 +292,10 @@ You can get both from command line. You will need [aws-cli](#aws-cli), to do thi
 ```
 eval $(make dev showenv | grep CONCOURSE_IP=)
 
-aws s3 cp "s3://${DEPLOY_ENV}-state/id_rsa" . && \
-chmod 400 id_rsa && \
-ssh-add $(pwd)/id_rsa
+aws s3 cp "s3://${DEPLOY_ENV}-state/id_rsa" ./deployer_id_rsa && \
+chmod 400 deployer_id_rsa
 
-ssh vcap@$CONCOURSE_IP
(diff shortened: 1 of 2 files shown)


alext:
I've updated the last commit to point at the github release now that the concourse PR has been merged.

keymon:
+1

PR #397, opened 2016-07-29
[#120100877] Fix terraform state issue

saliceti:
## What

Story: [Email Alerts from Pingdom](https://www.pivotaltracker.com/story/show/120100877)

Previously we used to download the state file from the bucket, reupload it after terraform run, then delete the local copy.

b01f941 changed this to use terraform built-in S3 storage. Unfortunately it still kept a local copy of the state file, and this confused terraform when we applied changes to different environments.

This PR deletes the local copy to avoid this issue and reverts the gitgnore change related to the state file.

## How to review
* In case of an issue, a terraform pingdom run in one environment was reconfiguring alerts from the previous environment.
* Run `make <env> pingdom` successively in different environments for example dev, ci, staging. It shouldn't do anything. 

## Who can review
Anyone but me

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 vagrant/.vagrant/
 bin/fly*
 .idea/
-.terraform/
--- a/terraform/scripts/set-up-pingdom.sh
+++ b/terraform/scripts/set-up-pingdom.sh
@@ -40,3 +40,6 @@ terraform apply \
 
 # Delete temporary directory
 rm -rf /tmp/terraform-pingdom
+
+# Delete local terraform state
+rm -rf .terraform


dcarley:
> this confused terraform when we applied changes to different environments

From meatspace: Colin's going to add more detail about what the effect of this was.

saliceti:
Commit message updated.

dcarley:
Tested against CI. The `url` attribute changes every time - is that expected?

Should we explicitly set it to `/`? There might also be an upstream issue, but I'm not sure how Terraform behaves with optional defaults.

First run:
```
Remote state management enabled
pingdom_check.paas_http_healthcheck: Refreshing state... (ID: 2247550)
pingdom_check.paas_db_healthcheck: Refreshing state... (ID: 2247551)
pingdom_check.paas_http_healthcheck: Modifying...
  url: "/" => ""
pingdom_check.paas_http_healthcheck: Modifications complete

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

Second run:
```
pingdom_check.paas_http_healthcheck: Refreshing state... (ID: 2247550)
pingdom_check.paas_db_healthcheck: Refreshing state... (ID: 2247551)
pingdom_check.paas_http_healthcheck: Modifying...
  url: "/" => ""
pingdom_check.paas_http_healthcheck: Modifications complete

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

saliceti:
Yes I've noticed the issue with the url as well. It's not related to this PR.
The terraform provider doesn't have a default for url, so it sends an empty string to the API. The API defaults to "/" and returns "/" when terraform GETs the resources. Terraform thinks the resources has changed so it tries to reapply the empty string.

We can workaround it by specifying "/" in our config. A better solution would be to set "/" as default in terraform.

dcarley:
I'll raise a separate PR.

Tested in prod too 👍

PR #404, opened 2016-08-03
[120765225] Disable healthcheck DB on DEV

richardc:
## What

Disable healthcheck DB on DEV

This commit disables provisioning of healthcheck DB on DEV environments to
save space,time and money :). It keeps it enabled on non-dev envs.

Changes consists of default setting in Makefile, that is passed via
pipeline upload script and new condition in deploy-healthcheck to test it.

## How to review

Deploy pipeline - during running of post-deploy.deploy-healthcheck note the new order of operations doesn't include provisioning a `healthcheck-db` service.

## Who can review

Cannot be reviewed by @combor or @richardc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
 	$(eval export SKIP_COMMIT_VERIFICATION=true)
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
+	$(eval export ENABLE_HEALTHCHECK_DB=false)
 	@true
 
 .PHONY: ci
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1762,6 +1762,8 @@ jobs:
             type: docker-image
             source:
               repository: governmentpaas/cf-cli
+          params:
+            ENABLE_HEALTHCHECK_DB: {{enable_healthcheck_db}}
           inputs:
             - name: paas-cf
             - name: config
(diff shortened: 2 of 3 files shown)


timmow:
What is the impact of this change on existing dev environments that have a healthcheck db? Do we need to remove the db? Will things still work when destroying an environment?

dcarley:
> What is the impact of this change on existing dev environments that have a healthcheck db? Do we need to remove the db? Will things still work when destroying an environment?

Good question and we should add the answer to the commit message.

We discussed in kick-off that the task which removes the healthcheck DB in `autodelete-cloudfoundry` and `destroy-cloudfoundry` would be sufficient in cleaning up existing environments. It looks like the work we already did to make it resilient to the DB creation failing will allow it to continue working, but I'll double check when testing.

richardc:
Your environment would be left with an orphaned service.  We could be more aggressive here and add an `else` clause that cleans up the service when ENABLE_HEALTHCHECK_DB is false.  Something like this commit:  1fe9781e40b26c0790b213241638b48ee35b55cf

keymon:
Can you document this in the README.md

alext:
> Your environment would be left with an orphaned service. We could be more aggressive here and add an else clause that cleans up the service when ENABLE_HEALTHCHECK_DB is false.

I'm not sure that's true, because the destroy pipelines attempt to delete the healthcheck DB if it's present. These are unchanged by this PR, so would continue to work as before.

richardc:
Commit message amended to reflect what will happen in an existing development environment.

dcarley:
I've tested against an environment that was autodeleted last night. `post-deploy` didn't create the service instance and it took 2mins instead of 11mins. Running `destroy-cloudfoundry` afterwards (which calls `remove-healthcheck-db`) worked.

Due to time constraints, I haven't tested the scenario where you have an existing service instance and `autodelete-cloudfoundry` should delete it, but based on the above and my reading of the code I'm pretty confident that it will work as expected.

Updated commit looks good 👍  Will merge when Travis has passed.

dcarley:
PS: Good PR number!

PR #413, opened 2016-08-09
Avoid unnecessary godep restore in platform-tests

alext:
## What

The availability tests have vendored dependenvies, so there's no need to
run `godep restore`. Not doing this will reduce the startup time for
this suite.

`godep restore` was previously run in the global `run_tests.sh` script.
This moves it into the run scripts for the individual suites that need it.

## How to review

Run the pipeline, and verify that all the tests start successfully.

## Who can review

Anyone but myself.


mtekel:
works correctly, merging

PR #416, opened 2016-08-10
[#127449817] Convert val_from_yaml tests to Ginkgo

dcarley:
## What

So that we can remove Aruba as described in 7f79755.

The significant differences between the old and new tests are:

- assertions on STDOUT *and* STDERR in each test because they are both
  important when the script is consumed in the pipeline
- re-worded test descriptions because they needed to be split into `Context`
  and `It` so that I had somewhere to assign variables

I did experiment with using a function like `run_with_fixture` within each
`It` block that took the key name and input file. However I settled on using
variables and a combination of `BeforeEach` and `JustBeforeEach` because I
think having named variables makes the test easier to read.

## How to review

1. Check that the Travis tests pass.
1. Optionally, run the tests locally with `make test`.

## Who can review

@keymon because he is working the rest of the referenced story.


keymon:
LGTM

PR #421, opened 2016-08-12
[#128084861] Add a max length check for DEPLOY_ENV to the makefile

benhyland:
## What

We have a length restriction and a restriction on characters
used in the DEPLOY_ENV variable, because we use this
variable to construct values used in AWS which themselves
have those restrictions.

To be more humane, we detect these cases and apologise to the
user with the unlucky name :/

We also remove the mention in the readme - it's better for
the makefile to be self-documenting since then we don't get out
of sync with the docs.

## How to review

Run `make check-env-vars` and play with your `DEPLOY_ENV`
var length and character composition,

It would be good to get this reviewed by someone with a non-bash shell.

## Who can review

Anyone except @benhyland or @paroxp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,14 @@ help:
 SHELLCHECK=shellcheck
 YAMLLINT=yamllint
 
+DEPLOY_ENV_MAX_LENGTH=15
+DEPLOY_ENV_VALID_LENGTH=$(shell if [ $${\#DEPLOY_ENV} -gt $(DEPLOY_ENV_MAX_LENGTH) ]; then echo ""; else echo "OK"; fi)
+DEPLOY_ENV_VALID_CHARS=$(shell if echo $(DEPLOY_ENV) | grep -q '^[a-zA-Z0-9-]*$$'; then echo "OK"; else echo ""; fi)
+
 check-env-vars:
 	$(if ${DEPLOY_ENV},,$(error Must pass DEPLOY_ENV=<name>))
+	$(if ${DEPLOY_ENV_VALID_LENGTH},,$(error Sorry, DEPLOY_ENV ($(DEPLOY_ENV)) has a max length of $(DEPLOY_ENV_MAX_LENGTH), otherwise derived names will be too long))
+	$(if ${DEPLOY_ENV_VALID_CHARS},,$(error Sorry, DEPLOY_ENV ($(DEPLOY_ENV)) must use only alphanumeric chars and hyphens, otherwise derived names will be malformatted))
 
 test: spec lint_yaml lint_terraform lint_shellcheck lint_concourse lint_ruby ## Run linting tests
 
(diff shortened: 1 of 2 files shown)


benhyland:
we got the branch name wrong :/ oh well.

README.md
```diff
@@ -65,7 +65,7 @@ state files).
 [instance profiles]: http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html
 [aws-account-wide-terraform]: https://github.gds/government-paas/aws-account-wide-terraform
 
-* Declare you environment name using the variable DEPLOY_ENV. It must be 18 characters maximum and contain only alphanumeric characters and hyphens.
+* Declare your environment name using the variable DEPLOY_ENV. It must be 15 characters maximum and contain only alphanumeric characters and hyphens.
```
dcarley: Should we remove this now that it's enforced in code? There's already a chance that people will skip over this because the README is so long and it's probable that we'd forget to update both numbers if the limit changes in future.

mtekel: I think it should be documented. I don't like finding about limits only when I hit them. When known upfront, I can work around them...

benhyland: We did think about this, and decided to go the easy route on the grounds that it's easy to assume the code is correct and to fix the docs if they ever get out of date.
I'd prefer to remove the 2 sources of truth for the value, but I wasn't certain that the makefile is the only user of the variable.
Happy to update the readme again, if you have a better steer?

Makefile
```diff
@@ -8,8 +8,12 @@ help:
 SHELLCHECK=shellcheck
 YAMLLINT=yamllint
 
+DEPLOY_ENV_MAX_LENGTH=15
+DEPLOY_ENV_VALID=$(shell if [ $${\#DEPLOY_ENV} -gt $(DEPLOY_ENV_MAX_LENGTH) ]; then echo ""; else echo "OK"; fi)
+
 check-env-vars:
 	$(if ${DEPLOY_ENV},,$(error Must pass DEPLOY_ENV=<name>))
+	$(if ${DEPLOY_ENV_VALID},,$(error Sorry, DEPLOY_ENV has a max length of 15, otherwise derived names will be too long))
```
alext: Instead of hardcoding '15' here, can it reference the above variable, otherwise this error message could get out of sync.

benhyland: good catch thanks

README.md
dcarley: Having looked at it again, I've realised that the README also says that you can only use "alphanumeric characters and hyphens". So we can't remove it as-is.

If you could extend the `Makefile` without too much difficulty to also check those, which would be beneficial, then I'd like the README entry removed. If it turns out to be too difficult though then don't worry and retain the README entry.

benhyland: Done - I think in a portable way, but might be as well for those with non-bash shells to check.

dcarley:
Looks good and works, thanks 👍

PR #422, opened 2016-08-12
[#128217001] Scale up the instance type of concourse

keymon:
https://www.pivotaltracker.com/story/show/128217001

What
----

We were using c4.large: 2x CPU, 3.5GB, ~$0.105 per Hour.
But we are having trouble with concourse consuming too much memory
and eventually collapses.

m4.large 2x CPU and 8GB is $0.12 per Hour, so it is a worth it to
upgrade.

How to review
----------

Deploy concourse as described in readme. Check VM type.

Who?
----

Anyone but @keymon


mtekel:
Indeed instance type changes to m4.large after deployment.

PR #423, opened 2016-08-12
[#128217001] Upgrade to stemcell 3262.5 with kernel 4.4

keymon:
https://www.pivotaltracker.com/story/show/128217001

What?
===

The new stemcell with kernel 4.4 will bring several performance improvements,  specially in the btrfs driver.

In the concourse channel the recommend upgrade, and they claim that they were running it for a while without issues.

Note that this image had issues with garden-linux, but not with garden-runc

How to review?
-----------

Deploy concourse. SSH and check the kernel version

Who?
---
anyone but @keymon

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -19,8 +19,8 @@ resource_pools:
   - name: concourse
     network: concourse
     stemcell:
-      url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3232.13
-      sha1: 477371a335d172f4728c7a8806448edb9703104c
+      url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3262.5
+      sha1: 13c73b82675742012ba3ded6528524b8f3ce9ec3
     cloud_properties:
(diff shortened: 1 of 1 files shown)


mtekel:
I had some initial troubles when testing this, but it seems to be related to rebuilding concourse in general (we lose ephemeral disks, but persist data about their contents in DB on persistent disk). After cleaning up DBs (see https://www.pivotaltracker.com/n/projects/1275640/stories/127256345/comments/145903969), my concourse started to behave properly...

keymon:
Yes, I confirm that I had issues with all the environments after deploying, without this change.

I think it  is again the issue https://github.com/concourse/concourse/issues/565 

The workaround, again:

```
for i in atc beacon tsa baggageclaim garden ; do monit stop $i; done
pkill -f runc
umount /var/vcap/data/baggageclaim/volumes /var/vcap/data/garden/graph
losetup -d /dev/loop1
rm -rf  /var/vcap/data/garden
rm /var/vcap/data/baggageclaim/volumes.img

/var/vcap/packages/postgresql_9.3/bin/psql -U atc -h 127.0.0.1 <<EOF
delete from volumes;
delete from image_resource_versions;
delete from containers;
EOF

for i in atc beacon tsa baggageclaim garden ; do monit start $i; done
```

PR #425, opened 2016-08-15
[#127339659] Stop `cat`ing bosh manifest in pipeline

Jonty:
## What

Currently we're `cat`ing the bosh manifest to STDOUT during the pipeline, which means we're accidentally logging all the credentials in the concourse logs. Which is bad.

This removes the :crying_cat_face:

After this is deployed we need to rotate all bosh credentials.

## How to review

Run the pipeline, observe that you no longer have a bosh manifest scrolling up your screen during the `render-bosh-manifest` task.

## Who can review

Anyone but @jonty.


mtekel:
I have to say that I have found this cat useful once. So it served it's purpose at least a bit.

Jonty:
@mtekel Me too, I did wonder about leaving it in for development environments.

dcarley:
> @mtekel Me too, I did wonder about leaving it in for development environments.

I don't think it's worth the danger and complexity. You can fetch it from the state bucket if required.

PR #426, opened 2016-08-16
[#127449817] Convert tag_release.sh test to Ginkgo

keymon:
[#127449817 Convert tests in concourse/scripts/ to Ginkgo + Gomega](https://www.pivotaltracker.com/story/show/127449817)

What
====

We want to convert all the tests to Ginkgo so that we can remove Aruba as described in 7f79755.

The new script uses ginkgo gexec library, and a bunch of helper methods has
been implemented to execute external commands. The most important is
`bashScript` and `runBashScript`, which will execute bash with the flags
`-e`, fail on error and `-u` fail on unassigned variable, and will check if
the command finished successfully.

This test, as the original one in rspec,  relies on executing external
scripts to do git operations on the repos, and check the resulting status.

The test will create a temporary directory.
 * `$HOME` will be set to such directory, as it is required to change the
global config of git for the tag_release.sh.
 * All required files must be copied there, including the script to test.
 * All the scripts will run in that directory.

How to review
------------

 * All the tests in travis should pass.
 * install golang 1.6 and [ginkgo](https://onsi.github.io/ginkgo/) and run
   the tests: `cd concourse/scripts; go test -ginkgo.v tag_release_test.go`
 * Check the code

Who?
----

Anyone but @keymon


dcarley:
Not the reviewer, but..

Could you add another commit to remove the Aruba dependencies and supporting code, now that all of the tests have been converted. We'll still need to keep in this dir because there are unit tests for some of the other scripts.

keymon:
sure

concourse/scripts/tag_release_test.go
```diff
@@ -0,0 +1,219 @@
...
+	})
+
+	AfterEach(func() {
+		cleanUpWorkingDirectory(workingDirectory)
+	})
+
+	Context("when there are multiple 'previous-*' tags to be promoted", func() {
+		BeforeEach(func() {
+			runBashScript(baseCmd, `
+        cd origin_repo
```
richardc: The indenting of these inline bash scripts seem a little off, making this hard to read.

In the original the code of the bash scripts are indented at one level lower than the surrounding code, so that would make this:

```
                BeforeEach(func() {
                        runBashScript(baseCmd, `
                                cd origin_repo

                                git commit --allow-empty -m "First commit"

                                git tag previous-0.0.1
                                git tag previous-0.0.2

                                git commit --allow-empty -m "Second commit"
                                git tag previous-0.0.3

                                # cannot push to repo while its checked out to a branch
                                git checkout refs/heads/master
                        `)
                })
```

keymon: Yeah, it is strange, in theory I am using spaces only and my vim dsplays it OK. I will check.

richardc: Mechanically transformed it to see what it'd look like - https://github.com/alphagov/paas-cf/blob/6b94a6f71fea199fdf62997c43371c8ac32074a7/concourse/scripts/tag_release_test.go

keymon: that is fixed. The issue is that gofmt uses tabs to indent, and so does my vim plugin for go. I got spaces for any other config.  gofmt does not indent strings, ofc.

I replaced the spaces in the strings with tabs.

richardc:
The test can't be run natively on OSX as it's cp(1) implements -a as `-pPR`, so it errors out on copying the script, as -r and -R are mutually exclusive:

```
$ cp -arf foo bar
cp: the -R and -r options may not be specified together.
```

keymon:
@richardc can you try now?

richardc:
@keymon That worked.  Now waiting on travis.

keymon:
OK. I used a shell out to bash to copy those files because implement a equivalent logic in golang is quite complicated, several lines of code that are not really readable. Other option is add a dependency. But as this script shells out everywhere anyway, I though that a `cp` would be good enough.

PR #428, opened 2016-08-17
[#127161593] Retitle CPU graphs

bleach:
## What

The title and axis label of the SPU graph still referred to jiffies and time. 

In d64ddcf and d135733 we switched to storing rates of jiffies and displaying
in grafana as the sum of the percentage use across all cores. For example, on
a 2 core machine the usage will be between 0 and 200%.

This commit tries to make this clear in the VM resource usage dashboard CPU
graph.

## How to review

- Either apply this to a dev environment OR take a copy of `manifests/cf-manifest/grafana/vm-resource-usage.json`, change the title and upload it to a grafana using the UI. 
- Check that the new title and axes label make sense
- Delete any temporary dashboard you created

## Who can review

Anyone but @bleach

### DIFF
--- a/manifests/cf-manifest/grafana/vm-resource-usage.json
+++ b/manifests/cf-manifest/grafana/vm-resource-usage.json
@@ -79,7 +79,7 @@
           ],
           "timeFrom": null,
           "timeShift": null,
-          "title": "CPU time",
+          "title": "CPU usage across all cores",
           "tooltip": {
             "msResolution": false,
             "shared": true,
@@ -92,7 +92,7 @@
           "yaxes": [
             {
               "format": "none",
-              "label": "Jiffies",
+              "label": "CPU % used",
               "logBase": 1,
               "max": null,
               "min": 0,


mtekel:
Did full deployment via BOSH and I can see the updated dashboard, which is now correctly labelled. Merging....

alext:
This change isn't correct. It's not a `%` graph, it's still the sum of Jiffies. As noted in the second commit referenced above (`d135733 `):

> The CPU graph reports values in Jiffies. We looked at configuring the CPU
plugin to report percentage values as described here -
https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_cpu.
The version of collectd that we're currently using (5.4.1) doesn't include
this feature, so we'd need to upgrade. We decided to leave this for a
later story.

So we haven't updated it to graph in percentage of all cores because that requires a newer version of collectd.

mtekel:
I am not sure the documentation is 100% exact. If you look at what collectd returns currently, it's a number between 0 and 100 for each CPU core and each of CPU usage types (user, sys, iowait, ...). I think it might have st. to do with https://github.com/alphagov/paas-cf/commit/d135733ed3eaf566387348d6fd49d58f7ae65c5d . Looks like it never reported metrics in jiffies?

alext:
Ok, so I've done some more digging:

From looking at the [collectd code](https://github.com/collectd/collectd/blob/collectd-5.4.1/src/cpu.c#L351-L407), this is just reporting the raw values from `/proc/stat`, which are in Jiffies.

However, it turns out that since 2.6 kernels, Jiffies exposed to userspace are always scaled to 100/sec (http://stackoverflow.com/a/32181955/983841). There are therefore c. 100 jiffies per second, so these numbers roughly correspond to the percentage used in each second. This is good enough for our needs.

PR #430, opened 2016-08-17
[#126965751] Test users not removed from CC DB

richardc:
## What

[#126965751](https://www.pivotaltracker.com/n/projects/1275640/stories/126965751) Test users not removed from CC DB

This series reworks the existing delete_admin task to be implemented in terms of `cf remove-user` rather than `uaac remove user` to prevent test users leaking, and then adds extra steps to the task to delete any users that match the test user prefix for this job.

## Dependencies and requirements

The commits marked NOKEEP should also be either dropped or reverted before merging, as they deliberately point at in-flight forks, and nobble the running of smoke-tests so we just test the user creation/deletion cycle.

You should review first  https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/61 and https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/62 to add jq to the cf-cli image.

> note: remember install all dependencies: `uaac` with `bundle install`, `awscli`, export AWS credentials. See README.md for details. Also install `jq`  https://stedolan.github.io/jq/

## How to review

Deploy pipeline to a dev environment using the master branch. We recommend disable the payload of the `smoket-tests` by adding a exit 0 as in https://github.com/alphagov/paas-cf/pull/430/commits/bf57c75b402c9aa44ecb1a1b97dc175d09924e35.

Run the smoke-tests several times. That will create several temp users, and delete them in uaa but not in CF. You can see the users created in uaa and CF by running.

note: remember install all dependencies: `uaac` with `bundle install`, `awscli`, export AWS credentials. See README.md for details. Also install `jq`  https://stedolan.github.io/jq/

```
export DEPLOY_ENV=....
UAA_PASS=$(aws --region eu-west-1 s3 cp s3://${DEPLOY_ENV}-state/cf-secrets.yml - | grep 

# Login into UAA 
uaac target https://uaa.${DEPLOY_ENV}.dev.cloudpipeline.digital --skip-ssl-validation
UAA_PASS=$(aws --region eu-west-1 s3 cp s3://${DEPLOY_ENV}-state/cf-secrets.yml - | grep uaa_admin_client_secret | awk '{print $2}')
uaac token client get admin -s $UAA_PASS
# Get users
uaac users | grep smoketest-user

# Login in CF
CF_ADMIN_PASS=$(aws --region eu-west-1 s3 cp s3://${DEPLOY_ENV}-state/cf-secrets.yml - | grep uaa_admin_password | awk '{print $2}')
cf login --skip-ssl-validation -a https://api.${DEPLOY_ENV}.dev.cloudpipeline.digital -u admin -p  ${CF_ADMIN_PASS}
cf curl /v2/users | jq -r ".resources[] | .entity.username | select(. != null)" | grep smoketest-user
```

You will see that there are users in CF that are not in UAA.

Now run this branch of the pipelines:

```
```

Run the `smoke-tests` job as it will complete quickly (thx to the disable commit). You can see that the users are now in sync, and additional users in UAA are deleted.

Additionally, you should be able to observe the number of cc users grows and then drops from the user impact dashboard in grafana (previous behaviour would have been only to grow).  

 

## Who can review

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1817,6 +1817,8 @@ jobs:
         ensure:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
+          params:
+            PREFIX: smoketest-user
 
   - name: acceptance-tests
     plan:
@@ -1922,6 +1924,7 @@ jobs:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
           params:
+            PREFIX: acceptance-test-user
             ENABLE_ADMIN_USER_CREATION: {{enable_cf_acceptance_tests}}
 
   - name: custom-acceptance-tests
@@ -1973,6 +1976,8 @@ jobs:
         ensure:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
+          params:
+            PREFIX: custom-acceptance-test-user
 
   - name: bosh-tests
     plan:
@@ -2081,6 +2086,8 @@ jobs:
         ensure:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
+          params:
+            PREFIX: performance-tests-user
 
   - name: tag-release
     plan:
@@ -2290,6 +2297,8 @@ jobs:
         ensure:
           task: remove-temp-user
           file: paas-cf/concourse/tasks/delete_admin.yml
+          params:
+            PREFIX: cont-smoketest-user
 
   - name: bosh-cli
     plan:
--- a/concourse/tasks/delete_admin.yml
+++ b/concourse/tasks/delete_admin.yml
@@ -34,3 +34,16 @@ run:
       echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
 
       cf delete-user "${NAME}" -f
+
+      echo "Removing leaked ${PREFIX} users"
+      url="/v2/users"
+      while [ "${url}" != "null" ] ; do
+        cf curl $url > users.json
+        # Delete users
+        cat users.json \
+           | jq -r ".resources[] | .entity.username | select(. != null) | select(test(\"^${PREFIX}\"))" \
+           | xargs -r -I {} cf delete-user {} -f
+
+        # See if there's another page of data.  If "next_url" is null we can stop
+        url=$(cat users.json | jq -r ".next_url")
+      done
--- a/concourse/tasks/delete_admin.yml
+++ b/concourse/tasks/delete_admin.yml
@@ -3,7 +3,7 @@ platform: linux
 image_resource:
   type: docker-image
   source:
-    repository: governmentpaas/cf-uaac
+    repository: governmentpaas/cf-cli
 inputs:
   - name: paas-cf
   - name: cf-secrets
@@ -14,21 +14,23 @@ run:
   path: sh
   args:
     - -e
+    - -u
     - -c
     - |
       NAME=$(cat admin-creds/username)
-      if [ "${ENABLE_ADMIN_USER_CREATION}" == "false" ]; then
+      if [ "${ENABLE_ADMIN_USER_CREATION:-}" == "false" ]; then
         echo "Temporary user creation is disabled (ENABLE_ADMIN_USER_CREATION=${ENABLE_ADMIN_USER_CREATION}). Skipping."
         exit 0
       fi
       ./paas-cf/concourse/scripts/import_bosh_ca.sh
 
       VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
-      UAA_ADMIN_CLIENT_PASS=$($VAL_FROM_YAML secrets.uaa_admin_client_secret cf-secrets/cf-secrets.yml)
-      UAA_ENDPOINT=$($VAL_FROM_YAML properties.uaa.url cf-manifest/cf-manifest.yml)
+      CF_ADMIN=admin
+      CF_PASS=$($VAL_FROM_YAML secrets.uaa_admin_password cf-secrets/cf-secrets.yml)
+      API_ENDPOINT=$($VAL_FROM_YAML properties.cc.srv_api_uri cf-manifest/cf-manifest.yml)
 
       echo "Removing user ${NAME}"
-      uaac target "${UAA_ENDPOINT}"
-      uaac token client get admin -s "${UAA_ADMIN_CLIENT_PASS}"
 
-      uaac user delete "${NAME}"
+      echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+
+      cf delete-user "${NAME}" -f
--- a/concourse/pipelines/failure-testing.yml
+++ b/concourse/pipelines/failure-testing.yml
@@ -140,6 +140,8 @@ jobs:
                 BOSH_FQDN: {{bosh_fqdn}}
             - task: remove-temp-user
               file: paas-cf/concourse/tasks/delete_admin.yml
+              params:
+                PREFIX: controller-smoketest-user
 
   - name: nats
     serial_groups: [ failure ]
@@ -182,6 +184,8 @@ jobs:
                 BOSH_FQDN: {{bosh_fqdn}}
             - task: remove-temp-user
               file: paas-cf/concourse/tasks/delete_admin.yml
+              params:
+                PREFIX: nats-smoketest-user
 
   - name: router
     serial_groups: [ failure ]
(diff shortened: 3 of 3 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1815,8 +1815,12 @@ jobs:
           file: paas-cf/concourse/tasks/smoke-tests-run.yml
 
         ensure:
-          task: remove-temp-user
-          file: paas-cf/concourse/tasks/delete_admin.yml
+          do:
+          - task: get-cf-cli-config
+            file: paas-cf/concourse/tasks/get-cf-cli-config.yml
+
```
mtekel: I think you should merge get-cf-cli-config into remove user:
* to avoid having to declare it multiple times like you do
* to avoid concourse having to launch more containers
* to have simpler, cleaner pipelines

concourse/tasks/delete_admin.yml
```diff
@@ -3,32 +3,31 @@ platform: linux
...
     - -e
     - -c
     - |
-      NAME=$(cat admin-creds/username)
       if [ "${ENABLE_ADMIN_USER_CREATION}" == "false" ]; then
         echo "Temporary user creation is disabled (ENABLE_ADMIN_USER_CREATION=${ENABLE_ADMIN_USER_CREATION}). Skipping."
         exit 0
       fi
       ./paas-cf/concourse/scripts/import_bosh_ca.sh
+      . ./config/config.sh
```
mtekel: replace this with
```
      CF_ADMIN=admin
      CF_PASS=$($VAL_FROM_YAML secrets.uaa_admin_password cf-secrets/cf-secrets.yml)
      API_ENDPOINT=$($VAL_FROM_YAML properties.cc.srv_api_uri cf-manifest/cf-manifest.yml)
```

concourse/tasks/delete_admin.yml
```diff
@@ -31,3 +32,6 @@ run:
       echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
 
       cf delete-user "${NAME}" -f
+
+      echo "Removing leaked users"
```
mtekel: I'd say st. like "Purging any ${PREFIX} leftover users". Just to put the prefix into the output so that it is clear that it's only removing those users.

richardc: 👍

concourse/tasks/delete_admin.yml
```diff
@@ -31,3 +32,6 @@ run:
...
+      cf curl /v2/users | jq -r ".resources[] | .entity.username | select(. != null) | select(test(\"^${PREFIX}\"))" | xargs -r -I {} cf delete-user {} -f
```
mtekel: A note about this: cf curl /v2/users would return 50 users (see https://apidocs.cloudfoundry.org/240/users/list_all_users.html) or 100 at most (if you specify `?results-per-page=100`). It is possible that there would be no ${PREFIX} users among them albeit many ${PREFIX} users still existing in CF. In my dev deployment, it returned 228 total results, 5 total pages. Out of those, I had 7 leftover `CATS` users. When specified 100 results per page, I got 9 `CATS` users. I expect prod can have more legitimate users.

I am not sure what determines ordering of the users returned by `/v2/users` (apidocs only mention asc or desc - but not what is it ordered by), but it's possible that after several iterations of cleanup, there would be no more ${PREFIX} users returned in 1st 100 users, even though they do exist...

We can either:
* leave it as is
* extend this to return 100 users (I +1 this)
* make this do full proper pagination (response contains `next_url`) and always get all users

If we don't do full pagination, we can check by hand when removing this workaround if it has really removed all the leftover users.

concourse/pipelines/create-bosh-cloudfoundry.yml
richardc: To my understanding `get-cli-config` was introduced to serve as a common step to enable the remove-healthcheck-db.yml common task in 56c2f5e8482fa4935b5888db1c872b8bfb4bb688.  Do you suggest that as a later change we merge those jobs too?

mtekel: regarding multiple parsing of same thing, I thikn we could perhaps login once and then export `~/.cf/config.json` and just include that into tasks that need a logged-in CLI. But that sounds like a "support cleanup/optimize" task or something for our [improvements spreadsheet](https://docs.google.com/spreadsheets/d/1034dgtWbchumi-xJuTP2n4Iq8KpxF20Vrwgy3txmeBg/edit#gid=1419569599)

In this case, doing parsing of information as a separate task  adds a lot of repetetion into our already quite huge pipelines...

concourse/tasks/delete_admin.yml
richardc: Thanks, it wasn't clear there was a default for `results-per-page` from the API documentation, and my development environment was too small to hit this, so I assumed unpaged by default.

I'll look at doing the pagination as I don't think it'll be very complex.

concourse/tasks/delete_admin.yml
dcarley: > replace this with

Can you include an explanation of why?

mtekel: That was a suggestion that we don't need to run a separate "get credentials" container/task every time we need to run this, but instead of running config.sh (produced by that container) that exports the variables above, we can export them directly here...

concourse/tasks/delete_admin.yml
```diff
@@ -3,32 +3,46 @@ platform: linux
...
       echo "Removing user ${NAME}"
-      uaac target "${UAA_ENDPOINT}"
-      uaac token client get admin -s "${UAA_ADMIN_CLIENT_PASS}"
 
-      uaac user delete "${NAME}"
+      echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+
+      cf delete-user "${NAME}" -f
+
+      echo "Removing leaked ${PREFIX} users"
```
dcarley: @richardc @keymon do you have any thoughts about whether we'd remove the following scripting once the previously leaked users have been removed from existing persistent environments?

If we do, can we do anything to make it easier for us to remove this in the future, such as adding a `FIXME: remove later` comment and/or breaking up the commits to allow us to revert this part (but retain the `cf delete-user` above)?

keymon: Personally, I would keep it, because the delete user can always fail because some random reason. 

The question then is why we do delete the user created explicitly , when we will delete it anyway when deleting all the tmp users.

dcarley: I'm not sure. It kind of overlaps with the admin user creation that also removes test users over a certain age. In fact, I think we missed in kick-off, have we checked that that deletes users correctly too?

If we're going to keep it then I feel like it's sufficiently complex and dangerous enough to deserve tests. We may also need to think more carefully about race conditions - what happens if this runs while the continuous smoke tests are running?

mtekel: continuous smoke test users should have their own prefix. But e.g. you can run 2 concurrent acceptance tests. In that case when 1st acceptance tests finish, this would delete the user from the 2nd acceptance tests job...

keymon: no, the check checks for prefix a the beginning of the text: `test("^${PREFIX}")` (uses `^`).

But it is true that if PREFIX="", then it wipes all the users :scream:

richardc: It should be hard to do this accidentally as we flipped the -u on, but a more explicit gate could be added.

richardc: Reverting just the cleanup should be straightforward as it's all contained within one commit in part for this reason.  Can put more words about that into the commit message, with a link to a "stop doing that now" ticket if one exists.

keymon: Maybe we can add some check to prevent deleting users accidentally. For instance, check that it contains the word "test" and does not contain a  "@"

`select(test("^mytestuser")  and test("test") and (test("@")|not))`

dcarley: @richardc @keymon @mtekel and I discussed this after standup. Concluded that:

- we'll only use this script for a short amount of time; revert after it has reached production
  - we're confident enough that it will do the right thing for the short amount of time it'll be present
  - so we don't need additional guards or automated tests
  - but we'll test it again in dev against an environment that has a sample of "real" users
- we'll extend `sync-admin-users` to also use `cf delete-user` in another story, which will take care of leaked test users in future
- we might want to split the functionality of `sync-admin-users` to two separate tasks in future:
  - remove orphaned (test?) users, which would run in dev and make it clearer to team members that we have this functionality
  - create admin users which doesn't run in dev

mtekel:
Note that this doesn't clean up "CATS" users and any other users created by acceptance/other tests. It only cleans up the temporary admins that we have created. We should also clean-up those non-admin users. The cleanup loop would address that when specifying prefix as 'CATS'.

richardc:
To support this we need to do one of:

* allow PREFIX to be an array of prefixes to cleanup
* split out the wildcard removal job, so we can use it multiple times with a distinct prefix

in addition to inspecting the jobs to see what prefixes they are using for their temporary users.

richardc:
Closing as per discussion.  We are not targeting the correct set of users to remove with this, will rework and propose a similar but different PR later.

PR #431, opened 2016-08-18
[#146888305] Generate secrets when they are missing

Jonty:
## What

This allows the secret generator to handle two cases we ran into while attempting to rotate passwords as part of [#146888305](https://www.pivotaltracker.com/n/projects/1275640/stories/127339659/comments/146888305):

1) Secrets that have a key but no value are not generated, so the following will result in no new secrets being generated:

````
  ---
  secrets:
    preserved_secret: yay
    will_not_be_generated:
````

2) A yaml file with a `secrets:` key but no values defined in it will result in the SecretGenerator crashing as it expects a Hash:
````
  ---
  secrets:
````
## How to review

In `manifests/shared` run `rspec` and observe the tests passing.

## Who can review

Not @jonty.

### DIFF
--- a/manifests/shared/lib/secret_generator.rb
+++ b/manifests/shared/lib/secret_generator.rb
@@ -34,21 +34,30 @@ class SecretGenerator
 
   def generate
     @required_secrets.each_with_object({}) do |(key, type), output|
+      # If a secret exists and is not nil or empty, keep it rather than
+      # generating a new one. This checks that existing_secrets is a hash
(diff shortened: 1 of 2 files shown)


mtekel:
Works fine.

Additional functionality (for next PR) I'd add around handling _orig:
* if _orig is different, re-generate md5 (or, if _orig exists, generate md5)
* if _orig is missing, re-generate, re-generate md5
* if hashed thing is missing, just re-generate it. Don't generate new orig as well.

PR #432, opened 2016-08-18
[#128048645] Add chrome_plugin to filter Pingdom status checks

keymon:
[#128048645 Distinguish PaaS and Notify pingdom alerts on the dashboard](https://www.pivotaltracker.com/story/show/128048645)

What?
-----

We want a dashboard for only PaaS checks and this shouldn't prevent Notify from having a dashboard for their checks

We wrote this is a small chrome plugin filters undesired checks. It workarounds the issue of using a shared Pingdom Account with checks from multiple teams: Pingdom only provides one status page per account, and we want to display only our checks in the dashboard.

It keeps only the checks which name starts with "PaaS"

It also strips away some of the cruft:
* The header a footer
* The page selector
* The page title

The script is also implemented as a userscript.

How to test?
------------

With chrome:

1. Download this directory onto the machine running the monitors.
2. Open [chrome://extensions](chrome://extensions).
3. Tick the "Developer mode" checkbox in the top right.
4. Click "Load unpacked extention..." and select this directory.
5. Refresh your pingdom status page.

With Firefox or Chrome with a greasemonkey extension:

1. Click [src/inject.user.js](src/inject.user.js) and your browser _should_ auto-install it.
3. If not, drag it onto the browser.
4. Refresh your pingdom status page.

Who?
---

Anyone but @keymon


richardc:
Errors on Chrome Version 52.0.2743.116 (64-bit) with

```
inject.user.js:46 Uncaught ReferenceError: GM_addStyle is not defined(anonymous function) @ inject.user.js:46
```

misc/chrome_plugins/pingdom_status_filter_checks/src/inject.user.js
```diff
@@ -0,0 +1,72 @@
...
+    var element = document.getElementsByClassName("largeTitle")[0];
+    if (element) {
+        element.parentNode.removeChild(element);
+    }
+
+    // Fix CSS for the checks when removing header
+    document.getElementById("checks").style.marginTop = "0px";
+});
+
+/*GM_addStyle ( "           \
```
richardc: Is this a magic comment or a leftover from development?

keymon: Leftover, sorry :angel:

PR #434, opened 2016-08-19
126965751 Test users not removed from CC DB

richardc:
## What

[#126965751](https://www.pivotaltracker.com/n/projects/1275640/stories/126965751) Test users not removed from CC DB

This series reworks the existing delete_admin task to be implemented in terms of `cf remove-user` rather than `uaac remove user` to prevent test admin users leaking, and then adds and additional task to the `cf-deploy` job to cleanup existing leaked users.

## How to review

Before deployment log into a developement environment with the `cf` tool and capture existing users with `cf curl /v2/users?results-per-page=1000 > users.json` (you may need to specify a larger value for results-per-page if this pages)

Deploy and run the pipeline.  After running repeat the `cf curl /v2/users?results-per-page=1000 > new_users.json`

Confirm that there are fewer/no users with a null username.

```
cat new_users.json | jq -r '.resources[] | select(.entity.username == null) | .metadata.guid'
```

Additionally, you should be able to observe the number of cc has dropped in your user impact dashboard in grafana (previous behaviour would have been only to grow).  

 
## Who can review

Anyone but @mtekel or @richardc or @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1574,6 +1574,44 @@ jobs:
                   bundle
                   bundle exec sync-admin-users.rb ${API_ENDPOINT} ../config/admin_users.yml "{{NEW_ACCOUNT_EMAIL_ADDRESS}}"
 
+        - task: purge-unnamed-users
+          config:
+            platform: linux
+            inputs:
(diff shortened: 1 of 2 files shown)


richardc:
This is the close cousin to #430 but with a more targetted cleanup of the CC DB.

PR #436, opened 2016-08-22
Remove orphaned tf outputs fixture

mtekel:
## What

We did some refactoring of our tests recently and this file is not used by anything any more. Clean up.

## How to review

`make test` + travis will also run the tests

## Who can review

anyone


dcarley:
Works 👍 

I would have found review easier if you'd included the SHA1 of the commit that orphaned this file. The tests passing is probably enough to confirm that it works, but I wanted to understand how it had happened. I originally thought you were referring to 878701f062700f038f319733d7f38247f1ab2d42 but I think it might have been unused as far back as 47cfa2d6e5a629a47c1057bfeeb315b8469b9f92.

PR #437, opened 2016-08-23
[#128160309] Concourse only ssh key

mtekel:
## What

[Rotate credentials](https://www.pivotaltracker.com/n/projects/1275640/stories/128160309)

Generate and use a separate, concourse only ssh key for deployer concourse. This improves security and our ability to rotate credentials specific only to concourse, instead of having to rotate shared SSH key, which in practice means additional re-deployment of bosh and every VM deployed by it.

## How to review

Run bootstrap concourse pipeline. Observe new key being generated, uploaded and concourse being re-deployed using this key. In AWS you can check the keypair name for rebuilt concourse instance. Try ssh-ing to concourse using new key.

## Merging & deploying
As a part of the rotate credentials story we also want to change concourse ATC passwords. These also require re-deployment of concourse. It is possible to do this change together with this PR, saving one concourse deployment. Review (TODO: link to credentials rotate PR) and deploy concourse using new passwords after this PR is merged.

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -67,6 +67,13 @@ resources:
       versioned_file: id_rsa
       region_name: {{aws_region}}
 
+  - name: concourse-ssh-private-key
+    type: s3-iam
+    source:
+      bucket: {{state_bucket}}
+      versioned_file: concourse_id_rsa
+      region_name: {{aws_region}}
+
   - name: git-ssh-public-key
     type: s3-iam
     source:
@@ -280,6 +287,28 @@ jobs:
         params:
           file: generated-concourse-cert/concourse-cert.tar.gz
 
+    - task: generate-concourse-ssh-key
+      config:
+        image: docker:///governmentpaas/curl-ssl
+        inputs:
+        - name: paas-cf
+        outputs:
+        - name: concourse-ssh-key
+        params:
+          AWS_DEFAULT_REGION: {{aws_region}}
+        run:
+          path: sh
+          args:
+          - -e
+          - -c
+          - |
+            apk add --update openssh
+            cd concourse-ssh-key
+            ssh-keygen -t rsa -b 4096 -f generated_concourse_id_rsa -N ''
+            ../paas-cf/concourse/scripts/s3init.sh {{state_bucket}} concourse_id_rsa generated_concourse_id_rsa
+            ../paas-cf/concourse/scripts/s3init.sh {{state_bucket}} concourse_id_rsa.pub generated_concourse_id_rsa.pub
+
+
     - task: terraform-apply
       config:
         image: docker:///governmentpaas/terraform
@@ -289,6 +318,7 @@ jobs:
         - name: concourse-terraform-state
         - name: generated-concourse-cert
         - name: git-ssh-public-key
+        - name: concourse-ssh-key
         params:
           VAGRANT_IP: {{vagrant_ip}}
           TF_VAR_env: {{deploy_env}}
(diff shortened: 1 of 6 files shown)


alext:
I think we should call the new key `deployer_id_rsa` or `deployer_concourse_id_rsa` to make it clear that it's specific to that concourse instance.

mtekel:
On the other side, we do call `deployer` `concourse` in many other places, e.g. in terraform, create-bosh-cf pipeline etc...

concourse/pipelines/create-deployer.yml
```diff
@@ -280,6 +287,28 @@ jobs:
...
+          path: sh
+          args:
+          - -e
+          - -c
+          - |
+            apk add --update openssh
+            cd concourse-ssh-key
+            ssh-keygen -t rsa -b 4096 -f generated_concourse_id_rsa -N ''
+            ../paas-cf/concourse/scripts/s3init.sh {{state_bucket}} concourse_id_rsa generated_concourse_id_rsa
+            ../paas-cf/concourse/scripts/s3init.sh {{state_bucket}} concourse_id_rsa.pub generated_concourse_id_rsa.pub
```
alext: Is there any reason this needs to be a separate task as opposed to including it in the existing `generate-ssh-keys` task?

mtekel: because in the bootstrap concourse task outputs are only accessible within the same job. generate-ssh-keys is in different job

dcarley: Can we use `put` to upload the generated key instead of `s3init`?

I know we do it in the other task, but it's a misuse of `s3init` that would be good for us to phase out.

mtekel: There's a lot of what could be improved/re-organised in the bootstrap pipeline. E.g. upgrading to same version as our main concourse. For this specific case, I did not want to introduce new pattern. We should fix up bootstrap pipeline properly based on what we have learned so far on the main concourse...

alext: > Can we use put to upload the generated key instead of s3init?

That gives us the bootstrapping problem. `s3init` is used here to only upload the key if it's not already in the state bucket.  Using a put would result in rotating the key on every deploy (unless we add some more logic around it).

mtekel: There is also advantage that s3init only uploads thing once and not every time like we do with put. We should probably fix that in the main pipeline as well.

manifests/concourse-manifest/spec/fixtures/vpc-terraform-outputs.yml
```diff
@@ -2,7 +2,6 @@
 terraform_outputs:
   environment: test
   infra_subnet_ids: subnet-12345678,subnet-23456789
-  key_pair_name: ssh_key_pair
```
mtekel: this will likely break BOSH deployment

alext:
> On the other side, we do call deployer concourse in many other places, e.g. in terraform, create-bosh-cf pipeline etc...

That's a fair point.  I note that it's called `concourse/0` in the AWS console, so I think it's fine to leave it as is.

alext:
I think this is good to go, but it should probably be reviewed by someone else as I'm also working on this story.

concourse/pipelines/create-deployer.yml
keymon: We should create a new resource to generate SSH keys. Now we are in concourse 1.5.1 we can implement custom resources easily.

For instance, this one:  https://github.com/redfactorlabs/concourse-smuggler-resource/blob/master/examples/ssh-keygen-s3/pipeline.yml

But  I think that is out of the scope of this story.

keymon:
Works as published and the changes make sense.

PR #438, opened 2016-08-23
[#128762469] Automate the rotation of a user's password

HenryTK:
## What

[#128762469](https://www.pivotaltracker.com/n/projects/1275640/stories/128762469)

We want an automated, repeatable, and standardised way of rotating a Cloud Foundry user's password and emailing it to them when they are unable to log in.
    
Common functions have been taken from the `create-user.sh` script, so they have been split into a separate file.
    
Run the script without arguments to see its documentation.

## How to review

There may be an advantage of testing this with a non-Mac user. In theory this should be tested in all environments, but production is the only likely place this script will need to be used.

* Choose an email address you have access to, it will be used for a test user.
* For the dev environment you will need to run `aws ses verify-email-identity --email-address` and pass the email address you have chosen. Follow the link in the email to verify the email with AWS.
* Create a user using the `create-user.sh` script with the email address you have chosen.
* Login using `cf login` to check the user is set up correctly, then log out (`cf logout`).
* Run the `rotate-user-password.sh` script to rotate the password. 
* Check you received the email and use the password in it to log in.
* Delete any users left in persistent environments (log in as admin and `cf delete-user $EMAIL`).

## Who can review

Anyone but myself

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+load_colors() {
+  if [ -t 1 ] ; then
+    export ESC_SEQ="\x1b["
+    export COL_RESET="${ESC_SEQ}39;49;00m"
+    export COL_LIGHT="${ESC_SEQ}1m"
+    export COL_DIM="${ESC_SEQ}2m"
+    export COL_BLINK="${ESC_SEQ}5m"
+    export COL_RED="${ESC_SEQ}31m"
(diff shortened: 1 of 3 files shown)


benhyland:
Seems to work fine (tested on Ubuntu 16.04 via Parallels).

Could possibly tidy up the shared code a bit further, but we're reaching the point of code reuse where we need to switch language to do much better.

I'll talk about automated testing a bit in the pivotal ticket - it will definitely need to be a separate story.

PR #439, opened 2016-08-23
Re-tidy secret-generators after lint auto-correct

alext:
## What

In the secret generates, we should have each secret definition on a line by itself to make it easier to read. The rubocop autocorrect went a bit OTT on these (in e24e653), which made
them a little harder to read.

## How to review

Review the changes and marvel at the magnificence of the new version.

## Who can review

Anyone but myself.


dcarley:
Looks mag-ni-fi-cent! It will also make diffs easier to read as we add more items.

PR #441, opened 2016-08-24
[#128160309] Bosh only ssh key

mtekel:
## What

[Rotate credentials](https://www.pivotaltracker.com/n/projects/1275640/stories/128160309)

## How to review

Run `create-bosh-cloudfoundry` pipeline. Observe new key being generated, uploaded to AWS by terraform and applied on BOSH server (which due to how bosh-init works will cause re-deployment). After deploying, you can check in AWS what key is BOSH using & compare that to key used by bosh deployed VMs.

## Implementation

Compared to https://github.com/alphagov/paas-cf/pull/437 , I have used (a different) mechanism of setting and getting the keys which is used in `create-bosh-cloudfoundry` pipeline, with a slight modification:
* use s3init to initialize empty files (so that we can `get` something instead of being blocked)
* use s3iam resource to `get` and `put` keys
* (new) only `put` keys when new ones are generated. This will cause upload jobs of generated keys to fail when there's nothing to upload, but it will not break the pipeline execution, as this is in `ensure` block. The advantages are:
  1. faster execution (as key only gets generated & uploaded when it is not present in state bucket)
  2. less concourse resources used (no need to create containers for put steps every time) 
  3. better caching - as there is not a "new" version of the resource after every put
  4. less storage used in the bucket (we don't re-upload the same file again) - less $ spent
  5. better auditability - bucked only contains file versions which have been used. We can track when new key was generated by upload times.

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -212,11 +212,18 @@ resources:
       region_name: {{aws_region}}
       versioned_file: {{bosh_manifest_state}}
 
-  - name: ssh-private-key
+  - name: bosh-ssh-private-key
     type: s3-iam
     source:
       bucket: {{state_bucket}}
-      versioned_file: id_rsa
+      versioned_file: bosh_id_rsa
+      region_name: {{aws_region}}
+
+  - name: bosh-ssh-public-key
+    type: s3-iam
+    source:
+      bucket: {{state_bucket}}
+      versioned_file: bosh_id_rsa.pub
       region_name: {{aws_region}}
 
   - name: git-ssh-private-key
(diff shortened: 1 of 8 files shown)


manifests/bosh-manifest/spec/fixtures/bosh-secrets.yml
```diff
@@ -1,6 +1,5 @@
 ---
 secrets:
-  bosh_postgres_password: BOSH_POSTGRES_PASSWORD
```
alext: This is currently being generated by `generate-bosh-secrets.rb`. If it's genuinely not used, then it should be removed from there as well, otherwise we should keep this fixture in sync with the generator.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -485,6 +497,47 @@ jobs:
           params:
             file: generated-bosh-secrets/bosh-secrets.yml
 
+      - task: generate-bosh-ssh-keypair
+        config:
+          platform: linux
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/curl-ssl
```
alext: Now that this is no longer using `s3init.sh`, There's no need to use this container, we could use a vanilla alpine container, or even a container with openssh already installed (eg our git-ssh container).

alext:
A few small comments about the implementation, but in general I like this approach. The error displayed in concourse isn't great, but I think the advantages of this approach outweigh that.

mtekel:
I have updated the container type and added a commit to get bosh DB PW directly from secrets...

keymon:
IMO it would be much better to implement this with a dedicated resource. As commented before, with this "smuggler resource" we could even implement a resource to generate the keys and encapsulate the logic of no creating a new key if it is not necessary.

https://github.com/redfactorlabs/concourse-smuggler-resource/blob/master/examples/ssh-keygen-s3/pipeline.yml

mtekel:
Good point, i'll add it to the improvements spreadsheet...

keymon:
Personally, I wonder if we should implement something like this now :-m

keymon:
I am OK with this PR, but I think we should implement this feature as a dedicated resource.

PR #442, opened 2016-08-24
[#128160309] Remove some unused entries from generated cf-secrets.

alext:
## What

While working on the secret rotation story, I noticed that these 2 entries are being generated, but are not used in the manifests anywhere. We should get rid of them to avoid confusion.

## How to review

Verify that these aren't referenced anywhere. Verify that the unit tests pass with the updated fixture.

## Who can review

Anyone but myself.


dcarley:
> Verify that these aren't referenced anywhere.

Is `grep <name> manifests/cf-manifest/` sufficient?

Unlike some CF properties there would have to be an explicit `grab <name>`, because we don't import the whole hash, right?

alext:
> Is grep <name> manifests/cf-manifest/ sufficient?

That should be. I used `ack` to search for the string throughout the whole repo, and found only these 2 entries. I'm not aware of anywhere that we're pulling in the whole secrets hash.

dcarley:
👍 

Is it worth removing these from persistent environments when you're doing secret rotation, to surface any possible problems early?

PR #443, opened 2016-08-24
[#128160309] Generate deployment key in deployer concourse

mtekel:
## What

Because we have now separate keys for concourse (#437) and BOSH (#441) we can move "deployment ssh key" generation to create-bosh-cf pipeline.

This is the key that BOSH uses for all the VMs it deploys. The key has thus be already available for BOSH creation, therefore we move it from VPC terraform (and outputs) to BOSH terraform. We generate it at the same time we generate other deployment credentials.

Advantage of having this key generated in `create-bosh-cloudfoundry` pipeline is that you don't have to start deployer concourse in order to generate new keys. Also, this key is not used by concourse any more (#437), meaning we don't have to handle it during the bootstrap already.

## How to review

Run bootstrap pipeline first. Observe key being removed from AWS, as it is removed from VPC terraform. Then run create-bosh-cf pipeline. Check key (be attepmted to) be generated in generate secrets job. Observe bosh terraform run uploading the same key under the same name back into AWS. (if you don't run bootstrap pipeline first, this step will fail as key with that name would already exist). The rest of pipeline should complete without any changes.

## Merging
#437 and #441 should be merged first. Then this PR re-based on master. Then it can be merged.

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -226,6 +226,21 @@ resources:
       versioned_file: bosh_id_rsa.pub
       region_name: {{aws_region}}
 
+  - name: ssh-private-key
+    type: s3-iam
+    source:
+      bucket: {{state_bucket}}
+      versioned_file: id_rsa
+      region_name: {{aws_region}}
+
+  - name: ssh-public-key
+    type: s3-iam
+    source:
+      bucket: {{state_bucket}}
+      versioned_file: id_rsa.pub
+      region_name: {{aws_region}}
+
+
   - name: git-ssh-private-key
     type: s3-iam
     source:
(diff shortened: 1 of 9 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -468,143 +485,183 @@ jobs:
...
+            try:
+              aggregate:
+                - put: bosh-ssh-private-key
+                  params:
+                    file: generated-bosh-ssh-private-key/bosh_id_rsa
+                - put: bosh-ssh-public-key
+                  params:
+                    file: generated-bosh-ssh-public-key/bosh_id_rsa.pub
+
+        - task: generate-deploymens-ssh-keypair
```
alext: Typo: 'deployment'

keymon:
+1, works as expected and does not redeploy the VMs

PR #445, opened 2016-08-25
[#128492081] Increase the RDS creation timeout

Jonty:
## What

While RDS instances take <10mins to create once the creation starts, we're seeing the creation itself not start for quite some time after it's requested. This is leading to the the tests failing and orphaned RDS instances being left behind.

Doubling the creation time should hopefully avoid this problem, but we should also query this with AWS themselves as we're frequently seeing extended creation times.

## How to review

Run the pipeline, observe the custom acceptance tests passing.

## Who can review

Not me, guv.

### DIFF
--- a/platform-tests/src/acceptance/rds_broker_test.go
+++ b/platform-tests/src/acceptance/rds_broker_test.go
@@ -38,7 +38,7 @@ var _ = Describe("RDS broker", func() {
 		// slow (several minutes).
 
 		const (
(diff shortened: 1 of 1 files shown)


keymon:
LGTM. Did not really test it, but it compiles and the change is minimal, I am 99.99999% confident the intended change works.

keymon:
https://en.wiktionary.org/wiki/guv

PR #447, opened 2016-08-25
secret_generator: s/ignores/removes/ in test desc

dcarley:
## What

Change the test description to make it clearer that existing passwords which
are no longer "requested" will be removed from the output. This is based on
a discussion with Alex when reviewing #442, where I asked if we needed to do
this manually and he pointed out that the code already does it.

## How to review

1. Confirm that the change and commit message makes sense.
1. Confirm that Travis build passes.

## Who can review

Not @dcarley


keymon:
LGTM, I will merge as soon as travis finishes.

PR #448, opened 2016-08-25
Use remote state and add ACTION variable to manual Terraform commands

dcarley:
## Notes

Sorry, no story ID. I'm running out of things to do and flying off piste 😈 

#### manually_upload_certs: Use Terraform remote state

This makes it consistent with the `set-up-pingdom.sh` script which was
modified in b01f941 to use Terraform's remote state functionality. This is
simpler and safer than downloading and uploading the statefile with aws-cli.

When testing this with `terraform plan` instead of `terraform apply` against
staging and production (it's not possible to use this in dev or CI) the
following error was produced on the very first run:

    # check password store and if varables are accesible
    Remote state cache has no remote info
    make: *** [manually_upload_certs] Error 1

After which a `remote` hash is written to the remote statefile and
subsequent runs work as expected.

#### Add ACTION variable to manual Terraform commands

So that we can run `plan` instead of `apply` to safely review changes. I
made a change similar to this to my local working copy when reviewing a
[credentials PR](https://github.gds/government-paas/credentials/pull/33)
because I didn't want to make any unexpected changes to the live config.

I thought about setting a safe default action of `plan`, but decided against
it because people might assume that it had run `apply` and changes wouldn't
have been applied. It would have also been possible to re-use the `ACTION`
variable within the scripts but I thought it was better to explicitly parse
a CLI argument.

I've updated the documentation accordingly, listing both commands that you
are likely to use.

## How to review

I'll describe using the least privileged environment that each target can be used in.

#### Pingdom

1. Export AWS keys and `DEPLOY_ENV=master` for CI.
1. Run the command with `ACTION` and confirm that it reports a useful error:

  ```
make ci pingdom
```
1. Run plan and confirm that it reports no changes or errors:

  ```
make ci pingdom ACTION=plan
```
1. Run apply and confirm that it reports no changes or errors:

  ```
make ci pingdom ACTION=apply
```

#### Certs

1. Export AWS keys and `DEPLOY_ENV=staging` for Staging.
1. Export `CERT_PASSWORD_STORE_DIR` to match where you have the `credentials-high` repo.
1. Run the command with `ACTION` and confirm that it reports a useful error:

  ```
make staging manually_upload_certs
```
1. Run plan and confirm that it reports no changes or errors:

  ```
make staging manually_upload_certs ACTION=plan
```
1. Run apply and confirm that it reports no changes or errors:

  ```
make staging manually_upload_certs ACTION=apply
```

## Who can review

Probably needs a Civil Servant with SC because it the targets depend on the `government-paas/credentials-high` repo.


HenryTK:
Was reviewing the Pingdom part of this one and encountered an issue. Using Terraform version 0.6.12 and running `DEPLOY_ENV=master make ci pingdom ACTION=plan` I get the following error:

```
Error loading local state: State version 3 not supported, please update.
```

Source code trace:
https://github.com/hashicorp/terraform/blob/v0.6.12/terraform/state.go#L1188
https://github.com/hashicorp/terraform/blob/v0.6.12/command/remote_config.go#L63

The state file from the CI bucket seems to indicate version 1, not version 3. The code also indicates [Terraform v0.6.12 uses state version 1](https://github.com/hashicorp/terraform/blob/v0.6.12/terraform/state.go#L20).

Would be good to know which version of Terraform the state file was produced by and whether anybody else has this issue.

richardc:
Version 3 would seem to track to the (0.7)[https://github.com/hashicorp/terraform/commit/706ccb7dfedb9807fd65643a6e4eefd6a8b4f8f0#diff-1eac40478f2df1bf18b772c48cd7ce9aR22] series of terraform.

HenryTK:
@richardc it does, but I don't even have that installed anymore. Wonder if something has been cached somewhere.

Jonty:
@HenryTK So I think that must be because of https://github.com/alphagov/paas-cf/pull/448/files#diff-a1cb0f2dcf908bac1ed895c8413bcdd0R7 which looks erroneous.

That statefile is version 4 on ci.

Jonty:
@HenryTK IGNORE. That is the serial and my eyes are malfunctioning. I can't find any tfstate with a  version other than 1.

Are you sure you don't have a local tfstate file that was confusing things during testing? The code looks correct.

Jonty:
@HenryTK DOUBLE IGNORE. Github hadn't refreshed and I hadn't seen your comment.

I give up.

Jonty:
I seem to remember @bazbremner saying something about them nuking local tfstate at the start of every run to prevent issues - is that still the case?

HenryTK:
@Jonty - the script now does `rm -rf .terraform` to remove the local copy of the remote state at the end, but if the script errors it does not get cleaned up. We probably want to consider removing it at the start, because if anyone has any errors that result in the state file being left behind it could cause problems the next time they run it (particularly if somebody else has modified the Pingdom checks since then). 

It seems like strange behaviour of `terraform remote config` to use the local cached copy in this way, especially when one of the main advantages of a remote config is the ability to share it across a team. I wouldn't consider it an issue, but perhaps a feature request to add `-disable-local-state` may be necessary. From [this comment](https://github.com/hashicorp/terraform/blob/v0.6.12/command/remote_config.go#L57-L59) they seem to trust the local copy more for some reason.

Jonty:
@HenryTK Yes, it's very odd that it functions this way, I've heard others discussing it.

I've added a statement to remove the terraform state before running to stop this happening in the future.

Jonty:
This needs someone with access to credentials-high to run it and check it over now, it all looks fine now henry's problem has been resolved.

alext:
> I've added a statement to remove the terraform state before running to stop this happening in the future.

Should we add this to the pingdom script as well as the certs script?

terraform/scripts/manually-upload-certs.sh
```diff
@@ -11,29 +13,28 @@ pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain.crt" > /dev/null
 pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain.key" > /dev/null
 pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain_intermediate.crt" > /dev/null
 
+# Ensure there's no local state before we kick off terraform, it prioritises it
+rm -rf .terraform
+
 WORKING_DIR=$(mktemp -d cf-certs.XXXXXX)
 trap 'rm -r "${WORKING_DIR}"' EXIT
```
alext: I don't think this `WORKING_DIR` is used any more... We should remove this.

alext:
Tested the certs upload on staging, and it all works fine. Other than the 2 small comments above, this looks good to me.

terraform/scripts/manually-upload-certs.sh
```diff
@@ -11,29 +13,28 @@ pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain.crt" > /dev/null
...
   -var apps_domain_intermediate_crt="$(pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain_intermediate.crt")" \
   terraform/cf-certs
-exit_status=$?
-set -e
-
-aws s3 cp "${WORKING_DIR}/cf-certs.tfstate" "s3://${DEPLOY_ENV}-state/cf-certs.tfstate"
 
-exit $exit_status
+# Delete new local terraform state
+rm -rf .terraform
```
benhyland: @Jonty @HenryTK The `rm` here won't get run if the `terraform` command fails. Is this intentional? Should we be removing the local state, rather than moving it out of the way? Mightn't we be causing a problem for the user if they are running other terraform commands from the same directory?

benhyland:
@Jonty @HenryTK @alext I added a commit to handle the local state differently. Does this look sensible?

benhyland:
Apart from the review I requested above, I think we've dealt with everything we needed to.

I have one remaining question to check I understand Dan's original PR comment:
He talks about the `least privileged environment` each target can be run in; these being CI for pingdom and Staging for the cert upload. Is that simply because we don't run pingdom checks on dev environments, and we use self-signed certs in dev and CI environments?

terraform/scripts/manually-upload-certs.sh
```diff
@@ -14,26 +16,23 @@ pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain_intermediate.crt" > /dev/nu
...
-terraform apply -var env="${DEPLOY_ENV}" \
+# Configure Terraform remote state
+terraform remote config \
+  -backend=s3 \
+  -backend-config="bucket=${DEPLOY_ENV}-state" \
+  -backend-config="key=${STATEFILE}" \
+  -backend-config="region=${AWS_DEFAULT_REGION}"
+
+terraform "${TERRAFORM_ACTION}" -var env="${DEPLOY_ENV}" \
   -var-file="terraform/${AWS_ACCOUNT}.tfvars" \
```
alext: This file can no longer be found (it needs the `PAAS_CF_DIR` prefix).

terraform/scripts/manually-upload-certs.sh
```diff
@@ -14,26 +16,23 @@ pass "certs/${AWS_ACCOUNT}/${DEPLOY_ENV}/apps_domain_intermediate.crt" > /dev/nu
...
 trap 'rm -r "${WORKING_DIR}"' EXIT
 
-if ! aws s3 ls "s3://${DEPLOY_ENV}-state/cf-certs.tfstate" --summarize | grep -q "Total Objects: 0"; then
-  aws s3 cp "s3://${DEPLOY_ENV}-state/cf-certs.tfstate" "${WORKING_DIR}/cf-certs.tfstate"
-else
-  echo "No previous cf-certs.tfstate file found in s3://${DEPLOY_ENV}-state/. Assuming first run."
-fi
+# Work in tmp dir to ensure there's no local state before we kick off terraform, it prioritises it
+PAAS_CF_DIR=$(pwd)
+cd ${WORKING_DIR}
```
alext: The `cd` here means that the above exit trap no longer works because `WORKING_DIR` doesn't contain an absolute path.

PR #449, opened 2016-08-26
WIP Shellcheck pipelines

richardc:
## What

## How to review

## Who can review

Describe who can review the changes. Or more importantly, list the people
that can't review, because they worked on it.

### DIFF
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -246,6 +246,7 @@ pipecleaner.py pipeline.yml [pipeline2.yml..]
         if flag == '--shellcheck':
             shellcheck = True
         if flag == '--shellcheck-fatal':
+            shellcheck = True
             shellcheck_fatal = True
 
     if not files:
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -44,7 +44,7 @@ class Pipecleaner(object):
         raw = re.sub('\{\{.*?\}\}', 'DUMMY', raw)
         return yaml.load(raw)
 
-    def shellcheck(self, job, task, args):
+    def shellcheck(self, shell, job, task, args):
         if not self.shellcheck_enabled:
             return
 
@@ -57,7 +57,7 @@ class Pipecleaner(object):
                 prefix = job + "." + task,
                 delete = False
         )
-        script.write("#!/bin/sh\n")
+        script.write("#!/bin/" + shell + "\n")
         for switch in args[:-2]:
             script.write("set " + switch + "\n")
         script.write(args[-1])
@@ -163,8 +163,8 @@ class Pipecleaner(object):
                             output_resources.add(i['name'])
 
                     if 'run' in item['config']:
-                        if item['config']['run']['path'] == 'sh':
-                            self.shellcheck(job['name'], item['task'], item['config']['run']['args'])
+                        if item['config']['run']['path'] in ['sh', 'bash', 'zsh']:
+                            self.shellcheck(item['config']['run']['path'], job['name'], item['task'], item['config']['run']['args'])
 
             overall_used_resources = overall_used_resources.union(used_resources).union(get_resources)
 
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -27,9 +27,7 @@ import re
 import sys
 import os
 import getopt
-import tempfile
 import subprocess
-
 
 class Pipecleaner(object):
     def __init__(self, shellcheck_enabled):
@@ -48,22 +46,26 @@ class Pipecleaner(object):
         if not self.shellcheck_enabled:
             return
 
-        if args[-2] != '-c':
-            print "Odd args " + job + "." + task
-            print args
-            return
+        script = ""
 
-        script = tempfile.NamedTemporaryFile(
-                prefix = job + "." + task,
-                delete = False
-        )
-        script.write("#!/bin/" + shell + "\n")
-        for switch in args[:-2]:
-            script.write("set " + switch + "\n")
-        script.write(args[-1])
-        script.close()
-        subprocess.call(["shellcheck", script.name])
-        os.unlink(script.name)
+        for switch in args[:-1]:
+            if switch != "-c":
+                script += "set " + switch + "\n"
+
+        script += args[-1]
+
+        process = subprocess.Popen(["shellcheck", "-s", shell, "-"],
+                                   stdin = subprocess.PIPE,
+                                   stdout = subprocess.PIPE,
+                                   stderr = subprocess.STDOUT)
+        out = process.communicate(script)[0]
+        exit = process.returncode
+
+        # TODO: exitcode non-zero == ERROR
+        #       output non-silent == WARNING
+
+        print "Exitcode: " + str(exit)
+        print "Output: " + out
 
     def check_pipeline(self, data):
         errors = {
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -172,7 +172,7 @@ class Pipecleaner(object):
                                 errors['shellcheck'].append({
                                     'job': job['name'],
                                     'task': item['task'],
-                                    'output': output,
+                                    '~': output,
                                     'fatal': self.shellcheck_fatal,
                                 })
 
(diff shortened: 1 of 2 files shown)


concourse/scripts/pipecleaner.py
```diff
@@ -40,6 +44,27 @@ def load_pipeline(self, filename):
...
+    def shellcheck(self, shell, job, task, args):
+        if not self.shellcheck_enabled:
+            return
+
+        if args[-2] != '-c':
+            print "Odd args " + job + "." + task
+            print args
+            return
+
+        script = tempfile.NamedTemporaryFile(
```
Jonty: This can be done much more cleanly by sending the shellcode to stdin instead of using temporary files.

Something a bit like:
`from subprocess import Popen, PIPE, STDOUT

p = Popen(['shellcheck'], stdout=PIPE, stdin=PIPE, stderr=STDOUT)    
stdout = p.communicate(input=item['config']['run']['args'])[0]
print(grep_stdout.decode())`

We should also be (loosely) parsing the response for shellcheck fatals and turning those into a pipecleaner fatal so we prevent tests from passing. Given we actually aim for zero warnings from shellcheck, perhaps any output from shellcheck at all should be a fatal?

concourse/scripts/pipecleaner.py
```diff
@@ -137,6 +162,10 @@ def check_pipeline(self, data):
                         for i in item['config']['outputs']:
                             output_resources.add(i['name'])
 
+                    if 'run' in item['config']:
+                        if item['config']['run']['path'] in ['sh', 'bash', 'zsh']:
+                            self.shellcheck(item['config']['run']['path'], job['name'], item['task'], item['config']['run']['args'])
```
Jonty: As above, we should be doing something with the output of shellcheck rather than just spewing it at stdout.

concourse/scripts/pipecleaner.py
```diff
@@ -27,9 +27,13 @@
 import sys
 import os
 import getopt
+import tempfile
+import subprocess
 
 
 class Pipecleaner(object):
+    def __init__(self, shellcheck_enabled):
+        self.shellcheck_enabled = shellcheck_enabled
```
Jonty: I wonder if we should perhaps check if the shellcheck binary exists and enable it based on that rather than having a flag? Unsure.

Regardless, the container that is now running pipecleaner in travis (and elsewhere) now needs shellcheck installing.

concourse/scripts/pipecleaner.py
richardc: Yeah, this is a note in the first commit message.

concourse/scripts/pipecleaner.py
richardc: It does already.  https://github.com/alphagov/paas-cf/blob/master/.travis.yml#L22-L25

The flag was more for development as it's an incomplete feature

concourse/scripts/pipecleaner.py
Jonty: In my defense I have barely drunk any of this coffee yet.

Jonty:
Tiny note: On IRC I said that you didn't need to worry about specifying the shell as shellcheck requires a shebang line, which is both true and clearly nonsense in the context of this PR as the exec line is specified in the pipeline. Sorry.

concourse/scripts/pipecleaner.py
richardc: The existing lint_shellcheck Make target just relies on the exitcode from shellcheck, so we should treat that as our fatal/nonfatal.  

shellcheck does have a json mode for output to resurface the information. https://github.com/koalaman/shellcheck#in-your-build-or-test-suites

concourse/scripts/pipecleaner.py
```diff
@@ -40,6 +44,27 @@ def load_pipeline(self, filename):
...
         return yaml.load(raw)
 
+    def shellcheck(self, shell, job, task, args):
+        if not self.shellcheck_enabled:
+            return
+
+        if args[-2] != '-c':
+            print "Odd args " + job + "." + task
+            print args
+            return
```
Jonty: Should this be a fatal error?

concourse/scripts/pipecleaner.py
Jonty: That seems sensible - I wouldn't worry too much about parsing it for now, simply dumping stdout from shellcheck if it exits nonzero is reasonable.

concourse/scripts/pipecleaner.py
richardc: Possibly;  it's only in a few places, so the consequence of merging this might be to un-weird it.  It just broke when I was hacking.

```
Odd args generate-secrets.generate-cf-secrets
Odd args generate-bosh-config.extract_terraform_outputs
Odd args bosh-cli.run-bosh-cli
Odd args init-bucket.try-fetch-bucket-state
```

dcarley:
Looks interesting

I was already a bit nervous that we don't have any tests for `pipecleaner.py`. What do you think about writing a test for this new functionality? We could use a framework similar to the tests in `concourse/scripts/`.

/cc @timmow when he returns

dcarley:
btw, I don't think it affects this PR, but for awareness: I raised concourse/fly#99 to look at getting some of the pipecleaner functionality upstream because they already have validations, tests and a model of the configuration. It hasn't had a reply though.

richardc:
Note to self,  remove `--shellcheck` option and instead use `--ignore_types`.  If `--ignore-types` contains `shellcheck` just don't collect it.

concourse/scripts/pipecleaner.py
```diff
@@ -137,6 +164,18 @@ def check_pipeline(self, data):
                         for i in item['config']['outputs']:
                             output_resources.add(i['name'])
 
+                    if 'run' in item['config']:
+                        if item['config']['run']['path'] in ['sh', 'bash', 'dash', 'ksh']:
```
mtekel: "ash" is the shell in alpine (which we use a lot)

richardc: Though we use it as `sh`, which is to say in ash_mode.

shellcheck doesn't have an `ash` mode.

~~~
$ shellcheck
No files specified.

Usage: shellcheck [OPTIONS...] FILES...
  -e CODE1,CODE2..  --exclude=CODE1,CODE2..  exclude types of warnings
  -f FORMAT         --format=FORMAT          output format
  -C[WHEN]          --color[=WHEN]           Use color (auto, always, never)
  -s SHELLNAME      --shell=SHELLNAME        Specify dialect (sh,bash,dash,ksh)
  -x                --external-sources       Allow 'source' outside of FILES.
  -V                --version                Print version information
~~~

I'll extract it to a `shellcheck_dialect` variable with a comment so it's obvious that's where the limitation comes from

mtekel: my point only was that if we call 'ash' in the pipeline instead of sh then the script won't get checked. maybe we should have some other test to ensure we only use 'sh' and 'bash'...

richardc: Ah right.  Yes makes sense, will think of how to do that, I think it's another class of check.

keymon:
What do we do with this PR?

I think the idea is great, but: shall we create a card a work on it?

richardc:
A card was created as the scope expanded to adding unit tests.  https://www.pivotaltracker.com/n/projects/1275640/stories/129536477

It's not been scheduled yet, but in the meantime we can close this PR and keep the branch.

keymon:
+1

jchesterpivotal:
We did [something similar](https://github.com/cloudfoundry/buildpacks-ci/tree/master/tasks/run-shellcheck) on Cloud Foundry Buildpacks team.

For reference, it's run in [this pipeline](https://buildpacks.ci.cf-app.com/teams/main/pipelines/buildpacks-ci)

Jonty:
@jchesterpivotal We implemented this here: https://github.com/alphagov/paas-cf/pull/493

Not quite the same thing, as we actually extract embedded scripts from the pipeline and test them, such as this example: https://github.com/alphagov/paas-cf/blob/c6f066cc8d8901377b8b375338043f62841b4885/concourse/scripts/spec/fixtures/pipecleaner_shellcheck.yml

jchesterpivotal:
I should've guessed!

I've personally become quite dark on embedded scripts and bash scripts generally, as they're fairly unwelcoming to testing. On buildpacks we're writing almost all our new tasks in Ruby and TDDing them.

Thanks for pointing me to pipecleaner -- I actually [asked for a change in validation behaviour](https://github.com/concourse/concourse/issues/709) recently.

PR #450, opened 2016-08-26
[#128160309] Fix missing ssh key input for terraform-destroy

combor:
## What

After separating SSH keys for Concourse and bosh:

https://github.com/alphagov/paas-cf/pull/441
https://github.com/alphagov/paas-cf/pull/437

destroy pipeline is missing those keys during terraform destroy phase.

## How to review

Run destroy pipeline and check if it works.

## Who can review

Not @combor


mtekel:
works, merging. sry I forgot about this...

PR #451, opened 2016-08-26
Only deploy the graphite nozzle if missing/changed

Jonty:
## What

Deploying the nozzle on every run means we lose some stats from the firehose while deployments are in progress.

This deploys the graphite-nozzle with the current nozzle git shorthash in the name. If the current version is missing from the cf deployment it is pushed, and all older versions are removed.

Carried out as a support task, taken from the PaaS Improvements doc.

## How to review

Run the pipeline. The graphite-nozzle app should not be pushed or restarted.

## Who can review

Not @jonty.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1779,7 +1779,7 @@ jobs:
                       PREFIX_JOB: "true"
                   EOF
 
-                  cf push
+                  ../paas-cf/concourse/scripts/cf_push_on_git_change.sh graphite-nozzle
 
         - task: update-kibana-timezone
           config:
--- a/concourse/scripts/cf_push_on_git_change.sh
+++ b/concourse/scripts/cf_push_on_git_change.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+
+APPNAME=$1
+VERSION=$(git rev-parse --short HEAD)
+CURRENT=$(cf apps | grep -E "^${APPNAME}-${VERSION}" | cat)
+OTHERS=$(cf apps | grep -o -E "^${APPNAME}[^ ]*" | grep -E -v "^${APPNAME}-${VERSION}" | cat)
+
+# Delete all other running instances of this app
+if [ "$OTHERS" != "" ]; then
+    echo "$OTHERS" | xargs -n1 cf delete -f
+fi
+
+# And if this version is missing, push it
+if [ "$CURRENT" = "" ]; then
+    cf push "${APPNAME}-${VERSION}"
+fi


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1779,7 +1779,11 @@ jobs:
                       PREFIX_JOB: "true"
                   EOF
 
-                  cf push
+                  # We only push if not running so stats are not disrupted
+                  # during deployments
+                  if ! cf app graphite-nozzle | grep started; then
```
dcarley: I don't think this implementation is correct because subsequent changes to the nozzle application will never be deployed.

Can we use blue-green here to prevent downtime like these stories for the healthcheck app? Or will it result in duplicate metrics?

- [SPIKE: Which zero-downtime plugin for cf push](https://www.pivotaltracker.com/story/show/124843741)
- [Zero-downtime deploy of healthcheck app](https://www.pivotaltracker.com/story/show/121934253)

Jonty: Multiple deployments will result in duplicate metrics AFAICT.

I realised that we wouldn't redeploy the graphite-nozzle, but figured that we touch it so infrequently a manual step to cause redeploy would be the lesser of the two evils - I just totally forgot to write that in the `What`. Sorry.

Ideally this should only re-deploy when the repo it builds from changes, but doing that in the current pipeline isn't feasible.

Jonty: Actually, I think I just figured out a way we may be able to do this. Please hold.

Jonty:
Ok, @dcarley, how about this?

It will still result in a very small amount of downtime when a new version of graphite-nozzle is deployed. By swapping over the last two statements in `cf_push_on_git_change.sh` we can opt for duplicate data rather than downtime if that would be preferred.

Jonty:
Slightly cleaned up, @alext thinks the approach is ok.

concourse/scripts/cf_push_on_git_change.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
richardc: This is causing errors as the cf-cli image doesn't have bash.

~~~
$ docker run -ti governmentpaas/cf-cli sh -i
/ # /bin/bash
sh: /bin/bash: not found
~~~

Or in the context of a pipeline run. 

~~~
sh: ../paas-cf/concourse/scripts/cf_push_on_git_change.sh: not found
~~~

richardc: With the fixup commit 881bb8eefaa7e1f87e29694529f8b0a2f37b15bc this works.

richardc: Overstates the 'works'.  Adding set -x to see where my failure is.

richardc:
Needs a squash, but this is now 👍  in my dev environment.

Jonty:
Squashed with a vengeance.

PR #452, opened 2016-08-26
Make the "small" VM type actually use t2.small

Jonty:
## What

We're currently using `m3.medium` for both "small" and "medium" VM types. Reducing the "small" VM's to being `t2.small` saves 62% of the cost per VM.

Carried out during support, task taken from the PaaS Improvements doc.

## How to review

* Deploy in an environment
* Observe all the m3.medium's being replaced with t2.small instances
* Check the platform still functions

## Who can review

Not @jonty.

### DIFF
--- a/manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
+++ b/manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
@@ -9,7 +9,7 @@ vm_types:
     network: cf
     env: (( grab meta.default_env ))
     cloud_properties:
-      instance_type: m3.medium
+      instance_type: t2.small
       ephemeral_disk:
         size: 10240
         type: gp2


dcarley:
Can you describe what jobs are affected by this change? Are there any that we should be concerned about switching to burstable CPU credits? Can you tell from git or Pivotal whether we have previously changed any of those from t2 to m3?

Jonty:
Jobs affected:
* Consul (3 instances)
* Queue (2 instances)
* Maintenance (1 instance)

Of that, the only one I'd be at all concerned about is Queue (a redis server used for buffering logstash). Perhaps make that a medium?

I'll spelunk for previous changes post-lunch, but it might be tricky given the level of refactoring this has gone through.

Jonty:
After searching around, I really can't find when we last changed these. I have a vague memory that we may have simply bumped the vm_size to m3.medium as we were unsure what sizes were appropriate for most instances at the time.

Personally I think this change should be fine, but I would like to bump Queue to medium.

mtekel:
I checked CPU utilization on queues in CI. It sits at around 0.6% with spikes to 1%. Also, we want to get rid of queues when we upgrade logsearch. Maintenance is mostly idle 0.4% on average with spikes to 0.6%. I'd be a bit afraid about consul, as it is a bit of alchemy how it behaves. T2 should be OK though, I'll test if there are any issues.

image

image

mtekel:
Update went fine (no downtime). Consuls have 2% CPU usage if peer and 3% if leader:
image

Also - kibana instance was affected as well, it normally uses 0.75% CPU, but has some spikes, probably when we use it. Nevertheless, this is very suitable use pattern for the bursty T2 instance type.
image

PR #453, opened 2016-08-26
[#128160309] Fix SSH to deployer concourse.

alext:
## What

#437 changed the deployer concourse to use its own ssh key. We
therefore need to update the Makefile to filter by the correct key name,
and update the instructions to fetch the correct key from S3, otherwise
access won't work.

## How to review

Verify that you can ssh to the deployer-concourse using the updated instructions/Makefile.

## Who can review

Anyone but myself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ pipelines: ## Upload pipelines to Concourse
 showenv: ## Display environment information
 	$(eval export TARGET_CONCOURSE=deployer)
 	@echo CONCOURSE_IP=$$(aws ec2 describe-instances \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -144,7 +144,7 @@ pipelines: ## Upload pipelines to Concourse
 showenv: ## Display environment information
 	$(eval export TARGET_CONCOURSE=deployer)
 	@echo CONCOURSE_IP=$$(aws ec2 describe-instances \
```
mtekel: All fine, I just wonder if we should add `export` in front of `CONCOURSE_IP` like all other env. vars produced by `showenv` do have.

mtekel:
As discussed elsewhere, `export` is out of scope of the PR. Merging.

PR #454, opened 2016-08-30
[#128160309] Fix query of concourse IP with renamed SSH key

keymon:
What
----

With the PR #437 we renamed the key associated to the concourse VM.

But that key name was used to retrieve the concourse IP when running
`make dev showenv DEPLOY_ENV=...`.

We change the name to fix the issue and be able to retrieve the IP from
concourse.

How to test?
------------

run  `make dev showenv DEPLOY_ENV=...` with and without this patch

Who?
---

Anyone but @keymon

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ pipelines: ## Upload pipelines to Concourse
 showenv: ## Display environment information
 	$(eval export TARGET_CONCOURSE=deployer)
 	@echo CONCOURSE_IP=$$(aws ec2 describe-instances \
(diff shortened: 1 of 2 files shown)


alext:
I've added a commit to also update the instructions in the README to account for this change.

mtekel:
what about https://github.com/alphagov/paas-cf/pull/453 ?

alext:
Actually, I forgot I had already opened a PR to address this - https://github.com/alphagov/paas-cf/pull/453

PR #456, opened 2016-08-30
[#128160309] Simplify key presence requirement for bosh destroy

mtekel:
## What

Simplify #450 : For destroy, TF doesn't really need the actual key to be present, as it never checks its contents against AWS. It is enough that empty file is present. This is similar to what we do in destroy-deployer pipeline (#455).

## How to review

Run destroy microbosh pipeline. Check that TF completes successfully.

## Who can review

not @mtekel


richardc:
👍  waiting on travis then will merge.

PR #457, opened 2016-08-30
[#128160309] Init deployments ssh-keys

mtekel:
## What

Without bootstrapping the ssh-key and ssh-key.pub, pipeline won't progress beyond init job, as next job needs to get these keys via S3 resource, which blocks until there are keys in the bucket...
Should have been part of #443. 

## How to review

delete your deployment keys by e.g. deleting your deployment, or by hand from s3 bucket. Run pipeline from start and observe they keys being initialized in the init job.

## Who can review

not @mtekel


alext:
I wonder whether it's worth adding a note to the commit message saying that this is a fix for d1428aa

mtekel:
done

PR #458, opened 2016-08-30
[#123003931] Gate running of availabity tests on healthcheck being deployed

richardc:
## What

[#123003931](https://www.pivotaltracker.com/n/projects/1275640/stories/123003931) Improve "User" (PaaS Operator) Experience for creating  new Environment

Here we introduce a new resource to the pipeline `healthcheck-deployed` which we use to determine if the healthcheck application has been deployed.  If it has not then we skip the task body in the `predeploy/wait-for-availability-test` and `availability-tests/run-tests` tasks.

It will be set `yes` in `post-deploy/deploy-healthcheck` on successful deployment of the healthcheck application, and set back to `no` in the delete and autodelete pipelines. 

We introduce it in two stages:
* Initial version in terms of s3-iam resource and shell scripting
* Encapsulated version implemented in terms of the enum resource type proposed in https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/68

## How to review

* Deploy to a development deployer concourse.
* Run the pipeline observing behavioural changes in the `predeploy/wait-for-availability-test` and `availability-tests/run-tests`.  On first run these tasks should be skipped.  On subsequent runs they should not be, until the enum is flipped or the environment is deleted.

Note: there is some discussion about which iteration of this change we find desirable.  Further comments in the pivotal ticket.

## Dependencies

The second stage depends on https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/68 being merged and built by dockerhub, though there is a `TESTING` temporary commit that points at the feature branch.

## Who can review

Not @richardc, @benhyland, or @HenryTK


richardc:
Premature merge.  I was perhaps not clear, but fd6ae75 is not intended to be merged, and alphagov/paas-docker-cloudfoundry-tools#68 should be reviewed/merged first.

PR #459, opened 2016-08-31
Adjust graphite instance

mtekel:
## What

Reduce graphite instance ephemeral disk and instance size. There's only 0,5G usage on ephemeral disk, reduce from 64G to 10. The instance is using about 7% cpu, reduce from m3.large to t2.medium. 

## How to review

Deploy. Observe graphite being updated. Update preserves all the metrics, but creates about 3 minute gap (which is how long the update takes).

## Who can review

not @mtekel

### DIFF
--- a/manifests/cf-manifest/cloud-config/040-graphite.yml
+++ b/manifests/cf-manifest/cloud-config/040-graphite.yml
@@ -11,7 +11,7 @@ vm_types:
     cloud_properties:
       instance_type: t2.medium
       ephemeral_disk:
-        size: 65536
+        size: 10240
         type: gp2
       elbs:
         - (( grab terraform_outputs.metrics_elb_name ))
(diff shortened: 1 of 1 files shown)


combor:
Its 1,5G for data on prod but still ok to go for 10G disk.
```
/dev/xvdb2       56G  1.6G   52G   3% /var/vcap/data
```

combor:
At the other hand those numbers do not match. I mean `avail` is 52G but `used` 1.6G? I can also see all logs being stored there and this device is shared between data and swap.
```
    Device Boot      Start         End      Blocks   Id  System
/dev/xvdb1               1    15326009     7663004+  82  Linux swap / Solaris
/dev/xvdb2        15326010   134207009    59440500   83  Linux
```

Swap in the new VM is 4G + 4G of space already used leaves 2G of free space. That might be not enough for the future. Shall we go for 15G instead of 10?

mtekel:
I see only 1.5G used, that leaves 4.5G free. Moreover, all our CF VMs (except API & derivates) use 10G ephemeral disk.

combor:
Thats why I said that numbers do not match. Look on `Size` and `Avail`.According to this it's 4G used.

combor:
Hm, most likely it's the `superblock` that is so big. It should be smaller on 10G disk so it should be fine to have 10G

PR #463, opened 2016-09-02
Cleanup bosh properties

mtekel:
## What

Over time we have accumulated various leftover properties. Many of them are not used any more. Others are set to default values for no specific reasons. This can create confusion when debugging problems and looking at those properties assuming they are significant. I have created a lot of smaller commits to explain why we can remove each given property. 

Also, this is a prequel to [Single file for BOSH manifest](https://www.pivotaltracker.com/n/projects/1275640/stories/128858813)

## How to review

Completely delete your CF deployment. Do `bosh cleanup --all` repeatedly until everything (packages, jobs, stemcells) have been cleaned up. Do a fresh deployment of CF (run the deploy-bosh-cf pipeline). Observe BOSH server being re-deployed and CF deployment and tests succeeding.

## Who can review

not @mtekel

### DIFF
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -57,7 +57,6 @@ jobs:
       disks:
         max_orphaned_age_in_days: 0
       cpi_job: (( grab cloud_provider.template.name ))
-      ignore_missing_gateway: "false"
       user_management:
         provider: local
         local:
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -58,7 +58,6 @@ jobs:
         max_orphaned_age_in_days: 0
       cpi_job: (( grab cloud_provider.template.name ))
       user_management:
-        provider: local
         local:
           users:
             - { name: admin, password: (( grab secrets.bosh_admin_password )) }
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -79,8 +79,6 @@ jobs:
         password: (( grab secrets.bosh_hm_director_password ))
       resurrector_enabled: false
 
-    ntp: (( grab meta.ntp ))
-
     agent:
       mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" terraform_outputs.bosh_fqdn ":4222" ))
 
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -33,7 +33,7 @@ jobs:
   - {name: director, release: bosh}
   - {name: health_monitor, release: bosh}
 
-  resource_pool: vms
+  resource_pool: bosh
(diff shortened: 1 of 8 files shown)


manifests/bosh-manifest/deployments/020-compiled-package-cache.yml
```diff
@@ -2,10 +2,7 @@ jobs:
 - name: bosh
   properties:
     compiled_package_cache:
-      provider: s3
```
keymon: I do like to be explicit, even if by default is s3. This makes it more obvious for any other developers and it would make it easier to parse with an automation tool or similar , without having to check the default in the spec.

manifests/bosh-manifest/deployments/010-bosh-template.yml
```diff
@@ -52,8 +52,6 @@ jobs:
       password: (( grab secrets.bosh_nats_password ))
 
     redis:
-      listen_address: 127.0.0.1
-      address: 127.0.0.1
```
keymon: The same than in https://github.com/alphagov/paas-cf/pull/463/commits/f6afdf6925fbaa055f10a163389881dc717a0d5c#r77513103, being explicit is also good. It would help you to easily know where is the redis server running, instead of having to go to the spec to learn the default.

manifests/bosh-manifest/deployments/010-bosh-template.yml
```diff
@@ -58,7 +58,6 @@ jobs:
         max_orphaned_age_in_days: 0
       cpi_job: (( grab cloud_provider.template.name ))
       user_management:
-        provider: local
```
keymon: As commented in https://github.com/alphagov/paas-cf/pull/463/commits/7d19a23b26e6051e1914a91c3c73cdd1b9350e88#r77513225, setting this explicitly  makes it easier to understand.

keymon:
LGTM, but I got some comments about removing  default values.

Setting some stuff explicitly helps to understand the manifest and its code, without having to do the effort to check the defaults. I think we should keep the default values for the blobstores (s3), and the user DB in bosh.

manifests/bosh-manifest/deployments/020-compiled-package-cache.yml
mtekel: I would instead _not include this_ as it is misleading. You can set this to webdav here or local or other random string, it will be still overridden to S3. Essentially, once you specify bucket name, this property is _ignored_. Automation tools should then not parse it at all. The default is in spec, but more importantly is overridden again in erb template.

manifests/bosh-manifest/deployments/010-bosh-template.yml
mtekel: You can expect some reasonable default & that it is working. We don't specify address for anything else but NATS because it is compulsory property. Moreover, for some reason, redis and cc spec use different property names (address and listen_address). If you don't want to rely on defaults (btw. redis will likely be removed in our next BOSH upgrade), then you should add test that the both are same.

keymon:
@mtekel +1, you convinced me.

keymon:
I have tested this with #464 that upgrades to v240. I am fine with it, and I would merge, but I prefer to wait for #464 to be merged first.

mtekel:
Maybe it's safer to merge this first? As this should be noop (well, bosh-init will deploy again, but there should be no changes to the bosh config).

PR #465, opened 2016-09-02
Improve display of stacked graphs on VM dashboard.

alext:
## What

Making the fill colour much stronger makes it much easier to see which
section of the stack correlated to which metric.

Before: 
screen shot 2016-09-02 at 16 23 40

After:
screen shot 2016-09-02 at 16 23 19

## How to review

Deploy these dashboards, and be amazed at the improvement.

## Who can review

Anyone but myself.

### DIFF
--- a/manifests/cf-manifest/grafana/vm-resource-usage.json
+++ b/manifests/cf-manifest/grafana/vm-resource-usage.json
@@ -20,7 +20,7 @@
           "datasource": "graphite",
           "editable": true,
           "error": false,
-          "fill": 2,
+          "fill": 8,
           "grid": {
             "threshold1": null,
             "threshold1Color": "rgba(216, 200, 27, 0.27)",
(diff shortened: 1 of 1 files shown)


mtekel:
I like the current design more. It has transparency which looks like future. The new pastel colors look like google material design. It would be nice if the color palette would be easily configurable so that everyone can select the colors that they like the most. Also, don't forget about the colorblind users...

alext:
This doesn't change the colours used, it just makes them bolder to help with clarity. For example looking at the CPU graph, with the current style, it's not clear whether the upper section is `system` or `steal` (or one of the others), because you're mostly relying on the line colours.

With the new style, it's much more obvious that it's `system` because you can look at the fill colour.

keymon:
I'm with @mtekel I the other palette more. I do not have a problem distinguish them.

The only issue might be that the colors are the same for some metrics (steal and nice) but it is the same issue in both cases.

alext:
I'm more concerned with the ability to interpret the data than the aesthetics.

For example, by looking at the top graph, what's using all the CPU at 14:53? For me, that's not clear with the current config, but it clear with the updated config.

PR #467, opened 2016-09-02
Makefile: add ssh_concourse target

mtekel:
## What

Add ssh_concourse target to makefile and a script it points to to ssh to concourse - for convenience.

## How to review

`make dev ssh_concourse`

## Who can review

not @mtekel

### DIFF
--- a/concourse/scripts/ssh.sh
+++ b/concourse/scripts/ssh.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+key=/tmp/concourse_id_rsa.$RANDOM
+trap 'rm -f $key' EXIT
+
+eval "$(make dev showenv | grep CONCOURSE_IP=)"
+
+aws s3 cp "s3://${DEPLOY_ENV}-state/concourse_id_rsa" $key && chmod 400 $key
+
+echo
+aws s3 cp "s3://${DEPLOY_ENV}-state/generated-concourse-secrets.yml" - | \
+ruby -ryaml -e 'puts "\Sudo password is " + YAML.load(STDIN)["secrets"]["concourse_vcap_password_orig"]' 
+echo
+
+ssh -i $key -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null vcap@"$CONCOURSE_IP"
--- a/Makefile
+++ b/Makefile
@@ -176,3 +176,5 @@ find_diverged_forks: ## Check all github forks belonging to paas to see if they'
 run_job: check-env-vars ##  Unbind paas-cf of $JOB in create-bosh-cloudfoundry pipeline and then trigger it
 	$(if ${JOB},,$(error Must pass JOB=<name>))
 	./concourse/scripts/run_job.sh ${JOB}
+ssh_concourse: check-env-vars ## SSH to the concourse server
+	@./concourse/scripts/ssh.sh


concourse/scripts/ssh.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -eu
+
+key=/tmp/concourse_id_rsa.$RANDOM
+trap "rm -f $key" SIGINT SIGTERM EXIT
```
alext: Are the `SIGINT` and `SIGTERM` necessary here?  I thought that an `EXIT` trap would always fire when the script exited regardless of the cause.

If they are necessary, we need to update the exit traps in several of our other scripts.

concourse/scripts/ssh.sh
```diff
@@ -0,0 +1,15 @@
...
+
+eval $(make dev showenv | grep CONCOURSE_IP=)
+
+aws s3 cp "s3://${DEPLOY_ENV}-state/concourse_id_rsa" $key && chmod 400 $key
+
+echo
+aws s3 cp "s3://${DEPLOY_ENV}-state/generated-concourse-secrets.yml" - | \
+ruby -ryaml -e 'puts "\Sudo password is " + YAML.load(STDIN)["secrets"]["concourse_vcap_password_orig"]'
```
alext: Any reason not to use `val_from_yaml.rb` here?

concourse/scripts/ssh.sh
```diff
@@ -0,0 +1,15 @@
...
+eval $(make dev showenv | grep CONCOURSE_IP=)
+
+aws s3 cp "s3://${DEPLOY_ENV}-state/concourse_id_rsa" $key && chmod 400 $key
+
+echo
+aws s3 cp "s3://${DEPLOY_ENV}-state/generated-concourse-secrets.yml" - | \
+ruby -ryaml -e 'puts "\Sudo password is " + YAML.load(STDIN)["secrets"]["concourse_vcap_password_orig"]' 
+echo
+
+ssh -i concourse_id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null vcap@$CONCOURSE_IP
```
alext: I think this needs the path to the keyfile here. At present, this doesn't work for me, I get the following:
```
$ make dev ssh_concourse
download: s3://alext-state/concourse_id_rsa to ../../../../tmp/concourse_id_rsa.4674

Sudo password is <REDACTED>

Warning: Identity file concourse_id_rsa not accessible: No such file or directory.
Warning: Permanently added '<REDACTED_IP>' (ECDSA) to the list of known hosts.
Unauthorized use is strictly prohibited. All access and activity
is subject to logging and monitoring.
Permission denied (publickey).
make: *** [ssh_concourse] Error 255
```

concourse/scripts/ssh.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -eu
```
alext: We could do with `-o pipefail` here as well. It's also probably more reliable to have a dedicated `set` line instead of relying on the `#!` line. That way they'll apply however this script is invoked (an be consistent with other scripts).

alext:
Thanks for doing this, I've been wanting something like this for a while now, just hadn't got around to it.

concourse/scripts/ssh.sh
mtekel: I wasn't sure exit catches all. I checked again and it does, updated script.

concourse/scripts/ssh.sh
mtekel: To have one less dependency. I wanted to make this as simple as possible & clear what it is doing.

concourse/scripts/ssh.sh
mtekel: updated

mtekel:
Should be working now.

PR #470, opened 2016-09-05
[#123003931] Revert "Convert healthcheck-deployed into an enum"

combor:
## What

This reverts commit a396e0908b33c5bf1328c2a9fe278c4c94994f04.

As we found that enum resource has problems with versions ordering we
will fall back to use s3 based check.

## How to review

*Upload* and run the fresh pipeline and check if availability tests are skipped. Next run it again to check if tests are executed. 

## Who can review

Not @combor


concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -65,13 +60,11 @@ resources:
       versioned_file: bosh-CA.tar.gz
 
   - name: healthcheck-deployed
```
alext: I'm seeing some odd behaviour when concourse has some existing versions for this resource from the enum version. I'm wondering whether we should rename the resource (but not the versioned_file) to force concourse to start from a clean state.

combor:
@alext It's done.

alext:
Looks good. I'm going to leave my pipeline running, and merge in the morning if all looks good then.

PR #472, opened 2016-09-06
DO NOT MERGE: [#128160407] Spike: password rotation pipeline

keymon:
## What

Create a pipeline that automates the process of rotating credentials.

## How to review

This a small implementation of the story [#128160407](https://www.pivotaltracker.com/n/projects/1275640/stories/128160407). Use this PR to comment on the code, but it is not intended to be merged.

Close the PR instead.

## Who can review

Anyone but  @keymon or @richardc

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2329,6 +2329,11 @@ jobs:
               - |
                 spruce merge \
                   cf-secrets/cf-secrets.yml \
+                  paas-cf/concourse/resources/cf-secrets-keep-grafana.yml \
+                  > intermediate.yml
+
+                spruce merge \
+                  intermediate.yml \
                   paas-cf/concourse/resources/cf-secrets-deleter.yml \
                   > modified-secrets/cf-secrets.yml
         on_success:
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -2304,6 +2305,132 @@ jobs:
         params:
           release: pipeline-pool
 
+  - name: reset-cloudfoundry-passwords
```
HenryTK: If we run this job twice and forget the manual step of updating Grafana password in the UI, then we lose the old Grafana password and can't log in. I don't know what recovery of that password is possible, would have to look under the hood of Grafana.

HenryTK:
Looks good to me. Despite the comment above I will probably close this pull request and leave it to implementors.

concourse/pipelines/create-bosh-cloudfoundry.yml
keymon: True. We were aware of that, but we did not really solve it.

HenryTK: Agreed. I think because it is low risk and low severity we don't need to do anything about it.

PR #475, opened 2016-09-07
[#130003999] upgrade terraform to 0.7.3

mtekel:
## What

Adjust pipeline, scripts and TF config to support TF 0.7.3. See https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/69 for TF container upgrade.

## How to review

Apply from bootstrap, then run create-bosh-cf pipeline. Run all destroy pipelines. After https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/69 is merged, remove TMP commit. Test creating pingdom alerts.

## Merging
https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/69 updates container in all environments which will break TF for them. Only merge that PR when you're ready to merge this and when there are no other jobs running in the pipelines, as this has to be 1st PR to go through pipelines after container upgrade.
## Who can review
not @mtekel or @combor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.6.15"
+    - TF_VERSION="0.7.3"
     - SPRUCE_VERSION="1.5.0"
     - DEPLOY_ENV="travis"
 
--- a/concourse/scripts/extract_terraform_state_to_yaml.rb
+++ b/concourse/scripts/extract_terraform_state_to_yaml.rb
@@ -4,8 +4,10 @@ require 'json'
 require 'yaml'
 
 outputs = JSON.load($stdin)
-terraform_outputs = {
-  'terraform_outputs' => outputs['modules'][0]['outputs']
+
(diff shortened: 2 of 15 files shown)


alext:
I see the following change when running the bootstrap pipeline:
```
aws_elb.concourse: Modifying...
  cross_zone_load_balancing: "false" => "true"
```

The default for this has changed to true with 0.7 - https://github.com/hashicorp/terraform/blob/v0.7.0/CHANGELOG.md#070-august-2-2016

alext:
The commit message "Update extract TF state 2 ymt to support new TF" seems to have a typo in it.  I can't actually work out what it's meant to be.

mtekel:
Regarding default change for the ELB - it is a sane default and in this case it doesn't change anything, as we use ELB to terminate SSL. On the other side - maybe we could remove it from other ELB settings. On the another side, for the other ELBs we depend on this option, so maybe it's better to leave it in, just in case terraform changes default again.

alext:
I think the ELB change is fine as is.  It would be good to fix the typo in the commit message, otherwise I think this is good to merge.

PR #479, opened 2016-09-12
[#130003999] Error on old format tfstate file in tfvars extract

alext:
## What

Previously, these scripts would output blank values when given an old
format tfstate file. This is bad, and led to some breakage on staging.

This updates the scripts to error when given an old format tfstate file.

## How to review

Code review. Verify the unit tests pass.
Run enough of the pipeline to be confident that these scripts still work correctly.

## Who can review

Anyone but me.


Jonty:
S'good.

PR #481, opened 2016-09-12
[#128932735] Initial implementation of Datadog

combor:
## What

Story: https://www.pivotaltracker.com/story/show/128932735

This PR brings initial support for Datadog agent on (almost) all out VMs.
Concourse VM has a separate story as this requires significant manifest changes.

- We add tags on on VMs so we can easily find them in Datadog UI.
- We had to disable Dogstatsd due to port conflict
- new terraform job provisions basic router alert 

## How to review
Test together with https://github.com/alphagov/paas-datadog-agent-boshrelease/pull/1
Deploy `create-bosh-cloudfoundry` pipeline from this branch with `ENABLE_DATADOG=true`
Check if tests pass and if router alert works.
Merge https://github.com/alphagov/paas-datadog-agent-boshrelease/pull/1, put a tag on `paas-datadog-agent-boshrelease` and update https://github.com/alphagov/paas-cf/blob/749ec4f581bfdf77346d187e43571a550813b9c9/manifests/cf-manifest/runtime-config/runtime-config-base.yml#L5

## Who can review
Not @mtekel or @combor

### DIFF
--- a/concourse/scripts/pipelines-bosh-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-bosh-cloudfoundry.sh
@@ -45,6 +45,7 @@ prepare_environment() {
   cf_aws_broker_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.aws-broker.version "${cf_manifest_dir}/050-rds-broker.yml")
   cf_os_conf_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.os-conf.version "${cf_manifest_dir}/../runtime-config/runtime-config-base.yml")
   cf_logsearch_for_cloudfoundry_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.logsearch-for-cloudfoundry.version "${cf_manifest_dir}/030-logsearch.yml")
+  cf_datadog_agent_version=$("${SCRIPT_DIR}"/val_from_yaml.rb releases.datadog-agent.version "${cf_manifest_dir}/../runtime-config/runtime-config-base.yml")
 
   if [ -z "${SKIP_COMMIT_VERIFICATION:-}" ] ; then
     gpg_ids="[$(xargs < "${SCRIPT_DIR}/../../.gpg-id" | tr ' ' ',')]"
@@ -56,6 +57,12 @@ prepare_environment() {
   get_git_concourse_pool_clone_full_url_ssh
 
   export EXPOSE_PIPELINE=1
+
+  if [ "${ENABLE_DATADOG:-}" = "true" ]; then
+    export PASSWORD_STORE_DIR=~/.paas-pass
+    datadog_api_key=$(pass datadog/api_key)
+    datadog_app_key=$(pass datadog/app_key)
+  fi
 }
 
 generate_vars_file() {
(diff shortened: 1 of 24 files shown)


concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -54,6 +55,14 @@ prepare_environment() {
 
   download_git_id_rsa
   get_git_concourse_pool_clone_full_url_ssh
+
+  if [ "${ENABLE_DATADOG:-}" = "true" ]; then
+    export PASSWORD_STORE_DIR=~/.paas-pass
```
benhyland: Do we have a list of env vars developers are expected to set or to not set as part of their development environment?
Is this something an individual dev might override?
If so, should we check whether it is already set before overriding it?

manifests/shared/deployments/datadog.yml
```diff
@@ -0,0 +1,14 @@
+---
+releases:
```
benhyland: Quite likely, I don't understand a subtlety of bosh/concourse here.
We specify a git resource for the agent here: https://github.com/alphagov/paas-cf/blob/fc4624c220d341cb8dc86c7435a5c763cd8be612/concourse/pipelines/create-bosh-cloudfoundry.yml#L101-L106
which we then use to create a binary and upload it to bosh.
But then, here, we rely on a manually created release tarball to define the release.
Aren't these achieving the same thing (an agent binary uploaded to bosh) in two different ways?

mtekel: yes, but for bosh-init we need to provide the tarball. That's the only reason we do it that way. Otherwise for CF we do it in the same way as all other releases. Once we have automated release builds, we can re-factor how we provide releases.

Also, https://github.com/cloudfoundry/bosh/issues/1364 - meaning we have to do some workarounds to upload releases specified in runtime/addons.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
mtekel: this is only now for current datadog dev/test account. Once we have a final account we might have a different set(s) of credentials. We'll have to update the credential getting part then.

regarding env. vars you are expected to set when deploying to dev - I think env. name is enough. You might want to add BRANCH if you want to deploy from a different branch.

manifests/shared/deployments/datadog.yml
mtekel: But good point is that now we are uploading the release in 2 ways, since the release with URL gets included in CF manifest (even though it is not needed there, but we merge the datadog file so that we can retrieve meta), so it will get uploaded. I'll remove the git resource and build&upload part then. This way concourse, bosh and all CF VMs will always deploy the same version.

terraform/datadog/datadog.tf
```diff
@@ -0,0 +1,29 @@
...
+
+provider "datadog" {
+    api_key = "${var.datadog_api_key}"
+    app_key = "${var.datadog_app_key}"
+}
+
+resource "datadog_monitor" "router" {
+  name = "${format("%s router hosts", var.env)}"
+  type = "service check"
+  message = "Missing router hosts in environment {{host.environment}}. Notify: @the-multi-cloud-paas-team@digital.cabinet-office.gov.uk"
```
benhyland: After deploying, this check (along with an ntp sync check) is present in the datadog webui, but it looks a bit strange.
Can you explain when {{host.environment}} is supposed to be replaced?
And what the `Notify: @the-multi-cloud-paas-team@digital.cabinet-office.gov.uk` achieves?

mtekel: This is datadog markup. When alert goes off, {{host.environment}} gets replaced by the value of the 'environment' tag of the host that triggered alert. `Notify` directive specifies where to send alert to.

Edit: looking at alerts received, {{host.environment}} gets replaced into ''. Nevertheless, the email mentions environment name explicitly, so {{host.environment}} can be removed.

benhyland: http://docs.datadoghq.com/monitoring/#notifications seems to suggest that the `Notify: ` is not needed - `@target-email` in the message is enough to trigger notification. Perhaps a more complete sentence would be nice, if you are making a change here anyway.

PR #482, opened 2016-09-12
[#130098529] Concourse template level properties

alext:
## What

This is necessary to support the upcoming datadog work. The datadog and
concourse bosh releases both have a tag property, but they're
incompatable (one's an array, and the other a hash etc). To avoid this
conflict the properties need to be specified per template.

Having properties at this level are a feature of the new Bosh 2.0
manifest schema. bosh-init has some limited support for this:

- it treats `instance_groups` as an alias for `jobs`, and within each
  job it treats `jobs` as an alias for `templates`. All the other keys
  within these are as per the 1.0 manifest schema.
- It supports specifying properties at the template(job) level.

Because of these limitations, this retains the 1.0 schema names for the
groups to avoid confusion (where this would look like a 2.0 manifest, but
behave like a 1.0 manifest), and moves all the properties to be
per-template.

It was also necessary to retain the job-level properties key as an empty
hash to prevent bosh-init erroring when merging the properties.

## How to review

Run the bootstrap pipeline. Verify that the updated deployer concourse works correctly (running the main pipeline is a good way to do this.

## Who can review

Anyone but myself.


manifests/concourse-manifest/concourse-base.yml
```diff
@@ -80,41 +132,7 @@ jobs:
...
-        graph_cleanup_threshold_in_mb: 3072
-        max_containers: 1024
-        network_pool: "10.254.0.0/20"
-      baggageclaim:
-        url: "http://127.0.0.1:7788"
-      collectd:
-        hostname_prefix: concourse_
-        interval: (( grab meta.collectd.interval ))
-        config: (( grab meta.collectd.config ))
+    properties: {}
```
mtekel: is this needed?

combor:
LGTM, I tested bootstrap + deployer concourse pipelines and they work. We need to run bootstrap after merging as concourse password has been moved in concourse manifest and that breaks bosh-cli-test.

PR #483, opened 2016-09-13
[#129847885] Upgrade deployer concourse to 2.1.0

alext:
## What

This upgrades the deployer concourse to 2.1.0. This includes upgrading garden-runc to the recommended version, and updating the stemcell to the latest version.

As part of the upgrade we had to address the following:

* Change in how pipelines are exposed
* Fix availability tests to handle new concourse API.
* Slightly update the Chrome plugin

We've noticed a small regression in how serial_groups behave with this version (see LINK for details). This issue won't impact normal usage, and is only an edge case when jobs are paused, so we don't consider this a blocker to upgrading.

## How to review

Run the bootstrap pipeline to upgrade an existing deployer concourse.
Verify that the pipelines work on the upgraded concourse.

## Who can review

Anyone but @HenryTK or myself.


mtekel:
It all works (even chrome plugin), I have no additional comments. Would merge in a controlled manner, as changes to scripts and avail tests will need new concourse to be deployed.

paroxp:
+1

PR #484, opened 2016-09-13
[#130186203] add datadog to concourse

mtekel:
## What

[Add datadog to concourse](https://www.pivotaltracker.com/n/projects/1275640/stories/130186203). This is now possible thanks to https://github.com/alphagov/paas-cf/pull/482. Use same approach and shared files as in #481 

## How to review

Run bootstrap pipeline. Then `make dev ssh_concourse`, sudo su and monit summary. Datadog agent should be present and running. Depending if you did `export ENABLE_DATADOG=true` before creating bootstrap pipelines (make dev bootstrap), the agent would either be "disabled" - logging 403s in the forwarder log (/var/vcap/sys/log/datadog-agent/forwarder.log) or sending metrics to our trial account in datadog, where you should be able to see a host with `concourse` tag (+aws account type + your envirnment name tags).

concourse-dog

## Who can review

not @mtekel or @combor or @paroxp

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -438,6 +438,8 @@ jobs:
       config:
         image: docker:///governmentpaas/spruce
         params:
+          AWS_ACCOUNT: {{aws_account}}
+          DATADOG_API_KEY: {{datadog_api_key}}
           CONCOURSE_MANIFEST_STUBS: |
             ./paas-cf/manifests/concourse-manifest/concourse-base.yml
             predefined-concourse-secrets/predefined-concourse-secrets.yml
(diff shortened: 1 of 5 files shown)


alext:
This is somewhat difficult to review while #481 is in progress. I'd suggest we get that one reviewed first, and then revisit this once it's been rebased.

mtekel:
Significant thing here is only the last commit (this PR will become just 1 commit after rebasing):
https://github.com/alphagov/paas-cf/pull/484/commits/3d3c1fe0e815a027c7d023d07df10391dc37c261

mtekel:
rebased after #481 was merged. Merging this can probably wait till we implement better credential passing and way to control enabling/disabling the agent.

PR #485, opened 2016-09-13
[#129847885] Use native git resource instead of our fork

HenryTK:
🚨 **DO NOT MERGE:**

## What

Concourse 2.1.0 has included our pull request to add GPG signature verification into the git resource, therefore we no longer need to use our fork.

Remove the custom resource type from the pipelines used by the deployer Concourse. As stated in the commit message, we will have to keep using the fork in our bootstrap Concourse as the version is fixed in the AMI.

## How to review

* Create a test branch off of this branch
* Remove `SKIP_COMMIT_VERIFICATION=true` from the dev section of the Makefile (line 89)
* Run the pipeline against a signed commit:
    * It can be any signed commit, doesn't have to be a merge commit: `git commit -S -m 'foo'`
    * Check the `paas-cf` resource is fetched `in` successfully. You only need to run the pipeline-lock job at the start.
* Run the pipeline against an unsigned commit. Repeat as above, but asserting its failure.

## Who can review

Anyone but @alext or me.

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -1,10 +1,5 @@
 ---
 resource_types:
-- name: git-gpg
-  type: docker-image
-  source:
-    repository: governmentpaas/git-resource
-
 - name: s3-iam
   type: docker-image
   source:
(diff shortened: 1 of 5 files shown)


mtekel:
This works. Previous issues were due to switching branch _and_ resource type at the same time. What happens still is that you might get the same version twice when switching the resource type. I guess this is OK. This also alleviates issue where resource doesn't pick the latest commit, but one before - but which is in this case the same version.

versions

PR #491, opened 2016-09-16
[#129847885] Fix run_job make target for new concourse.

alext:
## What

The API endpoints in concourse have changed with the introduction of
teams[1]. The `fly` command has also gained a `trigger-job` command, so
let's use it. This will insulate us against future changes to the API
layout.

[1] http://concourse.ci/downloads.html#v200

## How to review

Use the `run_job` make task and observe that it now works again.

## Who can review

Anyone but myself.


keymon:
Lovely, works!

PR #492, opened 2016-09-16
Make run_job work with concourse 2.1

mtekel:
## What

Concourse 2.1 changed API. Triggering of the job via curl did not work
any more. Luckily, there's new trigger-job command for fly. Also,
concourse team encourages using fly instead of API. Fix the script to
use new command.

## How to review

`JOB=<your-favourite-job> make dev run_job` & see it trigger it.

## Who can review

not @mtekel


mtekel:
this PR came 10 mins too late.

PR #493, opened 2016-09-16
Add shellcheck to pipecleaner

richardc:
## What

Here we add a new class of check to pipecleaner `shellcheck`

When enabled (and it is enabled by default) when linting we look for
tasks where the run property looks like it is an inline shell script,
and invoke `shellcheck` over it to see if the script looks valid.

When an issue is found, it will be reported like so:

    ==  spec/fixtures/pipecleaner_shellcheck.yml  ==
    WARNING * Shellcheck: job='shellcheck', task='bad-compare', ~='
    In - line 2:
    if [ "red" == "blue" ]; then
               ^-- SC2039: In POSIX sh, == in place of = is undefined.
               ^-- SC2050: This expression is constant. Did you forget the $ on a variable?

    '

We also add a ginkgo test for pipecleaner.py, with a test fixture of a
pipeline with an error we know shellcheck will pick up on.

There's some minor refactoring of pipecleaner here to reduce having to
explictly add error classes.

## How to review

* Review the code changes

* Test with the sample pipeline:

~~~
cd concourse/scripts
./pipecleaner.py spec/fixtures/pipecleaner_shellcheck.yml
~~~

* Check the unit tests work

~~~
cd concourse/scripts
ginkgo --focus=PipeCleaner
~~~

* From the top of the repo run `make lint_concourse` and see many shellcheck warnings.

## Who can review

Not @richardc

### DIFF
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -45,7 +45,7 @@ class Pipecleaner(object):
         raw = re.sub('\{\{.*?\}\}', 'DUMMY', raw)
         return yaml.load(raw)
 
-    def call_shellcheck(self, shell, args):
+    def call_shellcheck(self, shell, args, variables):
         """"Returns the exitcode and any output from running shellcheck"""
         if "shellcheck" in self.ignore_types:
             return 0, ""
@@ -55,6 +55,9 @@ class Pipecleaner(object):
         for switch in args[:-1]:
             if switch != "-c":
                 script += "set " + switch + "\n"
+
+        for name, value in variables.iteritems():
+            script += "export " + name + "='DUMMY'\n"
 
         script += args[-1]
 
@@ -170,8 +173,13 @@ class Pipecleaner(object):
                         # shellcheck expands its set of dialects.
                         shellcheck_dialects =  ['sh', 'bash', 'dash', 'ksh']
                         if item['config']['run']['path'] in shellcheck_dialects:
+                            params = item['config'].get('params', {}) or {}
+                            item_params = item.get('params', {}) or {}
+                            combined_params = params.copy()
+                            combined_params.update(item_params)
                             exitcode, output = self.call_shellcheck(item['config']['run']['path'],
-                                                                    item['config']['run']['args'])
+                                                                    item['config']['run']['args'],
+                                                                    combined_params)
                             if exitcode != 0:
                                 errors['shellcheck'].append({
                                     'job': job['name'],
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python
 """Pipecleaner is a tool for validating concourse pipelines.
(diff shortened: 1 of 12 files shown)


Jonty:
Works brilliantly, however I'm hesitant to merge it due to the sheer number of warnings it produces - 1290 lines of them! Unless we deal with them immediately it'll make pipecleaner output almost useless.

@timmow What say you?

richardc:
We could add it disabled.  `--ignore-types=shellcheck` in the Makefile

Jonty:
@richardc I kind of feel like we should run a short story to resolve all these and then merge it, a few of them are probably actual bugs.

Jonty:
We should perhaps disable this shellcheck test (SC1091):
````
In - line 5:
. terraform-variables/cf-secrets.tfvars.sh
^-- SC1091: Not following: terraform-variables/cf-secrets.tfvars.sh was not specified as input (see shellcheck -x).
````

**Comment by Jonty** [comment]
Ok, @richardc and I just walked through the warnings and we think disabling the following warnings is reasonable:
`export SHELLCHECK_OPTS="-e SC2086,SC1091,SC2046"`

**Comment by Jonty** [comment]
We've noticed a fair number of warnings about undefined variables (SC2154) that are defined as environment variables in a concourse `params:` block. @richardc is going to inject dummy variables for each of them to silence this.

File: concourse/scripts/pipecleaner.py [hunk 1]
```diff
@@ -171,7 +177,8 @@ def check_pipeline(self, data):
                         shellcheck_dialects =  ['sh', 'bash', 'dash', 'ksh']
                         if item['config']['run']['path'] in shellcheck_dialects:
                             exitcode, output = self.call_shellcheck(item['config']['run']['path'],
-                                                                    item['config']['run']['args'])
+                                                                    item['config']['run']['args'],
+                                                                    item['config'].get('params', {}))
```
richardc: The logic here isn't quite right.  You can also specify params at the task level. https://concourse.ci/task-step.html#task-params

richardc:
This should now be mergable.  Note there are two unfixed bugs in the pipelines that this shows up, both showing up missing parameters:

~~~
==  paas-cf/concourse/pipelines/create-bosh-cloudfoundry.yml  ==
WARNING * Shellcheck: job='cf-deploy', task='update-kibana-timezone', ~='
In - line 9:
export ES_HOST="${TF_VAR_logsearch_elastic_master_elb_dns_name}"
                ^-- SC2154: TF_VAR_logsearch_elastic_master_elb_dns_name is referenced but not assigned.

'

==  paas-cf/concourse/pipelines/destroy-deployer.yml  ==
WARNING * Shellcheck: job='destroy-concourse', task='destroy-concourse', ~='
In - line 6:
echo -n "${private_ssh_key}" > id_rsa
     ^-- SC2039: In POSIX sh, echo flags are undefined.
         ^-- SC2154: private_ssh_key is referenced but not assigned.

'
~~~

Jonty:
Pipeline tests pass after these changes, all good, merging!

PR #495, opened 2016-09-16
[#124720541] Configure metron_agent as a runtime-config addon

keymon:
[#124720541 Not all logs shipped to kibana/logsearch](https://www.pivotaltracker.com/story/show/124720541)

What
===

We want to get all the logs from all the machines in loggregator. Metron agent configures syslog to forward all the local logs, so we decided to install it in all the VMs.

To do so, we will add it as an addon.

To configure it as a addon, we had to refactor the manifest:
* We need to pass the terraform outputs to the runtime-config.yml  generation, so we can pass the environment config.
* We need to move some common config, like dropsonde_incoming_port to  a `common/metron.yml` file
* We need to move the definition of the cf release to  `common/releases.yml`

How to test?
===========

Travis should pass.

Deploy, it should pass the tests.

Verify that metron agent is running on the VMs and the logs are being ship to kibana

Who?
----

Anyone but @keymon or @richardc

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1363,6 +1363,7 @@ jobs:
                 ./paas-cf/manifests/shared/deployments/collectd.yml
                 ./paas-cf/manifests/shared/deployments/datadog.yml
                 ./paas-cf/manifests/cf-manifest/manifest/env-specific/{{cf_env_specific_manifest}}
+                ./paas-cf/manifests/cf-manifest/common/*.yml
               AWS_ACCOUNT: {{aws_account}}
               DATADOG_API_KEY: {{datadog_api_key}}
             run:
@@ -1408,6 +1409,7 @@ jobs:
                 repository: governmentpaas/spruce
             inputs:
               - name: paas-cf
+              - name: terraform-outputs
             outputs:
               - name: runtime-config
             params:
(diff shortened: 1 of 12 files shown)


keymon:
Rebased against maste

keymon:
@mtekel raised a concern that the approach of splitting the releases.yml is not so good, because means add all the releases to the runtime-config.

I will try a different approach, by setting up a `meta.releases` and fetch the right releases as required.

mtekel:
I prefer to have release defined close to where it is used, e.g. graphite release in 040-graphite.yml. I don't think we use (or intend to use) that release anywhere else...

keymon:
On the other hand, it is good to have common place for all releases with their versions, so you know where that is defined, instead of having to search all around to find where the version of X is defined.

keymon:
@mtekel and me are in a debate. I need to split the releases to a shared file, because it needs to be consumed by runtime-config.yml and the manifests.  To do so, I create a `meta.releases` and then I grab the releases I need in each case.

Doing this, I would like to aggregate every release definition in one place.

But @mtekel prefers to do that only if it is necessary.

For example, we can keep the release definition in the `manifest/050-graphite.yml`, or otherwise, put it in `common/releases.yml` and then do a `grab meta.releases.graphite`.

benhyland:
@keymon I prefer there to be a single way to define/reference releases.

I don't want to have to know that certain releases are special in some way.
So I want to list all releases in one place, and then refer to that list when they are needed elsewhere.

This also means it is easier to enforce invariants around our dependencies, e.g. grouping of releases that must be used together, or always using a single version of a dependency wherever it is needed.

keymon:
I added an additional PR were I move all the releases to the `releases.yml` file, so we can discuss what is the best approach.

keymon:
Rebased with master

keymon:
The PR has been completely refactor to do the very minimum to configure metron_agent as an addon.

PR #496, opened 2016-09-19
[#130182525] Control datadog deployment

mtekel:
## What

[Establish a better way of disabling datadog](https://www.pivotaltracker.com/n/projects/1275640/stories/130182525)

Implement a way that enables you to configure whether you want to have datadog agent deployed or not on 1. Concourse, 2. Bosh, 3. CF VMs

## How to review

Run bootstrap and bosh-cf pipelines with `DEPLOY_DATADOG_AGENT` set to false (default) and true. Verify that agents are being deployed or not. You can also test that setting deploy to false after true will remove the agent.

Note that this currently depends on https://www.pivotaltracker.com/story/show/130379909 being completed and merged before it will work. After that is done, we'll need to rebase and fix merge conflicts.

## Who can review

not @mtekel or @benhyland


richardc:
This appears to add a new way to enable datadog, without touching the old one.  In the context of these changes what is [`ENABLE_DATADOG`](https://github.com/alphagov/paas-cf/blob/control-datadog-deployment-130182525/concourse/scripts/pipelines-bosh-cloudfoundry.sh#L59) now intended to control?

mtekel:
Nothing, we leave it in so that we don't need to deploy changes to ci/staging/prod. We expect #130379909 to figure out the best way of passing credentials, which would involve removing the `ENABLE_DATADOG` and associated code.

richardc:
I'd expected this story to remove/rework `ENABLE_DATADOG`, but sure we can add that to the other story.

benhyland:
With a mini-review by @richardc we also spotted we hadn't dealt with the terraform apply.
We added some commits :)

manifests/cf-manifest/runtime-config/datadog-addon.yml
```diff
@@ -0,0 +1,6 @@
+addons:
+  - name: datadog-agent
```
keymon: I think we should add `- (( append ))`  before this, to be sure that he addons are merged properly. https://github.com/geofffranks/spruce#what-about-arrays

manifests/concourse-manifest/extensions/datadog.yml
```diff
@@ -0,0 +1,10 @@
+jobs:
+  - name: concourse
+    templates:
+      - name: datadog-agent
```
keymon: Again, I think we need `((append))`

mtekel:
Because of merge conflicts with https://github.com/alphagov/paas-cf/pull/500 , removing the last 2 commits:
* TF disabling should be handled in https://github.com/alphagov/paas-cf/tree/control-datadog-terraform - I have cherry picked 5c9c9fe02be0480501926bb6e75714082b561eea to that branch
* the last commit that removes old mechanism - which is also done in #500, so this is unnecessary

manifests/cf-manifest/runtime-config/datadog-addon.yml
mtekel: In this case we don't care about specific order, so we could use (( merge )). As it turns out

> If you don't specify a specific merge strategy, the array will be merged automatically; using keys if they exist (i.e. (( merge )), and array indices otherwise ((( inline ))).

This has been tested to work correctly without (( append )), so I don't see the need to include it here.

manifests/concourse-manifest/extensions/datadog.yml
mtekel: In this case we don't care about specific order, so we could use (( merge )). As it turns out

> If you don't specify a specific merge strategy, the array will be merged automatically; using keys if they exist (i.e. (( merge )), and array indices otherwise ((( inline ))).

This has been tested to work correctly without (( append )), so I don't see the need to include it here.

manifests/cf-manifest/runtime-config/runtime-config-base.yml
```diff
@@ -1,7 +1,4 @@
 ---
-meta:
```
mtekel: I suggest if other things need it they should add it explicitly and describe the need in the commit message.

manifests/cf-manifest/spec/support/manifest_helpers.rb
```diff
@@ -83,6 +83,7 @@ def load_runtime_config
     runtime_config = render([
       File.expand_path("../../../../shared/build_manifest.sh", __FILE__),
       File.expand_path("../../../runtime-config/runtime-config-base.yml", __FILE__),
+      File.expand_path("../../../runtime-config/datadog-addon.yml", __FILE__),
```
mtekel: this is to render runtime config. We have decided to test with this added. Ideally, we'd have matrix of tests where we test for every feature enabled/disabled. But for that, we at least need unified build/render scripts.

manifests/cf-manifest/runtime-config/datadog-addon.yml
keymon: The case I was describing was this one: https://github.com/geofffranks/spruce/issues/171 

But I agree that the issue should be solved by whoever encounters the problem.

keymon:
REbased with master

keymon:
LGTM

PR #497, opened 2016-09-20
Add guidance about disabling acceptance tests

Jonty:
## What

This just bit me.

## How to review

Agree with it. Ideally test that this is true.

## Who can review

Not @jonty.


keymon:
+1 but I would go further and put `SELF_UPDATE_PIPELINE=false` in example the command before: 

``` 
SELF_UPDATE_PIPELINE=false ENABLE_CF_ACCEPTANCE_TESTS=false make dev pipelines
```

Jonty:
@keymon Good point, resolved.

PR #498, opened 2016-09-20
Add make target to find textfiles with missing trailing newlines

benhyland:
## What

Add make target to find textfiles with missing trailing newlines
Despite the branch name, this doesn't yet enforce anything.

If people like it, we could add it to travis later on.

## How to review

Run `make lint_posix_newlines` and see whether it detects text files without trailing newline characters.

## Who can review

Anyone.


keymon:
For me is fine, but it has no associated story. ICan we add a story for pivotal, or associate it to something?

Apart of that, we should ignore vendored stuff (e.g. `/platform-tests/src/availability/vendor/` 

Then, to add this to travis, it is as easy as adding the new task as dependency of the `test` target.

Jonty:
This is really great! I took the liberty of pushing a tiny commit to ignore vendored deps, exit with a non-zero code when the test fails, and run it as part of the main test run.

Feel free to ignore or squash it at will.

keymon:
I would do the following improvements:

 1. The script can check multiple files at the same time.
 2. The script should return error if it fails,a nd print some informative message.
 3. We should avoid logic in the makefile (check for output)
 4. use git ls-files to find the files to check

keymon:
I did a commit with the proposal, feel free to squash or whatever

benhyland:
@Jonty @keymon thanks for picking this up.
I'm happy with whatever changes you see fit to make, but I'm not sure that the current version is correct - does `git ls-files` avoid testing binary files?

keymon:
true, `git ls-files` does not know about text files. But neither did the find.

Maybe a check using `file`: ``file --mime-type README.md` and grep for `text/

benhyland:
The first commit excluded non-text files with `grep -Iq .`

keymon:
Oh, I did not know that grep -I. I added it back.

combor:
Is there anything that needs to be fixed for this PR ?

keymon:
@combor I mention the lack of backing story to refer to. Then there are a pair of suggestions that jonty and me added, which maybe @benhyland wants to squash or not.

benhyland:
Personally I don't particularly want/need a backing story or to squash the commits. Are either of these a requirement to get this merged?

saliceti:
@benhyland can you squash the commits? I don't feel keep them separate is valuable in this case.

benhyland:
@saliceti squashed

PR #499, opened 2016-09-20
Add test for serial ingestor deploys

benhyland:
## What

Add test for serial ingestor deploys

I happened to be playing with the job_spec before story https://www.pivotaltracker.com/story/show/130730169 was picked up.

This PR adds a test for the expected workaround, and improves the feedback to the dev when trying to write similar tests using an incorrect job name.

## How to review

Implement a solution for https://www.pivotaltracker.com/story/show/130730169 and see if this test passes.

## Who can review
@mtekel

### DIFF
--- a/manifests/cf-manifest/spec/manifest/jobs_spec.rb
+++ b/manifests/cf-manifest/spec/manifest/jobs_spec.rb
@@ -12,7 +12,12 @@ RSpec.describe "the jobs definitions block" do
   let(:jobs) { manifest_with_defaults["jobs"] }
 
   def get_job(job_name)
-    jobs.select { |j| j["name"] == job_name }.first
+    job = jobs.select { |j| j["name"] == job_name }.first
+    if job == nil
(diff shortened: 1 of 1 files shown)


manifests/cf-manifest/spec/manifest/jobs_spec.rb
```diff
@@ -37,6 +44,13 @@ def get_job(job_name)
     end
   end
 
+  describe "in order to enforce metron dependency on ingestor" do
```
mtekel: how about "in order to ensure high availability of ingestor"

because this doesn't enforce any dependencies in any way

benhyland: gp, done

mtekel:
Cherry picked this into https://github.com/alphagov/paas-cf/commits/fix-VM-update-settings-130730169 and will make it part of PR for that story as it tests for changes made in that branch...

PR #503, opened 2016-09-22
[#130182525] Control datadog terraform

mtekel:
## What

We now have real datadog keys available for our environments, so we can get rid of "disabled" workaround. This PR also adds functionality to clean up datadog checks after datadog is disabled - when you run deploy pipeline again and when you run autodelete/destroy pipelines.

## How to review

Upload dev datadog keys to your dev env (`make dev upload-datadog-secrets`). Deploy with datadog enabled (`DEPLOY_DATADOG_AGENT=true make dev pipelines`). Run the deploy pipeline. Then disable datadog & upload your pipeline (`make dev pipelines`). Run deploy pipeline again & verify that terraform does clean up previously created checks. Then also check in autodelete/destroy pipelines that terraform tasks there run cleanup. You can do so by e.g. running these pipelines. (hint: edit your pipelines locally, place `exit 0` in real destroy tasks so that you skip them - if you don't really want to destroy your environment).

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -93,6 +93,7 @@ jobs:
           TF_VAR_datadog_api_key: {{datadog_api_key}}
           TF_VAR_datadog_app_key: {{datadog_app_key}}
           TF_VAR_env: {{deploy_env}}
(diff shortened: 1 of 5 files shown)


keymon:
:+1:

PR #506, opened 2016-09-23
[SUPPORT TASK] Concourse 2.2.1 upgrade

combor:
## What

This release fixes the problem with disk space that we had a lot of
trouble with:

> "BOSH-deployed workers will now be named after a frankenguid taking
parts from the BOSH instance ID and their hostname. This is to make it
so you can correlate the worker to the BOSH instance, while also
guaranteeing that when the worker is recreated it comes back under a new
name."

As we use bosh-init to deploy Concourse the name was always "unknown".

> "the issue is that it comes back with the same name every time
to concourse, a worker coming back with the same name means all the
volumes that were previously on it should still be there
so it probably gets confused when it sees the imported volume for it in
the db and then it's not there"

Release notes promises big improvements to resources candidate selection
algorithm that should bring CPU usage down.

http://concourse.ci/downloads.html#v221
http://concourse.ci/downloads.html#v220

## How to review

- run the bootstrap pipeline
- run the deployer pipeline
- all tests should pass

## Who can review

Not image

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -6,11 +6,11 @@ name: concourse
 
 releases:
   - name: concourse
-    url: https://github.com/concourse/concourse/releases/download/v2.1.0/concourse-2.1.0.tgz
-    sha1: 753bb48c73c77e8881040d4c60a7b882ec56b228
+    url: https://bosh.io/d/github.com/concourse/concourse?v=2.2.1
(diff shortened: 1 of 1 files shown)


keymon:
cool :+1:!

I am deploying it now

keymon:
```
$ ./bin/fly -t hector workers
name                                 containers  platform  tags  team
unknown-cd3a-4a2b-4ae7-b89def895956  16          linux     none  none
``` 

It has the new ID :)

keymon:
Seems good for me, merging.

PR #507, opened 2016-09-23
[#130921001] Add two new cells in zone 3, for a total of 6.

Jonty:
https://www.pivotaltracker.com/story/show/130921001

## What

Currently if we lose a zone we won't have enough capacity to support all deployed instances of apps on the platform. This adds two more cells in zone 3, thus meeting our requirement of being able to support all currently deployed apps on zone failure and lessening our reliance on z1 & z2.

This also adds a new makefile target that invokes @saliceti's script to show current cluster memory usage.

## How to review

* Edit `paas-cf/manifests/cf-manifest/manifest/env-specific/cf-default.yml` and increase the number of instances to 6 so that your development environment matches the new production configuration.
* Temporarily commit and deploy this to an environment.
* Log in to the AWS console and verify that your environment now has 6 cells spread across three zones.
* If you're feeling extra-cautious, log in to the BOSH CLI and verify the same information by running `bosh vms`.

## Who can review

Not I!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,3 +209,7 @@ tunnel: check-env-vars ## SSH tunnel to internal IPs
 
 stop-tunnel: check-env-vars ## Stop SSH tunnel
 	@./concourse/scripts/ssh.sh stop
+
+show-cf-memory-usage: ## Show the memory usage of the current CF cluster
+	$(eval export API_ENDPOINT=https://api.${SYSTEM_DNS_ZONE_NAME})
+	@./scripts/show-cf-memory-usage.rb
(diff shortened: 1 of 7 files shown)


keymon:
When doing a PR, can you add the story number to the card and the link? #130921001  https://www.pivotaltracker.com/story/show/130921001 

I updated this one.

keymon:
Can the commit be a little bit more descriptive, explaining that also add the network definition for which you require the terraform output?

keymon:
Seems to be OK and tests seem to be passing. I will merge on Monday if everything is OK.

keymon:
It works as expected and passes all the tests. 

I will merge, but if you don't mind please drop a pair of lines describing why the additional changes to setup the network and that it is one-off in this case.

PR #508, opened 2016-09-26
[#127970229] Enable RDS Broker to Create Final Snapshot

paroxp:
## What

We have updated the broker to create a final snapshot of a database instance
upon its deletion. We have added custom acceptance tests to check for this creation and assert a user can override this behaviour when provisioning a database instance, and by using `cf update-service`.

We had to fix some of the configuration, as it was misplaced. We've enabled the `allow_user_provision_parameters`, allowing to pass some parameters into the command line on execution. We've created our own helper and prepared a set of tests, checking the setup while the pipeline is being run.

## How to review

The review of this pull request is dependent on the merging of [pull request 13](https://github.com/alphagov/paas-aws-broker-boshrelease/pull/13) in the paas-aws-broker-boshrelease repository. Only once that pull request is merged can we remove the temporary commit `[TMP] use broker release from branch `.

The commit `[TMP] - pin cf-acceptance-tests image to version which has ginkgo` can be removed once [pull request 71](https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/71) on paas-docker-cloudfoundry-tools has been merged and the docker build completes.

* Deploy using this branch
* Check the output of the `cf-deploy` task in the `cf-deploy` job to ensure you have picked up the correct version of the RDS Broker release.
* Check the custom acceptance tests pass
* Check the failure testing pipeline still works. We have re-ordered the jobs and changed the container used by the *rds-broker* job.
* If you want to be really thorough you can repeat the two manual test cases documented in [pull request 17](https://github.com/alphagov/paas-rds-broker/pull/17) of the paas-rds-broker repository.

## Before merging

Make sure https://github.gds/government-paas/aws-account-wide-terraform/pull/59 was merged and applied to ci/staging/prod.

## Who can review

Anyone but: @combor, @benhyland, @paroxp, @mtekel or @HenryTK

### DIFF
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -19,7 +19,7 @@ meta:
 
 releases:
   - name: aws-broker
-    version: 0.0.9
+    version: 0.0.10
 
 jobs:
   - name: rds_broker
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -10,7 +10,7 @@ meta:
       multi_az: false
       publicly_accessible: false
       copy_tags_to_snapshot: true
-      skip_final_snapshot: true
+      skip_final_snapshot: false
       backup_retention_period: 7
       db_subnet_group_name: (( grab terraform_outputs.rds_broker_dbs_subnet_group ))
       vpc_security_group_ids:
@@ -39,6 +39,8 @@ jobs:
         job: rds_broker
         tags: (( inject meta.datadog_tags ))
       rds-broker:
+        allow_user_provision_parameters: true
+        allow_user_update_parameters: true
         aws_access_key_id: ""
         aws_secret_access_key: ""
         aws_region: "eu-west-1"
@@ -63,7 +65,7 @@ jobs:
                 providerDisplayName: "Amazon Web Services"
                 documentationUrl: "https://aws.amazon.com/documentation/rds/"
                 supportUrl: "https://forums.aws.amazon.com/forum.jspa?forumID=60"
-              plan_updateable: false
+              plan_updateable: true
               plans:
                 - id: "9b882524-ab58-4c18-b501-d2a3f4619104"
                   name: "M-dedicated-9.5"
--- a/platform-tests/src/acceptance/rds_broker_test.go
+++ b/platform-tests/src/acceptance/rds_broker_test.go
@@ -4,8 +4,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os/exec"
+	"regexp"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
@@ -15,9 +19,15 @@ import (
 	. "github.com/onsi/gomega/gexec"
 )
 
+const (
+	DB_CREATE_TIMEOUT = 30 * time.Minute
+)
+
 var _ = Describe("RDS broker", func() {
 	const (
-		serviceName = "postgres"
+		serviceName  = "postgres"
+		testPlanName = "M-dedicated-9.5"
+		region       = "eu-west-1"
 	)
 
 	It("should have registered the postgres service", func() {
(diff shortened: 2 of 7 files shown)


manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -37,6 +36,8 @@ jobs:
         job: rds_broker
         tags: (( inject meta.datadog_tags ))
       rds-broker:
+        skip_final_snapshot: false
```
mtekel: is this now a global parameter? was it ignored before? what is a default if the per-plan parameter is removed?

manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -18,7 +18,7 @@ meta:
 
 releases:
   - name: aws-broker
-    version: 0.0.9
+    version: latest
```
mtekel: is this a TMP or do we really want to unpin the version?

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -124,6 +127,72 @@ var _ = Describe("RDS broker", func() {
...
+			}, DB_CREATE_TIMEOUT, 15*time.Second).ShouldNot(Say(dbInstanceName))
+			fmt.Fprint(GinkgoWriter, "done\n")
+
+			dbInstance, err := rdshelper.NewRDSDBInstance(region)
+			Expect(err).NotTo(HaveOccurred())
+			snapshots, err := dbInstance.GetDBFinalSnapshots(rdsInstanceName)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).Should(ContainSubstring("DBSnapshotNotFound"))
+			fmt.Fprint(GinkgoWriter, "Final snapshots for %s:\n", rdsInstanceName)
+			fmt.Fprint(GinkgoWriter, snapshots)
```
mtekel: is this always expected to print empty list of snapshots? As error is expected to have occured and it is expected to habe DBSnapshotNotFound string in the error?

manifests/cf-manifest/manifest/050-rds-broker.yml
combor: Yes. Yes. `false`

manifests/cf-manifest/manifest/050-rds-broker.yml
combor: TMP. This is WIP PR. It needs cleaning.

platform-tests/src/acceptance/rds_broker_test.go
combor: No, it should be removed. WIP

combor: When it comes to error - we check if snapshot was created. If it doesn't exist you will get 404 with error string. If the returned error is not 404 or the string is different than the test should fail.

keymon:
The tests run fine and I am happy with then. 

They fail when the feature is not implemented (using old broker) and work when it is implemented (using new broker).

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -124,8 +133,108 @@ var _ = Describe("RDS broker", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(snapshots).Should(ContainSubstring(rdsInstanceName))
+
+			snapshotDeletionOutput, err := rdsClient.deleteDBFinalSnapshot(rdsInstanceName)
+			fmt.Fprintf(GinkgoWriter, "Snapshot deletion output for %s:\n", rdsInstanceName)
+			fmt.Fprint(GinkgoWriter, snapshotDeletionOutput)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
 })
```
Jonty: I can't help but feel that the following three functions could be refactored into one, however in the interest of getting The Neverending Story merged I think we should ignore this.

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -134,3 +243,48 @@ func quietCf(program string, args ...string) *Session {
...
+	rdssvc := rds.New(sess)
+	return &RDSClient{
+		region: region,
+		rdssvc: rdssvc,
+	}, nil
+}
+
+func (r *RDSClient) GetDBFinalSnapshots(ID string) (*rds.DescribeDBSnapshotsOutput, error) {
+	params := &rds.DescribeDBSnapshotsInput{
+		DBSnapshotIdentifier: aws.String(ID + "-final-snapshot"),
```
Jonty: This identifier should really be generated in one place rather than having logic duplicated on https://github.com/alphagov/paas-cf/pull/508/files#diff-ebbb4adbeb7f9ea59f6a7015a3a3afbaR281 to avoid future errors if we change it. Again, not enough to block merging this.

platform-tests/src/acceptance/run_tests.sh
```diff
@@ -5,7 +5,7 @@ set -eu
 godep restore
 
 if [ -n "${GINKGO_FOCUS:-}" ]; then
-  go test -timeout 30m -ginkgo.focus "${GINKGO_FOCUS}"
+  ginkgo -p -nodes=16 -focus="${GINKGO_FOCUS}"
```
HenryTK: To my surprise I found Ginkgo doesn't supply the same timeout parameter, we would be forced to refactor it so timeouts are set in the code. As a quicker fix should we set a timeout in the pipeline at task level?

mtekel: Reasoning behind this is: 
* if this times out, custom acceptance tests fail, job fails, pipeline is blocked
* if this takes forever, job keeps running, pipeline is blocked

The main negative effect here is blockage of the pipeline, which is the same in both cases. However without timeout, there's chance that test might finish after some time & unblock pipeline. Moreover it leaves logs behind which might help analyze why it took so long.

Jonty:
Ok, everything dependent is correctly merged, this PR has been rebased and cleaned, and FINALLY WE ARE DONE.

This was extremely unpleasant to review/merge.

PR #509, opened 2016-09-26
[SUPPORT] Fix CI healthcheck DB problem.

combor:
## What

This is temporary fix to allow provisioning of more security groups to
allow more than 96 pingdom IPs ( 32 per group ) to access our
infrastructure. Right now pingdom reports 108 IPs  and some of them (for instance Amsterdam and Düsseldorf ) can't access `/db` on healthcheck app failing the check. 

## How to review

Run the pipeline and check if `cf-terraform` created `aws_security_group.pingdom-probes-3`

## Who can review

Not @combor

### DIFF
--- a/terraform/cloudfoundry/router_elb.tf
+++ b/terraform/cloudfoundry/router_elb.tf
@@ -7,7 +7,8 @@ resource "aws_elb" "cf_router" {
     "${aws_security_group.web.id}",
     "${aws_security_group.pingdom-probes-0.id}",
     "${aws_security_group.pingdom-probes-1.id}",
-    "${aws_security_group.pingdom-probes-2.id}"
+    "${aws_security_group.pingdom-probes-2.id}",
+    "${aws_security_group.pingdom-probes-3.id}"
   ]
   access_logs {
     bucket = "${aws_s3_bucket.elb_access_log.id}"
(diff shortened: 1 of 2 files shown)


mtekel:
How about adding _4 as well? So that we don't have to solve the same problem (when _3 space runs out) again soon.

combor:
I think we need to fix it properly. This is temporary. I'll create pivotal story to re-factor it.

combor:
@mtekel story created: https://www.pivotaltracker.com/n/projects/1275640

benhyland:
The correct link for the follow up is https://www.pivotaltracker.com/story/show/131074821

benhyland:
@combor should we be using the new security group in `router_elb.tf`?

At the moment I think this PR just creates a new security group but doesn't associate it with any other AWS resource. Is there some other way this association happens that I haven't found?

combor:
@benhyland that's correct. Fixed.

terraform/cloudfoundry/router_elb.tf
```diff
@@ -8,6 +8,7 @@ resource "aws_elb" "cf_router" {
     "${aws_security_group.pingdom-probes-0.id}",
     "${aws_security_group.pingdom-probes-1.id}",
     "${aws_security_group.pingdom-probes-2.id}"
```
benhyland: Sorry dude, travis demands an additional comma and I don't have the authority to overrule it.

benhyland:
This looks good now.

PR #510, opened 2016-09-26
Allow login to fail during `remove-healthcheck-db`

richardc:
## What

If the API is up, it's still possible that you may not be able to login,
say if UAA is down.

Here we allow the `cf login` to fail also.

## How to review

Turn off UAA then delete a deployment.

## Who can review

Not @richardc


combor:
LGTM

PR #511, opened 2016-09-26
[#130814387] Metrics for concourse

mtekel:
## What

Send concourse metrics to datadog and create a simple graph that shows time difference in job runtime for each job.

## How to review

Upload datadog secrets to your dev environmnent if you haven't done so already (`make dev upload-datadog-secrets`) then `DEPLOY_DATADOG_AGENT=true make dev bootstrap`, apply bootstrap pipeline.

Run pipeline on your concourse several time (this is to get some metrics). Log in to datadog. You should see metrics starting with `<env_name>_concourse.`. Created graph (" <env> job runtime difference") should show differences in run times of your jobs.

## Who can review

not @mtekel or @saliceti

### DIFF
--- a/manifests/concourse-manifest/extensions/datadog.yml
+++ b/manifests/concourse-manifest/extensions/datadog.yml
@@ -1,3 +1,9 @@
+releases:
+  - name: riemann
+    url: https://github.com/alphagov/paas-cf/releases/download/riemann-1/riemann-gds-1.tgz
+    sha1: 8a4220a039ce6a2a0383aab8d4a76746501e58ec
+    version: gds-1
+
 jobs:
   - name: concourse
     templates:
@@ -8,3 +14,9 @@ jobs:
           tags:
             job: concourse
             tags: (( inject meta.datadog_tags ))
+      - name: riemann
+        release: riemann
+        properties:
+          riemann:
+            datadog:
+              api_key: (( grab meta.datadog.api_key ))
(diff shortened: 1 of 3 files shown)


manifests/concourse-manifest/extensions/datadog.yml
```diff
@@ -1,3 +1,9 @@
+releases:
+  - name: riemann
+    url: https://s3-eu-west-1.amazonaws.com/michael-state/riemann-2%2Bdev.1.tgz
```
benhyland: Could we stash this somewhere that isn't associated with a particular environment?

mtekel: Good point. Given that this is not a fork and we don't have our final bucket created yet (and pipeline to do automated builds in place), I'll create a release here on paas-cf and add the tarball there.

mtekel: done

benhyland:
@mtekel 

Sorry, I wasted time on this because I didn't set the branch when I ran the pipeline :/

Review still ongoing, but I'd be interested to know what you think the graph is useful for.

Without having tested it yet, it looks like datadog wont give me a graph over the discrete events, but instead averages over some time period (the smallest option being 1 hr). I don't think this is going to be helpful except for seeing very general trendlines. Can we do better?

benhyland:
We agreed offline that the graph may or may not turn out to be useful; that we might as well include it as a starting point and improve things later; and in any case, the terraform provider currently makes it difficult to specify good graphs - this may change soon or we can change it ourselves.

I've now seen metrics coming in so all looks good for merge.

PR #514, opened 2016-09-27
Add cell memory usage dashboard

Jonty:
## What

As part of [120268263](https://www.pivotaltracker.com/n/projects/1275640/stories/120268263) we want to have visibility over how much memory is currently in use in the cell cluster, how much we have available in total, and how much we'll have after losing a cell and losing an AZ.

This PR introduces a new default grafana dashboard that shows all of these metrics.

screen shot 2016-09-27 at 14 54 15

*Note:* We currently have no way of passing the number of zones into the manifest at build time so the scale factor representing the number of zones after a failure (2) is currently [hardcoded in this template](https://github.com/alphagov/paas-cf/pull/514/files#diff-a013cd33b85234fd1ce5f574c7a8196cR30) as "0.6666" for the AZ failure graph. We need to revisit this if we introduce another zone.

## How to review

Alter your `manifests/cf-manifest/manifest/env-specific/cf-dev.yml` to deploy 6 instances and temporarily commit it. Deploy this in an env.

Visit `https://metrics.${DEPLOY_ENV}.dev.cloudpipeline.digital/dashboard/file/cell-memory-usage.json` and view the new dashboard. Check it is correct.

## Who can review

Not I, nor any of my clones.

### DIFF
--- a/manifests/cf-manifest/grafana/cell-memory-usage.json
+++ b/manifests/cf-manifest/grafana/cell-memory-usage.json
@@ -1,0 +1,155 @@
+{
+  "id": 5,
+  "title": "Cell Memory Usage",
+  "originalTitle": "Cell Memory Usage",
+  "tags": [],
+  "style": "dark",
+  "timezone": "utc",
+  "editable": true,
+  "hideControls": false,
+  "sharedCrosshair": false,
+  "rows": [
+    {
+      "height": "600px",
+      "panels": [
+        {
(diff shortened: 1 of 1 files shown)


saliceti:
Please add auto refresh 30s (like the other dashboards).

The graph shows the memory *actually used* by apps. It would be interesting to know the memory *allocated* as well.

Jonty:
@saliceti Autorefresh added.

PR #517, opened 2016-09-29
[#120590777] Fix datadog hyphen bug

saliceti:
## What
Story: [Improve Visibility of Continuous Smoke Tests](https://www.pivotaltracker.com/story/show/120590777)

This has impacted me as the datadog terraform run failed for me because I have a hyphen in my environment name (`colin-sept`). We've noticed riemann actually converts hyphens to underscores automatically when it sends metrics to datadog. This does the same for our metrics, and change the queries accordingly.

## How to review
* Build an environment with hyphen(s) in the name
* It should not break in `datadog-terraform-apply`
* The dashboards should work in datadog

## Who can review

Anyone but @paroxp or myself


keymon:
Works as advertised.

PR #518, opened 2016-09-29
Pipeline shellcheck fixes

Jonty:
## What

This resolves all the remaining shellcheck errors that pipecleaner finds in the concourse pipelines and then makes shellcheck warnings fatal, as we aim for shellcheck zero.

This also fixes the deployer destroy pipeline which was broken when we refactored SSH key handling - the bug was spotted by shellcheck.

## How to review

1. Run the pipeline and ensure the tests pass, ideally in a fresh environment.

## Who can review

You.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1835,14 +1835,16 @@ jobs:
               - -e
               - -c
               - |
-                VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
-                SCRIPTS_PATH="./paas-cf/concourse/scripts"
-                TF_VARS_SCRIPT="extract_tf_vars_from_terraform_state.rb"
-                $SCRIPTS_PATH/$TF_VARS_SCRIPT < cf-tfstate/cf.tfstate | grep logsearch_elastic_master_elb_dns_name > ./logsearch_var.sh
-                . ./logsearch_var.sh
-                export ES_HOST="${TF_VAR_logsearch_elastic_master_elb_dns_name}"
+                SCRIPT_DIR="./paas-cf/concourse/scripts"
+
+                ruby "${SCRIPT_DIR}/extract_terraform_state_to_yaml.rb" \
+                  < cf-tfstate/cf.tfstate \
+                  > cf-tfstate.yaml
+
+                export ES_HOST
+                ES_HOST=$(ruby "${SCRIPT_DIR}/val_from_yaml.rb" terraform_outputs.logsearch_elastic_master_elb_dns_name cf-tfstate.yaml)
                 export ES_PORT="9200"
-                $(pwd)/paas-cf/concourse/scripts/kibana_set_utc.rb
+                ruby "${SCRIPT_DIR}/kibana_set_utc.rb"
 
         - task: datadog-terraform-apply
           config:
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1867,7 +1867,7 @@ jobs:
                 - -e
                 - -c
                 - |
-                  if [ -n "${TF_VAR_datadog_api_key}" -a -n "${TF_VAR_datadog_app_key}" -a "${ENABLE_DATADOG}" = "false" ]; then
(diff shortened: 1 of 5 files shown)


concourse/scripts/pipecleaner_test.go
```diff
@@ -77,8 +77,8 @@ var _ = Describe("PipeCleaner", func() {
 			})
 
 			It("should warn about the non-portable compare, and exit as though fatal", func() {
```
richardc: Rather than adjusting this test, it should be removed.

It was testing that a warning would fail in the context of "with --fatal-warnings", but it's not a warning, so the entire test is meaningless now.

Jonty: Oh what, I _did_ remove that test! Must have hit undo one-too-many-times.

richardc: 👍

richardc:
Not sure I I used the GH review thing wrong, the change I requested was done, but that was just a partial review of the changes to the testing of pipecleaner.  I haven't reviewed the rest

concourse/pipelines/destroy-deployer.yml
```diff
@@ -78,7 +87,7 @@ jobs:
             [ "$(cat concourse-bosh-state/concourse-manifest-state.json)" = "{}" ] \
               && echo Concourse bosh state empty, assuming already destroyed \
               && exit 0
-            echo -n "${private_ssh_key}" > id_rsa
```
mtekel: where was this variable coming from and does it need to be removed from there?

Jonty: It was removed in https://github.com/alphagov/paas-cf/commit/4e4c5fce2c85e42226aa801ebd6df54469dbe565 but destroy-deployer wasn't completely fixed at the time.

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1867,7 +1867,7 @@ jobs:
                 - -e
                 - -c
                 - |
-                  if [ -n "${TF_VAR_datadog_api_key}" -a -n "${TF_VAR_datadog_app_key}" -a "${ENABLE_DATADOG}" = "false" ]; then
+                  if [ -n "${TF_VAR_datadog_api_key}" ] && [ -n "${TF_VAR_datadog_app_key}" ] && [ "${ENABLE_DATADOG}" = "false" ]; then
```
mtekel: what exactly is the shellcheck error? Is this correctly evaluated for 'sh' ? the difference here is that && and || are evaluated by shell, vs -a and -o being evaluated by test...

concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -1835,14 +1835,16 @@ jobs:
...
               - |
-                VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
-                SCRIPTS_PATH="./paas-cf/concourse/scripts"
-                TF_VARS_SCRIPT="extract_tf_vars_from_terraform_state.rb"
-                $SCRIPTS_PATH/$TF_VARS_SCRIPT < cf-tfstate/cf.tfstate | grep logsearch_elastic_master_elb_dns_name > ./logsearch_var.sh
-                . ./logsearch_var.sh
-                export ES_HOST="${TF_VAR_logsearch_elastic_master_elb_dns_name}"
+                SCRIPT_DIR="./paas-cf/concourse/scripts"
+
+                ruby "${SCRIPT_DIR}/extract_terraform_state_to_yaml.rb" \
```
mtekel: why explicitly run this via `ruby` here, but not on line 1845?

concourse/pipelines/create-bosh-cloudfoundry.yml
Jonty: According to shellcheck the exact behaviour of `-a` and `-n` are not well specified between implementations, so `&&` and `||` are preferred.

concourse/pipelines/create-bosh-cloudfoundry.yml
Jonty: It was simply copied from elsewhere in the pipeline hence the inconsistency. I'll adjust it.

Jonty: Fixed.

PR #519, opened 2016-09-29
[#131355143] Rename datadog agent stub files and flags to increase readbility

keymon:
[#131355143 Add metron metrics to datadog](https://www.pivotaltracker.com/story/show/131355143)

What
----

As part of the story we will install the datadog nozzle, but the nozzle installation can be toggled.

We want to reuse the same variable for both datadog-agent and the datadog-nozzle. We also want to make the code more easy to understand.

We raise this PR independenly first as it is a refactor and improvement of the code, not a feature.

 * Rename datadog.yml stubs to datadog-agent.yml

   We are going to add datadog nozzle as an optional component in our CF installation. But the current naming of the stub files can be misleading and difficult to understand.

   We rename the stubs related to datadog-agent from datadog-agent.yml to make
   it more clear.

 * Rename DEPLOY_DATADOG_AGENT to ENABLE_DATADOG

   We want to use the same variable to enable/disable the isntallation of the  datadog agent and the datadog nozzle.

   There is no use case for enabling one and disabling the other, so we  decided to rename the variable and use it for both cases.

How to test
-----------

The tests should pass.

You can test only the job `cf-generate-config` and `bosh-generate-config` and confirm that enable/disable the setting changes the right parts of the manifest:

``` 
export DEPLOY_ENV=...
make dev pipelines ENABLE_DATADOG=false BRANCH=$(git rev-parse --abbrev-ref HEAD) 
make dev run_job JOB=generate-cf-config
# Wait for the job to run
aws s3 cp s3://hector-state/cf-manifest.yml /tmp/cf-manifest-a.yml

make dev pipelines ENABLE_DATADOG=true BRANCH=$(git rev-parse --abbrev-ref HEAD) 
make dev run_job JOB=generate-cf-config
# Wait for the job to run
aws s3 cp s3://hector-state/cf-manifest.yml /tmp/cf-manifest-b.yml

# compare
diff -Nur /tmp/cf-manifest-{a,b}.yml
```

Or if you have plenty of time, deploy the environment with `ENABLE_DATADOG=false`. It should not install agents or the nozzle. Enable it, it should install it.

Who?
----


combor:
Can you please rebase?

combor:
I compared generated manifest with the one from master branch and they are identical. Merging...

PR #522, opened 2016-09-30
[#131426461] Improve concourse metrics naming

saliceti:
## What
Story: [Improve concourse metrics naming](https://www.pivotaltracker.com/story/show/131426461)

The environment name is passed as a tag as `bosh-deployment` by #525. We can add `environment` for the metrics we push so we don't need to add the environment name in the metric name and we keep consistency.

As a consequence, we don't have to remove hyphens as it is fine to have hyphens in tags. We also put the concourse graphs in the same dashboard.

## Dependencies

#525 should be merged first

## How to review
* Deploy Concourse and run the CF pipeline
* Observe the metrics and dashboards make sense

## Who can review
Anyone but @paroxp or @keymon or @saliceti

PR #523, opened 2016-09-30
[#131426645] Restructure datadog terraform config

keymon:
[#131426645 Create datadog checks and graphs for gorouter](https://www.pivotaltracker.com/story/show/131426645)

What?
----

We are about to start writing a bunch of datadog checks.

Restructure beforehand to establish a convention that each check goes into its own file. For instance move gorouter datadog check definitions to specific file

Make a home for variables, and extract variable for the deploy env into a `null_data_source` to establish a convention for working around terraform's inability to interpolate in variable declarations.

See https://github.com/hashicorp/terraform/issues/4084

How to review this?
-------------------

Run the deployment with `DATADOG_ENABLED=true` in master and later in this branch.

```
SELF_UPDATE_PIPELINE=false ENABLE_DATADOG=true BRANCH=master make dev pipelines DEPLOY_ENV=...

SELF_UPDATE_PIPELINE=false ENABLE_DATADOG=true BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines DEPLOY_ENV=...
```

No changes should be applied to the datadog terraform.

Who?
---

Anyone but @benhyland or @keymon

### DIFF
--- a/terraform/datadog/concourse-jobs.tf
+++ b/terraform/datadog/concourse-jobs.tf
@@ -1,6 +1,5 @@
 resource "datadog_timeboard" "concourse-jobs" {
-
-  title = "${format("%s job runtime difference", var.env) }"
+  title = "${data.null_data_source.datadog.inputs.env} job runtime difference"
   description = "vs previous hour"
   read_only = false
 
@@ -8,7 +7,7 @@ resource "datadog_timeboard" "concourse-jobs" {
     title = "Runtime changes vs hour ago"
     viz = "change"
     request {
-       q = "${format("avg:%s_concourse.build.finished{*} by {job}", replace(var.env, "-", "_"))}"
+       q = "avg:${data.null_data_source.datadog.inputs.env}_concourse.build.finished{*} by {job}"
     }
   }
 }
--- a/terraform/datadog/concourse-pipeline.tf
+++ b/terraform/datadog/concourse-pipeline.tf
@@ -1,0 +1,13 @@
+resource "datadog_timeboard" "pipeline" {
+  title = "${data.null_data_source.datadog.inputs.env} - Concourse timeboard"
+  description = "Concourse metrics"
+  read_only = true
+
+  graph {
+    title = "Pipeline run time"
+    viz = "timeseries"
+    request {
+      q = "avg:${data.null_data_source.datadog.inputs.env}.pipeline_time{environment:${data.null_data_source.datadog.inputs.env}}"
+    }
+  }
+}
--- a/terraform/datadog/datadog-variables.tf
+++ b/terraform/datadog/datadog-variables.tf
@@ -1,0 +1,10 @@
+variable "datadog_api_key" {}
+variable "datadog_app_key" {}
+variable "env" {}
+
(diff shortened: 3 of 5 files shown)


mtekel:
looks ok, but not so easy to read as PR does 3 things:
1. splits it into multiple files
2. introduces null resource hack
3. changes {var.env} use to the resource hacked env. name, but which has - replaced with _. This means that for environments with - in name, various things will change. Is this desired? Do we also want to replace - with _ in e.g. human readable check description?

I think it would be better to split this into smaller, more explicit commits & think if every replacement of - with _ in var.env is desirable...

benhyland:
@mtekel I'm confused by your point 3. The change here is intended to be limited to datadog - other terraform files that use ${var.env} without going via the hack will not have the replacement applied.
So this affects datadog only, which we aren't yet using in anger since we haven't yet configured checks. It is probably a good idea to get the names right (including naming consistency) before we continue.

mtekel:
I'm not questioning whether we should do the var.env change or not but:
* it would be much easier to see where it was actually applied if it was in a commit that contains only that change. Now we have some things removed from common files and put into separate files + changed along the way.
* the PR message says this is noop and state should not change. But you are changing things, so this is not true
* I am not sure we always want to change - into _, e.g. in user readable descriptions of alerts (e.g. https://github.com/alphagov/paas-cf/pull/523/files#diff-f498ab84f007566d35d3cc1e2eb73ea2R2 )

benhyland:
The PR message is a little confusing - nothing should change because the datadog terraform provider uses ids rather than names to refer to its resources, and we aren't creating or destroying resources. That could be clearer.
I think it is better to be consistent in the use of env for datadog than to sometimes use `-` and sometimes `_` - this also should have been explained better.
On the desire to split the commits, I agree with hindsight it could have been better organised; however I don't see this as a good reason not to merge.

mtekel:
well, pr said `No changes should be applied to the datadog terraform.`. I wasn't sure if you realized you were actually modifying checks....

benhyland:
Sorry for the confusion!

mtekel:
For example, this might break the alert: https://github.com/alphagov/paas-cf/pull/523/files#diff-64a6d00ac0029c248347119636cbe97bR7

Have you checked if datadog agents also translate - to _ ? In general, from PR description and commit message it seems like this was an unintended change... I see this can be replaced in 2 places (in q in dashboards), but not sure everywhere elsewhere...

benhyland:
@mtekel thanks for the example.
I hadn't realised that was included in this changeset ><
We should get @saliceti to review, since he has an env with a `-`
As you said, the rieman metrics do have this replacement already but quite possibly the datadog agent does not.

timmow:
👍 for smaller commits. Its not obvious at the moment what has changed because of the file reorganisation

keymon:
@benhyland the PR #525 and #522 will make these change unnecessary. We won't need to keep that `data_null` resource, and all the PR will be only about move the definition of the router monitor from one file to other.

benhyland:
Good news.

PR #524, opened 2016-10-03
Fix the concourse link in smoke test emails

Jonty:
## What

The link to the continuous smoke tests was broken when the concourse upgrade added teams.

## How to review

Check that the link is now correct for a deployed concourse.

## Who can review

You! A friend! The person next to you! (As long as it isn't me)


mtekel:
yes, this matches the link in my latest concourse, merging

PR #526, opened 2016-10-04
More shellcheck fixes

mtekel:
## What

Add few more fixes for issues discovered during review of #518 . Most importantly, there should be no more errors or warnings from shellcheck after this, making it much easier to spot any new issues. To enforce that, we enable `--fatal-warnings` for pipecleaner.

## How to review

Make sure you have shellcheck 0.4.4.
* `make lint_concourse` should return no errors. 
* `make lint_shellcheck` should return no errors.
* bosh manifest generation in the pipeline should not be changed. You can test by temporarily removing your original bosh-ca from bucket & running pipeline again from init. Make sure generate-secrets picks the version of bosh-ca uploaded in init and not the cached one. You don't need to run your pipeline any further than this task. You can then compare generated bosh-ca tarball to the one you had before - mainly file names and paths. Once done, you can upload your original bosh-ca back. Or do nothing in case you let your pipeline run and finish.
* introduce a shellcheck warning by e.g. adding an output to some task that's not being used (e.g. into init) and checking that `make lint_shellcheck` exits with non-zero code.

## Who can review

anyone from the team but me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ lint_shellcheck:
 	find . -name '*.sh' -not -path '*/vendor/*' | xargs $(SHELLCHECK)
 
 lint_concourse:
-	cd .. && SHELLCHECK_OPTS="-e SC1091,SC2034,SC2046,SC2086" python paas-cf/concourse/scripts/pipecleaner.py paas-cf/concourse/pipelines/*.yml
+	cd .. && SHELLCHECK_OPTS="-e SC1091,SC2034,SC2046,SC2086" python paas-cf/concourse/scripts/pipecleaner.py --fatal-warnings paas-cf/concourse/pipelines/*.yml
 
 .PHONY: lint_ruby
 lint_ruby:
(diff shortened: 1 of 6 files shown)


Jonty:
Should we turn on `--fatal-warnings` now we have no errors?

mtekel:
Not sure, do we want to? What do other ppl think?

Jonty:
The warnings that are fatal are either errors or shellcheck, the remaining warnings are mostly "this thing isn't being used", but we probably shouldn't have them. There might be cases (as in the logging you removed) where working around those is tricky though.

benhyland:
I think we should turn it on. Static analysis that doesn't fail the build is just maintenance noise. If we cared enough to fix the warnings in this PR, we care enough to not let them happen again.

mtekel:
updated

Jonty:
👍

combor:
Looks good now. Merging.

PR #528, opened 2016-10-05
Shellcheck errors are now fatal in pipecleaner

Jonty:
## What

We recently changed shellcheck errors to be fatal in pipecleaner, but missed the documentation. This commit fixes it.

## How to review

Check that it really does say "Fatal" instead of "Warning" for shellcheck errors now.

## Who can review

You!


mtekel:
The change was implemented here: https://github.com/alphagov/paas-cf/pull/518/commits/f117653013bc7499cfcb6c36fc6984e338a8b00c

PR #531, opened 2016-10-06
[#131426645] gorouter graphs and alerts

mtekel:
## What

[Create datadog checks and graphs for gorouter](https://www.pivotaltracker.com/n/projects/1275640/stories/131426645)

## How to review

Deploy. Go to datadog, observe new graphs and monitors. To test monitors:
1. ssh to one of the routers & s`top. gorouter process running` and `gorouter healthy` monitors should trigger.
2. Remove more than 33% of your routes. If you have only 2 apps deployed, you can do so by stopping one app. `total routes difference` monitor should trigger.
3. Prevent the router connecting to nats (e.g. `iptables -A INPUT -s 10.0.17.11 -j DROP ; iptables -A INPUT -s 10.0.16.11 -j DROP ; iptables -A OUTPUT -s 10.0.17.11 -j DROP ; iptables -A OUTPUT -s 10.0.16.11 -j DROP`). `route update latency` monitor should trigger. Also `total routes discrepancy` monitor should trigger. If not, try make one router report significantly less or more routes than the other for at least 5 minutes. Then `total routes discrepancy` should trigger.

## Who can review

not @mtekel or @paroxp


terraform/datadog/router.tf
```diff
@@ -0,0 +1,165 @@
...
+
+resource "datadog_monitor" "total_routes_discrepancy" {
+  name = "${format("%s total routes discrepancy", var.env)}"
+  type = "query alert"
+  message = "Discrepancy in the amount of routes on routers. Check deployment status."
+  escalation_message = "Routers still have considerably different amount of total routes!"
+  no_data_timeframe = "5"
+  require_full_window = true
+
+  query = "${format("avg(last_5m):outliers(avg:cf.gorouter.total_routes{deployment:%s} by {ip}, 'dbscan', 3.0) > 0", var.env)}"
```
saliceti: Should we filter on {job:router} as well? This may be better for the outliers algorithm and this would be more consistent with other monitors.

terraform/datadog/router.tf
```diff
@@ -0,0 +1,165 @@
...
+
+resource "datadog_monitor" "total_routes_drop" {
+  name = "${format("%s total routes difference", var.env)}"
+  type = "query alert"
+  message = "Amount of the routes has decreased considerably, check deployment status."
+  escalation_message = "Total routes still dropping quickly. Check the deployment."
+  no_data_timeframe = "5"
+  require_full_window = true
+
+  query = "${format("pct_change(avg(last_1m),last_30m):avg:cf.gorouter.total_routes{deployment:%s} < -33", var.env)}"
```
saliceti: Should we add {job:router} for consistency?

terraform/datadog/router.tf
```diff
@@ -0,0 +1,165 @@
...
+
+resource "datadog_monitor" "route_update_latency" {
+  name = "${format("%s route update latency", var.env)}"
+  type = "metric alert"
+  message = "Route update latency too high, possibly serving stale routes."
+  escalation_message = "Route update latency still too high. Check the deployment."
+  no_data_timeframe = "5"
+  require_full_window = true
+
+  query = "${format("max(last_1m):avg:cf.gorouter.ms_since_last_registry_update{deployment:%s} by {ip} > 45000", var.env)}"
```
saliceti: Should we add {job:router} for consistency?

terraform/datadog/router.tf
```diff
@@ -0,0 +1,165 @@
+resource "datadog_monitor" "router" {
```
saliceti: Could you move this below with the other monitors

terraform/datadog/router.tf
```diff
@@ -0,0 +1,165 @@
...
+
+  require_full_window = true
+  tags {
+    "environment" = "${var.env}"
+    "job" = "router"
+  }
+}
+
+resource "datadog_timeboard" "gorouter" {
+  title = "${format("%s - GoRouter", var.env)}"
```
saliceti: The naming is different than the runtime difference dashboard. Can we make the same?
Like `title = "${format("%s gorouter", var.env)}"`

PR #534, opened 2016-10-07
[#120590777] Add Continuous Smoke Tests to concourse jobs dasboard

mtekel:
## What

[Improve Visibility of Continuous Smoke Tests](https://www.pivotaltracker.com/n/projects/1275640/stories/120590777)
Add a graph with continuous smoke test successes and failures into concourse job runtime dashboard.

## How to review
Run pipeline. Most importantly datadog terraform part which is in `cf-deploy` job. Go to datadog and observe new and strongly visible bar graph with continuous smoke tests successes and failures. If its empty then try running some continuous smoke tests. You can fail some of them to see the difference in the graph.

## Who can review

not @combor or @mtekel

### DIFF
--- a/terraform/datadog/concourse-jobs.tf
+++ b/terraform/datadog/concourse-jobs.tf
@@ -20,4 +20,23 @@ resource "datadog_timeboard" "concourse-jobs" {
     }
   }
 
+  graph {
+    title = "Continuous smoke tests"
+    viz = "timeseries"
+    request {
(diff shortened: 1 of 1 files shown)


terraform/datadog/concourse-jobs.tf
```diff
@@ -20,4 +20,23 @@ resource "datadog_timeboard" "concourse-jobs" {
     }
   }
 
+  graph {
+    title = "Continous smoke tests"
```
mtekel: fixed

PR #535, opened 2016-10-10
[#131635247] Install consul_agent in all VMs

keymon:
[#131635247 Metron error messages in logs](https://www.pivotaltracker.com/story/show/131635247)

What?
=====

Install consul_agent in all VMs

We have setup metron_agent in all VMs as a addon in #495.

Metron agent requires connect to etcd to discover the dopplers, and it is currently setup to discover the etcd by using `etcd.service.cf.internal` that is served by consul.

We did not setup the consul_agent in all the VMs, so the ones missing it had metron_agent installed, but unable to resolv etcd and, in consequence, discover the dopplers. Because that, the metron_agents were logging constantly the following error:

```
   {
      "file" : "/var/vcap/data/compile/metron_agent/loggregator/src/doppler/dopplerservice/finder.go",
      "timestamp" : 1475669685.3659,
      "source" : "metron",
      "line" : 139,
      "log_level" : "error",
      "message" : "Finder: Watch sent an error",
      "method" : "doppler/dopplerservice.(*Finder).run",
      "data" : {
	 "error" : "store request timed out"
      },
      "process_id" : 9989
   }
``` 

We will do some work in the future to reconsider this solution. Alternative approaches are:

 * Configure etcd using static IPs
 * Do not install metron_agent just to configure rsyslog. (see https://www.pivotaltracker.com/story/show/131975557)

How to review?
------------------

There's a test for this in the manifest, so travis should pass. 

Then you can double check by:  Deploy this, check that consul agent is running in all the VM. Check the logs ` /var/vcap/sys/log/metron_agent/metron_agent.stdout.log` or ` /var/vcap/sys/log/metron_agent/metron_agent.stderr.log` to check if the above errors is gone.

Who?
---

Anyone but @keymon

### DIFF
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -38,6 +38,8 @@ meta:
     release: (( grab meta.release.name ))
 
   clock_templates:
+  - name: consul_agent
+    release: (( grab meta.consul_templates.consul_agent.release ))
   - name: cloud_controller_clock
     release: (( grab meta.release.name ))
 
--- a/manifests/cf-manifest/manifest/040-graphite.yml
+++ b/manifests/cf-manifest/manifest/040-graphite.yml
@@ -1,5 +1,7 @@
 meta:
   graphite_templates:
+    - name: consul_agent
+      release: (( grab meta.consul_templates.consul_agent.release ))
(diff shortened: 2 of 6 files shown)


mtekel:
This works. The errors do disappear. In my dev that means 108K messages per hour (2.6M per day) less.

PR #537, opened 2016-10-10
[#131981965] Add more disk&memory to ES vm types

combor:
## What

[#131981965 Support: troubleshoot issues in logsearch/elasticsearch cluster](https://www.pivotaltracker.com/story/show/131981965)

This PR is a result of ES troubleshooting. 
We found out that all disks are almost full on all Elasticsearch VMs and ES process refuses to assign shards to any host that reached more than 90% of disk space used. As the volume of logs increases and we do more searches and indexing we need more memory and more CPUs to cope with the load. This PR updates ES VMs to m4.large that has 8GB of memory, 2 CPUs and increases disk space to 200GB. 

## How to review

- Deploy on top of existing environment. 
- check if you can still use Kibana
- check if you can access logs that were indexed before migration disks 

## Who can review

not @combor

### DIFF
--- a/manifests/cf-manifest/cloud-config/030-logsearch.yml
+++ b/manifests/cf-manifest/cloud-config/030-logsearch.yml
@@ -21,13 +21,16 @@ vm_types:
     network: (( grab vm_types.medium.network ))
     env: (( grab meta.default_env ))
     cloud_properties:
-      instance_type: (( grab vm_types.medium.cloud_properties.instance_type ))
+      instance_type: m4.large
+      ephemeral_disk:
+        size: 10240
+        type: gp2
       elbs:
         - (( grab terraform_outputs.logsearch_elastic_master_elb_name ))
 
 disk_types:
   - name: elasticsearch_master
-    disk_size: 102400
+    disk_size: 204800
     cloud_properties: {type: gp2}
   - name: queue
     disk_size: 102400


manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
```diff
@@ -23,6 +23,15 @@ vm_types:
         size: 10240
         type: gp2
 
+  - name: large
+    network: cf
+    env: (( grab meta.default_env ))
+    cloud_properties:
+      instance_type: m3.large
```
alext: Any reason not to use `m4.large` here.  They're fractionally cheaper than the `m3.large`, and give you an extra 0.5 gig of RAM.

manifests/cf-manifest/cloud-config/030-logsearch.yml
```diff
@@ -21,13 +21,13 @@ vm_types:
     network: (( grab vm_types.medium.network ))
     env: (( grab meta.default_env ))
     cloud_properties:
-      instance_type: (( grab vm_types.medium.cloud_properties.instance_type ))
+      instance_type: (( grab vm_types.large.cloud_properties.instance_type ))
```
mtekel: I think it would be simpler to put m3.large here instead of defining a new instance type, just so that you can grab a single property from it...

manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
combor: Hm, no reason. I'll change to m4.

manifests/cf-manifest/cloud-config/030-logsearch.yml
combor: Fair enough. Fixed.

combor:
> doesn't have any detail on why we need to also increase the instance size.

I added more details to the commit message.

mtekel:
This works. ES feels now faster even in dev. I did not lose any data during the migration.

PR #538, opened 2016-10-11
[#132060265] Check the git tag of the boshreleases prior upload

keymon:
[#132060265 Support: Wrong version of aws-broker deployed in CI](https://www.pivotaltracker.com/story/show/132060265)

What?
-----

How git-resource and concourse works, and how we implemented the build and upload of bosh-releases, it can happen that we build an older version of the bosh-release and upload it with the wrong version number.

The reason is that regardless the tag-filter configured in the git resource, the versions previously reported  still valid versions for concourse, and likely cached.
If the new version defined in the tag filter is not discovered (e.g. we forgot to add the tag or there was a typo) Concourse might checkout the latest valid version, but not the one defined in `tag-filter`.

We cannot add a safe-check in the git-resource, as the `get` might not be executed when the resource is cached.

Instead, and meanwhile we do not migrate to published final bosh releases built in a dedicated CI server, we will add a check before uploading the release, checking if the git tag matches the version.

Dependencies
------------

The PR https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/74 should be merged first. After it is merged and the new container is generated, you can delete the WIP commit in this branch.

How to review?
--------------

 1. run the pipeline, it should pass
 2. Disable the latest version in the git-resource of any bosh release. Rerun cf-deploy. It should fail.

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1507,21 +1507,27 @@ jobs:
                 - |
                   ./paas-cf/concourse/scripts/bosh_login.sh {{bosh_fqdn}} bosh-secrets/bosh-secrets.yml
 
+                  ./paas-cf/concourse/scripts/git_check_tag.sh {{cf_graphite_version}} graphite-statsd-boshrelease
                   paas-cf/concourse/scripts/bosh_create_and_upload_release.rb graphite \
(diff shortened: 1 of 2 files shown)


richardc:
https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/74 is now merged and published.  WIP commit deleted.


Looks good, will re-run in my environment and await travis builds

PR #539, opened 2016-10-11
[#127449817] Increase timeout for tag_release tests to 30s

keymon:
What?
-----

Similar to #388 and #435

This script sometimes takes more than Eventually's default timeout of 1sec to
return when running in travis. Probably git operations are too heavy for
the container based travis build.

This happened in travis several times[1]. Dan was able to reproduce it
by running `ginkgo --untilItFails`, but I couldn't.

I set the timeout to 30. Even with 3 seconds it timeouts sometimes, and it is
not really important to add more time if it eventually timeouts in a reasonable
time.

The failure looked like this:
``` 
  • Failure in Spec Setup (BeforeEach) [3.847 seconds]
  TagRelease
  /home/travis/build/alphagov/paas-cf/concourse/scripts/tag_release_test.go:147
    when there are multiple 'previous-*' tags to be promoted
    /home/travis/build/alphagov/paas-cf/concourse/scripts/tag_release_test.go:146
      and the before last tag of 'previous-*' is promoted [BeforeEach]
      /home/travis/build/alphagov/paas-cf/concourse/scripts/tag_release_test.go:115
        should promote the same tag version for 'next-*'
        /home/travis/build/alphagov/paas-cf/concourse/scripts/tag_release_test.go:89
        Timed out after 3.000s.
        /home/travis/build/alphagov/paas-cf/concourse/scripts/tag_release_test.go:161
```

[1] https://travis-ci.org/alphagov/paas-cf/builds/153593309

How to test?
-----------

Travis should pass, or see #426

Who?
----

Anyone but me

### DIFF
--- a/concourse/scripts/tag_release_test.go
+++ b/concourse/scripts/tag_release_test.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	ExecutionTimeout = 3 * time.Second
+	ExecutionTimeout = 30 * time.Second
 )
 
 var _ = Describe("TagRelease", func() {


saliceti:
This test is local and should never take a lot of time. There can be some overhead for initialisation but if the command takes more than 10-15s I would be worried. Could you set it to 15s at most?

keymon:
The only reason why there is a timeout at all is because we are forced to use `Eventually` in conjunction with `gexec.Exit(0)` as in https://github.com/alphagov/paas-cf/pull/539/files#diff-fe3bf3708b13219c511839e8e7834c17L36

The reason is that the exit code is not ready until the command finished, so you cannot check it in the test, you need to run a script that checks the execution in a given interval, like `Eventually` does.

We do not check the timeout of any other command, instead all of them they have an unlimited time to run.

This command is timing out because it is taking whatever time to run due some circumstances in travis which are out of our reach.

I will happily change from 30s to 15s if there is a good answer for: What do we win with 15s instead of 30s? if it timeouts, what will we do apart of rerun the run in travis, complain and maybe eventually increase it? Why set it to 15s and not 150s instead?

keymon:
Neither for you or me, I will agree to change the number to any value to the result of this: `echo $((RANDOM % (30-15) + 15))`

PR #541, opened 2016-10-11
[#131355243] Datadog: Add monitor for cell memory

mtekel:
## What

[Add datadog check of cell memory usage](https://www.pivotaltracker.com/n/projects/1275640/stories/131355243)
Create a monitor which check total available memory on cells. Alert if less than 50% is free.

## How to review

Deploy. Check that the new monitor is correctly created. Try triggering the alert by using a lot of memory on cells. You can try running e.g. `ruby -e 'foo = "123234314234" * 2000000000; sleep' &` on a cell VM directly. Or deploy lots of apps to CF, but which have to really consume memory (it doesn't matter what quota you assign to the app, only real memory consumption is counted).

See https://github.com/alphagov/paas-team-manual/pull/68 for information about alerting google groups - you need to to subscribe to receive alerts.

## Who can review

not @mtekel or @richardc

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -93,6 +93,7 @@ jobs:
           TF_VAR_datadog_api_key: {{datadog_api_key}}
           TF_VAR_datadog_app_key: {{datadog_app_key}}
           TF_VAR_env: {{deploy_env}}
+          TF_VAR_aws_account: {{aws_account}}
           ENABLE_DATADOG: {{enable_datadog}}
         ensure:
           put: datadog-tfstate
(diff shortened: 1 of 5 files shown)


benhyland:
This looks fine. Merging as soon as I can see the deploy in dev-land.

benhyland:
# 24GB in a single line for 2 mins, held by grep
`cat <(yes | tr \\n x | head -c $((1024*1024*1024*24))) <(sleep 120) | grep n`

https://app.datadoghq.com/monitors#998406?group=all&from_ts=1476267486361&to_ts=1476268838653

https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!topic/govpaas-alerting-dev/6w5_Mbk0IQA

PR #546, opened 2016-10-14
[#130844201] More distinctive concourses

mtekel:
## What

[Make it more difficult to run actions manually on Production Concourse](https://www.pivotaltracker.com/n/projects/1275640/stories/130844201)

To make our concourses more distinctive in order to avoid operator accidentally operating on different environments:
* add team with aws account specific name
* set login timeout to 5 minutes on prod

Note that this PR does not migrate the pipelines to the new team yet. We'll implement that later if we see the need.

## How to review

`make dev bootstrap`. Deploy. See team being created. Go to your concourse and try logging in using the new team. Password and user are the same (admin/usual PW). To test 5m timeout, edit pipeline & change the condition to check against 'dev' instead of 'prod' [here](https://github.com/alphagov/paas-cf/commit/6e80e34acc555b151ec3ed836dfd07527a5bc286#diff-5eac5d6fc041d1da10144e10c0fa2ed0R471), upload (`make dev bootstrap`), deploy, log in, wait 5 mins, click somewhere in pipeline (e.g. try to start a job) & be prompted to log in.

## Who can review

you!

### DIFF
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -442,6 +442,7 @@ jobs:
           AWS_ACCOUNT: {{aws_account}}
           DATADOG_API_KEY: {{datadog_api_key}}
           ENABLE_DATADOG: {{enable_datadog}}
+          CONCOURSE_AUTH_DURATION: {{concourse_auth_duration}}
           CONCOURSE_MANIFEST_STUBS: |
             ./paas-cf/manifests/concourse-manifest/concourse-base.yml
             predefined-concourse-secrets/predefined-concourse-secrets.yml
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -485,6 +485,7 @@ jobs:
     - get: concourse-ssh-private-key
     - get: concourse-manifest
       passed: ['concourse-prepare-deploy']
+    - get: paas-cf
 
     - task: concourse-deploy
       timeout: 30m
@@ -513,6 +514,30 @@ jobs:
         put: concourse-bosh-state
         params:
           file: concourse-deploy/concourse-manifest-state.json
+
+    - task: add-env-specific-team
+      config:
+        image: docker:///governmentpaas/self-update-pipelines
+        inputs:
+          - name: paas-cf
+          - name: concourse-manifest
+        params:
+          SYSTEM_DNS_ZONE_NAME: {{system_dns_zone_name}}
+          CONCOURSE_ATC_PASSWORD: {{concourse_atc_password}}
+          CONCOURSE_ATC_USER: admin
+          FLY_TEAM: {{aws_account}}
+          FLY_TARGET: {{deploy_env}}
+          FLY_CMD: ./paas-cf/bin/fly
+        run:
+          path: sh
+          args:
+            - -e
+            - -c
+            - |
+              export CONCOURSE_URL="https://deployer.${SYSTEM_DNS_ZONE_NAME}"
+              ./paas-cf/concourse/scripts/fly_sync_and_login.sh
+              echo y | ./paas-cf/bin/fly -t ${FLY_TARGET} set-team -n ${FLY_TEAM} \
+                         --basic-auth-username admin --basic-auth-password ${CONCOURSE_ATC_PASSWORD}
 
   - name: expunge-vagrant
     serial: true
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -73,7 +73,7 @@ jobs:
           databases:
             - name: atc
               role: atc
-              password: dummy-password
+              password: (( grab secrets.concourse_postgres_password ))
 
       - name: atc
         release: concourse
@@ -87,7 +87,7 @@ jobs:
             database: atc
             role:
               name: atc
-              password: dummy-password
+              password: (( grab secrets.concourse_postgres_password ))
 
       - name: groundcrew
         release: concourse
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -81,6 +81,7 @@ jobs:
           external_url: (( concat "https://" terraform_outputs.concourse_dns_name ))
           basic_auth_username: admin
           basic_auth_password: (( grab secrets.concourse_atc_password ))
+          auth_duration: (( grab $CONCOURSE_AUTH_DURATION ))
 
           postgresql:
             address: 127.0.0.1:5432
--- a/Makefile
+++ b/Makefile
@@ -146,6 +146,7 @@ prod: globals check-env-vars ## Set Environment to Production
 	$(eval export ENABLE_CF_ACCEPTANCE_TESTS=false)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=prod_deployment)
+	$(eval export CONCOURSE_AUTH_DURATION=5m)
 	@true
 
 .PHONY: bootstrap
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,7 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
 	$(eval export ENABLE_HEALTHCHECK_DB=false)
 	$(eval export ENABLE_DATADOG ?= false)
+	$(eval export CONCOURSE_AUTH_DURATION=48h)
 	@true
 
 .PHONY: ci
@@ -146,7 +147,6 @@ prod: globals check-env-vars ## Set Environment to Production
 	$(eval export ENABLE_CF_ACCEPTANCE_TESTS=false)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=prod_deployment)
-	$(eval export CONCOURSE_AUTH_DURATION=5m)
 	@true
 
 .PHONY: bootstrap
--- a/concourse/scripts/pipelines-deployer.sh
+++ b/concourse/scripts/pipelines-deployer.sh
@@ -32,6 +32,7 @@ system_dns_zone_name: ${SYSTEM_DNS_ZONE_NAME}
 aws_account: ${AWS_ACCOUNT:-dev}
 datadog_api_key: ${datadog_api_key:-}
 enable_datadog: ${ENABLE_DATADOG}
+concourse_auth_duration: ${CONCOURSE_AUTH_DURATION:-24h}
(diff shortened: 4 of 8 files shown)


manifests/concourse-manifest/concourse-base.yml
```diff
@@ -73,21 +73,22 @@ jobs:
           databases:
             - name: atc
               role: atc
-              password: dummy-password
+              password: (( grab secrets.concourse_postgres_password ))
```
mtekel: I guess that's specific to our postgres job that it doesn't require PW when connecting via localhost. We could argue that generating a password reduces attack surface. In the future, we might decide to make that posgres job safer or perhaps have some different setup where ATC is connecting on eth0 IP (or e.g. external workers need to connect to DB). 

This way or another, generating the password doesn't hurt anything so I think we should keep it.

keymon: I am split about this issue. On one hand I agree with @alext that it would be misleading to set a password that it is not being used. But on the other hand it is good to generate all the passwords.

But it is true that this line: `host   all                all                0.0.0.0/0     md5` might allow connect from other machines, so it is good to set the password.

Said that, the less effort would be leave this commit here.

concourse/pipelines/create-deployer.yml
```diff
@@ -467,6 +467,10 @@ jobs:
                               ./paas-cf/manifests/concourse-manifest/extensions/datadog-agent.yml
                               ./paas-cf/manifests/shared/deployments/datadog-agent.yml"
             fi
+            # shellcheck disable=SC2050
```
keymon: (Although I dislike it) We are keeping [these kind of settings centralised in the Makefile](https://github.com/alphagov/paas-cf/blob/6e80e34acc555b151ec3ed836dfd07527a5bc286/Makefile#L135-L148), shouldn't this value be set there? 

Otherwise, we can create a .yml for environment specific configuration, and merge the value with spruce depending the environment, as we do with the [CF manifest](https://github.com/alphagov/paas-cf/tree/6e80e34acc555b151ec3ed836dfd07527a5bc286/manifests/cf-manifest/manifest/env-specific) in the [cf pipeline](https://github.com/alphagov/paas-cf/blob/6e80e34acc555b151ec3ed836dfd07527a5bc286/concourse/pipelines/create-bosh-cloudfoundry.yml#L1368)

mtekel: That approach is possible, but it requires many more changes. Aim in this story was to quickly implement something and see how it does. If we want to have more changes (e.g. different timeouts for CI & staging), then we'll create another story. In that case, we can create 4 env. specific files (dev, ci, staging, prod) and always spruce them in. Currently I prefer concourse manifest to be simple and mostly contained in a single file only.

keymon: I disagree it is not more changes, it is the same amount, and it would be consistent with the other ways to define environment specific configuration.

keymon: In case of doubt these are the changes I propose:
```diff
diff --git a/Makefile b/Makefile
index 8b4734d..0f422ce 100644
--- a/Makefile
+++ b/Makefile
@@ -146,6 +146,7 @@ prod: globals check-env-vars ## Set Environment to Production
 	$(eval export ENABLE_CF_ACCEPTANCE_TESTS=false)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=prod_deployment)
+	$(eval export CONCOURSE_AUTH_DURATION=5m)
 	@true
 
 .PHONY: bootstrap
diff --git a/concourse/pipelines/create-deployer.yml b/concourse/pipelines/create-deployer.yml
index c4ff5eb..7ea357c 100644
--- a/concourse/pipelines/create-deployer.yml
+++ b/concourse/pipelines/create-deployer.yml
@@ -442,6 +442,7 @@ jobs:
           AWS_ACCOUNT: {{aws_account}}
           DATADOG_API_KEY: {{datadog_api_key}}
           ENABLE_DATADOG: {{enable_datadog}}
+          CONCOURSE_AUTH_DURATION: {{concourse_auth_duration}}
           CONCOURSE_MANIFEST_STUBS: |
             ./paas-cf/manifests/concourse-manifest/concourse-base.yml
             predefined-concourse-secrets/predefined-concourse-secrets.yml
@@ -467,10 +468,6 @@ jobs:
                               ./paas-cf/manifests/concourse-manifest/extensions/datadog-agent.yml
                               ./paas-cf/manifests/shared/deployments/datadog-agent.yml"
             fi
-            # shellcheck disable=SC2050
-            if [ {{aws_account}} = "prod" ] ; then
-              export AUTH_DURATION=5m
-            fi
             ./paas-cf/manifests/shared/build_manifest.sh $CONCOURSE_MANIFEST_STUBS \
               > concourse-manifest/concourse-manifest.yml
             ls -l concourse-manifest/concourse-manifest.yml
diff --git a/concourse/scripts/pipelines-deployer.sh b/concourse/scripts/pipelines-deployer.sh
index 83c8de9..f22f788 100755
--- a/concourse/scripts/pipelines-deployer.sh
+++ b/concourse/scripts/pipelines-deployer.sh
@@ -32,6 +32,7 @@ system_dns_zone_name: ${SYSTEM_DNS_ZONE_NAME}
 aws_account: ${AWS_ACCOUNT:-dev}
 datadog_api_key: ${datadog_api_key:-}
 enable_datadog: ${ENABLE_DATADOG}
+concourse_auth_duration: ${CONCOURSE_AUTH_DURATION:-24h}
 EOF
 }
 
diff --git a/manifests/concourse-manifest/concourse-base.yml b/manifests/concourse-manifest/concourse-base.yml
index fa07316..9bbc001 100644
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -81,7 +81,7 @@ jobs:
           external_url: (( concat "https://" terraform_outputs.concourse_dns_name ))
           basic_auth_username: admin
           basic_auth_password: (( grab secrets.concourse_atc_password ))
-          auth_duration: (( grab $AUTH_DURATION || "24h" ))
+          auth_duration: (( grab $CONCOURSE_AUTH_DURATION || null ))
 
           postgresql:
             address: 127.0.0.1:5432
```

concourse/pipelines/create-deployer.yml
```diff
@@ -514,6 +519,29 @@ jobs:
...
+          FLY_CMD: ./paas-cf/bin/fly
+        run:
+          path: sh
+          args:
+            - -e
+            - -c
+            - |
+              export CONCOURSE_URL="https://deployer.${SYSTEM_DNS_ZONE_NAME}"
+              ./paas-cf/concourse/scripts/fly_sync_and_login.sh
+              echo y | ./paas-cf/bin/fly -t {{deploy_env}} set-team -n {{aws_account}} \
```
keymon: I think you should use environment variables in the script rather the concourse substitutions. 

`FLY_TARGET: {{deploy_env}}` and `CONCOURSE_ATC_PASSWORD: {{concourse_atc_password}}` are already passed, but not used as environment variables. We should use those two and then also set something like `CONCOURSE_TEAM: {{aws_account}}` as environment variable.

keymon:
Does it makes sense to also create user different that `admin` to login?

concourse/pipelines/create-deployer.yml
mtekel: FLY_TARGET and CONCOURSE_ATC_PASSWORD are used by the login script. What's wrong with fly expansion? We use it everywhere, we would have to refactor all pipelines if we are only to use it in the definition of variables...

keymon: Because it is inconsistent with how we do it, even in that task itself. You use in one place `${FLY_TARGET}` and later `{{deploy_env}}` for the same purpose.

We try to use environment variables as it is easier to later extract the script, use external tasks with parameters and test.

keymon: Also, by using `CONCOURSE_TEAM_NAME: {{aws_account}}` it makes much more clear and explicit what is that. Currently one must go to the `fly set-team` documentation to know that `-n {{aws_account}}` intends to make the name of the team as the same than the name of the `aws_acount`.

mtekel: I could export FLY_TARGET in the script directly. This is for the login script and it has to be exported. Doing it as param is simpler (less code, less questioning). Passing team name and aws account as params adds 2 params which I only use once. Instead I can save 2 lines of code and use fly variables in the code directly...

keymon: Again, I disagree. 

It is not so much code, it is only one character more. And I insist as commented in https://github.com/alphagov/paas-cf/pull/546/files#r83616802 it makes sense to keep consistency, this is inconsistent with the task itself and the rest of the pipeline.

Also, we try to use environment variables in these cases, as the code can be extracted to a external script or a external task definition without big changes.

manifests/concourse-manifest/concourse-base.yml
```diff
@@ -73,21 +73,22 @@ jobs:
...
 
       - name: atc
         release: concourse
         properties:
           external_url: (( concat "https://" terraform_outputs.concourse_dns_name ))
           basic_auth_username: admin
           basic_auth_password: (( grab secrets.concourse_atc_password ))
+          auth_duration: (( grab $CONCOURSE_AUTH_DURATION ))
```
keymon: I recommend you to use `auth_duration: (( grab $CONCOURSE_AUTH_DURATION || null ))` or, otherwise, add a fixture and change the rspec of the manifest generation.

keymon:
Despite the comments https://github.com/alphagov/paas-cf/pull/546#discussion_r83614772 and the broken build in travis, the deployment works and the team is created.

keymon:
Can we add the 5m settings in the environments CI and staging? They also make sense to be protected.

manifests/concourse-manifest/concourse-base.yml
mtekel: I don't like empty properties. Also, setting default in 2 places (pipelines script & template). I'll add the variable to spec.

mtekel:
would it then make sense to set default to 5m and only override for dev?

keymon:
I have checked the implementation and works, so +1 for that.

We disagreed on the points, so I will let somebody else carry on with the review and decide about these points:
 
 * Use environment [variables or inline concourse interpolation in the variables](https://github.com/alphagov/paas-cf/pull/546#pullrequestreview-4445960). I wouldn't care to be merged if somebody else agrees.

 * Set the 5m timeout for CI and staging, in addition to production. https://github.com/alphagov/paas-cf/pull/546#issuecomment-254176924

keymon:
> would it then make sense to set default to 5m and only override for dev? 

Makes sense, but I am open to any approach.

mtekel:
I implemented both suggestions, hopefully it's OK to merge now.

keymon:
Thank you @mtekel that is great now :+1:

PR #547, opened 2016-10-14
Improve space usage of concourse cleaner chrome plugin.

alext:
## What

At present, there's some unnecessary padding at the top of the pipeline
display causing the scrollbars to appear etc.

This cleans thing up by allowing the main content block to use the full
space. As part of this, the enviuronment label has been moved to the
bottom, and floated on top to avoid it conflicting with the actual
pipeline elements - there's generally more room at the bottom of the
screen due to how the pipelines are layed out.

## How to review

* Add the plugin to your browser as described in the [README](../blob/master/misc/chrome_plugins/clean_concourse_pipeline/README.md). 
* Verify that the pipeline displays correctly, and the scrollbars don't appear.

## Who can review

Anyone but myself.


mtekel:
Works. I don't see any scrollbars when using this and accessing the pipelines overview page.

PR #550, opened 2016-10-17
[#127564379] Upgrade Logsearch to v203

Jonty:
## What

This is the minimal set of changes needed to upgrade our existing Logsearch install from v200 to v203. We took the approach of modifying our existing manifests to accommodate upstream changes rather than outright replacing them because we have previously heavily customised the Logsearch install to make it highly available.

Notable things:

* We no longer install HAProxy on the Kibana VM as it was totally unnecessary in our deployment with ELB's. As a result we needed to change the Kibana ELB port to correctly address Kibana itself rather than the HAProxy port.

* There may be a better way to set the Kibana timezone, however we chose not to implement it. Future versions of us may wish to look at: https://github.com/alphagov/paas-logsearch-for-cloudfoundry/blob/master/src/logsearch-config/src/es-mappings/logs-template.json.erb

* The drain/undrain process has totally changed in the new version of Logsearch and requires the administrator to re-enable shard reallocation after the upgrade, as the drain script disables it cluster-wide. An errand is shipped with Logsearch for this purpose, however we chose to simply duplicate it's single `curl` command in our pipeline for performance reasons - errands spin up a VM to run the command.

* We duplicate the Terraform state extraction as we can't reuse it from other jobs. We would simply extract the needed state as part of the new `curl` task, however we do not have a container available that both has `curl` and `ruby` available. This is ridiculous, and we should be refactoring our Terraform state extraction to be centralised and performed after each Terraform run.

* We needed to recreate the Parser VM's in all environments as the previous Logsearch deployment (v200) also deployed an Elasticsearch node on all the Parsers running as a client. It has been removed in the new version of Logsearch (v203) but Bosh will not remove the old Elasticsearch nodes, or upgrade them. The parser machines were oversized for their load, so we simply shrank them to `small` instances which will cause them to be recreated in all
environments.

## How to review

1. Deploy this branch in an existing environment.
2. Observe that the `parser` instances are recreated.
3. Visit `https://logsearch.$DEPLOY_ENV.dev.cloudpipeline.digital`, check that logs are still being received, and that no existing log data has been lost.
4. Visit `https://logsearch.$DEPLOY_ENV.dev.cloudpipeline.digital/app/kibana#/settings/about`, check that the kibana version is `4.4.0`.
5. `make dev bosh-cli` / `bosh ssh elasticsearch_master/0` / `curl http://localhost:9200` and check the version number outputted is `2.2.0`:
```
bosh_ewj1z4ls0@b90ea467-bcee-4849-bf05-5cef5de11d38:~$ curl http://localhost:9200
{
  "name" : "elasticsearch_master/0",
  "cluster_name" : "logsearch",
  "version" : {
    "number" : "2.2.0",
    "build_hash" : "8ff36d139e16f8720f2947ef62c8167a888992fe",
    "build_timestamp" : "2016-01-27T13:32:39Z",
    "build_snapshot" : false,
    "lucene_version" : "5.4.1"
  },
  "tagline" : "You Know, for Search"
}
```

🎉 

## Who can review

Not @benhyland or @jonty.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1483,6 +1483,30 @@ jobs:
           - get: datadog-tfstate
 
       - aggregate:
+        - task: extract-cf-terraform-outputs
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: ruby
+                tag: 2.2-slim
+            inputs:
+              - name: paas-cf
+              - name: cf-tfstate
+            outputs:
+              - name: cf-terraform-outputs
+            run:
+              path: sh
+              args:
+                - -e
+                - -c
+                - |
+                  SCPATH="./paas-cf/concourse/scripts"
+                  SCFILE="extract_tf_vars_from_terraform_state.rb"
+                  $SCPATH/$SCFILE < cf-tfstate/cf.tfstate > cf-terraform-outputs/cf.tfstate.sh
+                  ls -l cf-terraform-outputs/cf.tfstate.sh
+
         - task: upload-releases
           config:
             platform: linux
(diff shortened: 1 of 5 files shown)


mtekel:
Just a note, we do have a container with ruby & curl: self-update-pipelines. It would be a bit weird for reviewer to understand why do we need to use this special self-update-pipelines container, so the current situation is OK. It also highlights the need to fix the way we handle TF outputs.

mtekel:
Deployed OK. Didn't lose history. Got `{"acknowledged":true,"persistent":{},"transient":{"cluster":{"routing":{"allocation":{"enable":"all"}}}}}` on the shard relocation update. Versions of kibana & ELS match expected and logs still flow...

combor:
> Just a note, we do have a container with ruby & curl: self-update-pipelines.

The container with ruby and curl is also ... ruby container ( not the slim one )

PR #551, opened 2016-10-17
[#119717141] BOSH version bump

combor:
## What

Story: https://www.pivotaltracker.com/story/show/119717141

This is BOSH upgrade to the v257.14. Release notes:
https://github.com/cloudfoundry/bosh/releases/tag/stable-3262.14

Apart from [Redis removal](https://github.com/cloudfoundry/bosh/releases/tag/stable-3232) there are no other manifest changes mentioned neither by BOSH release notes nor by CF release notes. 
We bumped the version to the latest stable version as there are no recommendations in cf-release notes. Recommended stemcell version is 3262.X family - we are using 3262.9. 

## How to review

- apply this on top of existing env
- deploy env from scratch
- autodele pipeline
- all should work, no test failures

## Who can review

not @combor or @paroxp

### DIFF
--- a/manifests/bosh-manifest/deployments/010-bosh-template.yml
+++ b/manifests/bosh-manifest/deployments/010-bosh-template.yml
@@ -8,8 +8,8 @@ name: (( grab meta.environment ))
 
 releases:
 - name: bosh
-  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=255.8
-  sha1: 6b12652650b87810dcef1be1f6a6d23f1c0c13a7
+  url: https://bosh.io/d/github.com/cloudfoundry/bosh?v=257.14
+  sha1: b41821dccee78ad7bd44466cd0160920c183787a
 
 disk_pools:
 - name: disks
@@ -21,7 +21,6 @@ jobs:
 
   templates:
   - {name: nats, release: bosh}
-  - {name: redis, release: bosh}
   - {name: director, release: bosh}
   - {name: health_monitor, release: bosh}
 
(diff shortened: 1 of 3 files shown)


mtekel:
Suprisingly (after previous failed attempt, see https://github.com/cloudfoundry/bosh/issues/1263), all seems to work. I was able to upgrade, destroy and deploy without any issues. It also seems they have fixed the cleanup command (see https://github.com/cloudfoundry/bosh/issues/1294).

I think we should also update stemcell, as one used by bosh is getting a bit old, but I'd do that as another story...

mtekel:
Just noticed that VM 'name' tag in aws has changed. Before it was job/index, now is job/id. e.g cell/0 -> cell/fd6faeae-bb4e-4ae4-a03d-ebe67ddf5f8a. This is not so nice. I wonder if it has any further impact. E.g. when we enable datadog-aws integration.

mtekel:
Also, I got our well known error when doing 2nd delete:
```
  Started deleting instances > kibana/0 (08faeae2-e730-4177-b612-bf24b453844b)
   Failed deleting instances > doppler/0 (9ba621ff-9418-4d8d-a5d4-5e9d5c8953a8): PG::Error: ERROR:  update or delete on table "instances" violates foreign key constraint "rendered_templates_archives_instance_id_fkey" on table "rendered_templates_archives"
DETAIL:  Key (id)=(508) is still referenced from table "rendered_templates_archives".
```

alext:
> Also, I got our well known error when doing 2nd delete:

I wonder if that's something that might be fixed by upgrading the CPI.

mtekel:
So then we should consider if we make CPI and stemcell update a different story or extend this PR & re-test.

combor:
We've got stemcell upgrade story in pivotaltracker so I guess it can be prioritised.

mtekel:
With these stories, it's not really clear if they are:
* for concourse (uses garden-runC)
* for bosh (no garden)
* for CF (uses garden-linux)
* for all...

Impact of upgrade is different in each case...

mtekel:
We'll do that in another story, merging...

PR #552, opened 2016-10-18
[#132649921] Use patched gorouter with hotfix for logging CVE

keymon:
[#132649921 Respond to CVE 2016-6655: Utility script command injection](https://www.pivotaltracker.com/story/show/132649921)

What?
-----

Use patched version of gorouter v0.136.0-gds1

Temporary fix for CVE 2016-6655 [1]. This is a temporary fix to address the issue for gorouter while we work on upgrading CF.

[1]https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/thread/42YUJU2N27HBPFVMZR2QM7JI6YSEKORR/

How to review
-------------

Deploy it, tests should pass. Check that this command on router/{0,1} does not return a process in the gorouter group:

```
ps -feax | grep -B5 -e "system(lineWithDate)"
```

Who?
---

Anyone but @alext or @keymon

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -7,6 +7,10 @@ director_uuid: ~
 releases:
   - name: cf
     data: (( inject meta.cf-releases.cf ))
+  - name: routing
+    version: 0.136.0.gds1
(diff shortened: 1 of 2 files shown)


alext:
The updated routing release is the same version as the one in CF 240 with one additional commit. See https://github.com/alphagov/paas-routing-release/releases/tag/0.136.0.gds1

Jonty:
Works as described.

PR #553, opened 2016-10-19
No failure-testing pipeline by default

mtekel:
## What

Don't deploy failure-testing pipeline by default. We rarely run this pipeline and
* it consumes concourse resources for version checks
* it is a bit dangerous to be run on prod/staging, but is still present on that concourse...

## How to review

`make dev pipelines`. If you had it uploaded before, observe it being removed. To enable it, follow instructions in readme.

## Who can review

not me

### DIFF
--- a/concourse/scripts/pipelines-bosh-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-bosh-cloudfoundry.sh
@@ -131,8 +131,15 @@ update_pipeline() {
   pipeline_name=$1
 
   case $pipeline_name in
-    create-bosh-cloudfoundry|failure-testing)
+    create-bosh-cloudfoundry)
       upload_pipeline
+    ;;
+    failure-testing)
+      if [ "${ENABLE_FAILURE_TESTING:-}" = "true" ]; then
+        upload_pipeline
+      else
+        remove_pipeline
+      fi
     ;;
     destroy-*)
       if [ -n "${ENABLE_DESTROY:-}" ]; then
--- a/concourse/scripts/pipelines-bosh-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-bosh-cloudfoundry.sh
@@ -116,38 +116,40 @@ generate_manifest_file() {
     < "${SCRIPT_DIR}/../pipelines/${pipeline_name}.yml"
 }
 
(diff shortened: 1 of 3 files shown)


concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -116,38 +116,47 @@ generate_manifest_file() {
...
-    create-bosh-cloudfoundry|failure-testing)
-      bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
-        "${env}" "${pipeline_name}" \
-        <(generate_manifest_file) \
-        <(generate_vars_file)
+    create-bosh-cloudfoundry)
+      upload_pipeline
+    ;;
+    failure-testing)
+      if [ -n "${ENABLE_FAILURE_TESTING:-}" ] && [ "${ENABLE_FAILURE_TESTING:-}" = "true" ]; then
```
alext: In this line, the first test is redundant. This could simply be `if [ "${ENABLE_FAILURE_TESTING:-}" = "true" ];`

Having said that, see my comment below about the interpretation of these flags.

concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -116,38 +116,47 @@ generate_manifest_file() {
...
+        remove_pipeline
+      fi
     ;;
     destroy-*)
-      if [ -n "${ENABLE_DESTROY:-}" ]; then
-        bash "${SCRIPT_DIR}/deploy-pipeline.sh" \
-          "${env}" "${pipeline_name}" \
-          <(generate_manifest_file) \
-          <(generate_vars_file)
+      if [ -n "${ENABLE_DESTROY:-}" ] && [ "${ENABLE_DESTROY:-}" = "true" ]; then
```
alext: This is a change in behaviour (that's not called out in the PR summary). It also makes these variables inconsistent with some other variables in the `Makefile` (eg `SKIP_COMMIT_VERIFICATION`).  If we're going to change how these boolean flags are interpreted, I'd be keen that we do it for all of them.

mtekel:
Updated.

keymon:
:+1:

PR #554, opened 2016-10-19
Whilelist verify jenkins server

mtekel:
## What

Whitelist verify jenkins IP so that they can deploy onto CF. User for this jenkins has been already created.

## How to review

Double-check the ip address is one mentioned in #1050 support ticket in deskpro.

## Who can review

not @mtekel


keymon:
:+1:

PR #556, opened 2016-10-20
Add ssh_bosh script and make target

mtekel:
## What

Add a convenient `ssh_bosh` makefile target. This also helps to reduce time-to-fix in case you need to get to BOSH to investigate/fix.

## How to review

`make dev ssh_bosh`

## Who can review

anyone in the team


keymon:
The port 2222 usually conflicts with other things. For instance with a machine running in virtualbox.

I wonder if we can make  the `LOCAL_PORT=2222` to pick a random free port and dynamically. Something between a simple `LOCAL_PORT=22$((RANDOM % 99))` to a complex solution like: http://unix.stackexchange.com/a/55918/550

keymon:
You can also use the `ProxyCommand` as here: https://github.com/gds-attic/paas-alpha-tsuru-ansible/blob/master/ssh.config.template#L15

alext:
You don't even need to use netcat, you can use the `-W` option to ssh. eg https://github.com/alphagov/gds-boxen/blob/master/modules/gds_ssh_config/files/gds_ssh_config#L74

mtekel:
Updated...

keymon:
That is great!! :+1: 

Works like charm. But you forgot to update the README. update it and I will merge :)

mtekel:
updated, there was also one more typo

PR #558, opened 2016-10-20
Run bosh cleanup after deploy

mtekel:
## What

This way we can remove previous releases/packages/stemcells immediately after they stop being used. This is also important for long running environments where we would not run cleanup otherwise and only accumulate previous versions forever, which can also potentially impact BOSH performance and reliability.

## How to review

Apply this PR. Run deploy job. Observe it running bosh cleanup task.

## Who can review

not @mtekel


keymon:
Works and cleaned up a lot of things from my dev environment

PR #561, opened 2016-10-24
[#131975557] Configure log forwarding with syslog-release

paroxp:
## What

We're removing the old deployment of metron and consul agents, by reverting the previous commits. It will be replaced by our new approach.

We are adding a syslog_forwarder and syslog_cleanup addons to the base runtime config.

The syslog_forwarder addon configures syslog-release to forward syslog messages to the logsearch relp ingestors.

The syslog_cleanup addon configures os-config/cleanup_syslog_forwarding to remove the configuration leftover from the metron_agent.  See alphagov/paas-os-conf-release#3 for more information on this.
## How to review

This is dependant on [paas-os-conf-release pull request](https://github.com/alphagov/paas-os-conf-release/pull/3). Once the other PR is merged, we can remove the TMP commit from this PR and refer to the new release in this one.
- Deploy the pipelines
- Run the `create-bosh-cloudfoundry` pipeline
- See, if correct VMs have the syslog forwarder enabled
- See that VMs no longer have the old `/etc/rsyslog.d/00-syslog_forwarder.conf` present
## Who can review

Not @richardc or @paroxp.


manifests/cf-manifest/manifest/010-cf-jobs.yml
```diff
@@ -40,20 +42,22 @@ meta:
     release: datadog-for-cloudfoundry
   - name: cloud_controller_worker
     release: capi
+  - name: metron_agent
+    release: (( grab meta.release.name ))
 
   clock_templates:
-  - name: consul_agent
-    release: (( grab meta.consul_templates.consul_agent.release ))
```
alext: I think this needs to remain now that we've upgraded to CF 245.  Part of the changes are that several cf components now use consul to locate the etcd cluster.  See https://github.com/cloudfoundry/cf-release/commit/c432b73#diff-b9982126ee4438e20576a100670694d8R1275 for the upstream manifest change.

Jonty:
Have re-confirmed my findings yesterday, this branch appears to be broken on existing deployments due to a race condition restarting rsyslog. Both the new cleanup job (https://github.com/alphagov/paas-os-conf-release/pull/3) and the new syslog job restart syslog at the end of their pre-start script, and all addons are run in parallel meaning they can both attempt to restart it at the same time, leading to errors like this in concourse:

```
Error 450001: Action Failed get_task: Task 948908f3-cc48-47a1-7eb4-67a3178ef4b8 result: 1 of 4 pre-start scripts failed. Failed Jobs: cleanup_syslog_forwarding. Successful Jobs: set_mtu, syslog_forwarder, consul_agent.
```

It will either be `cleanup_syslog_forwarding` or `syslog_forwarder` that fails, depending on who won the race.

In the stderr log for the job that failed, you will either see `stop: Job has already been stopped: rsyslog` or `start: Job is already running: rsyslog` or one of several other exciting errors.

We can't merge this as it won't actually deploy unless we repeatedly re-run it.

richardc:
Reading the [bosh docs on the start order](https://bosh.io/docs/job-lifecycle.html#start), all pre-start actions happen in parallel.

We can maybe work around this by moving the `cleanup_syslog_forwarding` into a later stage by wrapping it up as a fake monit service.  This should solve the initial race, but may still clash with other rsyslog-fiddling jobs like `haproxy`.

We think we can avoid clashing with things like that by templating the monit to have the cleanup job `depend on haproxy` to add [service ordering]( https://mmonit.com/monit/documentation/monit.html#SERVICE-DEPENDENCIES) (spiking this now with https://github.com/alphagov/paas-os-conf-release/commit/b00a6001300f66b9814ef43cfc3d8ae8073cb952)

manifests/cf-manifest/manifest/010-cf-jobs.yml
richardc: Thanks for the catch, we'll address this.

keymon:
In #570 I added more options to the runtime-config for metron agent. Depending on what gets merged first, the other PR must be rebased and the commits merge, or, otherwise, this one must be rebased and and the commits removed.

manifests/cf-manifest/manifest/010-cf-jobs.yml
richardc: Added in ec76b04d66ca5f624610227d567ea3b9add47880

richardc:
@Jonty I believe I'd addressed the issue of racing rsyslogd restarts by revising https://github.com/alphagov/paas-os-conf-release/pull/3 to do it in a synthetic monit service, and addressed @alext's comment about `clock_templates`.  Should be good to retest.

Jonty:
Hurrah, all seems to work now, but not merging until Monday because really, no.

keymon:
@paroxp @Jonty the conflicts are likely due  some additional properties I had to add in order to make the `metron_agent` work as a addon in the new version of bosh. You can simply delete then.

On the other hand, be aware that because the same reason, now you might need some more properties for the syslog-forward job as a addon.

PR #562, opened 2016-10-24
[#133085655] Faster dev pipeline execution - no locking

mtekel:
## What

Improve dev pipeline execution by disabling pipeline locking by default and making disabling custom acceptance test optional via env. variable. Also refactor variable names, where VERB_something still meant yes/true if the variable was undefined previously. This doesn't fix all variables yet.

## How to review

1. `make dev pipelines`. Go for lunch (acceptance tests take 1h+). Check that pipeline didn't lock and unlock, acceptance and custom acceptance tests ran and healthcheck DB was not deployed.
2. Remove defaults from makefile:
```
--- a/Makefile
+++ b/Makefile
@@ -94,10 +94,8 @@ dev: globals check-env-vars ## Set Environment to DEV
        $(eval export SKIP_COMMIT_VERIFICATION=true)
-       $(eval export DISABLE_HEALTHCHECK_DB=true)
        $(eval export ENABLE_DATADOG ?= false)
        $(eval export CONCOURSE_AUTH_DURATION=48h)
-       $(eval export DISABLE_PIPELINE_LOCKING=true)
```
then `DISABLE_CF_ACCEPTANCE_TESTS=true DISABLE_CUSTOM_ACCEPTANCE_TESTS=true SELF_UPDATE_PIPELINE=false make dev pipelines`
Check that pipeline did lock and unlock, acceptance and custom acceptance were skipped and that healthcheckDB was created.

## Who can review

you can do it!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ staging: globals check-env-vars ## Set Environment to Staging
 	$(eval export ALERT_EMAIL_ADDRESS=the-multi-cloud-paas-team+staging@digital.cabinet-office.gov.uk)
 	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS=${ALERT_EMAIL_ADDRESS})
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
-	$(eval export ENABLE_CF_ACCEPTANCE_TESTS=false)
+	$(eval export DISABLE_CF_ACCEPTANCE_TESTS=true)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=staging_deployment)
 	@true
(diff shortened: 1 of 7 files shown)


keymon:
I am worried about the logic to disable the default check. It is a little of a hack that it relies on the task to return error to actually do the blockage. 

Apart of being counter intuitive, I see that there is a risk here: if there is any failure in the task, the job will pass OK and NOT lock the pipeline. That is risky.

I would suggest these changes:
 * Change the logic, lock the pipeline if the task returns error. That way, if it fails, it will always block.
 * Move the `if [ ${DISABLE_PIPELINE_LOCKING:-} = "true" ]` to a separated dedicated tasks, that is simplier and less likely to fail.

Also, a different alternative solution for this, would be create an alternative resource which would skip the locking logic if there is a property `disable: true` in the resource config. We could do a PR upstream

keymon:
Shall not the story refer to https://www.pivotaltracker.com/story/show/133085655

mtekel:
Updated

PR #563, opened 2016-10-24
[#128858813] Single file for BOSH manifest

mtekel:
## What

[Single file for BOSH manifest](https://www.pivotaltracker.com/n/projects/1275640/stories/128858813)
Put bosh manifest into single file instead of 10 that we had previously as a leftover from time when we also supported GCE.

## How to review

1. If you trust me, then run pipeline. There should be no deployment of bosh, as rendered manifest is exactly the same.
2. If you don't trust me, then pause bosh-deploy job and go to S3 bucket and compare rendered bosh manifest with previous version. There should be no changes.

## Who can review

you


richardc:
👍  merging paused until #564 is merged to avoid excessive churn.

PR #569, opened 2016-10-27
[#133184695] Create a 'Free' RDS broker service plan.

alext:
## What

This adds a new 'Free' service plan to the RDS broker. This plan is using the smallest available RDS instance with the smallest allowed storage allocation, and all snapshotting disabled. It's intended to be used for experimentation etc.

As part of this, I've also made the following changes:

* Updated the postgres version to 9.5.4 (from 9.5.2). (See commit message for details).
* Updated the acceptance tests to use this plan where possible (the snapshot related tests need to use one of the plans with snapshots available).
* Updated the healthcheck-db to use this Free plan (this includes a commit to re-create existing instances with the old plan that can be reverted once this has run everywhere).

## How to review

To test the updates to the healthcheck-db, you'll need an environment deployed form master with the DB created (remove the `$(eval export DISABLE_HEALTHCHECK_DB=true)` from the dev section of the makefile). Running the pipeline from this branch should then update this database instance to use the new Free plan.

Verify that the Instance on the Free plan has the correct parameters.

## Who can review

Anyone but myself.

### DIFF
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -79,6 +79,21 @@ jobs:
                 supportUrl: "https://forums.aws.amazon.com/forum.jspa?forumID=60"
               plan_updateable: true
               plans:
+                - id: "5f2eec8a-0cad-4ab9-b81e-d6adade2fd42"
+                  name: "Free"
+                  description: "5GB Storage, NOT BACKED UP, Dedicated Instance, Max 50 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.micro."
+                  free: true
+                  metadata:
+                    bullets:
+                      - "Dedicated Postgres 9.5 server"
+                      - "AWS RDS"
+                  rds_properties:
+                    inject: (( inject meta.rds_broker.default_postgres_rds_properties ))
+                    db_instance_class: "db.t2.micro"
+                    allocated_storage: 5
+                    backup_retention_period: 0
+                    skip_final_snapshot: true
+
                 - id: "b7d0a368-ac92-4eff-9b8d-ab4ba45bed0e"
                   name: "S-dedicated-9.5"
                   description: "20GB Storage, Dedicated Instance, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.small."
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -2,7 +2,7 @@ meta:
   rds_broker:
     default_postgres_rds_properties:
       engine: "postgres"
-      engine_version: "9.5.2"
+      engine_version: "9.5.4"
       storage_type: "gp2"
       auto_minor_version_upgrade: true
       multi_az: false
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -1,53 +1,31 @@
 meta:
   rds_broker:
+    default_postgres_rds_properties:
+      engine: "postgres"
+      engine_version: "9.5.2"
+      storage_type: "gp2"
+      auto_minor_version_upgrade: true
+      multi_az: false
+      publicly_accessible: false
+      copy_tags_to_snapshot: true
+      skip_final_snapshot: false
+      backup_retention_period: 7
+      db_subnet_group_name: (( grab terraform_outputs.rds_broker_dbs_subnet_group ))
+      vpc_security_group_ids:
+        - (( grab terraform_outputs.rds_broker_dbs_security_group_id ))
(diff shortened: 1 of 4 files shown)


manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -79,6 +79,25 @@ jobs:
...
               plan_updateable: true
               plans:
+                - id: "5f2eec8a-0cad-4ab9-b81e-d6adade2fd42"
+                  name: "Free"
+                  description: "5GB Storage, NOT BACKED UP, Dedicated Instance, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.micro."
+                  free: true
+                  metadata:
+                    costs:
+                      - amount:
+                          usd: 0.02
```
mtekel: free: true, but specifying costs?

alext: Good point.  I'm not sure what the best thing to do here is. This is how much these instances actually cost, but given we're not passing that cost on to the users, maybe this should be omitted.

Having said all that, the only way that a user can see these costs is by using `cf curl`.

mtekel:
I wonder about capital F in Free. It seems a bit unusual. It seems that 'free' is more common among other services/service providers.

alext:
> I wonder about capital F in Free.

I was thinking it looked better in sentence case, that's all. For example:

```
$ cf m -s postgres
Getting service plan information for service postgres as admin...
OK

service plan         description                                                                                                                                    free or paid
S-dedicated-9.5      20GB Storage, Dedicated Instance, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.small.                        paid
S-HA-dedicated-9.5   20GB Storage, Dedicated Instance, Highly Available, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.small.      paid
M-dedicated-9.5      20GB Storage, Dedicated Instance, Max 500 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.m4.large.                        paid
M-HA-dedicated-9.5   20GB Storage, Dedicated Instance, Highly Available, Max 500 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.m4.large.      paid
L-dedicated-9.5      20GB Storage, Dedicated Instance, Max 5000 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.m4.2xlarge.                     paid
L-HA-dedicated-9.5   20GB Storage, Dedicated Instance, Highly Available, Max 5000 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.m4.2xlarge.   paid
Free                 5GB Storage, NOT BACKED UP, Dedicated Instance, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.micro.          free
```

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -88,19 +89,6 @@ var _ = Describe("RDS broker", func() {
...
 			// Poll until destruction is complete, otherwise the org cleanup fails.
 			pollForRDSDeletionCompletion(dbInstanceName)
-
-			rdsClient, err := NewRDSClient(region)
-			Expect(err).NotTo(HaveOccurred())
-			snapshots, err := rdsClient.GetDBFinalSnapshots(rdsInstanceName)
-			fmt.Fprintf(GinkgoWriter, "Final snapshots for %s:\n", rdsInstanceName)
-			fmt.Fprint(GinkgoWriter, snapshots)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(snapshots).Should(ContainSubstring(rdsInstanceName))
```
alext: I've updated the tests to make this more clear.

manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -101,6 +79,25 @@ jobs:
                 supportUrl: "https://forums.aws.amazon.com/forum.jspa?forumID=60"
               plan_updateable: true
               plans:
+                - id: "5f2eec8a-0cad-4ab9-b81e-d6adade2fd42"
+                  name: "Free"
+                  description: "5GB Storage, NOT BACKED UP, Dedicated Instance, Max 200 Concurrent Connections. Postgres Version 9.5. DB Instance Class: db.t2.micro."
```
mtekel: max is 87, via `select * from pg_settings where name='max_connections';`
In parameter group, it's LEAST({DBInstanceClassMemory/9531392},5000) and
```
DBInstanceClassMemory
Returns the number of bytes of memory allocated to the DB instance class associated with the current DB instance, less the memory used by the Amazon RDS processes that manage the instance.
```
87 thus implies the DBInstanceClassMemory is 830 MiB. Meaning about 244MiB is reserved for the system.

mtekel:
All fine, passed all the tests, migrated healthcheckDB. I'll merge after description update.

manifests/cf-manifest/manifest/050-rds-broker.yml
alext: Updated to say 50. That leaves plenty of headroom.

PR #571, opened 2016-10-28
[#133184695] Revert "Update plan of existing healthcheck instances"

alext:
## What

In #569 we switched the healthcheck DB to use the new Free plan. This required a temporary script to do this. Now that this has run everywhere, it can be removed.

## How to review

Verify that the healthcheck DB is using the Free plan in all persistent environments. Verify that this revert looks sensible.

## Who can review

Anyone but myself.

This reverts commit fccc839907ec9ed07cd23a3e6cee418d7b004864.


keymon:
LGTM, it is updated in ci, master and prod.

PR #572, opened 2016-10-28
[#132082633] Use version 0.0.5 of paas-haproxy-release

henrytk:
## What

[#132082633 - Router HAProxy restarts rsyslog without reason](https://www.pivotaltracker.com/n/projects/1275640/stories/132082633)

We want to reduce the number of unnecessary restarts of rsyslog, as the amount of time it takes to restart has been known to cause deployment failures.

This pull request changes our version of the HAProxy release to 0.0.5. This version changes the start script so it only restarts rsyslog if the configuration changes.

## How to review

* Deploy from this branch.
* Check tests pass

## Who can review

Anyone but me


keymon:
Everything good :+1:

PR #565, opened 2016-10-26
[#133003305] Upgrade stemcell 3263 and migrate to garden-runc

keymon:
[#133003305 Upgrade to stemcells with kernel 4.4 and garden-linux to garden-runc](https://www.pivotaltracker.com/story/show/133003305)

and

[#132909443 Respond to CVE-2016-5195 kernel vulnerability (AKA upgrade to stemcell 3263 and garden-runc)](https://www.pivotaltracker.com/story/show/132909443)

What?
-----

In the story #132909443 we apply the patch for [a vulnerability that allows escalation of privileges CVE-2016-5195 (aka dirtycow)](http://dirtycow.ninja/) in the linux kernel that has been reported recently. Our systems are affected, but specially worrying are the cells and the containers running on it.

Despite being reticent, Cloud Foundry community included a [patched version in a older series with 3.3](https://cloudfoundry.slack.com/archives/bosh/p1477113219004741) on 3262.22. That has been done in the PR #560

But in this PR we want to upgrade to stemcell 3263 and migrate to garden-linux:

 * We learn that the stemcell 3263 includes the patch, but it runs the kernel 4.4. But this kernel can cause problems with garden-linux.
 * Also, garden-linux is not supported anymore.

This change represents some risk in the cells running the garden containers. To allow a quick manual rollback if required, we keep in the manifest the definitions of the garden-linux release and stemcell 3262. We need to add a dummy job definition `cell_3_3` to avoid the stemcell and release being deleted by the bosh-clean job (see #558)

How to test?
------------

 * Review that the changes make sense.
 * Deploy this release. All test should pass OK.
 * You can test the rollback procedure described here in [the summary](https://www.pivotaltracker.com/story/show/133003305/comments/153730755)

Note: As it is a stemcell change, all VMs will be reprovision.

Note: We should send a mail to the tenants, in the review or approval. See the card for details.

Who?
----

Anyone but @keymon or @benhyland

PR #575, opened 2016-10-31
[#133431043] Remove old garden-linux release, stemcell and cell_3_3 job

keymon:
[#133431043 Clean-up stemcell+release from upgrade stemcell and garden-run](https://www.pivotaltracker.com/n/projects/1275640/stories/133431043)

What?
-----

After #565 has been deployed and it run on prod without issues for a while, we can remove the references to the old garden-linux, the stemcell 3262 and the temporary job cell_3_3.

How to test?
------------

Nothing really, simply check it makes sense.

Who?
----

Anyone but @keymon


alext:
Confirmed this is a no-op deploy, and the `bosh cleanup` removes the unused stemcell, and the garden-linux release.

PR #578, opened 2016-10-31
[#132160889] Replace datadog tag naming 

paroxp:
## What

With the new features in datadog agent, we don't need to create a `job` tag adhoc, as `bosh-job` is provided to us by the datadog bosh-release: https://github.com/onemedical/datadog-agent-boshrelease/pull/8.

We have refactored all the graphs and monitors, to use the new tag `bosh-job` and `bosh-deployment` instead.

Note: This does not apply for the metrics reported by datadog-nozzle, which will be tagged with `deployment` and `job` instead.

The datadog agent sets the bosh-job tag automatically. We're removing the specific tags added to each one of the jobs, as they have been deprecated.

As `environment` and `bosh-deployment` were actually the same concept, for consistency, we will delete `environment` and  use only `bosh-deployment`.

Test that the tags are defined globally in the runtime config instead of job by job. Only check that `aws_account` is set.

All the metrics are tagged as `deployment` or `bosh-deployment` and it makes sense to keep the same convention for the tags of the monitors.

## How to review

Run the deployment, check if all datadog graphs and monitors respond. Note that it will take time to get all the metrics in place.

## Who can review

Neither @keymon nor @paroxp

### DIFF
--- a/manifests/bosh-manifest/extensions/datadog-agent.yml
+++ b/manifests/bosh-manifest/extensions/datadog-agent.yml
@@ -6,7 +6,5 @@ jobs:
   properties:
     datadog: (( inject meta.datadog ))
     tags:
-      job: bosh
       bosh-job: bosh
       bosh-az: (( grab terraform_outputs.bosh_az_label ))
-      tags: (( inject meta.datadog_tags ))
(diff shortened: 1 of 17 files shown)


Jonty:
Rebasing this branch and fixing the single conflict (the test being moved, no biggie) has broken the branch totally. Bouncing back to @paroxp and @keymon.

keymon:
We actually changed and moved one test from the file `jobs_spec.rb`, to `shared_config_spec.rb`. 

But I think that during the rebase the old one was added again. 

Fixed.

PR #580, opened 2016-11-01
[#132312331] Use latest aws-broker release from Github release.

alext:
## What

This updates to the latest release of rds-broker. This is actually a
no-op change because there were some changes introduced to the
rds-broker[1] that were then reverted[2]. It's worth updating so that
the pipeline is up to date, making future changes easier.

This also changes the way we consume the release. We were previously
building a dev release in the pipeline from the git sources. This is
somewhat inefficient, and also has issues with the way that the version
is pinned in the git resource, especially when we have to go back a
version. Building a release once and uploading to a git release makes
this much simpler, and also means that we consume the release in the
same way as 3rd-party releases that are published to bosh.io.

[1]https://github.com/alphagov/paas-rds-broker/pull/18
[2]https://github.com/alphagov/paas-rds-broker/pull/20

## How to review

This is a no-op change to the rds-broker. It should therefore be sufficient to verify that this is true by comparing the rds-broker code that the 2 different versions point at. There should be no differences at all (even though they're different commits).

Verify that the pipeline changes are correct. This can be done by running the pipeline up to and including cf-deploy.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -112,12 +112,6 @@ resources:
       uri: https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease.git
       tag_filter: {{cf_datadog_for_cloudfoundry_version}}
 
-  - name: aws-broker-boshrelease
-    type: git
-    source:
-      uri: https://github.com/alphagov/paas-aws-broker-boshrelease.git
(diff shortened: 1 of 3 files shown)


alext:
Note: because the source code between 0.0.10 and 0.0.12 is identical (and therefore has the same checksums), bosh won't need to compile anything, and won't need to update the VMs. It will just update its internal references as to which version is deployed.

PR #581, opened 2016-11-02
Parameterise hardcoded domain and emails

andrewdriver123:
## What

We are now getting stuck into our epic for doing Paas-CF on AWS and our first story is to find a solution whereby we can override the default GDS values for domain names and email addresses in a way that does not displease the GDS folks.

This PR adds a macro function into the Makefile which is used to attempt to look up domain/email values in `paas-pass`, and if not found it will default to the currently used GDS values

If we've done everything correctly, then this should have no impact on existing GDS workflow, but will make it easier for other non-GDS people to make their own deployments.

New values which can now be configured in `paas-pass` are as follows:

├── paas-cf
│   ├── dev-app-domain.gpg
│   ├── dev-sys-domain.gpg
│   ├── ci-app-domain.gpg
│   ├── ci-sys-domain.gpg
│   ├── staging-app-domain.gpg
│   ├── staging-sys-domain.gpg
│   ├── prod-app-domain.gpg
│   ├── prod-sys-domain.gpg
│   ├── dev-alert-email-address.gpg
│   ├── ci-alert-email-address.gpg
│   ├── prod-alert-email-address.gpg
│   └── staging-alert-email-address.gpg

## How to review

Review changes to the `Makefile`
Follow the normal deployment procedure

## Who can review

Anyone on the GDS team.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,9 @@ list_merge_keys: ## List all GPG keys allowed to sign merge commits.
 		fi;\
 	done
 
+# pass password store unhelpfully dumps it's not-found message to stdout, so we're using a macro to grep the output and remove it
+get_from_pass_or_default = $(eval export $(1)=$(shell paas-pass $(2) | grep -v not\ in\ the || echo $(3)))
+
 .PHONY: globals
 PASSWORD_STORE_DIR?=${HOME}/.paas-pass
 globals:
@@ -90,8 +93,8 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export AWS_ACCOUNT=dev)
 	$(eval export ENABLE_DESTROY=true)
 	$(eval export ENABLE_AUTODELETE=true)
-	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
-	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
+	$(call get_from_pass_or_default,SYSTEM_DNS_ZONE_NAME,paas-cf/dev-sys-domain,${DEPLOY_ENV}.dev.cloudpipeline.digital)
+	$(call get_from_pass_or_default,APPS_DNS_ZONE_NAME,paas-cf/dev-app-domain,${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
 	$(eval export SKIP_COMMIT_VERIFICATION=true)
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
 	$(eval export DISABLE_HEALTHCHECK_DB=true)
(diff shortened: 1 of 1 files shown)


henrytk:
The Travis build failure currently showing affected my pull request as well. It seems like a transient error with a third party dependency which is now resolved (for me anyway). Re-running should be sufficient.

bleach:
Hello,

Thanks for submitting this. We've set aside some time in our current iteration (ends 15th Nov) to do a technical review.

In the mean time, we were wondering how you came to choose `paas-pass` as a configuration mechanism. 

For the Government PaaS threat model, the domain and email address wouldn't need to be encrypted at rest, is that different for you? It also requires people to use [pass](https://www.passwordstore.org/), which works for our team, but not sure if it does for anyone else. 

Also, we were trying to tell if [self-updating of concourse pipelines](https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-bosh-cloudfoundry.yml#L400) would still work after this change. Is that something you've managed to test.

jimconner:
Thanks for setting aside some time to review this. `paas` (or `paas-pass` if you're at GDS) is generally a great solution for storing all manner of YAML snippets whether they're super-secret or not as it means that we can pass config values into code without having to paste in a dozen different environment variables whenever we need to invoke a command. I thought we had agreed on using `pass` for this when we met up a few weeks ago. If there's some other more preferable approach you'd like us to take, please let us know and we'll implement that mechanism too.

We haven't yet done a full deployment using this modified code as we've got issues with accessing any AWS-deployed VMs at the moment (corporate firewall stuff) - however I don't think this is going to cause problems with the self-updating pipelines as the pipeline steps don't ever call back to the Makefile to fetch those domain or alert-email settings - Updating the alert-email in `pass` would be enough to get the pipeline reconfigured with the new alert-email until a `make <env> pipelines DEPLOY_ENV=<something>` step was run again... but that's currently the case with the hardcoded values in the Makefile anyway.

alext:
> I don't think this is going to cause problems with the self-updating pipelines as the pipeline steps don't ever call back to the Makefile to fetch those domain or alert-email settings

I'm afraid this job does call back into make (https://github.com/alphagov/paas-cf/blob/master/concourse/scripts/self-update-pipeline.sh#L29), and would therefore attempt to call paas, which won't exist in the container. It's necessary to call the Makefile because that's the source of truth for the domains etc, and those details are needed to generate the pipelines.

bleach:
@jimconner I thought there was going to be a more detailed proposal of the mechansim via a github issue first and didn't think we'd settled on an implementation, so a little bit of confusion. 

We should take some time in the review to see if we can work out an approach that suits you and others. Your use case `pass` make sense to me, seems like there's an opportunity to allow people who pull config values from other stores to do so via the same mechanism.

jimconner:
Oh dear! There we were thinking that the Makefile was going to be the lowest of the low-hanging fruit to deal with... (far easier than the Terraform manifests). We'll do some more thinking around this problem and see if we can come up with an idea that's compatible with the self-updating pipelines.

srbry:
We have been spitballing some ideas here (most of which have fallen flat on their faces). The frontrunners are currently abstracting the config parameters away into a file stored in a separate git repo or s3 bucket. This would enable config changes to be accessible by concourse and also be triggers. The reason for a separate git repo is to remove the requirement to fork this repo just to tweak a config value.

Any feedback/ issues with this approach would be appreciated.

bleach:
I've got the nod to push our review up the priority a bit, so we'll let whoever picks it up at our end have a think about it.

combor:
> frontrunners are currently abstracting the config parameters away into a file stored in a separate git repo or s3 bucket. This would enable config changes to be accessible by concourse and also be triggers.

@srbry We had some discussion about this and that sounds like more flexible and simpler solution that is not mixing credential store with data store.  It will also allow us to provide config file as a Concourse resource so jobs can consume values easily. I would imagine yaml file to have a list of deployments like:

```
deployments:
- name: dev
  system_dns_zone_name: dev.exaple.com
  alert_email_address: dev@example.com
  ...
- name: ci
  system_dns_zone_name: ci.example.com
  alert_email_address: ci@example.com
  ...
``` 
 > The reason for a separate git repo is to remove the requirement to fork this repo just to tweak a config value.

👍 We can use Concourse resource to pull that file from s3 or git repo. 

I will close this PR for now ( and related one as well ). Thanks!

PR #582, opened 2016-11-02
[#121483833] Capture API calls during pipeline tests

henrytk:
## What

[#121483833 - Capture api calls made in pipeline tests](https://www.pivotaltracker.com/story/show/121483833)

We want to be able to capture the CF_TRACE output during Cloud Foundry tests in the pipeline, because it provides useful debugging information when investigating failed tests. We do not want to output this to the pipeline, as we cannot be certain it is free of secrets, so we will store it in a bucket.

## How to review

* Deploy from this branch
* Check the `upload-test-artifacts` task succeeds for the following jobs:
    * acceptance-tests
    * custom-acceptance-tests
    * smoke-tests
* Check the bucket `gds-paas-$DEPLOY_ENV-test-artifacts` for three files, which should be TAR files from the jobs listed above.
* Make one of the tests fail and check it still uploads the file.
* If you are really curious you can download them, un-TAR them and look inside.

## Who can review

Anyone but me.

### DIFF
--- a/concourse/tasks/smoke-tests-run.yml
+++ b/concourse/tasks/smoke-tests-run.yml
@@ -9,6 +9,8 @@ inputs:
   - name: cf-release
   - name: test-config
   - name: bosh-CA
+outputs:
+  - name: artifacts
 run:
   path: bash
   args:
@@ -16,7 +18,8 @@ run:
     - -c
     - |
       ./paas-cf/concourse/scripts/import_bosh_ca.sh
-
+      mkdir -p /var/vcap/sys
+      ln -s $(pwd)/artifacts /var/vcap/sys/log
       mkdir -p /var/vcap/jobs/smoke-tests/bin/ /var/vcap/packages/smoke-tests/src/github.com/cloudfoundry
       ln -snf $(pwd)/test-config/config.json /var/vcap/jobs/smoke-tests/bin/config.json
       ln -snf $(pwd)/test-config/run /var/vcap/jobs/smoke-tests/bin/run
--- a/concourse/tasks/upload-test-artifacts.yml
+++ b/concourse/tasks/upload-test-artifacts.yml
@@ -1,0 +1,25 @@
+---
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: governmentpaas/awscli
+inputs:
+  - name: paas-cf
+  - name: artifacts
+run:
+  path: sh
+  args:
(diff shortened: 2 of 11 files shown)


henrytk:
The Travis build is currently failing due to an unrelated issue with Go dependencies.

saliceti:
I'm not sure the `s3init` refactor is necessary. "s3init" has a special meaning emphasizing that a file is uploaded only once. You don't need this logic here and could do simply `aws s3 cp`.

concourse/tasks/smoke-tests-run.yml
```diff
@@ -24,3 +26,8 @@ run:
       ln -snf $(pwd)/cf-release/src/smoke-tests /var/vcap/packages/smoke-tests/src/github.com/cloudfoundry/cf-smoke-tests
       set -o pipefail
       /var/vcap/jobs/smoke-tests/bin/run
+
```
saliceti: This block will not be executed if the smoke test fails. Can we do the upload in an `ensure` block?

henrytk:
I have removed the refactoring of the s3init.sh script and used AWS CLI instead. I have also wrapped the upload task in an ensure block.

concourse/tasks/upload-test-artifacts.yml
```diff
@@ -0,0 +1,24 @@
...
+run:
+  path: sh
+  args:
+    - -e
+    - -c
+    - |
+      echo "Archiving test artifacts..."
+      RAND=$(hexdump -n 2 -e '/2 "%u"' /dev/urandom)
+      ARTIFACT_NAME="test-artifact-$(date '+%Y-%m-%d-%H-%M-%S')-$RAND.tgz"
+      tar -C artifacts -czvf "${ARTIFACT_NAME}" ./
```
saliceti: It looks like tar doesn't like "." :

```
$ tar xvzf test-artifact-2016-11-03-12-19-56-17410.tgz
x ./: Can't update time for .
x ./acceptance_tests/
x ./acceptance_tests/CATS-TRACE-CATS-1.txt.log
x ./acceptance_tests/CATS-TRACE-CATS-2.txt.log
x ./acceptance_tests.log
x ./acceptance_tests.err.log
tar: Error exit delayed from previous errors.
```

I would suggest something more explicit such as:

```
(cd artifacts && tar -czvf "${ARTIFACT_NAME}" *)
```

PR #583, opened 2016-11-02
[#support] Perform a fly sync after login

keymon:
What?
----

Our fly_sync_and_login.sh file will download the fly command from the
fly server if the local file is older than the one in the server.

But if the file has been locally updated with a different version more
recently, for example because one environment has a different concourse
version, the curl command will not update the binary.

But concourse fly command provides a `fly sync` that will update the
binary if there is version mistmatch. We can execute this fly sync
command just after login.

How to test?
-----------

Download a different version of the fly command, for instance 2.4.0 from
the concourse ci itself:

```
curl 'https://ci.concourse.ci/api/v1/cli?arch=amd64&platform=linux' -O bin/fly && chmod +x bin/fly
```

run bosh-cli and check that gets updated

```
make dev bosh-cli DEPLOY_ENV=...
```

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/scripts/fly_sync_and_login.sh
+++ b/concourse/scripts/fly_sync_and_login.sh
@@ -16,3 +16,6 @@ chmod +x "$FLY_CMD"
 echo "Doing fly login"
 echo -e "${CONCOURSE_ATC_USER}\n${CONCOURSE_ATC_PASSWORD}" | \
   $FLY_CMD -t "${FLY_TARGET}" login -k --concourse-url "${CONCOURSE_URL}"
+
+echo "Doing fly sync"
+  $FLY_CMD -t "${FLY_TARGET}" sync


alext:
I seem to remember that we moved away from doing this because it resulted in downloading the binary every time, even if it hadn't changed. I'll try to dig out the history.

alext:
See https://github.com/alphagov/paas-cf/commit/ce799487649c62753c3f8445f443ab2e26c10333, and https://github.com/alphagov/paas-cf/commit/bcb785b6ceae8b9c9bb084b9c24524d8ca1ae181 for more details.

alext:
I see from your comment on ce79948 that the `fly sync` behaviour has now been improved so that it no longer downloads the binary unless needed. We should call this out in the commit message.

Also, I think it's probably worth backing out some of the changes in those 2 commits as they're no longer necessary. We should only need to use curl if the binary doesn't already exist locally.

keymon:
As commented IRL, I think it is good enough how it is now and, anyway, going back will end with almost the same implementation.

PR #584, opened 2016-11-03
use paas-pass to remove hardcoded values

srbry:
## What

We are now getting stuck into our epic for doing Paas-CF on AWS and our first story is to find a solution whereby we can override the default GDS values for domain names and email addresses in a way that does not displease the GDS folks.

This PR adds a function to the `Vagrantfile` that will attempt to get credentials from `paas-pass` or return a default if not found in `paas-pass`.

If we've done everything correctly, then this should have no impact on existing GDS workflow, but will make it easier for other non-GDS people to make their own deployments.

New values which can now be configured in paas-pass are as follows:

```
├── paas-cf
│   ├── dev-subnet-id.gpg
│   ├── dev-security-group.gpg
│   ├── ci-subnet-id.gpg
│   ├── ci-security-group.gpg
│   ├── staging-subnet-id.gpg
│   ├── staging-security-group.gpg
│   ├── prod-subnet-id.gpg
│   ├── prod-security-group.gpg
```

## How to review

Review changes to the `Vagrantfile`
Follow the normal deployment procedure

## Who can review

Anyone on the GDS team.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -1,25 +1,31 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+def secretFromPass(path, default)
+  secret = `paas-pass 2>/dev/null #{path}`
+  return default if secret == ""
+  secret.strip
+end
+
 Vagrant.require_version ">= 1.8.0"
 
 AWS_ACCOUNT = ENV.fetch("AWS_ACCOUNT", "dev")
 AWS_ACCOUNT_DATA = {
   "dev" => {
-    subnet_id: "subnet-d55defe8", # us-east-1e in default VPC, 172.31.32.0/20 range
-    security_group: "sg-99e0c0e0", # "Bootstrap Concourse" security group
+    subnet_id: secretFromPass('paas-cf/dev-subnet-id', 'subnet-d55defe8'), # us-east-1e in default VPC, 172.31.32.0/20 range
+    security_group: secretFromPass('paas-cf/dev-security-group', 'sg-99e0c0e0'), # "Bootstrap Concourse" security group
   },
   "ci" => {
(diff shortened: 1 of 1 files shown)


keymon:
In order to handle this configuration differences, I want to propose we do the following:

 * We can have a special repository for configuration: e.g. paas-cf-config-gds
    * Will contain GDS and environment specific configuration: dev, ci, staging, prod specific. GDS specific, etc.
    * Examples:
       * domains
       * # of instances in different jobs.
       * disk sizes, memory sizes, and other parameters.
 
 * All config must be modeled in a .yml format, using a spec, similar to the releases. We can use that spec to evaluate that the config is valid and that all the required config is present.

 * We should create a bunch of simple tools to read these YAML in a consistent way. Preferrable in golang to bundle the tools with the containers. Examples:
    * val_from_yaml command line tool: Extract a value from a yaml.
    * yaml_to_terraform_vars: Create a bunch of `TF_VAR_whatever` to be consumed by terraform

 * The different pipelines will read from config repository directly. So, a change in the config of prod will bypass the deployment in CI and staging. This will allow us to operate the platform and do configuration changes that are not being tested anyway in ci/staging (e.g. # of cells).

 * The different config in the git repo should be tagged, to ease track changes.

I think it should be a quick win, fairly easy to implement.

Opinions?

srbry:
Sounds like the same sort of thing we were thinking of here, our opinion was to try and replicate something like BOSH's `specs` would be a good way of showing what the YAMLs could be. We are going to start working on it now.

keymon:
@srbry Glad you like it :) 

Be aware that is a personal proposal, I did not get too much feedback from our team.

combor:
I have commented on related PR already. We kinda agree that we should have separate yaml file with  organisation specific values in it that we can provide as Concourse resource and parse with small scripts.

https://github.com/alphagov/paas-cf/pull/581#issuecomment-259696910

I will close this now.

PR #587, opened 2016-11-03
[#132736941] introduce dashboard for datadog overview

benhyland:
## What

Introduce dashboard for PaaS monitor overview

We are using [Smashing](https://github.com/Dashing-io/smashing) to provide a dashboard.
We chose this as it is already in use by Verify, it seems simplest and easiest out of
existing tools, and it is easy to read at a distance.

We add `service:foo_env` tags to all monitors. We were advised to do this by Datadog since the
`service` prefix is treated differently from most tags and allows the filter in the UI to work properly.

We deploy the new dashboard to the prod env in the main pipeline.
There's a temporary commit to allow the reviewer to see the thing work in dev if they want to.

## How to review

To run the dashboard locally,
```
cd tools/paas_dashboard
bundle install
DD_API_KEY=<api key> DD_APP_KEY=<app key> bundle exec smashing start
```
The keys can be retrieved from the integrations -> api section of the datadog webui.

The dashboard by default looks for monitors with service tags for `prod_monitors`, `staging_monitors`, `master_monitors` - it won't find any until after this is merged.
You can play with the dashboard by deploying the branch and running the code locally with
a modification in `tools/paas_dashboard/jobs/datadog.rb` to target different service tags.

Review https://github.com/alphagov/paas-team-manual/pull/70 at the same time.

Once you're happy and merged, on the mac-mini terminal, do something like:
```
curl https://github.com/dsingleton/frame-splits/archive/gh-pages.zip > frame-splits.zip
unzip frame-splits.zip
google-chrome ./frame-splits/build.html
```
Figure out how you want to display the various dashboards we have.
You should probably use the local frame-splits for the concourse pipeline split as well.

## Who can review

Not @benhyland or @saliceti - perhaps @paroxp could take a look ;)


tools/paas_dashboard/dashboards/paas-overview.erb
```diff
@@ -0,0 +1,19 @@
...
+<div class="gridster">
+  <ul>
+    <li data-row="1" data-col="1" data-sizex="1" data-sizey="1">
+      <a href="https://app.datadoghq.com/monitors#manage?query=service:prod_monitors"><div data-title="Prod" data-id="prod_counts" data-view="Health"></div></a>
+    </li>
+    <li data-row="1" data-col="2" data-sizex="1" data-sizey="1">
+      <a href="https://app.datadoghq.com/monitors#manage?query=service:staging_monitors"><div data-title="Staging" data-id="staging_counts" data-view="Health"></div></a>
+    </li>
+    <li data-row="1" data-col="3" data-sizex="1" data-sizey="1">
+      <a href="https://app.datadoghq.com/monitors#manage?query=service:master_monitors"><div data-title="CI" data-id="ci_counts" data-view="Health"></div></a>
```
paroxp: Could we add `target="_blank"` attribute to the `a` tag?

IMO, it's easier to navigate, if the link would open in a new tab, instead of taking the current state.

tools/paas_dashboard/jobs/datadog.rb
```diff
@@ -0,0 +1,46 @@
...
+
+def get_and_emit_data_for_env(service_tag:, data_id_prefix:)
+  results = get_monitor_results(service_tag)
+  puts results.to_json
+  critical_count, warning_count = get_counts(results)
+  send_event("#{data_id_prefix}_counts", criticals: critical_count, warnings: warning_count)
+end
+
+def get_monitor_results(service_tag)
+  dog.get_all_monitors[1].select { |monitor| monitor['tags'].include?(service_tag) }
```
paroxp: Rather, than filtering the results ourselves, we should probably figure out, how does the ruby package work.

I'm worried about the fact, we gather 68 results, instead of 16. It may be nothing right now, but could be potentially an issue in the future.

FYI, this functionality works as expected in the python library.

tools/paas_dashboard/config.ru
```diff
@@ -0,0 +1,21 @@
+require 'dashing'
+
+configure do
+  set :auth_token, 'YOUR_AUTH_TOKEN'
```
paroxp: Is that required? I don't think it's being used at the moment.

tools/paas_dashboard/dashboards/paas-overview.erb
```diff
@@ -0,0 +1,19 @@
+<script type='text/javascript'>
+$(function() {
+  Dashing.numColumns = 3
+  Dashing.widget_base_dimensions = [600, 500]
```
paroxp: I still think, we should have a responsive implementation instead.

Different scenarios, will handle this differently, which may cause some grief.

tools/paas_dashboard/dashboards/paas-overview.erb
benhyland: sure

tools/paas_dashboard/dashboards/paas-overview.erb
paroxp: Sure, but that's JS implementation. We could have done that with CSS perhaps, which would be a top up to the JS Smashing uses?

tools/paas_dashboard/jobs/datadog.rb
benhyland: I have an issue open with Datadog on this. I don't think we should delay merging on their response though.

tools/paas_dashboard/config.ru
benhyland: It's used to secure requests to the API that provides a push interface for updating data bound to widgets. We don't use the push mechanism at the moment, and we don't need to secure it, so I left it as the default value.

benhyland:
I just had a response from Datadog which advises that we should use the manual filtering workaround we already adopted. They point to the `silenced key`  `notify_no_data` fields for a more refined determination of what is or is not triggered - we could adjust this so we don't display warnings on non-triggered 'No Data' monitors.
They also promise to improve the api docs.

PR #588, opened 2016-11-03
[#132331007] Document how to rotate the manually uploaded certs.

alext:
## What

This process is somewhat gnarly due to limitations in terraform and the
way we've had to split things up.

Having said that, it's still valuable to have this documented until we
can come up with another approach (possibly not using terraform for this
specific bit).

This PR also includes the removal of a shellcheck disable that's no longer needed now that we've upgraded shellcheck.

## How to review

Read the new docs, and verify they make sense.

Verify Travis passes.

## Who can review

Anyone buy @henrytk or myself.

### DIFF
--- a/doc/non_dev_deployments.md
+++ b/doc/non_dev_deployments.md
@@ -38,39 +38,57 @@ should use "prod".
 ## Manual upload of SSL certificates
 
 For some environments (e.g. prod) we want to use purchased valid certificates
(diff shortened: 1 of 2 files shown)


bleach:
Very clear docs and a nice tidy-up of the shellcheck ignore.

bleach:
Merged without a GPG sig so it won't deploy. (It doesn't need to, it only affects travis and docs)

PR #591, opened 2016-11-04
Clarify documentation for datadog dev proceedure

Jonty:
## What

If you don't have dev keys (or have expired ones) in your environment when enabling datadog you will get strange errors due to a terraform bug and a datadog bug. This clarifies the documentation to ensure you won't forget to publish keys.

## How to review

Check that the procedure described makes sense.

## Who can review

You! Certainly not @jonty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,8 +283,10 @@ need to run the pipeline from start in order to bring the changes forward.
 
 ## Optionally deploy datadog agent in dev environment
 
-If you want to deploy datadog agent in your dev environment, set ENABLE_DATADOG
-variable to true:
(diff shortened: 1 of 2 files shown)


benhyland:
lgtm, will merge. Was there anything we could do from make or in the pipeline to verify the keys in use before we actually try to use them?

Jonty:
@benhyland We could `curl` an endpoint that returns a 200|403 at the beginning of the pipeline, or in the Makefile when turning Datadog on. I'm not sure if this is something that impacts anyone regularly enough to be worth implementing - perhaps worth keeping an eye out for and revisiting if it hurts someone again?

PR #592, opened 2016-11-04
[#133748125] make target to restart concourse and clear volumes

keymon:
[#133748125  Automate execution of clear volumes procedure of concourse](https://www.pivotaltracker.com/story/show/133748125)

What?
----

We want to automate the procedure to restart and clear all the volumes in concourse. We run this often enough to automate it.

For several versions of concourse, we had multiple problems mostly related to baggageclaim volumes[1][2][3][4]

The development team in concourse says being working on it and refactoring the logic related to volume management.

Meanwhile the problem is not completely fixed, we found that the procedure commented here: https://github.com/concourse/concourse/issues/565#issuecomment-237845601 helps a lot. Basically it will:
 * stop concourse services and workers.
 * kill all containers.
 * clear container network.
 * clean up the volumes and images in disk.
 * remove the images and volumes references in concourse DB.
 * restart concourse.

The concourse team insisted that this is not a supported procedure and that the bugs must be fixed, but it did work for us several times and we consider it is a fair workaround until concourse fixes the issues.

[1] https://github.com/concourse/concourse/issues/565
[2] https://github.com/concourse/concourse/issues/737
[3] https://github.com/concourse/concourse/issues/682
[4] https://github.com/concourse/concourse/issues/507

How to test?
------------

Run `make shake_concourse_volumes DEPLOY_ENV=...`. All the concourse service will be restarted. Check that it still works.

Double check that `make ssh_concourse DEPLOY_ENV=...` and `make ssh_concourse DEPLOY_ENV=... SSH_CMD="ls /"` work

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/scripts/shake_concourse_volumes.sh
+++ b/concourse/scripts/shake_concourse_volumes.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+cat <<"EOF"
+
+This operation will restart concourse and dropping all the volumes...
+
+                 ,-.
+       _,.      /  /          ,--.!,
+      ; \____,-==-._  )    __/   -*-
+      //_    `----' {+>  ,d08b.  '|`
+      `  `'--/  /-'`(    0088MM
+            /  /         `9MMP'       BLOW THAT CONCOURSE!!!
+      dew   `='       hjm
+                               via http://ascii.co.uk/art/plane
+                               and http://ascii.co.uk/art/bomb
+
+EOF
+
+set -e
+
+sudo -v -p "Type sudo password to continue: "
+
+echo "Stopping all the concourse services..."
+for i in atc beacon tsa baggageclaim garden ; do
+   sudo -i /var/vcap/bosh/bin/monit stop $i;
+done
+sleep 10
+
+echo -n "Killing all runc containers..."
+sudo pkill -f runc # Be sure that all containers are stopped
+while pgrep -f runc; do
+   sleep 1
+   echo .
+done
+echo
+
+echo "Flushing all iptables rules from previus containers..."
+sudo iptables -F
+sudo iptables -X
+sudo iptables -t nat -F
+sudo iptables -t nat -X
+
+echo "Deleting all baggageclaim volumes and garden data"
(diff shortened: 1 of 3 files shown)


concourse/scripts/shake_concourse_volumes.sh
```diff
@@ -0,0 +1,65 @@
...
+sudo iptables -X
+sudo iptables -t nat -F
+sudo iptables -t nat -X
+
+echo "Deleting all baggageclaim volumes and garden data"
+if mount | grep -q /var/vcap/data/baggageclaim/volumes; then
+   sudo umount -fl /var/vcap/data/baggageclaim/volumes
+fi
+if mount | grep -q /var/vcap/data/baggageclaim/volumes; then
+   sudo umount -fl /var/vcap/data/garden/graph
```
alext: Copy/paste error in the `if`. Should be `/var/vcap/data/garden/graph`

concourse/scripts/shake_concourse_volumes.sh
```diff
@@ -0,0 +1,65 @@
...
+   sudo losetup -d /dev/loop0
+fi
+if sudo losetup -a | grep -q /dev/loop1; then
+   sudo losetup -d /dev/loop1
+fi
+sudo rm -rf /var/vcap/data/garden /var/vcap/data/baggageclaim/volumes.img
+
+echo "Dropping volume, containers and images references from concourse DB..."
+# Remove all the tables related to containers
+sudo /var/vcap/packages/postgresql_9.3/bin/psql -U atc -h 127.0.0.1 <<EOF
```
alext: This doesn't need `sudo`

concourse/scripts/shake_concourse_volumes.sh
alext: I've also ran into what looks like a race condition here. I fixed the script to have the correct path, and got an error:
```
umount: /var/vcap/data/garden/graph: not mounted
```
I think this directory gets automatically unmounted by something when garden is stopped, and the last container has exited. We might need to do something more robust after the `pkill` above.

PR #594, opened 2016-11-04
Upgrade to concourse 2.4.0

keymon:
[#133513945 Upgrade concourse 2.4.0](https://www.pivotaltracker.com/story/show/133513945)

What?
-----

We want to upgrade concourse to keep the latest version for support and because they claim if fixes issues managing volumes.

No changes in the config but some additional features and bugfixes.

To highlight:

>   feature
>
> Baggageclaim will now be more durable to corrupt volumes. Previously
> a borked metadata file would effectively wedge the Baggageclaim API,
> making the worker unrecoverable. You would see an error like "failed
> to list volumes" in your builds. Baggageclaim will now pretend these
> volumes don't exist in the API, and reap them from the disk.

[1] https://concourse.ci/downloads.html#v240

I also use the same version of stemcell than CF and bosh.

How to test?
-----------

Deploy concourse as explained in the README. Use it for a while.

Who?
----

Anyone but @keymon

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -22,8 +22,8 @@ resource_pools:
   - name: concourse
     network: concourse
     stemcell:
-      url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3263.5
-      sha1: a4a3b387ee81cd0e1b73debffc39f0907b80f9c6
+      url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3263.8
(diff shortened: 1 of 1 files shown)


alext:
> I also use the same version of stemcell than CF and bosh.

This makes sense, however, 3263.10 is also available. There's an argument to say that we should use the latest available so that we avoid having to upgrade this machine again in future.

keymon:
Hehe, that was released after this PR.

Mmm, I got mix feelings about this. The changes in that version are not really relevant for us, and I doubt we will need to upgrade to that version (see https://github.com/cloudfoundry/bosh/releases/tag/stable-3263.10) and it is nice to have the same version everywhere.

alext:
> The changes in that version are not really relevant for us, and I doubt we will need to upgrade to that version

That makes sense. I'd actually failed to find the release notes, so that was more of a question really.

PR #595, opened 2016-11-07
Use CF_PUSH_TIMEOUT for any `cf push` or `cf start` in our tests.

keymon:
# What

Currently we run several tests in parallel, and the load might make the operation last longer, so the test fails due timeout. It already happened several times in our CI environment

We use in most cases CF_PUSH_TIMEOUT (2m) for `cf push` or `cf start` currently 2m, instead of `DEFAULT_TIMEOUT` (30s) in the rest of operations.

In this PR we apply it to all the cases.

# How to review?

Double check that makes sense. I think it is quite safe to merge.

# Who?

Anyone but @keymon

### DIFF
--- a/platform-tests/src/acceptance/http_closed_test.go
+++ b/platform-tests/src/acceptance/http_closed_test.go
@@ -28,7 +28,7 @@ var _ = Describe("Http client", func() {
 			"-p", "../../example-apps/static-app",
 			"-d", config.AppsDomain,
 		).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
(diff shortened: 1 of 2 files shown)


mtekel:
Indeed, 30s was on the edge, in my dev:
```
• [SLOW TEST:33.946 seconds]
Http client
/tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_closed_test.go:49
  that tries to connect to apps
  /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_closed_test.go:37
```

PR #597, opened 2016-11-07
[#133855001] Avoid - and _ in generated passwords.

alext:
## What

We've recently hit an issue with a generated password that started with
a - tripping up one of the CLI tools we use (`cf create-service-broker`).

This therefore changes the password generation scheme we use to use hex
instead. It also increases the length from 12 to 18 chars so that the
generated passwords continue to represent the same number of random
bytes.

Note: because passwords are only generated when missing, this won't affect any existing passwords, only new ones.

## How to review

Code review.
Run the pipeline as far as generate-secrets.
* Existing passwords should be unchanged.
* New passwords should now be 18 chars of hex.

## Who can review

Not me.

### DIFF
--- a/manifests/shared/lib/secret_generator.rb
+++ b/manifests/shared/lib/secret_generator.rb
@@ -3,11 +3,11 @@ require 'openssl'
 require 'digest/md5'
 
 class SecretGenerator
-  PASSWORD_LENGTH = 12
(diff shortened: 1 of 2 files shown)


mtekel:
works. I only wonder if some component won't now complain about password being too long...
to test, i'll re-deploy from scratch (which I needed to do anyway).

PR #585, opened 2016-11-03
[#131975557] Fix rsyslog log forwarding and parsing

keymon:
[#131975557 Alternate solution to configure syslog forwarding without metron agent](https://www.pivotaltracker.com/story/show/131975557)

What?
-----

After merging #570 and #561, we have broken the log forwarding to logsearch.

See the story and commits for details.

We fix the issues in this PR

Dependencies
------------

https://github.com/alphagov/paas-syslog-release/pull/1 should be merged first, the code tagged, a new release built and the commit in this PR updated.

Both can be tested together.

How to review?
--------------

 * Deploy.
 * Check that the logs are being sent to loggregator and parsed properly
 * Check that in the VMs:
    * the `/etc/rsyslog.d/rsyslog.conf` using the template format with `[job=jobname index=X]`
    * the `/var/vcap/jobs/cleanup_syslog_forwarding/bin/cleanup_syslog_forwarding_ctl` script contains the code `# Cleanup metron_agent's rsyslog config, if present` and the file `/etc/rsyslog.d/00-syslog_forwarder.conf` is not there anymore.

Who?
----

Anyone but @keymon

PR #598, opened 2016-11-07
[#131975557] Add adhoc bosh ssh task to clean old rsyslog conf

keymon:
[#131975557 Alternate solution to configure syslog forwarding without metron agent](https://www.pivotaltracker.com/story/show/131975557)

What?
-----

In #561 and #585 we added some logic to remove the old rsyslog config created by metron_agent. But this solution did not work completely due some reason, and some VMs keeped the old file.

As it is quite slow and tedious the approach of creating an adhoc task to delete the file, we will use a concourse task that runs a `bosh ssh` on all the VMs.

This is simplier, quicker to develop and review, and easier to test and later remove. I also prevent unnecesary bosh deployments and restart of services after removing the task.

Currently this task will do bosh ssh on each VM, check if the file `/etc/rsyslog.d/00-syslog_forwarder.conf` is present, and if it is, remove it and restart rsyslogd.

This commit can be later reverted.

How to test
-----------

Fake the situation by dropping  this file on some VMs:

```
bosh ssh access/1 'sudo touch /etc/rsyslog.d/00-syslog_forwarder.conf'
bosh ssh api/1 'sudo touch /etc/rsyslog.d/00-syslog_forwarder.conf'
```

Apply this pipeline, and run the task

```
export DEPLOY_ENV=...
BRANCH=$(git rev-parse --abbrev-ref HEAD) JOB=post-deploy make dev run_job
```

You should see in the output of post-deploy/bosh-adhoc that the script is executed in all the VMs, and the ones that had the file did remove it and restart rsyslog.

Who?
---

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2117,6 +2117,7 @@ jobs:
           - get: cf-secrets
             passed: ['cf-deploy']
           - get: bosh-CA
+          - get: bosh-secrets
 
       - task: retrieve-config
         config:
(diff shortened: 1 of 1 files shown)


henrytk:
The bosh-adhoc task fails with:

```
Logged in as 'admin'
Target set to 'my-bosh'
sh: can't create updated-cf-manifest/cf-manifest.yml: nonexistent directory
```

PR #599, opened 2016-11-07
[#133106357] Switch rsyslog from RELP to TCP

saliceti:
## What
Story: [Fix rsyslog high cpu usage & getting stuck on restart](https://www.pivotaltracker.com/story/show/133106357)

There is an issue in our configuration when sending logs via RELP protocol from syslog to logstash via an ELB. If the ingestor is temporarily disconnected, the messages aren't logged anymore. Rsyslog tries to reconnect, opens many connections and consumes a lot of CPU.
We may investigate the root cause but in the meantime, we revert to the default TCP configuration because it doesn't show this behaviour.

## How to review
* Deploy
* SSH to `cell/0` for example
* Observe rsyslog is not connected to the ingestor ELB via RELP: `netstat -an | grep :2514`
* Observe rsyslog is connected to the ingestor ELB via TCP: `netstat -an | grep :5514`
* In AWS console, locate the ingestor ELB and remove the ingestor backends
* Observe rsyslog is disconnected: `netstat -an | grep :5514`
* Observe kibana doesn't show new logs
* Reconnect the ingestor backends
* Observe rsyslog reconnects automatically: `netstat -an | grep :5514`
* Observe kibana shows new logs

## Who can review
Not @paroxp or me

### DIFF
--- a/manifests/cf-manifest/runtime-config/runtime-config-base.yml
+++ b/manifests/cf-manifest/runtime-config/runtime-config-base.yml
@@ -21,6 +21,6 @@ addons:
     properties:
       syslog:
         address: (( grab terraform_outputs.logsearch_ingestor_elb_dns_name ))
-        port: 2514
-        transport: 'relp'
+        port: 5514
+        transport: 'tcp'
         log_template: 'metron_agent'


benhyland:
The 'how to review' instructions do nothing to verify cpu usage doesn't spike, and don't advise comparison with a revision exhibiting the problem. We should assume logsearch is tested well enough that their valid config is OK to deploy.

This sort of thing will be addressed as part of our process review but for now it's enough to believe the analysis in the story and confirm the config change looks sensible.

I'm happy with the syslog transport config change, but not sure why we still refer to the `ingestor_relp` in the `ingenstor_<zone>` jobs - do we still need this?

paroxp:
We didn't go into hassle of removing all `relp` references, as an [issue](https://github.com/logstash-plugins/logstash-input-relp/issues/18) has been raised upstream and we may wish to switch back to `relp` after that has been resolved. Is that ok?

benhyland:
So long as it was considered, that's fine, although I think it's quite likely we never return to this issue if TCP works out in the long term. Good to merge.

alext:
> We didn't go into hassle of removing all relp references, as an issue has been raised upstream and we may wish to switch back to relp after that has been resolved.

I'm not sure that's the best approach. I think it's good for hygiene to clean this sort of thing out, especially when we're not sure whether they're going to be needed in future or not. If they are needed in future, they're only a `git revert` away.

paroxp:
I see. Perhaps, a separate story may need to be played?

PR #593, opened 2016-11-04
WIP: [#133741599] Remove the cleanup_syslog_forwarding addon

keymon:
[#133741599 Remove syslog cleanup job](https://www.pivotaltracker.com/story/show/133741599)
[#131975557 Alternate solution to configure syslog forwarding without metron agent](https://www.pivotaltracker.com/story/show/131975557)

What?
-----

in #561 and 561cb6b we had to add a addon to remove the rsyslog config
file created by metron_agent.

The change has been deployed in all the environments, so we can remove
this cleanup job.

How to review
-----------

Check that makes sense.

Who?
---

Anyone but @keymon

PR #601, opened 2016-11-08
[#133741599] Revert "Add adhoc bosh ssh task to clean old rsyslog conf"

keymon:
[#133741599 Remove syslog cleanup job](https://www.pivotaltracker.com/story/show/133741599)

Dependencies
------------------

Wait for #598 to be deployed in prod!

What?
-----

Continuation of #593

We added an adhoc task to remove a spurios rsyslog config from the metron_agent.

The previous approach from #561 and 561cb6b to add a addon to remove
the rsyslog config file created by metron_agent did not work.

This reverts commit 82c8fc8743fe6145561bab96d1a60045a7f9ecde. The job has run in all the environments, deleting the undesired config and restarting rsyslog.

How to review?
---------------

check that makes sense

Who?
----

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2117,7 +2117,6 @@ jobs:
           - get: cf-secrets
             passed: ['cf-deploy']
           - get: bosh-CA
-          - get: bosh-secrets
 
       - task: retrieve-config
         config:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-bosh-cloudfoundry.yml
```diff
@@ -2151,32 +2151,6 @@ jobs:
                 done > config/config.sh
 
       - aggregate:
```
henrytk: The aggregate block now only has one step in it, and is therefore redundant.

keymon: Yeah, but it is handy to have it like that, don't you think so?

I won't insist to much on that, anyway, if you prefer it without it, I will revert all the merge.

henrytk: Meh, keep it. I suppose it promotes aggregation by default.

PR #606, opened 2016-11-15
Fix issues created by displaying grafana PW

mtekel:
## What

We do call the environment script in our pipeline and we run it in a container that does not have 'aws' tool installed - or even keys configured. Add same condition that we use for ATC password and adjust the relevant task in the pipeline to skip trying to retrieve grafana password

## How to review

Run bosh-cli tests task and make sure it passes.

## Who can review

not @mtekel

### DIFF
--- a/concourse/scripts/environment.sh
+++ b/concourse/scripts/environment.sh
@@ -43,13 +43,6 @@ if [ -z "${CONCOURSE_ATC_PASSWORD:-}" ]; then
   fi
 fi
 
-if [ -n "${DECRYPT_CONCOURSE_ATC_PASSWORD:-}" ]; then
-  GRAFANA_PASSWORD=$(pass "${DECRYPT_CONCOURSE_ATC_PASSWORD}/grafana_admin_password")
-else
-  GRAFANA_PASSWORD=$(aws s3 cp "s3://${DEPLOY_ENV}-state/cf-secrets.yml" - | \
-                     ruby -ryaml -e 'puts YAML.load(STDIN)["secrets"]["grafana_admin_password"]')
-fi
-
 cat <<EOF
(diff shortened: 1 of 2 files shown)


concourse/scripts/environment.sh
```diff
@@ -43,11 +43,13 @@ if [ -z "${CONCOURSE_ATC_PASSWORD:-}" ]; then
   fi
 fi
 
-if [ -n "${DECRYPT_CONCOURSE_ATC_PASSWORD:-}" ]; then
-  GRAFANA_PASSWORD=$(pass "${DECRYPT_CONCOURSE_ATC_PASSWORD}/grafana_admin_password")
-else
-  GRAFANA_PASSWORD=$(aws s3 cp "s3://${DEPLOY_ENV}-state/cf-secrets.yml" - | \
-                     ruby -ryaml -e 'puts YAML.load(STDIN)["secrets"]["grafana_admin_password"]')
+if [ -z "${GRAFANA_PASSWORD:-}" ]; then
+  if [ -n "${DECRYPT_CONCOURSE_ATC_PASSWORD:-}" ]; then
```
alext: This variable name is somewhat confusing, it wasn't clear what the concourse ATC password had to do with the Grafana password. I had to look in the makefile to understand what's going on here. I think this could be better named (maybe `PASS_DEPLOY_ENV_DIR` or similar).

I realise that this hasn't been introduced in the PR, but this seems like a good opportunity to address this.

mtekel:
renamed...

alext:
I think this is still going to fail in the self-update-pipelines task when on a non-dev environment because that uses the other path in the script, and therefore will try to use `pass` to lookup the password.

mtekel:
Updated again. Now only part of `showenv`.

alext:
:+1: This looks good to me.

PR #607, opened 2016-11-15
[#133751193] Enable datadog BOSH healthmonitor integration and add some monitors on healthy metric]

keymon:
[#133751193 Enable datadog BOSH healthmonitor integration](https://www.pivotaltracker.com/story/show/133751193)

NOTE: This is WIP because we are not sure about the cost of adding the metrics in datadog.

What?
----

[The BOSH health monitor runs on the director VM and receives heartbeats every minute from all bosh agents via the Bosh NATS bus](https://bosh.io/docs/monitoring.html). The heartbeats contain system metrics (CPU, Memory, load, etc) and if all the monit services are running on the VM `bosh.healthmonitor.system.healthy` .

We are mainly interested in this `bosh.healthmonitor.system.healthy` metric, and we will ignore the other metrics that are already being sent by the datadog agents on each VM.

[Bosh health_manager can be configured to send the metrics to datadog](https://github.com/cloudfoundry/bosh/blob/master/release/jobs/health_monitor/spec#L185-L194).

We will create datadog monitors each job/vm defined in the manifest using Terraform. The monitor will alert based on the threshold of the sum of `bosh.healthmonitor.system.healthy` metric for each job type. We set `warning` if we loose one instance and `critical` if we loose 2.

Jobs with only one instance are not supposed to be critical, we will receive only a warning. We do not monitor jobs with 0 instances in the manifest.

During deployment, if a VM is replaced we will receive a temporary warning, as the services get restarted. It makes sense as redundancy is reduced.

If bosh or NATS are redeployed, we will not get metrics at all. The monitors will alert with `No data` after 30 min, to cover the time to redeploy bosh.

Dependencies
------------

Merge https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/77

How to review?
-------------

Redeploy everything with datadog enabled:
```
ENABLE_DATADOG=true  BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines DEPLOY_ENV=...
```

Check that the monitors are working and defined.

Who?
----

Anyone but @keymon or @saliceti

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -79,6 +79,10 @@ jobs:
         user: hm
         password: (( grab secrets.bosh_hm_director_password ))
       resurrector_enabled: false
+      datadog_enabled: (( grab meta.datadog.enabled ))
+      datadog:
+          api_key: (( grab meta.datadog.api_key ))
+          application_key: (( grab meta.datadog.application_key ))
 
     agent:
       mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" terraform_outputs.bosh_fqdn ":4222" ))
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.7.3"
+    - TF_VERSION="0.7.10"
     - SPRUCE_VERSION="1.5.0"
     - DEPLOY_ENV="travis"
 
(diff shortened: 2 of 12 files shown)


manifests/bosh-manifest/bosh-manifest.yml
```diff
@@ -79,6 +79,10 @@ jobs:
         user: hm
         password: (( grab secrets.bosh_hm_director_password ))
       resurrector_enabled: false
+      datadog_enabled: true
```
keymon: true

mtekel:
All works, but there is a catch: The warn threshold is <. That is, if you have 2 VMs for a job and one is down, it's still ok, because warn is < 1 :
image

mtekel:
All works fine, merging.

PR #610, opened 2016-11-21
Correct failover when not logged into CF CLI

paroxp:
## What

We've ran into an issue, where the script failed with the message requesting correct AWS credentials, whilst not being logged in.

In order to avoid the confusion, we've placed an extra check just before the AWS check which should failover nicely.

## How to review

Attempt to run the script, whilst being logged out.

## Who can review

Not @paroxp

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -89,6 +89,10 @@ check_params_and_environment() {
     abort_usage "Org must be defined"
   fi
 
+  if ! cf orgs >/dev/null 2>&1; then
+    abort "You need to be logged into CF CLI"
+  fi
+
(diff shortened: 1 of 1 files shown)


scripts/create-user.sh
```diff
@@ -89,6 +89,10 @@ check_params_and_environment() {
     abort_usage "Org must be defined"
   fi
 
+  if ! cf orgs >/dev/null 2>&1; then
+    abort "You need to be logged into CF CLI as admin user"
```
saliceti: This doesn't actually test if the user is admin or not. Can we change the message so it deals only with logged in or not?

PR #611, opened 2016-11-23
[pipecleaner] Don't assume dummy variables are safe.

alext:
## What

When shellchecking scripts in the pipelines, pipecleaner inserts dummy
`export` statements for each of the `PARAMS` from the pipeline to prevent
shellcheck complaining about undefined variables. Because these were set
to the static string 'DUMMY', shellcheck wouldn't complain if they were
used unquoted in the scripts because it assumed their value was safe.
Inserting a call to `date` in these prevents shellcheck from assuming
they're safe.

## How to review

Code review.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/scripts/pipecleaner.py
+++ b/concourse/scripts/pipecleaner.py
@@ -57,7 +57,9 @@ class Pipecleaner(object):
                 script += "set " + switch + "\n"
 
         for name, value in variables.iteritems():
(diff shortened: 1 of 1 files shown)


Jonty:
Excellent fix, nice work. Will merge when the tests pass.

PR #614, opened 2016-11-24
[#134551329] Parallelise acceptance tests

combor:
## What

We want to speed up acceptance tests by enabling higher parallelisation.
The most stable configuration with huge reduction of tests run time was achieved with m4.xlarge VM type and 10 test nodes. Also we introduce 90 sec sleep time for `acceptance-tests` to spread initial load over time. Such config allows us to finish all tests suites in ~20 mins.

All test results and more reasoning: https://www.pivotaltracker.com/story/show/134551329 

## How to review

Read, run bootstrap to re-provision concourse and run the CF deploy pipeline twice.
Second run of `acceptance-tests` should take ~20min.  

## Who can review

Not @saliceti or @combor

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2332,7 +2332,7 @@ jobs:
             outputs:
               - name: artifacts
             run:
-              path: sh
+              path: bash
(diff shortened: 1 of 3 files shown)


dcarley:
Awesome! 🍻

PR #615, opened 2016-11-24
[#134920125] Don't run acceptance by default in dev

mtekel:
## What

[Dont run acceptance tests in dev](https://www.pivotaltracker.com/n/projects/1275640/stories/134920125)

TODO: Let's first consider if we still want to do this, in the light of #614
## How to review

`make dev pipelines`. Check that the acceptance tests don't run.

## Who can review

not @mtekel


alext:
As you mention, now that #614 is merged, the acceptance tests take about the same amount of time as the custom acceptance tests (which this wouldn't disable). I don't think this is therefore worth doing.  What do you think?

alext:
Closing this as we think #614 will sufficiently address the speed problems for now.

PR #616, opened 2016-11-25
Change dev bootstrap AZ to avoid capacity issues.

alext:
## What

We've been running into the following error fairly regularly when
creating bootstrap instances:

```
InsufficientInstanceCapacity => We currently do not have sufficient
m3.large capacity in the Availability Zone you requested (us-east-1e).
Our system will be working on provisioning additional capacity. You can
currently get m3.large capacity by not specifying an Availability Zone
in your request or choosing us-east-1b, us-east-1c, us-east-1a.
```

Using a different AZ (which is determined by the subnet being used)
normally avoids this, which suggests that the capacity problems are
limited to a single AZ.

This PR also removes the IP ranges from the comments, as they're not useful.

## How to review

Spin up a bootstrap concourse in dev. Verify this works without issue.

I don't think it'll be possible to know for sure whether this solves the issue without running it for a while.

## Who can review

Anyone but myself.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -6,19 +6,19 @@ Vagrant.require_version ">= 1.8.0"
 AWS_ACCOUNT = ENV.fetch("AWS_ACCOUNT", "dev")
 AWS_ACCOUNT_DATA = {
   "dev" => {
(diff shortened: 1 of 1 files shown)


mtekel:
Could it be because m3 is old instance type and AWS wants to phase them out?

benhyland:
lgtm

PR #617, opened 2016-11-25
Fix dashboard to match warn states in monitors

benhyland:
We had a typo (and no tests)


saliceti:
Tested successfully with dashing outside the pipeline. Merging.

PR #620, opened 2016-11-28
whitelist valtech ip address

benhyland:
## What

Added Holly-requested valtech ip to tenant whitelist in prod.
For off-site usability testing on 2016-12-02.

## How to review

Holly or her testee needs to check connectivity from the remote site before the user testing starts.


mtekel:
IP matches that from story, merging

PR #622, opened 2016-11-28
[#130188061] Use rds-broker boshrelease from the build concourse.

alext:
## What

Now that we have the build concourse building these releases, we should update the pipeline to consume them instead of the existing temporary solution.

Note: as part of the work to move it to the build concourse, we also
renamed the boshrelease to match the broker.

## How to review

Deploy. Verify it downloads the release successfully from the new location.

## Who can review

Anyone but myself.

### DIFF
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -28,10 +28,10 @@ meta:
       allocated_storage: 20
 
 releases:
-  - name: aws-broker
-    version: 0.0.12
-    url: https://github.com/alphagov/paas-rds-broker-boshrelease/releases/download/0.0.12/aws-broker-0.0.12.tgz
-    sha1: 5de975ce69085df22870fc695986fc4331a37881
+  - name: rds-broker
+    version: 0.1.1
+    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/rds-broker-0.1.1.tgz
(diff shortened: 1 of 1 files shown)


Jonty:
Works as described!

mtekel:
Should we have a tag/release in https://github.com/alphagov/paas-rds-broker-boshrelease/releases ? Otherwise it's hard to tell what this is built from...

alext:
> Should we have a tag/release in https://github.com/alphagov/paas-rds-broker-boshrelease/releases

I don't think so. The use of Github releases was a temporary thing. The build concourse is now the place to see what a given release was build from.

mtekel:
What I mean is that it's not clear what code this is built from. Should the code be tagged at least?

alext:
Yeah, it could do that. At the moment, this is very much the MVP (the scope was pretty large as it was). I've noted several potential follow-ups in the story. I'll add git tagging to the list.

mtekel:
I thought that that was a must-have and very essential thing. BOSH release creation process is bound to git and will e.g. alert if you have uncommitted changes. The fact that you can't tell (from a tag in bosh release repo) what was the release built from was one of the arguments against local builds and for CI...

alext:
> The fact that you can't tell (from a tag in bosh release repo) what was the release built from was one of the arguments against local builds

True, but in this case you can tell by looking at the build history in the build concourse. It's not super convenient, but the information is available.

I'd also note that this is not really the right PR to be having this discussion about as this is merely updating paas-cf to consume the generated release.

PR #624, opened 2016-11-29
[#SUPPORT] Temporally disable reserved characters tests

combor:
## What
We know that java-buildpack 3.9 doesn't pin tomcat version. Tomcat starting from 8.0.39 has changed which character are allowed in URL. We need to work with upstream to fix this. In the time being we need to unblock the pipeline. 

## How to review

Run acceptance tests from this branch. 

## Who can review

not @combor


bleach:
After reading this PR and the commit message I'm unclear what the problem is and which commit introduced it.

alext:
I've confirmed that the changes work correctly. The tests pass, and only one extra test has been disabled.

combor:
Updated PR and commit comment.

combor:
BTW. The bug most likely is somewhere here: https://github.com/cloudfoundry/java-buildpack/blob/b05095483d44bba001009f295ba2ff563140bff3/lib/java_buildpack/util/tokenized_version.rb#L37

combor:
And that's the list of available tomcat versions: https://java-buildpack.cloudfoundry.org/tomcat/index.yml

combor:
Issue created: https://github.com/cloudfoundry/java-buildpack/issues/355

PR #626, opened 2016-11-29
[#130821719] User research org cleanup script

benhyland:
[#130821719 Automate reset of User Research organisation after Lab Session](https://www.pivotaltracker.com/story/show/130821719)

## What

Introduce a script to wipe and recreate the `paas_user_research` org
and to reset the passwords of the user testing users.

This will be run manually after each user research session, to ensure that the next session starts with a clean environment.

See https://www.pivotaltracker.com/story/show/130821719 and https://github.com/alphagov/paas-team-manual/pull/75

## How to review

Either:
- review code and story summary and trust us that it works, or
- create an org in dev with some apps, services, and users, and run the script with appropriate modifications to see it for yourself, or
- ask @benhyland or @alext for a walkthrough

## Who can review

Anyone but @benhyland and @alext

### DIFF
--- a/scripts/reset-org.py
+++ b/scripts/reset-org.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+import subprocess
+import sys
+import time
+
+class SubprocessException(Exception):
+    pass
+
+class OrgReset(object):
+    def __init__(self, org, initial_space, org_managers, users, users_are_org_managers, users_are_space_managers, users_are_space_developers, org_quota):
+        self.org = org
+        self.initial_space = initial_space
+        self.org_managers = org_managers
+        self.users = users
+        self.users_are_org_managers = users_are_org_managers
+        self.users_are_space_managers = users_are_space_managers
+        self.users_are_space_developers = users_are_space_developers
+        self.org_quota = org_quota
+
+    def confirm(self):
+        if not self.confirm_user_input('Are you sure you want to delete org %s? (y/n): ' % self.org):
+            sys.exit("'No future change is possible.' (phew, not deleted)")
+
+    def confirm_user_input(self, question):
+        response = raw_input(question)
+        return response == "y"
+
+    def whoami(self):
+        self.user = self.parse_current_user(self.run(['cf', 'target']))
+
+    def delete_org(self):
+        # require org to be deleted in advance? have this check for org and and advise, so we don't use the `-f` flag
+        # Prompt for confirmtaion...
+        self.run_with_retry(['cf', 'delete-org', self.org, '-f'], 20, lambda err:
+            'one or more resources within could not be deleted' in err.output
+        )
+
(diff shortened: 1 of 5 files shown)


keymon:
The PR was missing the [#130821719] in the title and the link in the body. I added then.

scripts/reset-user-testing.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+
+# shellcheck disable=SC1090
+source "${SCRIPT_DIR}/common.sh"
+
+check_aws_account_used prod
```
keymon: It would be convenient to have a variable that can be easy to override to specify the environment:

```
TARGET_AWS_ACCOUNT=${TARGET_AWS_ACCOUNT:-prod}
``` 
and then use `$TARGET_AWS_ACCOUNT` instead of hardcoded prod everywhere.

The same for the `--no-email` option, that was added in a previous commit but not used. You could use a variable like:

```
"${SCRIPT_DIR}"/rotate-user-password.sh -e prod -u holly.challenger+1@digital.cabinet-office.gov.uk ${NO_EMAIL:+--no-email}
```

So if `$NO_EMAIL` is set, it will add the variable there.

keymon:
Having a glance to the code seems OK. I wouldn't use python, I think it makes the script more complicated and simple shell script would be OK, but it is good anyway. It would be easy to change some things as commented to allow test the script easy, but being so small it is easy to change.

I will test it in a dev environment and merge, unless you want to do some of the proposed changes.

benhyland:
Better to merge asap imo, I'll do some tweaks in a separate PR

scripts/reset-user-testing.sh
benhyland: I don't really want to encourage people to use this script for manual testing. Are you sure these changes would be valuable in the longer term, or did you think of them for a more convenient review?

scripts/reset-org.py
```diff
@@ -0,0 +1,120 @@
...
+    parser.add_argument('--org-managers', help='Add the OrgManager role for the listed team members', action='store_true', default=False)
+    parser.add_argument('--space-managers', help='Add the SpaceManager role for the listed team members in the initial space', action='store_true', default=False)
+    parser.add_argument('--space-developers', help='Add the SpaceDeveloper role for the listed team members in the initial space', action='store_true', default=False)
+    parser.add_argument('--quota', help='Name of the org quota to use')
+
+    args = parser.parse_args()
+
+    org_reset = OrgReset(args.o, args.s, args.m, args.u, args.org_managers, args.space_managers, args.space_developers, args.quota)
+    org_reset.confirm()
+    org_reset.whoami()
```
keymon: I think it makes more sense print `whoami` and then ask for confirm.

PR #628, opened 2016-11-30
[#133247209] Datadog: don't false alarm during deployment

mtekel:
## What

[Datadog: don't false alarm during deployment](https://www.pivotaltracker.com/n/projects/1275640/stories/133247209)
Reconfigure datadog checks so that we don't get false positives (alerts, no_data notifications) during BOSH deployments (when you e.g. rotate a stemcell).

## How to review

Deploy environment with datadog enabled. Apply this branch. Then bump stemcell (e.g. change the number to 3263.10 in `manifests/cf-manifest/manifest/000-base-cf-deployment.yml` - you'll need to commit. To test that, create a branch from this branch and commit there, or you can add a temporary commit into this branch). Apply. There should be no false positive alerts.

Note that we have tested stemcell bump (to 3236.10) several times during development, with passing acceptance and custom acceptance tests. I think we could leave the stemcell bump a part of this branch, as the reviewer will have to test it as well. And then it becomes tested by both developer and reviewer, so no reason to drop it and do upgrade in separate story.

## Who can review

not @mtekel or @henrytk

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -42,6 +42,6 @@ stemcells:
 update:
   canaries: 0
   max_in_flight: 1
-  canary_watch_time: 30000-600000
-  update_watch_time: 5000-600000
+  canary_watch_time: 30000-420000
+  update_watch_time: 5000-420000
   serial: false
--- a/terraform/datadog/cell.tf
+++ b/terraform/datadog/cell.tf
@@ -3,7 +3,7 @@ resource "datadog_monitor" "cell-available-memory" {
   type               = "query alert"
   message            = "${format("Less than {{threshold}}%% memory free on cells. There is only {{value}} %% memory free on average on cells. Review if this is temporary or we really need to scale... @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
   escalation_message = "There is only {{value}} % memory free on average on cells. Check the deployment!"
-  no_data_timeframe  = "5"
+  no_data_timeframe  = "7"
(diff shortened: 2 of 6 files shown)


alext:
Won't increasing these time windows increase the time it takes to alert for a genuine problem?

Is there not a mechanism in datadog whereby we can specify planned downtime for a host and silence the alerts for that period only (Something like the "scheduled downtime" in Nagios/Icinga)?

henrytk:
> Won't increasing these time windows increase the time it takes to alert for a genuine problem?

Yes, but weigh that against the reduced noise and increased chance of us noticing and acting on an alert and I think it is worth the increase.

> Is there not a mechanism in datadog whereby we can specify planned downtime for a host and silence the alerts for that period only (Something like the "scheduled downtime" in Nagios/Icinga)?

There are ways of silencing monitors and hosts for periods of time (although you can't mute hosts as part of a cluster check for some reason). The problem is in how it could be automated in an intelligent fashion. It is likely you would end up silencing Datadog alerts for most if not all of the deployment.

keymon:
You set the `update_watch_time` for bosh to 7m, and then the no_data timeout also to 7 minutes.

Wouldn't it make sense to set the no_data to slightly more, like 8m. You actually mention that in the story itself.

mtekel:
Reality of update speed is that worst case scenario is 5, thus 7 gives some reserve. For disk check we set 10 mins in the end. All others that are set to 7, do update in practice much faster (e.g. 2mins, 3 mins). We could speculate what is the best number, but for now 7 seems good compromise for both timeouts and bosh update delay. Moreover, in case VM is not re-created, it is quite accurate.

In general, we should re-think our checks and decide which concepts we want to apply. This might include removing no_data notification on most of the monitors...

keymon:
This works as expected. I tested it a pair times and it is fine. There are not alerts sent, only warnings from bosh healthy during deployment, which is expected and desired. 

But I think we can do a separate PR for the stemcell ;)

Merging.

PR #629, opened 2016-11-30
[#135207089] Use java offline buildpack

keymon:
What?
-----

The acceptance tests started failing due the java-buildpack using a newer minor version for tomcat (from 8.0.38 to 8.0.39), which changed the behaviour with special chars in the request. See #624 for details.

In this PR we want to use the java-offline-buildpack, which pins all the versions of the dependencies instead of allowing minor upgrades. This will allow us to better control the upgrades.

The commit will enable again the test.

We also simplify the buildpack a little bit.

How to test?
------------

 * Code review
 * The pipeline should pass. Trust us, let it run in CI ;)

Who?
---
Anyone but @keymon or @combor

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -301,9 +301,7 @@ properties:
         total_routes: 1000
     default_quota_definition: default
 
-    user_buildpacks: []
-    system_buildpacks: (( grab properties.cc.default_buildpacks ))
-    default_buildpacks:
+    install_buildpacks:
       - name: staticfile_buildpack
         package: staticfile-buildpack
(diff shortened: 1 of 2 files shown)


alext:
Does this change have any implications for tenants deploying Java apps?

keymon:
[Commented in the story](https://www.pivotaltracker.com/story/show/135207089/comments/156789927):

> *  No tenant intervention is required. The staged apps will still use the droplet with whatever version they use.
> *  If an app is restaged, it might pull a different version (older, but at minor version) than the one that it was staged. For instance, maybe the app is using tomcat 8.0.39 and then with the offiline buildpack goes back to 8.0.36. But that is fine as with the "online" one we accepted minor version changes. Also we also accept minor version changes during CF upgrades.

PR #630, opened 2016-11-30
[#127792523] Enable SSH for CF

paroxp:
## What

We'd like to enable the ssh functionality for our CF. The flag has been switched, and the tests are being corrected.

This also enables the upstream ssh tests in cf-acceptance-tests.

We've updated the `ssh-proxy oauth` config to match docs. This looks to have changed in newer releases since this was initially implemented. This scope is what the [documentation](http://docs.cloudfoundry.org/running/config-ssh.html#cf-manifest) says should be set.

We use `vm_extensions` to specify `access` VM ELB. This avoids the need to have an extra vm_type just to add the VMs to an ELB.

We make the `access` use a small VM. It's only proxying SSH connections, so doesn't need to be very big. Some crude load testing shows that this VM size is plenty (we managed to get a peak load average or 0.2)

## How to review

  - Run the pipeline from this branch
  - Create/Push an app
  - Attempt to `cf ssh` into it (You may need to have `SpaceDeveloper` role set)
  - **Optional**: If you've got app already:
    - Establish if you can SSH
    - If not, run `cf enable-ssh`
    - Retry ssh'ing.

## Who can review

Neither @alext nor @paroxp

### DIFF
--- a/manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
+++ b/manifests/cf-manifest/cloud-config/020-cf-vm-types.yml
@@ -104,16 +104,6 @@ vm_types:
         - (( grab terraform_outputs.cf_router_elb_name ))
 
 # Diego below
-  - name: access
-    network: cf
-    env: (( grab meta.default_env ))
-    cloud_properties:
-      instance_type: m3.medium
-      ephemeral_disk:
-        size: 10240
-        type: gp2
-      elbs:
-        - (( grab terraform_outputs.cf_ssh_proxy_elb_name ))
 
   - name: cell
     network: cell
--- a/manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
+++ b/manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
@@ -1,0 +1,8 @@
+---
+
+vm_extensions:
+
+  - name: ssh_proxy_elb
+    cloud_properties:
+      elbs:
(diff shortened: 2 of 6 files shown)


mtekel:
I wonder about the scopes: what is this ssh-client used for exactly? Why does it need admin scope and which functionality does it enable that it didn't previously have?

paroxp:
I don't think you need to be an admin. The Developer role, should be sufficient enough.

It simply allows a developer, to ssh into the application VM, mess around with the system if a desire appears.

mtekel:
I was asking about this: https://github.com/alphagov/paas-cf/blob/30bca251d4d00d7937985b31d3f85cbac86d022f/manifests/cf-manifest/manifest/020-cf-properties.yml#L492

The added scope enables this ssh-proxy client to act as admin... I just wonder what is this needed for. I had impression that we fixed it previously and then decided to disable it, meaning it was working without that scope. But that was in previous CF release.

alext:
> The added scope enables this ssh-proxy client to act as admin

I'm not sure why it needs that, but that's what the documentation specifies - http://docs.cloudfoundry.org/running/config-ssh.html#cf-manifest. It seemed good to follow the docs, otherwise we may end up with hard-to-diagnose issues in future.

combor:
Tested and it LGTM. Merging.

mtekel:
That docs look like someone maybe copy-pasted it wrong. I just hope that this doesn't open a possible privilege escalation hole. It seemed to work before without it, so I wondered why it does need such strong/escalated scope now...

PR #631, opened 2016-11-30
[#135224045] Disable RDS auto minor version upgrade

saliceti:
## What
Story: [Upgrade RDS instance versions](https://www.pivotaltracker.com/story/show/135224045)

We are not sure how the next upgrade will impact us and we are planning a major upgrade anyway in #135224045. This removes the unknown for now and we will decide if we want to re-enable it or not in that story.

## How to review
* Deploy
* Check in the AWS console that the setting `Auto Minor Version Upgrade` is set to `No` for both BOSH and CF RDS instances

## Who can review
Anyone but me


keymon:
works

PR #633, opened 2016-12-01
[#135024433] Whitelist IP for DWP carers trial.

alext:
## What

https://www.pivotaltracker.com/story/show/135024433 - new trial users need this IP added to the whitelist so that they can access the API etc.

## How to review

Verify that the IP matches the story, and that the change is well-formed.

## Who can review

Anyone but myself.


benhyland:
lgtm

PR #636, opened 2016-12-02
Fix showenv on old bash, and include uaa_admin_password

alext:
## What

The `^^` uppercasing variable substitution is only available in bash 4+.
OS X still ships with bash 3.2. Changing this to use `tr` is less elegant,
but more portable.

This PR also adds `uaa_admin_password` to the list of variables returned.
This is used to login to dev deployments, and so is useful to have present here.

## How to review

Verify that `make dev showenv` works without error.

## Who can review

Anyone running bash 3.2.


mtekel:
works fine, merging

PR #637, opened 2016-12-02
[#130821719] Fix create_user.sh with the --no-email flag

alext:
## What

This was a copy/paste error from the rotate-user-password.sh script,
which uses USERNAME instead of EMAIL.

This also updates the printing of password to avoid any possible confusion around quotes and .

## How to review

Test that you can create users in a dev deployment using the `--no-email` flag. eg:
```sh
./scripts/create_user.sh -o foo -e foo@example.com --no-email
```

## Who can review

Anyone but myself.

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -189,7 +189,7 @@ send_mail() {
 }
 
 print_password() {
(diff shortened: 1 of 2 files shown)


mtekel:
now it works, merging. also, copy-paste of password is now nice and easy...

PR #638, opened 2016-12-02
[#133604325] Create the CVE monitor

paroxp:
## What

In order to respond with grace to any upcoming CVE, we need the CVE Notifier do its job properly.

We're creating the monitor alert, which will send an email to our DeskPro account, creating the CVE ticket for the support to deal with.

This should be deployed to production only and include the necessary tags, to be considered on our PaaS Dashboard (smashing).

## How to review

  - Push pipelines from this branch with `ENABLE_DATADOG=true` and `ENABLE_CVE_NOTIFIER=true` set
  - Check if new monitor has been created
  - You may want to trigger some fake event, in order to send an email

## Who can review

Not @paroxp


concourse/scripts/pipelines-bosh-cloudfoundry.sh
```diff
@@ -110,6 +110,7 @@ datadog_api_key: ${datadog_api_key:-}
 datadog_app_key: ${datadog_app_key:-}
 enable_datadog: ${ENABLE_DATADOG}
 enable_cve_notifier: ${ENABLE_CVE_NOTIFIER:-false}
+enable_cve_monitor: ${ENABLE_CVE_MONITOR:-false}
```
keymon: As I see more and more of these settings, more I think that we should split the config as [commented here](https://github.com/alphagov/paas-cf/pull/584#issuecomment-258907304). 

We should have one place (e.g. .yaml per environment) where we enable or disable this, and that can be read consistently across the pipeline, instead of explicitly passing it each time.

But not an issue to solve in this PR.

keymon: Anyway, why create a different variable? Why not use `$ENABLE_CVE_NOTIFIER` ?

keymon: I vote for having only one. I doubt we will enable one without enabling the other.

BUT, maybe you can parametrise the support email instead. In dev/staging you can use the alert maillists. This way, in development, we don't spam support.

keymon: From what I see, we do not pass the `-var-file` for each environment when calling terraform: https://github.com/alphagov/paas-cf/blob/ace39811b938419861b3f43b0d04df808680bb4a/concourse/pipelines/create-bosh-cloudfoundry.yml#L2033-L2035

combor: Regarding first comment: we've got Vault story scheduled for this sprint.

terraform/datadog/cve-notifier.tf
```diff
@@ -19,3 +19,23 @@ resource "datadog_monitor" "cve-notifier" {
...
   }
 }
+
+resource "datadog_monitor" "cve-notifier-reporter" {
+  count             = "${var.enable_cve_notifier}"
+  name              = "New CVE Detected"
+  type              = "event alert"
+  message           = "${format("New CVE has been found on https://pivotal.io/security @%s", var.support_email)}"
+  no_data_timeframe = "180"
+  query             = "events('priority:all tags:service:pivotal').rollup('count').last('1h') >= 1"
```
keymon: I would put 24h

keymon:
LGTM

terraform/datadog/variables.tf
```diff
@@ -7,3 +7,8 @@ variable "job_instances" {
   description = "List of pairs <job_name>:<job_count> of expected bosh job instance count"
   default     = []
 }
+
+variable "support_email" {
+  description = "DeskPro email address"
+  default     = "govpaas-alerting-dev@digital.cabinet-office.gov.uk"
```
mtekel: you can omit default. this will require the address to be always defined explicitly. which is sometimes a good form of check...

PR #642, opened 2016-12-05
[#122026001] Ensure that personal data is deleted as described in our policy 

combor:
## What

We want to be sure that when a user stops using the service, their personal data is removed, and that any records of their actions/personal data in logs is automatically (programmatically) purged to facilitate a maximum 2-year retention period.

This PR adds tests around our retention period settings to ensure that when we change them such changes are reflected in our Privacy Policy document. 

## How to review

Run the pipelene and check if deployment is no-op and all tests pass. 

## Who can review

not @combor or @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -2396,6 +2396,8 @@ jobs:
           file: paas-cf/concourse/tasks/custom-acceptance-tests-run.yml
           params:
             DISABLE_CUSTOM_ACCEPTANCE_TESTS: {{disable_custom_acceptance_tests}}
+            DEPLOY_ENV: {{deploy_env}}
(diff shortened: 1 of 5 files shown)


alext:
This fails when run within the pipeline with the following error:
```
• Failure [0.001 seconds]
S3 logs bucket
/tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/s3_retention_test.go:42
  should have retention period set to 30 days [It]
  /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/s3_retention_test.go:41

  Expected error:
      <*awserr.baseError | 0xc820015000>: {
          code: "MissingRegion",
          message: "could not find region configuration",
          errs: nil,
      }
      MissingRegion: could not find region configuration
  not to have occurred

  /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/s3_retention_test.go:38
```

PR #644, opened 2016-12-06
Add script to handle cf admin login in dev.

alext:
## What

In development we need to login as the cf admin user, which has an
autogenerated password which needs to be pulled out of the state bucket.
This script simplifies this process.

This is deliberately restricted to operate on dev only because we
should be using our individual user accounts in other environments.

## How to review

* Verify that the script correctly logs you into a dev CF deployment.
* Verify that it fails with a helpful error when using AWS keys for a different account.

## Who can review

Anyone but myself.


mtekel:
works fine, thanks for a nice helper script

PR #647, opened 2016-12-07
[#133036569] Migrate to new ELB security policies

combor:
## What

Story: https://www.pivotaltracker.com/story/show/133036569

The default behaviour of Terraform is for the ELBs to get the
AWS default security policy at the time that the ELB is created. This
has led to our long-lived deployments having an older security policy,
which raises warnings in Trusted Advisor etc.

This therefore updates all ELBs that terminate SSL to use the latest
built-in AWS policy (2016-08).

## How to review

Code sanity check. 
Run the pipeline and check if `cf-terraform` job applied new policies. 
Run bootstrap and check if Concourse ELB has been updated. 

## Who can review

Not @alext or @combor

### DIFF
--- a/terraform/cloudfoundry/cf_api_elb.tf
+++ b/terraform/cloudfoundry/cf_api_elb.tf
@@ -31,6 +31,15 @@ resource "aws_elb" "cf_cc" {
   }
 }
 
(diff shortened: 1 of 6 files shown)


keymon:
works fine.

alext:
I'll take care of running the bootstrap pipelines in ci, staging and prod.

PR #648, opened 2016-12-07
[#134423349] Monitor concourse getting stuck

mtekel:
## What

["Monitor concourse"](https://www.pivotaltracker.com/n/projects/1275640/stories/134423349)
Monitor concourse load as a mean to detect that it is broken.
Monitor continuous smoke tests duration as a mean that concourse is too slow and about to break, or to indicate that pipeline can be blocked.

## How to review

Deploy. Apply huge load on your concourse (>200). Verify that monitors trigger. Start running continuous smoke tests. Then stop them. After 30 mins you should get notification.

## Who can review

not @mtekel


keymon:
Code looks good... deploying to see it in action...

terraform/datadog/concourse.tf
```diff
@@ -0,0 +1,90 @@
...
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "concourse"
+  }
+}
+
+resource "datadog_monitor" "continuous-smoketests" {
+  name               = "${format("%s continuous smoketests runtime", var.env)}"
```
keymon: May I suggest we add the work "concourse" on this description. That way is easier to find the monitor.

keymon:
This works fine. If you answer the comment in https://github.com/alphagov/paas-cf/pull/648#discussion_r91483228 I am happy to merge.

terraform/datadog/concourse.tf
```diff
@@ -0,0 +1,90 @@
+resource "datadog_monitor" "concourse-load" {
+  name               = "${format("%s concourse load", var.env)}"
+  type               = "query alert"
+  message            = "${format("Concourse load is too high: {{value}}. Check VM health. @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
+  escalation_message = "Concourse load still too high: {{value}}."
+  notify_no_data     = false
+  query              = "${format("avg(last_5m):avg:system.load.5{bosh-job:concourse,bosh-deployment:%s} > 200", var.env)}"
```
keymon: The load.5 is already an average of 5m, then we average again in concourse. But I don't think this is an issue at all, and it works fine.  I am not sure how to express it differently, neither.

PR #651, opened 2016-12-08
Make dashboard link to triggered alerts

bleach:
## What

Link to triggered monitors, rather than the managed monitors, so I can quickly see a list of things making an environment not be green.

Add instructions for running the dashboard locally because I wanted to do that to make this change.

## How to review

- Follow the links and make sure they link to the right places
- Test the instructions for running locally

## Who can review

Not me

### DIFF
--- a/tools/paas_dashboard/README.md
+++ b/tools/paas_dashboard/README.md
@@ -31,3 +31,10 @@ See [dashing.io](http://dashing.io) and the [smashing wiki](https://github.com/d
 
 We deploy our smashing app as part of the `create-bosh-cloudfoundry` pipeline.
 It should be available at https://paas-dasboard.cloudapps.digital
+
+### Running locally
+
+Obtain API and app keys from https://app.datadoghq.com/account/settings#api
+
+```bundle install
+ DD_API_KEY=<api key> DD_APP_KEY=<app key> bundle exec smashing start```
--- a/tools/paas_dashboard/dashboards/paas-overview.erb
+++ b/tools/paas_dashboard/dashboards/paas-overview.erb
@@ -8,17 +8,17 @@ $(function() {
 <div class="gridster">
   <ul>
     <li data-row="1" data-col="1" data-sizex="1" data-sizey="1">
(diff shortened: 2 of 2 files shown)


saliceti:
I think there was an issue with the triggered monitors view as we couldn't filter based on the environment. That is why the managed triggers view was used instead.
@benhyland maybe remembers better?

benhyland:
I only remember a bit.

There is not a perfect correspondence between the monitors api and the various UI views. The UI view which most consistently came close to showing what was represented by the api response was the Manage Monitors view, not the Triggered Monitors view. I think if we use the Triggered Monitors view we will regularly click through from a non-green dashboard and not see the monitor we were looking for. Unfortunately I don't remember exactly why - it might be to do with silencing/no data, which we plan to fix anyway.

It's probably worth a try with the new link if Graham is happy with it, we can always change it back if we don't like it.

bleach:
The discrepancy I've noticed is that triggered seems to show "per host" and managed shows "per monitor", which also seems to be the way the dashboard counts them.

screen shot 2016-12-09 at 17 51 54

screen shot 2016-12-09 at 17 52 09


So, right now we have 3 elasticsearch nodes in "warn" state for disk space. On the dashboard they show as 1 warning, on the triggered view they show one per host, 3 warnings. I'm totally fine with this in return for not seeing all the "OK" monitors, but maybe not everyone else is.

I'm also think that there's a possibility, for the disk space alerts, that you could not notice that there are multiple distinct problems. I'm guessing that if there were multiple disk space warnings, "triggered" would show each host and "managed" would show just one warning.

So, I think, but am not sure, that the triggered view is still more useful.

benhyland:
I think sometimes monitors detected by the dashboard show up in the managed monitors view but not in the triggered monitors view.
It's plausible that this happens only with certain causes of the 'No Data' state, but I don't recall.
I expect we'll see some kind of discrepancies until DD give us an api which allows us to address triggered hosts, instead of monitors.
@saliceti I'm merging this, since this conversation seems to be becoming about whether to merge, which was not at all my intent when I originally commented.

PR #652, opened 2016-12-08
Increase elasticsearch data storage

bleach:
## What

I've increased the size of our elasticsearch persistent disks because they are fairly steady at 80% full (we alert at 75% full). I think we'd prefer to be at something more like 65% full.

## How to review

Technical review (optional):

1. `df -h /var/vcap/store` on `elasticsearch_master/0` (and 1 and 2 if you are very thorough)
1. deploy this change
1. Repeat step 1 and verify that the size has increased
1. Check that all your elasticsearch data didn't disappear

## Who can review

Anyone but @bleach

### DIFF
--- a/manifests/cf-manifest/cloud-config/030-logsearch.yml
+++ b/manifests/cf-manifest/cloud-config/030-logsearch.yml
@@ -31,7 +31,7 @@ vm_types:
 
 disk_types:
   - name: elasticsearch_master
-    disk_size: 307200
+    disk_size: 377856
     cloud_properties: {type: gp2}
   - name: queue
     disk_size: 102400


bleach:
Graph of average disk usage of /var/vcap/store on elasticsearch_master, showing that usage has levelled off recently

screen shot 2016-12-09 at 09 17 27

bleach:
@keymon commented on slack that we probably want to rework this to only increase the disk space in prod & I agree

mtekel:
another thing to think about is by how much we increase: we recently bumped to 300G. If we really need disk space to store more logs, because we have more usage on platform, then we should bump to 500 or 600G so that we don't have to do this frequently...

rboulton:
@bleach One thing to consider when calculating disk space requirements for elasticsearch is - what happens if one node goes down, and all the data on that node gets rebalanced to other nodes?  If you have a 3 node cluster, then that would increase the space used on each remaining node by 50%.

I've seen clusters make themselves run out of disk space when this happens; for a 3 node cluster, something with sufficient margin below 66.6% should be considered the maximum safe disk usage, for this reason.

bleach:
Thanks @rboulton - I had to make a new PR, #658 because I forced push this branch after closing this PR. I've linked to your comment from there

PR #653, opened 2016-12-09
[#133036569] Set ELB security policies (take 2)

alext:
## What

The default behaviour if this is unspecified is for the ELBs to get the
AWS default security policy at the time that the ELB is created. This
has led to our long-lived deployments having an older security policy,
which raises warnings in Trusted Advisor etc.

This therefore updates all ELBs that terminate SSL to use the latest
built-in AWS policy (2016-08).

We previously attempted this in #647, but had to revert. This was due
to a misunderstanding about how these policies work. An ELB has one or
more policies defined for it. Each SSL listener then has one of these
policies attached to it. The previous commit failed because we were
attempting to attach a policy to the listeners that didn't exist on the
ELB.

## How to review

To test this properly, it's necessary to have an ELB configured with an old policy (and with no new policies attached at all):
* Deploy from master as far as cf-terraform
* In the web UI, change the SSL policy for one of the ELBs to an old policy (Using the "change" link for the Ciphers in the listener properties).
* Delete the now unused policy from the elb (substitute cf-router if changing a different one):
    `aws elb delete-load-balancer-policy --load-balancer-name ${DEPLOY_ENV}-cf-router --policy-name ELBSecurityPolicy-2016-08`

Now deploy from this branch (again as far as cf-terraform).
Verify that the ELB has had it's policy updated to the current one (it may be necessary to refresh the web UI before the change is visible).

## Who can review

Anyone but myself.

### DIFF
--- a/terraform/cloudfoundry/cf_api_elb.tf
+++ b/terraform/cloudfoundry/cf_api_elb.tf
@@ -31,6 +31,17 @@ resource "aws_elb" "cf_cc" {
   }
 }
 
+resource "aws_lb_ssl_negotiation_policy" "cf_cc" {
+  name          = "paas-${var.default_elb_security_policy}"
+  load_balancer = "${aws_elb.cf_cc.id}"
+  lb_port       = 443
+
+  attribute {
+    name  = "Reference-Security-Policy"
+    value = "${var.default_elb_security_policy}"
+  }
+}
+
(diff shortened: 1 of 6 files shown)


keymon:
This works and LGTM, although it LGTM before as well :)

Hope it works this time. I will monitor it.

PR #654, opened 2016-12-09
[#135801295] Upgrade stemcell for BOSH and Concourse

mtekel:
## What

[Apply USN-3151-2 everywhere](https://www.pivotaltracker.com/n/projects/1275640/stories/135801295).

To apply the security fix, we need to upgrade stemcell used by concourse and BOSH. We have decided to upgrade to the same version as CF uses currently.

## How to review

Apply. You can either watch (and trust) the output from bosh-init, which should say `Uploading stemcell 'bosh-aws-xen-hvm-ubuntu-trusty-go_agent/3263.12'...` or ssh to bosh/concourse (`make dev ssh_bosh`/`make dev ssh_concourse`) and watch logs of bosh agent, which reports which stemcell it thinks the VM is running - and trust this information. Logs are located in `/var/vcap/bosh/logs` and it should look like `[App] 2016/12/09 11:35:15 INFO - Running on stemcell version '3263.12' (git: 41edff6464246431d6440b2446bb9cc5f872167f+)`

## Who can review

not @mtekel

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -154,8 +154,8 @@ resource_pools:
 - name: bosh
   network: private
   stemcell:
-    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3263.8
-    sha1: 95ecc2709ac62f21d0a1c6cac023585c3919b825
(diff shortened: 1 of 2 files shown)


alext:
:+1: I'm going to merge this on Monday because the corresponding container change hasn't built in Dockerhub yet.

PR #655, opened 2016-12-09
[#134345035] Add Cloud Controller process checks and monitors

keymon:
[#134345035 Monitor cloud controller job](https://www.pivotaltracker.com/story/show/134345035)

What?
-----

We want to monitor elementary health metrics of the cloud controller to help us diagnose the issues and warn us in advance if there are problems with platform. This story is about monitoring on the VM directly, there are other stories to monitor API on high level (e.g. externally).

This PR uses the feature from https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/7 to implement datadog monitors for the Cloud Controller API main process and workers and the CC HTTP endpoint.

We add some monitors for the Cloud Controller metrics from the CF datadog
nozzle:

 * `cf.cc.failed_job_count.total`: alert if there were an increase of more than 5 errors in the last 30m.

 * `cf.cc.log_count.error`: In a similar way, alert if there is an increase of more than 5 errors in the last 30m.

  From the docs:
  > Number of log messages of severity “error.” Error is the most severe level. It is used for failures and during error handling. Most errors can be found under this log level, eg. failed unbinding a service, failed to cancel a task, Diego app crashed error, staging completion errors, staging errors, and resource not found. The count resets when the Cloud Controller process is restarted. Emitted every 30 seconds per VM.

 * `cf.cc.job_queue_length.total`: warning when is > 6, alert >10. This can detect high load in the APIs or slowness in any component of the platform when deploying.

Dependencies?
-------------

This depends on https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/7. Once merged, the relevant commit must be updated in this PR.

There is a TEMPORARY commit in this  PR for testing. Remove it before merge.

How to review?
------------

I added a TEMPORARY commit to this branch that will:
 * Install datadog agent on the Cloud Controller VMs, without needing reinstall all the environment with DATADOG_ENABLED.
 * Enable the datadog nozzle
 * Includes a directory with the new monitors to test in terraform.

So to review:

 1. Deploy as usual with the datadog credentials in the S3 bucket. it is NOT required to have `DATADOG_ENABLED`
 2. apply the monitors manually:

```
cd terraform/datadog/tmp_check_cc_monitors/ && terraform apply -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var aws_account=dev  -var env=hector
```

You can simply check if the monitors makes sense, but if you want to trigger then:
 * Stop API cloudcontroler and the workers and check alerts from the process & HTTP monitors.
 * Stop the rds-broker and try to create instances. It will increase the error count.
 * Stop several CC workers and try to push several apps at the same time.

 Note: I was not able to make it increase the failed jobs counter.

 Who?
 ----

 Anyone but @keymon

### DIFF
--- a/terraform/datadog/cloud_controller.tf
+++ b/terraform/datadog/cloud_controller.tf
@@ -64,3 +64,66 @@ resource "datadog_monitor" "cc_api_healthy" {
     "job"        = "api"
   }
 }
+
+resource "datadog_monitor" "cc_failed_job_count_total_increase" {
+  name                = "${format("%s Cloud Controller API failed job count", var.env)}"
+  type                = "query alert"
+  message             = "Amount of failed jobs in Cloud Controller API grew considerably, check the API health."
+  escalation_message  = "Amount of failed jobs in Cloud Controller API still growing considerably, check the API health."
+  require_full_window = false
+
+  query = "${format("change(max(last_1m),last_30m):max:cf.cc.failed_job_count.total{deployment:%s} > 5", var.env)}"
+
+  thresholds {
+    warning  = "3"
+    critical = "5"
+  }
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "api"
+  }
+}
+
+resource "datadog_monitor" "cc_log_count_error_increase" {
+  name                = "${format("%s Cloud Controller API log error count", var.env)}"
+  type                = "query alert"
+  message             = "Amount of logged errors in Cloud Controller API grew considerably, check the API health."
+  escalation_message  = "Amount of logged errors in Cloud Controller API still growing considerably, check the API health."
+  require_full_window = false
+
+  query = "${format("change(max(last_1m),last_30m):sum:cf.cc.log_count.error{deployment:%s} > 5", var.env)}"
+
+  thresholds {
+    warning  = "3"
+    critical = "5"
+  }
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "api"
+  }
+}
+
+resource "datadog_monitor" "cc_job_queue_length" {
+  name                = "${format("%s Cloud Controller API job queue length", var.env)}"
+  type                = "query alert"
+  message             = "Job queue in Cloud Controller API grew considerably, check the API health."
+  escalation_message  = "Job queue in Cloud Controller API still too big, check the API health."
+  require_full_window = false
+
+  query = "${format("max(last_30m):max:cf.cc.job_queue_length.total{deployment:%s} > 10", var.env)}"
+
+  thresholds {
+    warning  = "6"
+    critical = "10"
+  }
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "api"
+  }
+}
--- a/terraform/datadog/cloud_controller.tf
+++ b/terraform/datadog/cloud_controller.tf
@@ -1,0 +1,66 @@
+resource "datadog_monitor" "cc_api_master_process_running" {
+  name                = "${format("%s Cloud Controller API master process running", var.env)}"
+  type                = "service check"
+  message             = "Cloud Controller API master process is not running. Check deployment state."
+  escalation_message  = "Cloud Controller API master process is still not running. Check deployment state."
+  notify_no_data      = false
+  require_full_window = true
+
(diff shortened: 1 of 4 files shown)


keymon:
I added a commit to remove old references to datadog-for-cloudfoundry repo.

terraform/datadog/cloud_controller.tf
```diff
@@ -0,0 +1,129 @@
...
+}
+
+resource "datadog_monitor" "cc_log_count_error_increase" {
+  name                = "${format("%s Cloud Controller API log error count", var.env)}"
+  type                = "query alert"
+  message             = "Amount of logged errors in Cloud Controller API grew considerably, check the API health."
+  escalation_message  = "Amount of logged errors in Cloud Controller API still growing considerably, check the API health."
+  require_full_window = false
+
+  query = "${format("change(max(last_1m),last_30m):max:cf.cc.log_count.error{deployment:%s} > 5", var.env)}"
```
saliceti: The metric is per host. I think you need to add `by {index}` to create an alert per host.
Or you could sum across all hosts instead and create only one alert.

terraform/datadog/cloud_controller.tf
```diff
@@ -0,0 +1,129 @@
+resource "datadog_monitor" "cc_api_master_process_running" {
+  name                = "${format("%s Cloud Controller API master process running", var.env)}"
+  type                = "service check"
+  message             = "Cloud Controller API master process is not running. Check deployment state."
+  escalation_message  = "Cloud Controller API master process is still not running. Check deployment state."
+  notify_no_data      = false
+  require_full_window = true
+
+  query = "${format("'process.up'.over('bosh-deployment:%s','process:cc_api_master').last(4).count_by_status()", var.env)}"
```
saliceti: Could you change to `over(..."bosh-job:api"...)`?
Without it I've had some other hosts reported in the DD API

terraform/datadog/cloud_controller.tf
```diff
@@ -0,0 +1,129 @@
...
+
+resource "datadog_monitor" "cc_api_worker_process_running" {
+  name                = "${format("%s Cloud Controller API worker process running", var.env)}"
+  type                = "service check"
+  message             = "Cloud Controller API worker process is not running. Check deployment state."
+  escalation_message  = "Cloud Controller API worker process is still not running. Check deployment state."
+  notify_no_data      = false
+  require_full_window = true
+
+  query = "${format("'process.up'.over('bosh-deployment:%s','process:cc_api_worker').last(4).count_by_status()", var.env)}"
```
saliceti: Could you change to `over(..."bosh-job:api_worker"...)`?
Without it I've had some other hosts reported in the DD API

terraform/datadog/cloud_controller.tf
```diff
@@ -0,0 +1,129 @@
...
+
+resource "datadog_monitor" "cc_api_healthy" {
+  name                = "${format("%s Cloud Controller API healthy", var.env)}"
+  type                = "service check"
+  message             = "Large portion of Cloud Controller API master unhealthy. Check deployment state."
+  escalation_message  = "Large portion of Cloud Controller API master still unhealthy. Check deployment state."
+  no_data_timeframe   = "7"
+  require_full_window = true
+
+  query = "${format("'http.can_connect'.over('bosh-deployment:%s','instance:cc_endpoint').by('*').last(1).pct_by_status()", var.env)}"
```
saliceti: Could you change to `over(..."bosh-job:api"...)`?
Without it I've had some other hosts reported in the DD API

saliceti: Why use cluster alert instead of check alert here?

terraform/datadog/cloud_controller.tf
keymon: I think there is a bug in DD. It shows more hosts in the  "Monitored Hosts" but the right hosts in "Monitor History"

terraform/datadog/cloud_controller.tf
keymon: Same as before, `cc_api_worker` process runs on the `api` and `api_worker`, we should monitor both. If I do that, I would have to create 2x monitors.

terraform/datadog/cloud_controller.tf
keymon: Again, as before.

saliceti:
This commit https://github.com/alphagov/paas-cf/pull/655/commits/6528942ba10890477d730b926a51335459258523 removes datadog-cc-worker that was added in the first commit. can we remove it altogether?

keymon:
Yes, remove it, sorry

PR #656, opened 2016-12-09
[#133036569] Fix TF destroy for Metrics ELB.

alext:
## What

The destroy failed because both listeners were referencing the same policy.
Destroying the first resource failed because it attempted to delete the
policy as well as the attachment, but the policy was still in use by the
other listener.

Using a policy per listener avoids this conflict.

## How to review

Deploy from this branch as far as cf-terraform. Run the destroy-cloudfoundry pipeline, and verify it succeeds.

## Who can review

Anyone but myself.


benhyland:
Can't really argue with this, lgtm

PR #657, opened 2016-12-09
[#126126973] Tooling to use STS tokens once we enforce MFA

Jonty:
## What

In [another PR](https://github.gds/government-paas/aws-account-wide-terraform/pull/68) we enforce users to have MFA enabled using an IAM policy that blocks all access to AWS if a token has not been registered. Because regular AWS keys are not MFA'd, any request made using them will be blocked by the policy.

The solution to this is to move to [security tokens](http://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html) which are MFA'd, and are issued using the regular AWS keys. These tokens are time-limited with a maximum lifetime of 36 hours, however we set the expiry to 12 hours so they last a working day.

This PR introduces tooling that both checks you have valid keys before running Make tasks, and tooling to issue & cache tokens.

We acknowledge that the interaction with this tooling may not be perfect yet, however we would like to see how people use it before integrating it more deeply or adding more magic ✨.

It is suggested that developers create bash aliases to both create and source the correct credentials for an environment. Such as:

    alias fetch_token="${PAAS_CF_DIR}/scripts/create_sts_token.sh" && \
        source "${HOME}/.aws_sts_tokens/${AWS_ACCOUNT}_${DEPLOY_ENV}.sh"

## How to review

1. Run `make showenv` and observe that you are warned about not having credentials.
2. Run `./scripts/create_sts_token.sh` to create an STS token in `~/.aws_sts_tokens/`.
3. Run `source ~/.aws_sts_tokens/$AWS_ACCOUNT_$DEPLOY_ENV.sh` (Assuming you have `AWS_ACCOUNT` set to `dev` or similar).
4. Run `make showenv` and observe it working using the temporary credentials.
5. Optionally wait 12 hours and try `make showenv` again, then observe a notice telling you your credentials are now invalid.

## Who can review
Preferably not @benhyland or @jonty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,8 +172,8 @@ bootstrap-destroy: ## Destroy bootstrap
 bosh-cli: ## Create interactive connnection to BOSH container
 	concourse/scripts/bosh-cli.sh
 
-.PHONY: pipelines check-aws-credentials
-pipelines: ## Upload pipelines to Concourse
+.PHONY: pipelines
+pipelines: check-aws-credentials ## Upload pipelines to Concourse
 	concourse/scripts/pipelines-bosh-cloudfoundry.sh
 
 .PHONY: showenv
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ check-env-vars:
 check-tf-version:
 	@./terraform/scripts/ensure_terraform_version.sh $(MIN_TERRAFORM_VERSION)
 
+check-aws-credentials:
+	@./scripts/validate_aws_credentials.sh
+
 test: spec lint_yaml lint_terraform lint_shellcheck lint_concourse lint_ruby lint_posix_newlines ## Run linting tests
 
 spec:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -169,12 +172,12 @@ bootstrap-destroy: ## Destroy bootstrap
 bosh-cli: ## Create interactive connnection to BOSH container
 	concourse/scripts/bosh-cli.sh
 
-.PHONY: pipelines
+.PHONY: pipelines check-aws-credentials
 pipelines: ## Upload pipelines to Concourse
```
alext: I guess you meant to add `check-aws-credentials` as a dependency of this task, in which case it needs to be on the line below.

benhyland:
Fixed the typos @alext found

dcarley:
> an STS token that is cached on the local filesystem

Out of interest, is it necessary to cache it on disk? Is there an overhead or limit in requesting new tokens?

Jonty:
@dcarley The overhead is that you need to re-enter a 2FA code every time you request one, the caching is to make using STS less irritating.

PR #658, opened 2016-12-09
Increase elasticsearch disk space in prod

bleach:
## What

Rework of #652

I've increased the size of our elasticsearch persistent disks in prod because they are fairly steady at 80% full (we alert at 75% full). I think we'd prefer to be at something more like 65% full.

I originally planned to change it everywhere, but @keymon convinced me it would be worth saving money by only changing it in prod.

## How to review

Technical review (optional):

1. deploy this change and check that the size has not increased in your dev environment
1. create a temporary branch that is a copy of this one and switch your env to pull from it
1. add a commit to set `ENV_SPECIFIC_CF_MANIFEST=cf-prod.yml` in Makefile:101
1. deploy
1. check that `/var/vcap/store` on elasticsearch_master got bigger
1. Check that all your elasticsearch data didn't disappear

## Who can review

Anyone but @bleach

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1349,6 +1349,7 @@ jobs:
                 ./paas-cf/manifests/cf-manifest/cloud-config/*.yml
                 ./terraform-outputs/*.yml
                 ./cf-secrets/cf-secrets.yml
+                ./paas-cf/manifests/cf-manifest/cloud-config/env-specific/{{cf_env_specific_manifest}}
             run:
               path: sh
               args:
@@ -1357,7 +1358,7 @@ jobs:
                 - |
                   echo "Generating cloud config..."
 
-                  ./paas-cf/manifests/shared/build_manifest.sh $MANIFEST_STUBS > cloud-config/cloud-config.yml
+                  eval ./paas-cf/manifests/shared/build_manifest.sh $MANIFEST_STUBS > cloud-config/cloud-config.yml
           on_success:
             put: cloud-config
             params:
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -1349,7 +1349,7 @@ jobs:
                 ./paas-cf/manifests/cf-manifest/cloud-config/*.yml
                 ./terraform-outputs/*.yml
                 ./cf-secrets/cf-secrets.yml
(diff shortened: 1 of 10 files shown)


bleach:
@rboulton commented on the original PR https://github.com/alphagov/paas-cf/pull/652#issuecomment-266360597

bleach:
> @bleach One thing to consider when calculating disk space requirements for elasticsearch is - what happens if one node goes down, and all the data on that node gets rebalanced to other nodes? If you have a 3 node cluster, then that would increase the space used on each remaining node by 50%.

> I've seen clusters make themselves run out of disk space when this happens; for a 3 node cluster, something with sufficient margin below 66.6% should be considered the maximum safe disk usage, for this reason.

mtekel:
Not sure default should be 300G. We bumped that recently from 200G because of prod disk space. Also, 377 vs 300 seem like a modest increase. If we want to avoid doing bumps frequently, we should go to e.g. 500+.

bleach:
> @bleach One thing to consider when calculating disk space requirements for elasticsearch is - what happens if one node goes down, and all the data on that node gets rebalanced to other nodes? If you have a 3 node cluster, then that would increase the space used on each remaining node by 50%.
>
> I've seen clusters make themselves run out of disk space when this happens; for a 3 node cluster, something with sufficient margin below 66.6% should be considered the maximum safe disk usage, for this reason.

@rboulton I'm not actually clear on what happens right now if a node goes down

- we set `cluster.routing.allocation.enable` to `all` https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-bosh-cloudfoundry.yml#L1957
- `number_of replicas` is 1 for all our indices

We replace nodes every few weeks by destroying the EC2 instances, creating a new instance & mounting the EBS data volume. Most recently we did this last Thursday, when the cluster data volumes were all at ~80% usage & we didn't appear to run out of space.

Because we use EBS to store data, in most cases, if a node does go away it's mostly not the case that the data has gone away, so I wonder if we really want any automatic reallocation, apart from when we know we've changed cluster size or if we lose an EBS volume.

bleach:
Noticed that this entire PR seems slightly irrelevant, as the disk space is dropping faster than I first thought.

screen shot 2016-12-12 at 10 09 08

keymon:
The changes make sense, and I think that the additional changes proposed (only one file for env specific settings and separate prod ES disk size) are good to have.

I would merge this, but I would suggest to:
 * Option 1: increase the disk size even more. For instance 450G. 
 * Option 2: do not increase the disk size at all in prod, but merge the other changes for the future.

Let me know what you think and if you want to change

bleach:
I'm closing this PR because the problem is fixing itself. I'll leave the branch around in case we want it.

PR #659, opened 2016-12-12
[#122026001] Remove Privacy Policy info from tests

combor:
## What

This is follow up for: https://www.pivotaltracker.com/story/show/122026001/comments/157906483

As we agreed that we are not going to mention our current retention
period in Privacy Policy it would be confusing to leave such info in
tests. Instead tests mention that we expect to have retention period set
to N days.

## How to review

Check if changes are sane. This should be no-op deployment so we can leave it for CI Concourse to test. 

## Who can review

not @combor


mtekel:
LG, merging

PR #661, opened 2016-12-12
[#135943277] Whitelist IPs for MOD paas trial

benhyland:
## What

Confirmed ips belong to MOD with `whois`.

For story https://www.pivotaltracker.com/story/show/135943277

We should merge this straight away.
We shouldn't continue with the story to create org and users until it is merged, since the users would not be able to access the system.

Review by anyone not @mtekel / @benhyland


alext:
Confirmed that the IPs do indeed belong to the MoD, and match the ones requested in the story.

PR #662, opened 2016-12-12
[#131375807] Add Route Emitter process and metrics checks and monitors

keymon:
[#131375807 Check health of route emitter component](https://www.pivotaltracker.com/story/show/131375807)

What?
----

We want to monitor the route_emitter as we do with other processes.

There are also [several metrics from the CF Datadog nozzle](https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/8). After testing we decided to monitor:

 * The process is running
 * The process opens the debug port.
 * The consul lock is held:
   * Local check in each route_emitter, querying http://localhost:8500/v1/kv/v1/locks/route_emitter_lock
   * The metric sent by CF Nozzle: `cf.route_emitter.LockHeld.v1-locks-route_emitter_lock`
 * ~~The gorouter and the route_emitter total routes match~~
   * ~~We need to take into account LRPs that are not running. They are in route-emitter but not in gorouter.~~
   * Note: I actually decided to drop this monitor as explained in https://github.com/alphagov/paas-cf/pull/662#issuecomment-266715172

 * ~~The route-emitter is still getting new routes from the Registered and Unregistered events in BBS, and the bulk route sync. To do so, we check if the related metrics are being updated.~~

Dependencies
-----------

This PR depends on https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/8 which must be reviewed and merged first. After that, the commit referring to `datadog-for-cloudfoundry` bosh-release must be updated with the new final release details.

There is a WIP commit in this PR for testing purposes, which should be deleted before merge.

How to review?
--------------

I added a TEMPORARY commit to this branch that will:

 * Install datadog agent on the route-emitter VMs, without needing reinstall all the environment with DATADOG_ENABLED.
 * Enable the datadog CF nozzle
 * Includes a directory with the new monitors to test in terraform.

So to review:

 * Deploy as usual with the datadog credentials in the S3 bucket. it is NOT required to have DATADOG_ENABLED
 * apply the monitors manually:
   `cd terraform/datadog/tmp_check_cc_monitors/ && terraform apply -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var aws_account=dev  -var env=hector`

To test the monitors:

 * You can simply check if the monitors makes sense.
 * If you want to trigger then:
   * stop & start route_emitter, consul, etc.
   * go to the nats boxes, and block the route_emitter with iptables.
   * go to the route_emitter boxes and block access to BBS. That would stop the routes from being updated. (e.g. `iptables -A OUTPUT -p tcp --destination-port 8889 -j DROP`, clear it with `iptables -F OUTPUT`)

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/datadog/route-emitter.tf
+++ b/terraform/datadog/route-emitter.tf
@@ -62,3 +62,31 @@ resource "datadog_monitor" "route_emitter_consul_lock" {
     "job"        = "route_emitter"
   }
 }
+
+resource "datadog_monitor" "route_emitter_lock_held_once" {
+  name                = "${format("%s route-emitter lock held exactly once (%s)", var.env, element(list("upper", "lower"), count.index))}"
+  type                = "query alert"
+  message             = "There is not exactly one route-emitter holding the lock."
+  escalation_message  = "There is still not exactly one route-emitter holding the lock."
+  notify_no_data      = false
+  require_full_window = true
+
+  count = 2
+
+  query = "${
+    format(
+      "min(last_5m):sum:cf.route_emitter.LockHeld.v1_locks_route_emitter_lock{deployment:%s}.fill(last, 60) %s",
+      var.env,
+      element(list("> 1", "< 1"), count.index)
+    )}"
+
+  thresholds {
+    critical = "${element(list("1", "1"), count.index)}"
+  }
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "route_emitter"
+  }
+}
--- a/terraform/datadog/route-emitter.tf
+++ b/terraform/datadog/route-emitter.tf
@@ -1,0 +1,64 @@
+resource "datadog_monitor" "route_emitter_process_running" {
+  name                = "${format("%s route-emitter process running", var.env)}"
+  type                = "service check"
+  message             = "route-emitter process not running. Check router state."
+  escalation_message  = "route-emitter rep process still not running. Check router state."
+  notify_no_data      = false
+  require_full_window = true
+
+  query = "${format("'process.up'.over('bosh-deployment:%s','process:route-emitter').last(4).count_by_status()", var.env)}"
+
+  thresholds {
+    ok       = 1
+    warning  = 2
+    critical = 3
+  }
+
+  tags {
+    "deployment" = "${var.env}"
+    "service"    = "${var.env}_monitors"
+    "job"        = "route_emitter"
+  }
+}
+
+resource "datadog_monitor" "route_emitter_healthy" {
+  name                = "${format("%s route-emitter healthy", var.env)}"
+  type                = "service check"
+  message             = "Large portion of route-emitter unhealthy. Check deployment state."
+  escalation_message  = "Large portion of route-emitter still unhealthy. Check deployment state."
+  no_data_timeframe   = "7"
+  require_full_window = true
+
+  query = "${format("'http.can_connect'.over('bosh-deployment:%s','instance:route_emitter_debug_endpoint').by('*').last(1).pct_by_status()", var.env)}"
+
+  thresholds {
+    critical = 50
+  }
+
+  tags {
+    "deployment" = "${var.env}"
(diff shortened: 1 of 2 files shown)


henrytk:
> We need to take into account LRPs that are not running. They are in route-emitter but not in gorouter.

Just out of curiousity, why are there routes in route-emitter but not in gorouter? I don't understand what is going on there.

keymon:
There is very little info about this, but from what I read in the code, the route emitter won't emit routes that don't have a working backend, so they never get into the gorouter.

henrytk:
@keymon does that mean a user has probably pushed an unhealthy application and just left it (perhaps as a trial or dev application)?

keymon:
> @keymon does that mean a user has probably pushed an unhealthy application and just left it (perhaps as a trial or dev application)?

yes, that happens often.

Also, as from our conversation in the real world: The monitor is wrong. I was assuming that "DesireLRPs - RunningLRPs" would be "Number of Routes without Backend". But it is not.  If there is a Desired but not Running LRP which has != 1 routes associated (no routes, or 2 or more), the "DesireLRPs - RunningLRPs" will be 1 more, but "Number of Routes without Backend" != 1

So we need to refactor that monitor.

henrytk:
@keymon Is it safe to continue to review and possibly merge https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/8/files while this is being refactored? Or, should I wait?

keymon:
@henrytk: Yes please

henrytk:
Now that https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/8 has been merged a new 'final' release has been built for the datadog integrations:

```
releases:
  - name: datadog-for-cloudfoundry
    version: 0.1.3
    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/datadog-for-cloudfoundry-0.1.3.tgz
    sha1: e20e2518d1d677cd5e731d39bd8e286397f95fa8
```

henrytk:
The message in commit https://github.com/alphagov/paas-cf/pull/662/commits/03ff2ea0e8467e0f2c3fd27a96a3e97fe1509f84 is missing the `[1]` reference.

terraform/datadog/route-emitter.tf
```diff
@@ -90,3 +90,31 @@ resource "datadog_monitor" "route_emitter_lock_held_once" {
...
+resource "datadog_monitor" "route_emitter_fresh_registered_routes" {
+  name                = "${format("%s route-emitter has refreshed registered routes from BBS", var.env)}"
+  type                = "query alert"
+  message             = "route-emitter is not refreshing the routes with BBS"
+  escalation_message  = "route-emitter is still not refreshing the routes with BBS"
+  notify_no_data      = true
+  no_data_timeframe   = "7"
+  require_full_window = true
+
+  query = "${
```
henrytk: This is an interesting query because we are essentially implementing a no_data check and don't care about the metric. I wonder if Datadog provide a more built-in support for declaring certain metrics as critical. Just a thought, no need to change anything.

keymon: true, you are right. I played for a while and  Iwas not able to to implement anything. For instance, I tried a function that replaces nulls with 0, and then alert on count, but it does not work :(

henrytk:
Spoke with @keymon offline. The `route-emitter is syncing routes with BBS` and `route-emitter has refreshed registered routes from BBS` alerts are going to no_data immediately without alerting. I think there is enough covered by the other monitors to justify removing these two monitors. @keymon will raise a ticket with datadog to ask why the `count_not_null` function isn't behaving as expected.

If @keymon drops these two monitors and removes the WIP commit I will merge.

keymon:
I removed that monitors and push the changes in the branch feature/131375807_monitor_route_emitter_more_monitors.

I will open a ticket with datadog

PR #666, opened 2016-12-13
Remove aws cred check from pipelines make target

benhyland:
This check is useful in dev, but breaks permanent envs since
self-update-pipelines runs the make target directly.

We can figure out how to gracefully run this check in dev only
some other time.


mtekel:
yes, this works, I can run self-update-pipelines in dev

PR #667, opened 2016-12-13
Remove usage of the check aws credentials task

paroxp:
## What

Remove the usage of the `check-aws-credentials` task, as it breaks more things than it should.

## How to review

Run your everyday routine and see if you're not bothered by some error message contained in `scripts/validate_aws_credentials.sh`.

No more task usage occurrences should appear.

## Who can review

Not @paroxp


benhyland:
lgtm

PR #669, opened 2016-12-14
[#130728371] acceptance-tests: Improve XFF header tests

dcarley:
## Preface

I was meant to rewrite an existing story for this before I went on leave, but I didn't and I now can't find it. While looking for that story I noticed that we have a new story for our incorrect `X-Forwarded-For` header behaviour (Pivotal ID in subject). These changes should make it easier to do that work in the future. I'm not going to move the card in Pivotal and reviewing other stories should take precedence to reviewing this.

## What

The existing tests for individual indexes of the XFF slice made it quite
hard to understand the complete behaviour. For example, the real egress IP
is the second-to-last element rather than the last element.

They were also susceptible to false-positives because we were checking that
the second element didn't belong to a 10/8 private network, rather than
asserting the actual value of the second-to-last element. So the tests would
continue to pass if an intermediate proxy inserted more elements before our
real egress IP.

Make the tests clearer and more strict by obtaining our real egress IP and
comparing the entire contents of the header. Because we're checking the
entire contents, rather than individual elements, the two existing tests
have been combined into one. This also makes it slightly faster because we
only have to push the app once.

Unfortunately we need to use a third-party service to get the egress IP. I
chose `canhazip.com` because it's reputable (been around for a while, run by
someone well known) and the response format is simple. There's some more
information about it here:

- https://major.io/icanhazip-com-faq/

I have left a FIXME comment next to the `127.0.0.1` element. This IP is
inserted by the hop between HAproxy and gorouter. It's undesirable, because
tenants should expect to get the IP from the last element, and we hope to
remove it in the future. In the meantime it is intentionally documented by
the test.

## How to review

1. Ensure the code changes are clear and correct.
1. You can run the tests yourself, but I don't think it's essential:
```
git fetch && git checkout bugfix/130728371-improve_xff_test
make dev run_job JOB=custom-acceptance-tests
```

## Who can review

Not @dcarley

### DIFF
--- a/platform-tests/src/acceptance/http_forwarded_for_test.go
+++ b/platform-tests/src/acceptance/http_forwarded_for_test.go
@@ -10,16 +10,36 @@ import (
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
 
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net"
+	"net/http"
 	"strings"
 )
 
 var _ = Describe("X-Forwarded headers", func() {
-	var headers struct {
-		X_Forwarded_For []string `json:"X-Forwarded-For"`
-	}
+	const (
+		egressURL   = "https://canhazip.com/"
+		fakeProxyIP = "1.2.3.4"
+	)
 
-	var _ = BeforeEach(func() {
+	var (
+		egressIP string
+		headers  struct {
+			X_Forwarded_For []string `json:"X-Forwarded-For"`
+		}
+	)
+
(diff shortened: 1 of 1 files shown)


mtekel:
This works fine. Removes one test and replaces other. Previously both tests were 61 seconds each, now this single test run in 50s in my dev. Code cleanup is also welcome. We'll see about external service reliability and will solve it if it becomes an issue. Merging.

dcarley:
Thanks @mtekel

PR #672, opened 2016-12-16
WIP: [#134345035] Do not notify "No Data" for CC log error monitor

keymon:
[#134345035 Monitor cloud controller job](https://www.pivotaltracker.com/story/show/134345035)

What?
-----

In the monitor added in #655 we alert when there is no data after
2 minutes (the default).

But this metric is not generated if no log entries are produced, so it
might happen that we get No Data alerts in periods of no activity,
specially in CI or staging.

How to review?
------------

We need to first discuss: https://github.com/alphagov/paas-cf/pull/645#issuecomment-265200333

Who?
---

Anyone but @keymon


paroxp:
Could you refer to the comments here: https://github.com/alphagov/paas-cf/pull/645#issuecomment-265200333 ?

keymon:
Ok, I was not aware of that. Then we should wait :-m

PR #673, opened 2016-12-16
[#131876783] Split destroy-concourse job in destroy pipeline.

alext:
## What

As part of the process for migrating to paas-bootstrap we need to
destroy the concourse instance, but not run the concourse terraform
destroy. Splitting this job into 2 separate jobs will allow us to
accomplish this by pausing the destroy-terraform job.

## Options for review

Code review.
Fire up a bootstrap-concourse and check the pipeline looks correct.

## Who can review

Anyone but @dcarley or myself.


benhyland:
Pipeline and code makes sense as described.

PR #675, opened 2016-12-16
[#SUPPORT] Add the Host header to the HAProxy logs

keymon:
Currently we capture the access log of the HAProxy in front of our gorouters.

Today I was troubleshooting some spike in the request latency, and I wanted to check the access log from the HAProxy in logsearch. But I found out that I could not correlate the requests with the applications because the log messages do not include the Host header.

That is valuable information, but in order to identify for which application each request is, we need to log the Host header.

We will use the paas-haproxy bosh release PR in https://github.com/alphagov/paas-haproxy-release/pull/7

We also migrate this bosh release to the PR based builds (see https://www.pivotaltracker.com/story/show/132745311)

Dependencies
-----------

Depends on https://github.com/alphagov/paas-haproxy-release/pull/7 which should be merged first and the relevant commit updated in this PR.

But both can be tested together.

How to review?
--------------

 1. Deploy this.
 2. Access any app `curl -k https://healthcheck.hector.dev.cloudpipelineapps.digital`
 3. go to your logsearch and check that the haproxy logs have a new field: `haproxy.captured_request_headers == healthcheck.hector.dev.cloudpipelineappseapps.digital`

Who?
---

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -93,12 +93,6 @@ resources:
       tag_filter: {{paas_cf_tag_filter}}
       commit_verification_key_ids: {{gpg_ids}}
 
-  - name: paas-haproxy-release
-    type: git
-    source:
-      uri: https://github.com/alphagov/paas-haproxy-release.git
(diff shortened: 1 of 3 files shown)


saliceti:
Tested successfully, commit updated.

PR #676, opened 2016-12-16
[#136189211] Revert "Merge pull request #674 Increase number of CC workers to 3 per VM"

keymon:
Revert PR #674 as it did not reduce the number of jobs in the queue.

We will try to change the monitor instead.

This reverts commit b14f8f523dce125e06a13f08e32f77867736aee0, reversing
changes made to b40bde8017fc5d17839feb3a1dcd61146babe0d1.

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -190,9 +190,7 @@ properties:
       droplet_upload:
         timeout_in_seconds: ~
       generic:
-        number_of_workers: 3
-      local:
-        number_of_workers: 3
+        number_of_workers: ~
 
     app_events:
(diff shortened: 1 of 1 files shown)


benhyland:
We discussed this last week, the suggestion was to not revert, since the default is not particularly good and we might prefer to be explicit about config we've had to look at once already. I thought this was an actual decision but just in case I'll leave it to @dcarley or @timmow to close or merge.

bleach:
A benefit of reverting would be that we'd have a record in the commit history that increasing the number of workers didn't have the effect we desire.

A benefit of keeping is that we'd have a little more API worker capacity (although we're not sure it solves this particular problem, there isn't any harm to it either).

Discussed with @timmow and @dcarley and we've decided to leave things as they are.

PR #677, opened 2016-12-19
[#SUPPORT] Only enable routes drop monitor in dev and prod

keymon:
What?
-----

The `router_routes_drop` monitor get easily triggered when running the tests in environments with little routes, like ci or staging.

We do not need to monitor this on those environments, but it is still useful in prod. We can also keep it in dev for testing/development.

> Note: as we add `count`, it will actually recreate the monitor in prod, because the ID changes (add a .0 suffix). That is fine and expected.

How to test?
------------

You can apply these monitors in dev like this:

```
PAAS_CF_DIR=$(pwd)
mkdir /tmp/testing_tf
cd /tmp/testing_tf
ln -s ${PAAS_CF_DIR}/terraform/datadog/{datadog.tf,variables.tf,router.tf} .

terraform apply -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var aws_account=dev  -var env=hector -var-file  ${PAAS_CF_DIR}/terraform/dev.tfvars
```

You will see that it creates the monitor.

Now run with a different vars file:

```
terraform apply -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var aws_account=dev  -var env=hector -var-file  ${PAAS_CF_DIR}/terraform/ci.tfvars
```

It will destroy the old monitor.

Remember do `terraform destroy`

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/datadog/router.tf
+++ b/terraform/datadog/router.tf
@@ -66,6 +66,12 @@ resource "datadog_monitor" "route_update_latency" {
   }
 }
 
+variable "datadog_monitor_total_routes_drop_enabled" {
+  description = "Selector to enable/disable the named resource"
+
+  default = 1
+}
+
 resource "datadog_monitor" "total_routes_drop" {
   name                = "${format("%s total routes difference", var.env)}"
   type                = "query alert"
@@ -73,6 +79,9 @@ resource "datadog_monitor" "total_routes_drop" {
   escalation_message  = "Total routes still dropping quickly. Check the deployment."
   no_data_timeframe   = "7"
   require_full_window = true
+
+  # Conditionally enable this resource based on var.aws_account.
+  count = "${var.datadog_monitor_total_routes_drop_enabled}"
 
   query = "${format("pct_change(avg(last_1m),last_30m):avg:cf.gorouter.total_routes{deployment:%s,job:router} < -33", var.env)}"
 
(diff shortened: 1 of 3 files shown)


terraform/datadog/router.tf
```diff
@@ -66,6 +66,22 @@ resource "datadog_monitor" "route_update_latency" {
...
+# The resource will refer this variable using a `lookup` function in the count
+# attribute.
+#
+# To enable the resource for a given environment, add an entry here.
+variable "__datadog_monitor_total__routes_drop__enabled" {
+  description = "Selector to enable/disable the named resource"
+
+  default = {
+    "dev"  = 1
+    "prod" = 1
```
dcarley: There was some discussion in Slack yesterday about moving this logic to the `tfvars` files where we already change some behaviour based on the AWS account. What was the conclusion?

Also, should the default be enabled, and we explicitly disable it for CI and staging? That would match the other flags we use.

keymon: It is really strange, I was pretty sure I changed this. I don't know what happened here. I will change it again.

I will simplify to only use a simple variable with tfvars, and explicitly disable it for CI and staging with default enabled.  

Not sure what you mean with "That would match the other flags we use."

keymon: Changed. I guess I did not commit the changes I did, sorry :)

saliceti:
Can we rename the variable `__datadog_monitor_total__routes_drop__enabled` to `datadog_monitor_total_routes_drop_enabled`?
This would be more in line with the rest of the style.

PR #679, opened 2016-12-20
[#SUPPORT] Plot summarised router response times

keymon:
What?
-----

In the grafana dashboard we plot the 90 percentile of the both routers, so we can spot deviations in the response time.

But that graph has several problems that make it not useful:
 * The metric is calculated and generated every 10 seconds, but we plot a total of 12 hours. The graph is really spiky and noisy.
 * Sometimes there are very little requests and some really slow but expected requests happen, that appear in the graph. For instance, the /events in a dashing dashboard which takes 900s. As we plot linearly, that causes the graph to get out of scale.
 * We don't plot the maximum request time, only 90 percentile.

In this commit we propose the following changes:

 * summarise all routers using max. The intention of the graph is highlight an issue, if one router is misbehaving, we can go and spot
   it later.

 * summarise all the values as the "max" of 1 minute window. That will make the graph be far less noisy, and we won't miss any issue as we use
   max.

 * Use Log scale to plot the response time. This way we can plot together high values (max) and average values.
   Our area of interest, 10ms - 2000ms gets the maximum resolution.  Also set minimum to 10ms.

 * Plot the Max, 90 percentile, mean 90 and median, so we get an idea of the distribution of the response times.

 * We also change some colors and labels, add info of max, min and avg values, etc.

Screenshot Before the changes
---------------------------
Before the changes, useless due having little metrics at night and a request was 15m:

screenshot from 2016-12-20 11 56 01

So in this other I changed the window to latest 6h. Note that for 12h is even more noisy.
screenshot from 2016-12-20 11 56 19

Screenshot after the changes
-------

This one is the same than the first image above. You can tell the outsiders, the median, average, etc.

screenshot from 2016-12-20 11 32 21

> Note, I removed the side labels in the code,

How to test?
-----------

Go to any grafana and import the file `manifests/cf-manifest/grafana/user-impact.json` as dashboard (click on the grafana logo > Dashboards > Import).

Who?
----

Anyone but @keymon

### DIFF
--- a/manifests/cf-manifest/grafana/user-impact.json
+++ b/manifests/cf-manifest/grafana/user-impact.json
@@ -196,7 +196,9 @@
       "height": "250px",
       "panels": [
         {
-          "aliasColors": {},
+          "aliasColors": {
+            "count": "#CCA300"
+          },
           "bars": false,
           "datasource": "graphite",
           "editable": true,
@@ -206,22 +208,26 @@
             "threshold1": null,
             "threshold1Color": "rgba(216, 200, 27, 0.27)",
             "threshold2": null,
-            "threshold2Color": "rgba(234, 112, 112, 0.22)"
+            "threshold2Color": "rgba(234, 112, 112, 0.22)",
+            "thresholdLine": false
           },
           "height": "250",
           "id": 4,
           "isNew": true,
           "legend": {
             "avg": false,
-            "current": false,
-            "max": false,
-            "min": false,
+            "current": true,
+            "max": true,
+            "min": true,
             "show": true,
             "total": false,
-            "values": false
+            "values": true,
+            "alignAsTable": false,
+            "rightSide": false,
+            "sideWidth": null
           },
           "lines": true,
-          "linewidth": 2,
+          "linewidth": 1,
           "links": [],
           "nullPointMode": "null",
           "percentage": false,
@@ -235,14 +241,27 @@
           "targets": [
             {
               "refId": "A",
-              "target": "aliasByNode(stats.timers.cfstats.router.*.http.responsetimes.http.upper_90, 4)"
+              "target": "alias(summarize(maxSeries(stats.timers.cfstats.router.*.http.responsetimes.http.upper), '1m', 'max', false), 'upper')"
+            },
+            {
+              "refId": "C",
+              "target": "alias(summarize(maxSeries(stats.timers.cfstats.router.*.http.responsetimes.http.upper_90), '1m', 'max', false), '90 percentile')",
+              "textEditor": false
+            },
+            {
+              "refId": "D",
+              "target": "alias(summarize(maxSeries(stats.timers.cfstats.router.*.http.responsetimes.http.mean_90), '1m', 'max', false), '90 avg')"
+            },
+            {
+              "refId": "E",
+              "target": "alias(summarize(maxSeries(stats.timers.cfstats.router.*.http.responsetimes.http.median), '1m', 'max', false), 'median')"
             }
           ],
           "timeFrom": null,
           "timeShift": null,
-          "title": "Router 90th percentile response times",
+          "title": "Router response times",
           "tooltip": {
-            "msResolution": false,
+            "msResolution": true,
             "shared": true,
             "value_type": "cumulative"
           },
@@ -254,9 +273,9 @@
             {
               "format": "ms",
               "label": null,
-              "logBase": 1,
+              "logBase": 10,
               "max": null,
-              "min": 0,
+              "min": 10,
               "show": true
             },
             {
@@ -267,7 +286,8 @@
               "min": 0,
               "show": true
             }
-          ]
+          ],
+          "decimals": 0
         },
         {
           "aliasColors": {},


dcarley:
@keymon Is the screenshot from after the changes? Could you upload a screenshot of before the changes? The comparison would make it easier to review this PR.

keymon:
After the changes. I will upload before.

keymon:
@saliceti was doubting about the need to include the mean and the median in this graph. I will elaborate why I think each value is really useful in this case.

 * The mean and the median, together, would allow us to know the shape and skew of the graph. Presuming a normal distribution, we can say:
     * If mean and median are similar: the function is close to a normal distribution with no skew.
     * If mean is lower than median: The function is skew to the right, that is, there are a lot of request that are slower than the mean, a several requests that are really fast.
     * If mean is higher than median: the function is skew to the left, so most of the requests are faster than the mean, but there are several requests that are quite slow.

 * The 90th percentile allow us  to know how much of the population is skewed to the right. This also gets an idea of the most common worse scenario. 
 * The Maximum allow us to get an idea of the length of the leg of graph, how bad the outsiders are.

Here goes some graphs that explain it a little bit better, from https://www.quora.com/How-do-outliers-affect-normal-distribution-in-statistics:

main-qimg-80ab1b71dd969696e8a5ea598768db89-c

Also be aware that In our case, as sometimes we got so little requests, the graph will display very similar values for the mean, median and 90th percentile. It is worth it to note that for normal web requests, the distribution is usually no skew and skew to the right.

In conclusion: IMO we should display all these metrics, I find then useful as explained, and I do think we all will eventually gain the intuition to find then it useful. But for that, we must display them! :smile: .

keymon:
Other note: I compute the 90th percentile average, which means the average of the 90% of the lowest values. The reason is that currently we got too little traffic and the outsiders, which are really big, have too much weight in the mean.

saliceti:
Thanks for the explanation. It makes sense but for me [less is more](https://en.wiktionary.org/wiki/less_is_more). I prefer showing fewer metrics that are super useful than many metrics that might be useful one day.
It's my opinion though. What is in this PR is much better than what we have today anyway.

keymon:
> I prefer showing fewer metrics that are super useful than many metrics that might be useful one day.

Just to keep it clear, for me all the 3: max, upper_90, mean_90 and median are all useful right now, because the reasons stated before. I used those metrics to troubleshoot issues and understand our traffic during support.

paroxp:
LGTM - However, I'd like to gather more people from the team, to agree on these. After all:

> The needs of the many outweigh the needs of the few.

henrytk:
This looks good so I am going to merge it. I never found that graph very useful anyway. There are some very useful explanations in here from @keymon, which I would like to somehow preserve for future reference (I can't think how so I am just bookmarking it in my browser :)

mtekel:
So I had to look at grafana dashboard definition to understand what the metrics mean (hard to tell by name). These are actually all _maximums_ over 1 minute interval of:
* maximum reported by router
* maximum 90 percentile
* mean 90 percentile (but mean over what period?)
* median

I understand why we might want to show maximum of maximum and 90th percentile. But why show _max_ or mean/median? That makes it appear response times are worse than they are in practice...

keymon:
> but mean over what period?

All those metrics are computed by statsd, which uses a default window of 10s, unless configured otherwise.

> why we might want to show maximum of maximum and 90th percentile

I highly recommend read the comments above, that explain most you are asking about.

The 90, 95, 99p are very common metrics to remove outsiders.

Highest values (bigger than 90 percentile) often add noise. But in our case, given that we got very little samples, as we have very low traffic, it is good to know the maximum and plot it.

If we had 200reqs/10s or more, it would be OK to ignore that.

> That makes it appear response times are worse than they are in practice...

mtekel:
The metrics are computed by gorouter internally and emitted to metron. Then input via graphite nozzle. Where is there statsd? Also, before gorouter emits the average (mean) it has to compute it, I am not sure over what period.

Also, I said I understand why we plot maximums of maximums, but why _max_ of median and average? You only explain why you want to plot these two, but not why plot MaxSeries of the 2...

keymon:
> The metrics are computed by gorouter internally and emitted to metron.

I think Graphite Nozzle emit the time as a timer to statsd. Saddly  the project does a overuse of the language and calls "graphite" to "statsd" but from what I have seen, statsd is doing the averages. 

See here: https://github.com/pivotal-cf/graphite-nozzle#httpstartstop

> HTTP requests passing through the Cloud Foundry routers get recorded as HTTPStartStop Events. graphite-nozzle takes these events and extracts useful information, such as the response time and status code. These metrics are then sent through to StatsD. The following table gives an overview of the HTTP metrics graphite-nozzle handles:

But please, pass me the reference if you think that is not the case. 

Also, that changes very little about the graph that is being proposed in this PR.

> You only explain why you want to plot these two, but not why plot MaxSeries of the 2...

Actually true, we have 2 different series, so it gets complicated. In this case:

 * the maximum values of 2 series is the max of both series.
 * It is impossible to compute the 90p of 2 series, because the value is already aggregated per series. But you can approximate by using the maximum. That makes sense, because you want to know the worse scenario.

There is little we can do about this, because each router reports the metrics independently. Maybe we can send the metrics with the same key and remove the router ID. But that is out of the scope of this small PR, that only intents improve the visualisation based on the existing data.

mtekel:
So it turns out 'graphite nozzle' is actually 'statsD nozzle'. Good to know. What I meant was to use `averageSeries` instead of `maxSeries` for mean_90 and median. Alternatively, you can use `movingMedian` function to calculate median of medians. But I see no point in showing 'maximum' of all medians or averages. upper_90 and upper are to spot spikes, so maximum (maxSeries) only make sense there...

keymon:
`averageSeries` for those metrics is tricky. Imagine that one router is behaving really badly and the other is OK. The bad one has a 90p of 50ms, but the other is 1s. In that case averageSeries would be ~500ms, which hides the issue.

Moving median gets a window of values in a given window. That is the same than `summarise`. But same problem applies, you need the maximum, because median or mean would hide high values.

The point of showing the maximum is because the series are already aggregated (mean, 90p, etc), so there is nothing statistically meaningful that you can do but get the worse case. And the worse case is the slowest case, that is the maximum time.

mtekel:
upper and upper_90, which are maxSeries should help you spot broken routers...
median and mean_90 should help you spot issues on the platform, e.g. when everything gets slower because of new update or some issue with CF.

Also, when something is called mean or median, I expect it to be it, not the max of all medians or means. This can make debug really confusing...

keymon:
But we **cannot** provide a unified metric because the metrics are reported and aggregated per router. This PR is about improving a dashboard that provides little information and was not useful to something that provides more.

But as I said, this is my opinion. I explained and justified all across this thread in this PR, please read this comments. This does not mean that I am against reverting this change and going back to the original graph if only me finds this useful. 

Please feel free to propose it.

mtekel:
I think this shouldn't have been merged until we had an agreement. Perhaps this should have been a story with proper kickoff. I find it really confusing as it is now, I had to look into grafana json to find our what the new labels really mean. I think the maximums for upper and upper_90 are useful to spot outliers, however I think maximums for median and mean_90 are misleading.

henrytk:
As the reviewer I considered it mergeable. I agree that it can be really difficult to extract meaning from this kind of data. Perhaps our problem is not with this graph, but more generally a lack of transparency of the underlying data, and a lack of education on our part on how metrics are collected, aggregated, and stored.

mtekel:
my main confusion on the 1st sight came from the screenshot. By definition, all these metrics should be averages over some interval (and there was even discussion about it). But then you have `90 percentile` and `90 avg` metrics, indicating that 90 percentile, or all non-avg are not averages, which was weird to me. I wanted to see what's the difference between the avg and non-avg 90 percentile metrics, just to find out these are now all maximums. This is good for some metrics (extremes), but not for other. And it should be at least labelled like what it is.

keymon:
If this is the issue, lets do use the naming from statsd:

 * upper 90: Upper of the 90% lowest values.
 * avg 90: average of the 90% lowest values.

Then, because it is complicated to do any computation across series that are already aggregated, lets put in the title: "(Maximum values across all routers)"

Would that simplify it?

PR #680, opened 2016-12-20
[#SUPPORT] Plot summarised app requests rate

keymon:
What?
----

In the grafana dashboard we plot the tenant's app and CC requests rate and 4xx and 5xx rate, so we can understand the normal traffic rate and spot rise in errors.

The metric is generated with a resolution of 10s, but we plot a total of 12 hours. Because that the graph is really spiky and noisy.

A solution would summarise the metric in a 1 minute window, creating a more smooth and easy to read metric.

In the case of app requests, to do so, we use the count of requests instead of the rate. We sum the values from all the gorouters, and then summarise in 1m window using sum. Finally we scale to seconds, as the metric is reported every 10s.

Screenshoot before
-----------------
screenshot from 2016-12-20 13 47 55

and
screenshot from 2016-12-20 14 30 17

Screenshoot after
----------------

screenshot from 2016-12-20 13 48 22

and 

screenshot from 2016-12-20 14 38 10

How to test
-----------

Go to any grafana and import the file `manifests/cf-manifest/grafana/user-impact.json` as dashboard (click on the grafana logo > Dashboards > Import).

Who?
---

Anyone but @keymon

### DIFF
--- a/manifests/cf-manifest/grafana/user-impact.json
+++ b/manifests/cf-manifest/grafana/user-impact.json
@@ -15,7 +15,9 @@
       "height": "150px",
       "panels": [
         {
-          "aliasColors": {},
+          "aliasColors": {
+            "5xx": "#BF1B00"
+          },
           "bars": false,
           "datasource": "graphite",
           "editable": true,
@@ -33,13 +35,13 @@
           "isNew": true,
           "legend": {
             "alignAsTable": false,
-            "avg": false,
-            "current": false,
-            "max": false,
+            "avg": true,
+            "current": true,
+            "max": true,
             "min": false,
             "show": true,
             "total": false,
-            "values": false
+            "values": true
           },
           "lines": true,
           "linewidth": 2,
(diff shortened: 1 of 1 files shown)


saliceti:
It does look better for long time range like 12 hours. However, we can't use the dashboard for debugging as we loose a lot of detail.

Example on one hour time range:
Before
screen shot 2016-12-21 at 15 13 16
After
screen shot 2016-12-21 at 15 14 17

keymon:
In order to use the graph for debugging: 

 1. Go to the board.
 2. click on the graph
 3. change the window from 5m to 1m 

IMO the idea of the dashboard is to be a dashboard: Display a highlight view of the state to be able to spot issues and analyse the health of the platform by examining the screen.

paroxp:
As per comment here: https://github.com/alphagov/paas-cf/pull/679#issuecomment-268816244

---

LGTM - However, I'd like to gather more people from the team, to agree on these. After all:

> The needs of the many outweigh the needs of the few.

PR #682, opened 2016-12-20
Fix link to paas-dashboard

timmow:
Small typo in the link to the paas dashboard. Also changed it to a link
so when viewing in web interface it is clickable

### DIFF
--- a/tools/paas_dashboard/README.md
+++ b/tools/paas_dashboard/README.md
@@ -30,7 +30,7 @@ See [dashing.io](http://dashing.io) and the [smashing wiki](https://github.com/d
 ### Our deployment
 
 We deploy our smashing app as part of the `create-bosh-cloudfoundry` pipeline.
-It should be available at https://paas-dasboard.cloudapps.digital
+It should be available [here](https://paas-dashboard.cloudapps.digital)
 
 ### Running locally
 


mtekel:
Works. Also intentional unsigned merge, so that this doesn't trigger pipeline run.

dcarley:
> Also intentional unsigned merge, so that this doesn't trigger pipeline run.

Could you explain why in more detail?

I'm not sure about the convention. It feels like having inconsistent git history and a red dashboard (which distracts the team) is worse than an unnecessary deployment.

mtekel:
Pipeline run takes some time. There could be other things that we want to deploy with priority. Running such "noop" deploy (which only changes readme) blocks promotion of real changes. Red dashboard is not nice though. I expected the git resource to not pick unsigned merge at all...

keymon:
We shouldn't do that, it breaks the flow and can break other parts (e.g. bosh-cli is broken now)

keymon:
In order to avoid trigger the pipeline, the git-resource includes the option of specify `[ci-skip]` that is more clear and explicit, as it explicitly states that it was the desire of the person merging to NOT trigger the pipeline.

We informally discussed this in the office some time ago regarding similar situations and we concluded that was the best option. 

In order to get a signed commit in master:  `git commit --allow-empty -m "Empty signed commit for merge #662 [ci skip]" -S`

PR #683, opened 2016-12-21
[#131802383] Remove MTU settings

paroxp:
## What

As the problem with MTU has been fixed by amazon already we can remove this temporary fix and maintain one release less.

This reverts 8edc941 and 9ef009f. More details about MTU problem in referenced commit descriptions.

This also fixes problems with upcoming IPSEC release that we discovered. IPSec adds additional headers and authentication data so increases packet size. Some requests were exceeding MTU and than dropped causing connectivity issues.

[More info can be found here.](http://www.cisco.com/c/en/us/td/docs/interfaces_modules/services_modules/vspa/configuration/guide/ivmsw_book/ivmvpnb.html)

## How to review

Obtain your ELB A record from AWS console and export it into `ELB` env variable.

```
export ELB="piotr-dev-cf-loggregator-1885377316.eu-west-1.elb.amazonaws.com"
```

Once, your custom ELB has been established, run the following curl command:

```
curl -k -m 20 -v  https://$ELB:443/apps/3bf24a38-cfdb-4e55-ad38-83e1ad7823f7/containermetrics -H 'Host: doppler.cloud.service.gov.uk:443' -H 'Accept: ' -H 'User-Agent: Go-http-client/1.1'   -H "Authorization: bearer $(yes = | head -1227 | tr -d "\n")" -H 'Accept-Encoding: gzip' ; echo
```

You should **a** response.

## Who can review

Anyone but @combor or @paroxp

### DIFF
--- a/concourse/pipelines/create-bosh-cloudfoundry.yml
+++ b/concourse/pipelines/create-bosh-cloudfoundry.yml
@@ -100,13 +100,6 @@ resources:
       tag_filter: {{cf_graphite_version}}
       branch: gds_master
 
-  - name: os-conf-boshrelease
-    type: git
-    source:
-      uri: https://github.com/alphagov/paas-os-conf-release.git
-      tag_filter: {{cf_os_conf_version}}
-      branch: gds_master
-
   - name: graphite-nozzle
     type: git
     source:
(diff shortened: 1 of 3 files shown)


timmow:
I dont understand from the commit message what relevance this has to #684 - I can guess something about ipsec and MTU, but I think it would be good if #684 caused us to revert the MTU setting to explain why

combor:
I have updated commit message.

saliceti:
Note the `curl` command has to be run from inside the VPC as the problem is due to packets leaving via the NAT gateway.
I tested successfully from `api/0`, going out to each ELB IP.

PR #684, opened 2016-12-21
[#131802383] Enable IPSec

paroxp:
## What

We've integrated the `SAP/ipsec-release` into our CF and enabled its use for all of our nodes, to ensure encrypted communication between them.

The stubs that you will find in a `93e7662` commit, are not being used. They've caused a little confusion, whilst debugging ipsec implementation.

This highly relies on: #683 being merged first.
This also depends on https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/10 . Once that is merged, 8e06c54fc35ff09d2700a638185f0d1f6466da61 should be updated with links to new final release build.

## How to review

Deploy the pipeline from this branch and run `create-bosh-cloudfoundry` pipeline.

It should go through without any issues.

After logging into `cell`, run `tcpdump` command (see below) and check if there is any traffic between `cell` and `router` and if the IPSec is enabled on the calls.

```
tcpdump -n -i eth0 net 10.0.48.0/22
```

You should encounter the following outcome, which will consist of header, encrypted content and finally the key exchange.

```
16:31:09.873951 IP 10.0.48.101 > 10.0.32.4: ESP(spi=0x0be0677a,seq=0x6cf), length 84
16:31:09.874404 IP 10.0.48.101 > 10.0.32.4: ESP(spi=0x0be0677a,seq=0x6d0), length 84
16:31:09.874437 IP 10.0.32.4 > 10.0.48.101: ESP(spi=0x001c1bb1,seq=0x6d2), length 84
16:31:09.943974 IP 10.0.48.101.500 > 10.0.32.4.500: isakmp: phase 2/others ? inf[E]
16:31:09.944240 IP 10.0.32.4.500 > 10.0.48.101.500: isakmp: phase 2/others ? inf[E]
```

## Follow-up

This PR sets the IPSec level to `use` [here](https://github.com/alphagov/paas-cf/pull/684/files#diff-9fe5d38520f7f34353ffcd0816bf98d1R620). This is not recommended level for everyday security... However, it will prevent any downtime in production.

Therefore, please be aware, of the other PR #685, which needs to be applied after these changes have been run through `ci`, `staging` and `prod` successfully.

## Who can review

Anyone but @combor or @paroxp or @mtekel


timmow:
I commented on #683 that I think if removing the MTU fix is important to implementing the IPSec release, we should call that out in the commit message for removing the MTU fix

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -613,3 +613,8 @@ properties:
 
   garden:
     log_level: error
+  racoon:
+    certificate_authority_cert: (( grab secrets.bosh_ca_cert ))
+    certificate_authority_private_key: (( grab secrets.bosh_ca_key ))
```
saliceti: I don't feel comfortable of distributing our CA private key on the VMs. It even ends up in a [world readable file](https://github.com/SAP/ipsec-release/blob/1d27c9115adc838d0cfac402a6615d0dbf16a912/jobs/racoon/templates/etc/racoon/certs/ca_pkey.pem.erb).
I wish we could generate the certificates upfront and provide them via properties. That would require forking this release.
In the meantime, we could create a dedicated CA for IPsec. At least if the key leaks, only IPSec would be affected...

bleach:
To summarise what was agreed at standup, we're not going to merge this just before holidays.

bleach:
Premature send, sorry.

We're going to raise an implementation story to cover monitoring, debugging and other considerations.

In the mean time I think there's definitely value in continuing to review this as a SPIKE.

manifests/cf-manifest/manifest/020-cf-properties.yml
bleach: I share these concerns.

combor: Yes, that's reasonable. However we should do that for any other private key that we've got currently in our manifest.
```
cf-manifest/manifest$ grep -R _key .
./020-cf-properties.yml:    host_key_fingerprint: (( grab secrets.ssh_proxy_host_key.public_fingerprint ))
./020-cf-properties.yml:    agent_key: (( grab secrets.consul_agent_key ))
./020-cf-properties.yml:    encrypt_keys: (( grab secrets.consul_encrypt_keys ))
./020-cf-properties.yml:    server_key: (( grab secrets.consul_server_key ))
./020-cf-properties.yml:    client_key: ""
./020-cf-properties.yml:    peer_key: ""
./020-cf-properties.yml:    server_key: ""
./020-cf-properties.yml:      client_key: (( grab properties.etcd.client_key ))
./020-cf-properties.yml:      server_key: ~
./020-cf-properties.yml:      client_key: (( grab properties.etcd.client_key ))
./020-cf-properties.yml:      client_key: ~
./020-cf-properties.yml:      client_key: (( grab properties.etcd.client_key ))
./020-cf-properties.yml:      client_key: (( grab properties.etcd.client_key ))
./020-cf-properties.yml:    db_encryption_key: (( grab secrets.cc_db_encryption_key ))
./020-cf-properties.yml:      resource_directory_key: (( concat meta.environment "-cf-resources" ))
./020-cf-properties.yml:      app_package_directory_key: (( concat meta.environment "-cf-packages" ))
./020-cf-properties.yml:      droplet_directory_key: (( concat meta.environment "-cf-droplets" ))
./020-cf-properties.yml:      license_key: ~
./020-cf-properties.yml:      buildpack_directory_key: (( concat meta.environment "-cf-buildpacks" ))
./020-cf-properties.yml:      verification_key: (( grab secrets.uaa_jwt_verification_key ))
./020-cf-properties.yml:      signing_key: (( grab secrets.uaa_jwt_signing_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key  ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key  ))
./020-cf-properties.yml:      active_key_label: keylabel1
./020-cf-properties.yml:      encryption_keys:
./020-cf-properties.yml:      server_key: (( grab secrets.bbs_server_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.consul_agent_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key ))
./020-cf-properties.yml:      host_key: (( grab secrets.ssh_proxy_host_key.private_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key  ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key ))
./020-cf-properties.yml:        client_key: (( grab secrets.bbs_client_key ))
./050-rds-broker.yml:        aws_access_key_id: ""
./050-rds-broker.yml:        aws_secret_access_key: ""
./050-rds-broker.yml:        state_encryption_key: (( grab secrets.rds_broker_state_encryption_key ))
```

manifests/shared/deployments/ipsec.yml
```diff
@@ -0,0 +1,9 @@
+meta:
+  ipsec:
+    targets:
+    - 10.0.32.0/24
```
saliceti: Can we avoid hard coding values? Maybe get from terraform outputs?

manifests/cf-manifest/manifest/020-cf-properties.yml
dcarley: > we should do that for any other private key that we've got currently in our manifest.

Could you clarify what you mean by "do that"? We do currently generate those other private keys upfront and provide them as properties. Although I suspect there may be other examples of world-readable files.

The reason it's a particular problem to distribute the BOSH CA this way is that, should somebody compromise a machine that's using IPsec, they can go on to sign new certificates that will be trusted by the rest of the machines, e.g. impersonate the SSH proxy. Whereas if one of the individual certificates is leaked then the damaged is limited to that one service.

combor: I meant to not store private keys on VMs at all.

saliceti: Actually the CA private key is deployed [when the job is updated](https://github.com/SAP/ipsec-release/blob/1d27c9115adc838d0cfac402a6615d0dbf16a912/jobs/racoon/templates/etc/racoon/certs/ca_pkey.pem.erb) but [deleted when racoon starts](https://github.com/SAP/ipsec-release/blob/master/jobs/racoon/templates/bin/racoon_ctl.erb#L36-L54) (and the local certificate is recreated).
So it's not that bad.

dcarley: I'm not sure how that would be possible. Can you explain?

saliceti:
Tested successfully. The traffic is encrypted only between router and cells and all tests pass.

manifests/cf-manifest/manifest/020-cf-properties.yml
combor: @saliceti yes, that's correct. I forgot about it..but anyway...I've got commit with new CA for ipsec so I'm gonna test it now.

combor: @dcarley Sorry, I missed your comment as it's after Colins comment even if it's younger. 
We can go case by case, analyse why we need private keys, check if it's possible to not provide them and if they are necessary make sure they are deleted after usage.

combor:
So now this uses new IPSEC-CA to generate VM certs. Also, we no longer hardcode subnets but getting them from terraform outputs. Needs re-testing.

manifests/shared/deployments/ipsec.yml
combor: I've made it but I strongly dislike that change. Now we have to synchronise 3 files if we want to add one more subnet for cells. Previously it was two.

combor: Actually, it's not 3 it's four:
- https://github.com/alphagov/paas-cf/blob/master/terraform/cloudfoundry/variables.tf#L25
- https://github.com/alphagov/paas-cf/blob/ipsec_release_131802383/terraform/cloudfoundry/outputs.tf#L149
- https://github.com/alphagov/paas-cf/blob/ipsec_release_131802383/manifests/shared/spec/fixtures/terraform/terraform-outputs.yml#L48
- https://github.com/alphagov/paas-cf/blob/ipsec_release_131802383/manifests/shared/deployments/ipsec.yml#L9

Looks ugly.

saliceti:
Suggestion to simplify in my branch: `ipsec_release_131802383_colin`

saliceti:
LGTM

manifests/cf-manifest/manifest/010-cf-jobs.yml
```diff
@@ -301,6 +305,11 @@ jobs:
         agent:
           services:
             gorouter: {}
+      racoon:
+        verify_certificate: "on"
```
dcarley: Seems like this would live better in the global `properties.racoon` alongside `certificate_authority_cert`. Could you move it and remove it from the cell job?

dcarley:
> I'm a bit confused. Does this enable IPsec on all nodes, or only gorouters and cells?
> 
> The commit message which adds racoon says "all nodes" and that seems to match with my understanding of the global properties.racoon.disabled: false.
> 
> But the story says "gorouters and cells" which matches properties.racoon.ports in the router and cell jobs.

`disabled` property was removed (because it was redundant) and commit message says that it's only applied to two job types. Makes sense to me now 👍 

Small nitpick. The commit message says:

> We've decided, the most important ones that should have it, are router and cell nodses.

Could you change that to explain why - we're introducing IPsec between these two types of nodes because they are one of the few communications within Cloud Foundry that don't have any transport encryption of their own and we want to secure tenant HTTP traffic as close to the application container as possible.

manifests/cf-manifest/manifest/010-cf-jobs.yml
paroxp: Certainly.

paroxp:
I've followed @combor's tactics on this and copied your comment into the place of my commit message. Sorry about the lack of precision :(

Thanks!

alext:
This has been merged with the [TMP] commit still in place pointing at the dev release. Also https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/10 (which the dev release relates to) hasn't been merged yet.

mtekel:
Given that this is in prod, it's the best test/verification that the changes in that PR are OK and can be merged. Commit to use final release will then be added to #685 - but the release itself would be the same, so no real changes.

Jonty:
@alext Yes, this was realised just after the merge, but we decided to let it run through without fixing this until today. My bad.

PR #685, opened 2016-12-21
[#131802383] Make sure we require the IPSec at all times

paroxp:
## What

Great! It looks like IPSec has been deployed in `ci`, `staging` and `prod`, right?

As a follow up to #684, we'd like to require the use of IPSec between the enabled nodes.

In the previous PR, IPSec level has been set to `use`, which means we can avoid any downtime. It is recommended to be set to `require`, hence this PR.

Level `require`, forces the traffic to use IPSec but it also drops any connection not using it.

## How to review

- **Make sure you're happy with the IPSec being already successfully deployed to all the environments.**

## Who can review

Anyone but @combor or @paroxp


bleach:
As per https://github.com/alphagov/paas-cf/pull/684#issuecomment-268796311 we're not going to merge this just yet and we're going to create implementation story to follow up in the new year.

PR #686, opened 2016-12-21
Allow apostrophes in tenant email addresses

bleach:
## What

When creating a user with the email address firstname.o'lastname@domain
this script would previously report that the email address was invalid.

Since I wanted to create such a user I needed to make the script accept email addresses with apostrophes in them.

## How to review

Ask a friendly colleague with an apostrophe in their email address if they mind getting a bunch of emails and test against your development environment.

## Who can review

Anyone but @bleach

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -85,7 +85,7 @@ check_params_and_environment() {
     abort_usage "Email must be defined"
   fi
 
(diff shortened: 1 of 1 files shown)


bleach:
In review, @mtekel found that this change didn't work properly on Linux. I've corrected it and there's a brief explanation of what we found out about differences in the regex engines on Linux and Mac OS X in the commit message.

mtekel:
This works now correctly for me as well, merging.

PR #687, opened 2016-12-23
[#136504507] Upgrade stemcells to 3263.14

timmow:
What 
----

[Response to USN-3156-1](https://www.pivotaltracker.com/story/show/136504507)

This contains a fix for [USN-3156-1][1]. The fix was present in the
3263.13 stemcell release, but there seems to be no reason not to update
to the latest stemcell

[1] https://www.ubuntu.com/usn/usn-3156-1/

How to review
-------------

 * Check that the change makes sense (urls, hash)
 * redeploy concourse and redeploy the environment and check that the changes apply and acceptance tests pass

Who can review
-------------

Anyone but me

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -154,8 +154,8 @@ resource_pools:
 - name: bosh
   network: private
   stemcell:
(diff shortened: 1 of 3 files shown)


combor:
Code makes sense. I'll try to redeploy.

PR #689, opened 2016-12-28
[#136605385] Auto-deploy PaaS Roadmap

paroxp:
## What

We'd like to include roadmap deployment, in our production pipeline by default.

In any other environment, it should default to a false value, preventing it from being deployed without a need.

**Also, as part of this PR**

As part of #136550911 we will be replacing our CVE Notifier with a better approach.

We've decided, to remove the notifier sooner rather than later, in order to prevent it from spamming our support email.

## How to review

**Note:** It may prove itself to be difficult, if you're reviewing changes in the `Files changed` tab rather than separate `Commits`. Github may think I am replacing `cve-notifier` with `paas-dashboard`.

- Deploy from this branch, with `DEPLOY_ROADMAP=true` applied.
- Visit `https://paas-roadmap.${DEPLOY_ENV}.dev.cloudpipelineapps.digital/` and see if everything in order.

**Optionally extra checks**

- Deploy from this branch, with `ENABLE_CVE_NOTIFIER=true` applied.
- After running `cf apps`, make sure `cve-notifier` app does not exist.

## Who can review

Anyone but @paroxp


terraform/globals.tf
```diff
@@ -149,3 +149,8 @@ variable "assets_prefix" {
   description = "Prefix for global assests like S3 buckets"
   default     = "gds-paas"
 }
+
+variable "deploy_roadmap" {
```
saliceti: Do we need this variable?

saliceti:
Manual steps for prod:

* Remove paas-roadmap before this change is applied to prod to avoid a route conflict
* Remove CVE notifier whenever

terraform/globals.tf
paroxp: I suppose we don't and do in the same time. We don't really need it to be deployed to staging or master. On some occasion, you'd like to deploy it to dev just to see if it works or checkout the future changes.

PR #694, opened 2016-12-31
[#136340183] Change tracker token management

saliceti:
## What

Story: [Productionnecker](https://www.pivotaltracker.com/story/show/136340183)

The previous behaviour broke in production as the token was never
uploaded to the bucket and the deploy rubbernecker task would wait
forever.
We now use the same approach as for the datadog secrets: the token is
uploaded manually via a make command then retrieved and set as a
pipeline parameter.

## How to review

* Code review or:
* Standard deploy. It shouldn't deploy rubbernecker
* Upload pivotal tracker token: `make dev upload-tracker-token`
* Deploy with `DEPLOY_RUBBERNECKER=true`. It should deploy it.

## Who can review
Not me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,6 +194,9 @@ upload-datadog-secrets: check-env-vars ## Decrypt and upload Datadog credentials
 	$(if $(wildcard ${DATADOG_PASSWORD_STORE_DIR}),,$(error Password store ${DATADOG_PASSWORD_STORE_DIR} does not exist))
 	@scripts/upload-datadog-secrets.sh
 
+upload-tracker-token: check-env-vars ## Decrypt and upload Pivotal tracker API token to S3
+	pass pivotal/tracker_token | aws s3 cp - "s3://${DEPLOY_ENV}-state/tracker_token"
+
 .PHONY: manually_upload_certs
 CERT_PASSWORD_STORE_DIR?=~/.paas-pass-high
 manually_upload_certs: check-tf-version ## Manually upload to AWS the SSL certificates for public facing endpoints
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -194,6 +194,9 @@ upload-datadog-secrets: check-env-vars ## Decrypt and upload Datadog credentials
 	$(if $(wildcard ${DATADOG_PASSWORD_STORE_DIR}),,$(error Password store ${DATADOG_PASSWORD_STORE_DIR} does not exist))
 	@scripts/upload-datadog-secrets.sh
 
+upload-tracker-token: check-env-vars ## Decrypt and upload Pivotal tracker API token to S3
+	pass pivotal/tracker_token | aws s3 cp - "s3://${DEPLOY_ENV}-state/tracker_token"
+
```
paroxp: It's simple and it works as I did check. I'm not sure however, if this is the desired approach, by looking at "how are datadog secrets handled".

I'm not asking to change it, just making a note for other Reviewer, who may perhaps have an opinion on this.

saliceti: * Lack of .PHONY: .PHONY us useful for the case you might have a file in the same directory with the same name as the task. For `upload-tracker-token` it's highly unlikely so I assume we can avoid it.
* Same file approach: we usually create a separate script when the command is more complex. For a one-liner, it didn't seem necessary.
* Using the pass assuming it's the one: I used the same approach as in [environment.sh](https://github.com/alphagov/paas-cf/blob/4d5f8dad330b6bcae94c977c246c2f71e1453047/concourse/scripts/environment.sh#L40) where `$HOME/.paas-pass` is assumed by default

mtekel:
...why upload the token from secrets to S3 and then get it from there? could we input it from secrets directly in the pipelines script?

alext:
> could we input it from secrets directly in the pipelines script?

I think that would break the self-update-pipelines step.

mtekel:
too bad, looks like another good use case for vault...

PR #697, opened 2017-01-03
 [#131876783] Use deploy_env instead of bosh-deployment

keymon:
[#131876783 Migrate paas-cf to paas-bootstrap](https://www.pivotaltracker.com/story/show/131876783)

What?
----

With the migration to paas-bootstrap we want to manage multiple deployments from the same bosh. That prevents us from using bosh-deployment tag to uniquely identify the environment.

We added instead the tag deploy_env which identifies the environment.

In this commit we stop using the bosh-deployment tag, and we use deploy_env.

The metrics from the datadog agent will have the tag from the changes done in [1] and the concourse ones from [2]

[1] https://github.com/alphagov/paas-cf/pull/696y
[2] https://github.com/alphagov/paas-bootstrap/pull/8

Dependencies
------------

It is convenient that https://github.com/alphagov/paas-cf/pull/696 has been deployed some time before in all environments.

Remove the temporary commit before merge.

How to test this?
----------------

Code review. 

You can test it with the metrics from CI thx to the temporary commit which allows define a different name for the monitors. Run this, changing `env2` as needed:

```
terraform apply \
  -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) \
  -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) \
  -var aws_account=ci  \
  -var env2=hector \
  -var env=master  \ 
  -var-file ../ci.tfvars
```

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/datadog/concourse.tf
+++ b/terraform/datadog/concourse.tf
@@ -4,7 +4,7 @@ resource "datadog_monitor" "concourse-load" {
   message            = "${format("Concourse load is too high: {{value}}. Check VM health. @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
   escalation_message = "Concourse load still too high: {{value}}."
   notify_no_data     = false
-  query              = "${format("avg(last_5m):avg:system.load.5{bosh-job:concourse,bosh-deployment:%s} > 200", var.env)}"
+  query              = "${format("avg(last_5m):avg:system.load.5{bosh-job:concourse,deploy_env:%s} > 200", var.env)}"
 
   thresholds {
     warning  = "150.0"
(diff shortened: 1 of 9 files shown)


mtekel:
What about https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-cloudfoundry.yml#L2198 ?

terraform/datadog/concourse.tf
```diff
@@ -61,7 +61,7 @@ resource "datadog_timeboard" "concourse-jobs" {
     viz   = "timeseries"
 
     request {
-      q = "${format("avg:concourse.pipeline_time{bosh-deployment:%s,pipeline_name:create-cloudfoundry}", var.env)}"
+      q = "${format("avg:concourse.pipeline_time{deploy_env:%s,pipeline_name:create-cloudfoundry}", var.env)}"
```
mtekel: for this to work, https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-cloudfoundry.yml#L2198 needs to use deploy_env tag

keymon: I changed it, added one commit.

mtekel:
works, merging

PR #698, opened 2017-01-04
[#SUPPORT] Revert "#679 Plot summarised router response times"

keymon:
What?
----

This reverts https://github.com/alphagov/paas-cf/pull/679, as not all
members of the team agree with the implementation.

How to review?
-------------

Discuss with the rest of the team what is the best approach. If no agreement is achieved, simply merge this to go back to the original graphs.

See #679 for more info.

Who?
----

All team but @keymon

### DIFF
--- a/manifests/cf-manifest/grafana/user-impact.json
+++ b/manifests/cf-manifest/grafana/user-impact.json
@@ -218,9 +218,7 @@
       "height": "250px",
       "panels": [
         {
-          "aliasColors": {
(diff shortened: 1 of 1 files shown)


dcarley:
Closing per discussion in Slack:

- it's a step in the right direction
- we might want to change the graph or axis labels to make it clear that it's a rolling max
- we can improve on it further after using it for a sprint

PR #700, opened 2017-01-04
[#131354359] Improve cell memory usage graph 

henrytk:
## What

We want it to be clear what the graph is showing, so we make three changes:
    
1. Re-label 'AZ Failure' to 'Available memory after failure of one AZ'
2. Re-label 'Cell Failure' to 'Available memory after failure of one cell'
    
These first two changes should clear up any confusion over what those lines
depict.
    
3. Change the Y-axis unit to 'megabytes'.
    
Grafana was treating this as a 'short' number, meaning it would display
2000 as 2K, which in the context of memory makes it look like kilobytes.
Now it knows it is a measure of megabytes and should display values in
gigabytes.

Also, fix a little bug I found along the way.

## How to review

Import the JSON file into any Grafana UI (preferably not production) to view it. Check it makes sense. No bikeshedding allowed.

## Who can review

Anyone but me

### DIFF
--- a/manifests/cf-manifest/grafana/cell-memory-usage.json
+++ b/manifests/cf-manifest/grafana/cell-memory-usage.json
@@ -10,107 +10,108 @@
   "sharedCrosshair": false,
   "rows": [
     {
+      "collapse": false,
+      "editable": true,
       "height": "600px",
       "panels": [
         {
-          "title": "Cluster Memory Usage",
-          "error": false,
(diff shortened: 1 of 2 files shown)


keymon:
Can you attach a screenshot of after and before?

henrytk:
@keymon done

paroxp:
LGTM.

PR #703, opened 2017-01-05
[#131876783] Move runtime-config from paas-cf to paas-bootstrap

keymon:
[#131876783 Migrate paas-cf to paas-bootstrap](https://www.pivotaltracker.com/story/show/131876783)

What?
-----

Second part of the PR https://github.com/alphagov/paas-bootstrap/pull/34

Here we delete all the related code from paas-cf

We also configure the syslog ingestor DNS alias, to which the syslog-forwarders are configured to send the logs to.

Dependencies
------------

Depends on: https://github.com/alphagov/paas-bootstrap/pull/34

How to review?
-------------

Review it with https://github.com/alphagov/paas-bootstrap/pull/34

Check that everything that has to be deleted is deleted.

Who?
----

Anyone but @keymon or @alext


terraform/cloudfoundry/dns-records.tf
```diff
@@ -82,6 +82,14 @@ resource "aws_route53_record" "logsearch" {
   records = ["${aws_elb.logsearch_kibana.dns_name}"]
 }
 
+resource "aws_route53_record" "logsearch_ingestor" {
+  zone_id = "${var.system_dns_zone_id}"
+  name    = "logsearch-ingestor.${var.system_dns_zone_name}."
+  type    = "CNAME"
```
mtekel: AWS recommends to use alias type for ELBs...
http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html

alext: Ok, but in this case I think it's better to be consistent. If we want to use alias records, then that's a change that we should make to all records pointing at ELBs. Having a mixture will cause people reading this later to have to try to figure out why some use an alias and some use a cname.

mtekel:
LG, merging...

PR #704, opened 2017-01-06
[#126126973] Fixes for require mfa

combor:
## What

Story: [Ensure all AWS IAM accounts have MFA enabled](https://www.pivotaltracker.com/story/show/126126973)

This PR brings adjustments required to be compliant with changes introduced in https://github.com/alphagov/paas-bootstrap/pull/16. It changes the way how CONCOURSE_ATC_PASSWORD is fetched. Now we get it from `concourse-sectrets.yml` from S3 bucket. Also, it unifies the way we get ATC password in create CF pipeline ( where possible )

## How to review

- make sure `make dev showenv` displays correct ATC password
- run pipeline to check if all jobs are OK

## Who can review

Not @combor or @saliceti 

You can review it now and don't have to wait for https://github.com/alphagov/paas-bootstrap/pull/16 to be merged.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -123,13 +123,6 @@ resources:
       region_name: {{aws_region}}
       versioned_file: bosh-CA.tar.gz
 
-  - name: concourse-manifest
-    type: s3-iam
-    source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
-      versioned_file: concourse-manifest.yml
-
   - name: ssh-private-key
     type: s3-iam
     source:
@@ -267,7 +260,6 @@ jobs:
         - get: paas-cf
           trigger: {{auto_deploy}}
         - get: git-ssh-private-key
-        - get: concourse-manifest
       - task: init-pipeline-pool
         config:
           platform: linux
@@ -339,7 +331,6 @@ jobs:
               repository: governmentpaas/self-update-pipelines
           inputs:
             - name: paas-cf
-            - name: concourse-manifest
           params:
             DEPLOY_ENV: {{deploy_env}}
             BRANCH: {{branch_name}}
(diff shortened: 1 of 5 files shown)


concourse/scripts/self-update-pipeline.sh
```diff
@@ -23,7 +23,7 @@ else
   echo "Self update pipeline is enabled. Updating. (set SELF_UPDATE_PIPELINE=false to disable)"
 
   VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
-  CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" jobs.concourse.templates.atc.properties.basic_auth_password concourse-manifest/concourse-manifest.yml)
+  CONCOURSE_ATC_PASSWORD=$("$VAL_FROM_YAML" secrets.concourse_atc_password concourse-secrets/concourse-secrets.yml)
```
keymon: Do we actually need this anymore? because in environment.sh we download the secrets directly.

The good thing of using the resource is that it is more explicit. The bad thing is that it is more code to maintain.

The funny thing is that once we migrate to a shared store for secrets, all of this will disappear :)

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -259,6 +252,13 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-rubbernecker
 
+  - name: concourse-secrets
```
keymon: See comment https://github.com/alphagov/paas-cf/pull/704#discussion_r94952388

concourse/scripts/self-update-pipeline.sh
keymon: Actually, all this PR can be about remove all the logic of retrieving the concourse password and only have it in the environments.sh

keymon:
Despite my comment, LGTM and I could merge.

keymon:
LGTM, merging

keymon:
The readme here is wrong, it stills says that is generated: https://github.com/alphagov/paas-cf/blob/8714ea55f61436204b608d32697878da244c4e40/README.md#L259-L264

PR #705, opened 2017-01-06
[#136433665] DATADOG: monitor concourse disk

mtekel:
## What

We have previously removed concourse specific monitor in
b2f63b7f2befd0a1063c24cb5d44e3dc97d070b7 but forgot to remove condition to
exclude concourse from generic disk monitor. Remove that exclusion to cover
concourse disks as well

## How to review

Deploy. Check disk monitor covers concourse disks as well.

## Who can review

not @mtekel


keymon:
LGTM

PR #706, opened 2017-01-06
[#137092871] Increase thresholds for job queue length monitor

henrytk:
## What

Since increasing the parallelisation of our acceptance tests we have
seen higher spikes in the `cf.cc.job_queue_length.total` metric, resulting
in our monitor for CI being triggered (staging and production do not
currently run the acceptance tests).

We increase the thresholds to eliminate these false alerts. Some minor
investigation was done[1] to make sure the relevant components (api,
cells) are coping with the increased load and the alerts really are
false.

[1] https://www.pivotaltracker.com/story/show/137092871

## How to review

The code review required is almost nil. However, the real test is that we don't get false alerts during deployments. I recommend waiting for several deploys to happen and for us to retrospectively look over Datadog to ensure the monitors weren't triggered.

## Who can review

Anyone but me.


keymon:
LGTM

PR #710, opened 2017-01-10
[#136351753] Rotate BBS encryption key

combor:
## What

This is the first step of BBS encryption key rotation. It replaces the hardcoded passphrase with generated one. Old pass is not removed as in this phase it will be used to decrypt already encrypted data. New one will be than used to re-encrypt everything and put back in ETCD. There will we a follow up PR to remove old pass after this change is deployed to all environments. 

## How to review

- code sanity check
- deploy and wait for acceptance tests to pass 

## Who can review

not @combor

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -495,10 +495,12 @@ properties:
       dropsonde_port: (( grab properties.metron_agent.dropsonde_incoming_port ))
 
     bbs:
-      active_key_label: keylabel1
+      active_key_label: key-2017-01
       encryption_keys:
         - label: keylabel1
           passphrase: "a secure passphrase"
+        - label: key-2017-01
+          passphrase: (( grab secrets.bbs_encryption_key ))
       ca_cert: (( grab meta.secrets.bbs_ca_cert ))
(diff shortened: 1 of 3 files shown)


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -495,10 +495,10 @@ properties:
       dropsonde_port: (( grab properties.metron_agent.dropsonde_incoming_port ))
 
     bbs:
-      active_key_label: keylabel1
+      active_key_label: key-2017-01
       encryption_keys:
-        - label: keylabel1
-          passphrase: "a secure passphrase"
```
alext: In the the commit messages, and PR description you mention that the old key is retained for now to enable migration, yet it's being removed here.

combor: Uh, that seems to be commit squash problem :( . Good catch. Fixed now.

dcarley:
Needs rebasing against `master` to take the Travis fix from #712.

combor:
Rebased.

alext:
The code change looks good.  I haven't had a chance to test it though.

Github doesn't provide a "my comments have been addressed" option for a review, only "approve review"

PR #714, opened 2017-01-11
Update accidental dev datadog boshrelease pointer.

Jonty:
## What

8e06c54 should not have been merged to master as it points to a development release. This commit updates it to point to the final datadog boshrelease.

## How to review

Check the release URL is valid & hash is correct, if you must.

## Who can review

Not @jonty.


keymon:
LGTM

keymon:
I hope this will do a noop deployment, the ID of the packages should be the same.

PR #718, opened 2017-01-12
[#133198093] Document digicert TLS certificate process

timmow:
How to obtain and rotate TLS certs from digicert

## What

[#133198093](https://www.pivotaltracker.com/story/show/133198093)

Document how to rotate TLS keys/certs
## How to review

Log into digicert and check the instructions make sense. 
It probably doesn't make sense to rotate or obtain any certificates, but you should be able to see enough for review from the web interface

## Who can review

Anyone with access to the high creds store - I think thats @bleach @henrytk @dcarley and @alext

### DIFF
--- a/doc/non_dev_deployments.md
+++ b/doc/non_dev_deployments.md
@@ -35,6 +35,82 @@ For staging and production a `DEPLOY_ENV` is required even though it's not used
 in domain names. Staging should use a `DEPLOY_ENV` of "staging", and production
 should use "prod".
 
+## Obtaining new SSL certificates
+
+We use digicert for our SSL certificates. The credentials for this are stored in
+the high creds store - this is a global account but with these credentials you
+should create your own account under Account - Users. You should only do this if
+you have access to the high creds store, and do not create accounts for users
+without access to this.
+
+Once you have this account, and access to a suitable payment method, you can
+order a new SSL certificate by going to "Request a Certificate". The certificate
+type needed is a "Wildcard Plus" certificate.
+
+
+## Rotating SSL certificates in digicert
+
+To rotate an existing certificate, go to Certificates - Orders. Then click on
+the order with the Common Name you want to rotate. You must then use the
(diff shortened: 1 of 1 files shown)


doc/non_dev_deployments.md
```diff
@@ -35,6 +35,78 @@ For staging and production a `DEPLOY_ENV` is required even though it's not used
...
+            --passphrase '' \
+            --country GB \
+            --organization 'Government Digital Service' \
+            --common-name '*.example.com'
+```
+
+replacing example.com with the required common name
+
+In the commit message for the certificates in the high cred store, you should
+include the version of the container used to generate the certs and the go
```
dcarley: Could we include a brief explanation about why we do this so that people understand the importance? Something about being able to verify the source if there are future vulnerabilities in the libraries that we used to create the keys.

doc/non_dev_deployments.md
```diff
@@ -35,6 +35,78 @@ For staging and production a `DEPLOY_ENV` is required even though it's not used
...
+version. This information can be found by running
+
+`docker inspect governmentpaas/certstrap`
+
+You should also include the full
```
dcarley: I think this sentence duplicates the previous paragraph. Could we change "This information can be found by running" to "You can do this by pasting the output of the following command into the commit message"?

timmow: In the previous commit messages we've explicitly included the revision and the go version, and then pasted the output of the docker inspect command which is what I was trying to get at...

dcarley: Ahh. All of that information appears to be in the `docker inspect` output by way of environment variables. Do you think that's enough?

timmow:
@dcarley I have updated - what do you think?

dcarley:
> @dcarley I have updated - what do you think?

LGTM, thanks! Will merge when Travis has finished.

PR #719, opened 2017-01-12
[#136725001] Add plans with Storage Encryption enabled

keymon:
[#136725001 Enable at rest encryption for broker created RDS DB instances](https://www.pivotaltracker.com/n/projects/1275640/stories/136725001)

What?
-----

We want provide the option of provision DBs with Storage Encription enabled. This is a requirement for DIT onboarding.

At rest encryption has been mentioned as a good thing to turn on by our AWS account manager.

As explained in the ADR: https://github.com/alphagov/paas-team-manual/pull/88 , we decided to provide additional explicit plans with Encryption enabled, and keep the old ones. We will add logic in the broker to prevent updates between plans with or without encryption. We only add to the HA plans, as removing plans later is painful.

In this PR we add two plans based on the Medium and Large HA plans.

We upgrade the RDS broker to ensure that it is not possible update between plans with/without encryption. We check that the manifest defined plans enable/disable the encryption as expected.

Dependencies
------------

 - Related: https://github.com/alphagov/paas-team-manual/pull/88
 - Must be merged first and this PR updated: https://github.com/alphagov/paas-rds-broker/pull/34 https://github.com/alphagov/paas-rds-broker-boshrelease/pull/28

How to review
---------

 - Code review
 - Travis should pass
 - Create a DB in a encrypted and other in un-encrypted plans.
 - Check the settings in aws RDS.
 - Try to update between plans with/without: should fail
 - Try to update within plans with and without: should work

Who?
----

Anyone but @keymon

### DIFF
--- a/manifests/cf-manifest/manifest/050-rds-broker.yml
+++ b/manifests/cf-manifest/manifest/050-rds-broker.yml
@@ -6,6 +6,7 @@ meta:
       storage_type: "gp2"
       auto_minor_version_upgrade: true
       multi_az: false
+      storage_encrypted: false
       publicly_accessible: false
       copy_tags_to_snapshot: true
       skip_final_snapshot: false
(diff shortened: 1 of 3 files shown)


manifests/cf-manifest/manifest/050-rds-broker.yml
```diff
@@ -88,6 +89,8 @@ jobs:
                     allocated_storage: 5
                     backup_retention_period: 0
                     skip_final_snapshot: true
+                    multi_az: false
+                    storage_encrypted: false
```
alext: These 2 are unnecessary because they already come from the defaults.

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -498,6 +498,8 @@ properties:
     bbs:
       active_key_label: key-2017-01
       encryption_keys:
+        - label: keylabel1
+          passphrase: "a secure passphrase"
```
alext: This looks to have been accidentally reintroduced.

manifests/cf-manifest/spec/manifest/rds_broker_spec.rb
```diff
@@ -75,78 +75,200 @@
...
-          it { is_expected.to include("multi_az" => false) }
+          it_behaves_like "backup enabled plans"
+          it_behaves_like "non-HA plans"
+          it_behaves_like "Encryption disabled plans"
         end
 
         describe "M-HA-dedicated-9.5" do
           let(:plan) { pg_plans.find { |p| p["name"] == "M-HA-dedicated-9.5" } }
-          subject { plan.fetch("rds_properties") }
+          subject { plan }
```
alext: With this refactoring, there's no real need to have a separate `let(:plan)` and `subject`. They could be combined into: `subject { pg_plans.find { |p| p["name"] == "M-HA-dedicated-9.5" } }`.

PR #721, opened 2017-01-13
[#136100681] Add Postgres DB for BBS

saliceti:
# What

Story: [Migrate Diego to use a Relational Datastore](https://www.pivotaltracker.com/story/show/136100681)

Create a new RDS DB for BBS and configure BBS to use it.

Triggers the data from ETCD to be migrated to the DB.

# How to review
* Deploy from master
* Deploy from the branch
* The migration takes a few minutes when the database VM is being updated
* Check for `bbs.migration-manager.finished-migrations` in BBS log. If you use logsearch, look for `migrations`.
* Check for lrps in BBS DB ([see comment](https://github.com/alphagov/paas-cf/pull/721#issuecomment-273091264) on where to run this from):
`psql -d bbs -h "${db_address}" -U bbs -W -c 'select * from actual_lrps '`
* Check via `cfdot` by `bosh ssh database/0` and running `source /var/vcap/jobs/cfdot/bin/setup && cfdot desired-lrps`
* Check all tests pass

# Who can review
Not @combor or me

### DIFF
--- a/manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
+++ b/manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
@@ -6,3 +6,9 @@ vm_extensions:
     cloud_properties:
       elbs:
         - (( grab terraform_outputs.cf_ssh_proxy_elb_name ))
+
+  - name: cf_rds_client
+    cloud_properties:
+      security_groups:
+        - (( grab terraform_outputs.default_security_group ))
+        - (( grab terraform_outputs.cf_rds_client_security_group ))
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -220,6 +220,8 @@ jobs:
       - name: cf
     update:
       serial: true
+    vm_extensions:
+      - cf_rds_client
 
   - name: uaa
     azs: [z1, z2]
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -511,6 +511,9 @@ properties:
         client_cert: (( grab secrets.consul_agent_cert ))
         client_key: (( grab secrets.consul_agent_key ))
         require_ssl: false
+      sql:
+        db_connection_string: (( concat "postgres://bbs:" secrets.cf_db_bbs_password "@" terraform_outputs.cf_db_address ":5432/bbs" ))
+        db_driver: postgres
       dropsonde_port: (( grab properties.metron_agent.dropsonde_incoming_port ))
       require_ssl: true
 
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -514,6 +514,7 @@ properties:
       sql:
         db_connection_string: (( concat "postgres://bbs:" secrets.cf_db_bbs_password "@" terraform_outputs.cf_db_address ":5432/bbs" ))
         db_driver: postgres
+        max_open_connections: 20
       dropsonde_port: (( grab properties.metron_agent.dropsonde_incoming_port ))
       require_ssl: true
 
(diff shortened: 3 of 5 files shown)


dcarley:
Travis failed with:
```
        - $.properties.diego.bbs.sql.db_connection_string: Unable to resolve `secrets.cf_db_bbs_password`: `$.secrets.cf_db_bbs_password` could not be found in the datastructure
```

Do you need to add it to the fixtures file?

Make me wonder actually, why do we use a fixtures file here instead of `generate-cf-secrets.rb`?

saliceti:
@dcarley good point, let me try to use `generate-cf-secrets.rb` instead

manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
```diff
@@ -6,3 +6,9 @@ vm_extensions:
     cloud_properties:
       elbs:
         - (( grab terraform_outputs.cf_ssh_proxy_elb_name ))
+
+  - name: cf_rds_client
+    cloud_properties:
+      security_groups:
+        - (( grab terraform_outputs.default_security_group ))
```
alext: I don't think it's necessary to add the default security group here, it'll already be added to the base VM. (It wasn't necessary for the `ssh_proxy_elb` vm_extension above)

alext: Actually, I'm getting confused between security groups and elbs. This needs to be here.

saliceti:
@dcarley I've added the fixture. As discussed IRL, using `generate-cf-secrets.rb` is probably the way to go but it should be a separate story.

manifests/cf-manifest/cloud-config/021-cf-vm-extensions.yml
saliceti: As discussed, this list of security groups overrides the default list from bosh, so we need to add the default group again here.

dcarley:
> @dcarley I've added the fixture. As discussed IRL, using generate-cf-secrets.rb is probably the way to go but it should be a separate story.

ACK. Could you create one?

saliceti:
Story created: https://www.pivotaltracker.com/story/show/137698283

dcarley:
Deployed OK and pipeline tests passed, however..

I ran the availability tests from my laptop at the same time because I was interested in whether it affected applications and the `availability-tests` in the pipeline finishes before the migration has. It went something like this:

16:03:25 `cf-deploy` started
16:04:53 `database/1` logs: `bbs.migration-manager.finished-migrations`
16:08:11 `database/0` logs: `bbs.migration-manager.running-migrations`
16:08:11 `database/0` logs: `bbs.migration-manager.finished-migrations`
16:12:11 `cf-deploy` finished
16:13:~16:14 the local availability tests had 277 dropped requests

Now I'm not confident that it wasn't my broadband or the VPN that dropped the traffic, but I am suspicious of the timing. I might need to pair with you tomorrow to look into it further.

dcarley:
Couple of notes from reviewing:

- where it says "connecting to the bbs VM" I think it means `bosh ssh database/0` which is where the bbs job is running
- to connect the Postgres DB I had to `apt-get install postgresql-client-9.3` on `database/0` in my development environment and get the RDS hostname and password from the `ps` output of `bbs`. Not sure if there's an easier way?

saliceti:
Tip: you can run the `cf-terraform` job and hijack the `init-db` task. It has `psql` installed and credentials available.

dcarley:
I rolled back to `master`, modified the `availability-tests` in the pipeline to continue running and log a timestamp, then deployed this branch again. Once again it dropped some requests but I've **now** figured out why - the healthcheck app is redeployed after `cf-deploy` and doesn't use any zero-downtime plugins ([#121934253](https://www.pivotaltracker.com/story/show/121934253)).

tl;dr: This is good to merge when you've completed your performance testing for the RDS instance size. Taking my name off it in the meantime in case I'm not available when it becomes ready.

alext:
:+1: This looks good to merge. We're going to hold off merging until the morning so that it hits production while we're in the office.

dcarley:
As discussed in Slack: Assigning to myself and merging, so that we can progress the story while Alex is out of the office this morning.

PR #722, opened 2017-01-13
[#136100681] Remove ETCD configuration for BBS

saliceti:
⚠️  Depends on PR https://github.com/alphagov/paas-cf/pull/721 ⚠️ 

# What

Story: [Migrate Diego to use a Relational Datastore](https://www.pivotaltracker.com/story/show/136100681)

BBS has now migrated to postgres, we don't need ETCD anymore.

# How to review
* After https://github.com/alphagov/paas-cf/pull/721 is merged, rebase this branch on master
* Deploy from master
* Make sure all tests pass

# Who can review
Not @combor or me

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -506,10 +506,10 @@ properties:
       auctioneer:
         api_url: http://auctioneer.service.cf.internal:9016
       etcd:
-        machines: [(( grab properties.etcd.advertise_urls_dns_suffix ))]
-        ca_cert: (( grab meta.secrets.consul_ca_cert ))
-        client_cert: (( grab secrets.consul_agent_cert ))
-        client_key: (( grab secrets.consul_agent_key ))
+        machines: []
+        ca_cert: ""
+        client_cert: ""
+        client_key: ""
         require_ssl: false
       sql:
         db_connection_string: (( concat "postgres://bbs:" secrets.cf_db_bbs_password "@" terraform_outputs.cf_db_address ":5432/bbs" ))


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -506,11 +506,14 @@ properties:
       auctioneer:
         api_url: http://auctioneer.service.cf.internal:9016
       etcd:
-        machines: [(( grab properties.etcd.advertise_urls_dns_suffix ))]
-        ca_cert: (( grab meta.secrets.consul_ca_cert ))
-        client_cert: (( grab secrets.consul_agent_cert ))
-        client_key: (( grab secrets.consul_agent_key ))
+        machines: []
+        ca_cert: ""
```
mtekel: remove if this is empty? or is there some other default cert?

saliceti: Unfortunately diego-release [expects it](https://github.com/cloudfoundry/diego-release/blob/v0.1487.0/jobs/bbs/templates/bbs_ca.crt.erb) and there is [no default](https://github.com/cloudfoundry/diego-release/blob/v0.1487.0/jobs/bbs/spec#L68-L69)

mtekel: you could change it then to st like "not_used" - to make it obvious that this has to be here, but is not being used

otherwise that looks like many of those null variables we have around manifest...

keymon: +1 for a more clear text if not used, if possible.

alext: Or maybe a comment...

saliceti: It's already mentioned in the commit message. Should I add an inline comment?

combor: We used to use commit messages as a place for such comments and we preferred them over inline comments.

mtekel: the thing with empty quotes is that it looks like there was something supposed to be (e.g. fetched from some file or generated) but something silently failed and now you have empty quote. If you put there 'not_used', then you make it clear that it is not used, as 'not_used' is also clearly not a valid cert, so deployment would fail if it was to be used...

alext: I think the best thing would be to add a `FIXME` comment to this whole etcd block saying we should remove it when we upgrade to a diego release where this is no longer mandatory.

dcarley: I think it's fine as-is, personally. We shouldn't get too hung up on it.

Presumably it will be removed at some point in the future when etcd support goes from bbs.

alext: > I think it's fine as-is, personally. We shouldn't get too hung up on it.

Agreed. I'm going to merge this as is.

PR #723, opened 2017-01-16
Change user facing email address in user scripts

mtekel:
## What

When we create a user or rotate their password, we want to send email from
the support address, so that when they respond, it goes to support and there
is someone to deal with it.

## How to review

You can try creating users or rotating some passwords. Either create a dummy user or you can even rotate your own PW. **NEVER rotate PWs of other people** unless requested! Or you can trust me. You can also check that the support address is now verified: `aws ses get-identity-verification-attributes --identities gov-uk-paas-support@digital.cabinet-office.gov.uk`

## Who can review

not @mtekel

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -9,7 +9,7 @@ source "${SCRIPT_DIR}/common.sh"
 ###########################################################################
 # Defaults
 DEFAULT_SPACE=sandbox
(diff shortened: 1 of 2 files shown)


dcarley:
> aws ses get-identity-verification-attributes --identities gov-uk-paas-support@digital.cabinet-office.gov.uk

Has this been added to all of the AWS accounts?

Could you make a note in the commit message that it's been done?

mtekel:
good point. I have added it to other accounts and adjusted the commit message.

PR #724, opened 2017-01-16
[#137698283] Test using generated secrets instead of fixtures

alext:
## What

To avoid us having to update the fixture every time a new secret is added/removed, this instead uses the secret generator to generate a secrets file for use in the tests.

I've also taken this opportunity to do refactor some other test refactorings. See individual commits for details.

## How to review

Code review. Verify that tests pass in Travis.

## Who can review

Anyone but myself.


dcarley:
Commits look good and tests pass 👍

PR #726, opened 2017-01-17
[#137698283] CF manifest tests cleanup

alext:
## What

This is a follow-up to #724:

Now that we're generating the secrets for the tests, there's no need to separately generate them for the yaml lint check.

I've taken the opportunity to separate the cloud-config from the manifest in the tests to make it clearer what comes from where.

## How to review

Code review. Verify tests pass in travis.

## Who can review

Anyone but myself.


dcarley:
👍 👍

PR #728, opened 2017-01-17
[#119089085] Stop graphite nozzle sg cleanup

mtekel:
## What

After #727 is merged and applied, we no longer need to keep unbinding the graphite nozzle sg. This PR fixes that.

## How to review

Checkout, make dev pipelines, run cf-deploy job, observe that graphite-nozzle group is no longer being removed from global running and staging SGs.

## Who can review

not @mtekel


alext:
:+1: This looks good to merge once #727 has reached production.

PR #730, opened 2017-01-18
[support] Add default env. names for ci, staging and prod

mtekel:
## What
Add default env. names for staging and prod.

This means you don't have to re-export DEPLOY_ENV every time you switch from
your dev to work on ci, staging or prod. In our current version of scripts,
you have to provide correct DEPLOY_ENV name, as that is used to access
environment's bucket and if it's wrong, showenv will fail, failing any
subsequent steps.

The error that you get when DEPLOY_ENV is wrong:
```
/usr/lib/ruby/2.2.0/psych.rb:370:in `parse': (<unknown>): mapping values are not allowed in this context at line 1 column 79 (Psych::SyntaxError)
	from /usr/lib/ruby/2.2.0/psych.rb:370:in `parse_stream'
	from /usr/lib/ruby/2.2.0/psych.rb:318:in `parse'
	from /usr/lib/ruby/2.2.0/psych.rb:245:in `load'
	from concourse/scripts/val_from_yaml.rb:10:in `load_yaml'
	from concourse/scripts/val_from_yaml.rb:46:in `<main>'
```

## How to review

Set your DEPLOY_ENV to your dev env. name (or some other random value). With ci/staging/prod credentials loaded, do `make ci/staging/prod showenv`.

## Who can review

not @mtekel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
-.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck lint_concourse check-env-vars
-
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck lint_concourse check-env
 .DEFAULT_GOAL := help
 
 help:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -116,6 +116,7 @@ ci: globals check-env-vars ## Set Environment to CI
 	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS=${ALERT_EMAIL_ADDRESS})
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
 	$(eval export ENABLE_DATADOG=true)
+	$(eval export DEPLOY_ENV ?= master)
```
dcarley: > Ci only has default name, as you could decide to operate on build instead of master.

Would you use `alphagov/paas-cf` against the build environment? Isn't it managed by `alphagov/paas-release-ci`?

dcarley:
👍  the idea.

Would this fix the problem you had last week of `make prod showenv` giving you the Kibana/Grafana password from your dev environment?

mtekel:
yes, that would fix the problem. I did this because I had to set env manually way too often when working on support and for ci/staging/prod, the env. names are set and not easy/likely to change, so why not have them in makefile directly...

dcarley:
Works, but there is an oddity that you still need to set `DEPLOY_ENV` in your shell because the `check-env-vars` target requires it (and validates something that isn't used):
```
➜  paas-cf git:(easy-showenv) ✗ printenv DEPLOY_ENV
➜  paas-cf git:(easy-showenv) ✗ make prod showenv
Makefile:17: *** Must pass DEPLOY_ENV=<name>.  Stop.
➜  paas-cf git:(easy-showenv) ✗ DEPLOY_ENV=notused make prod showenv
export AWS_ACCOUNT=prod
export DEPLOY_ENV=prod
…
```

I'm not sure if that matters. Any thoughts?

Jonty:
@dcarley I think that can be resolved by removing `check-env-vars` from the `ci`, `staging`, and `prod` Make targets?

dcarley:
> @dcarley I think that can be resolved by removing check-env-vars from the ci, staging, and prod Make targets?

Yep, sounds reasonable. We should also rename it to `check-deploy-env` to make it clear that it's not doing more.

mtekel:
I might want to add AWS credentials check to ensure you are really operating on the account you intend, we already have some scripts for that...

mtekel:
Did a bigger re-factor. Let me know if that works for you...

mtekel:
One more thing - I think it's OK for someone to take over this and do any further improvements you find useful...

dcarley:
> One more thing - I think it's OK for someone to take over this and do any further improvements you find useful...

We should be careful about adding further features. The reason I've been slow to review this is because the scope changed and there's now more to review.

mtekel:
dropped .phony re-factor

mtekel:
I bumped into this (wrong deploy env and wrong account) 4 more times while on support this week. Would be good to get this reviewed and merged...

dcarley:
Looks good and works. Sorry it's taken me *so* long to review.

alext:
This missed out the `setup_cdn_instances` target that has been added since this PR was opened, and is still referencing the `check-aws-credentials` target.

I'll fix this as part of #796

PR #731, opened 2017-01-18
[#121401583] Use vm_extensions instead of dedicated VM types.

alext:
## What

This avoids lots of duplication in the vm_type definitions. Many of them
were duplicates of the medium type with added security_groups, elbs,
instance_profiles etc.

Switching to vm_extensions[1] reduces this duplication. It can also make
it much clearer what's applied to a given job (assuming you name the
extensions well).

This also introduces a manifest test to ensure that all referenced
vm_extensions exist.

I have not updated the router and cell vm_types because these are more
specialised, and I didn't feel that using vm_extensions would make
things clearer (if anything it might be less clear).

[1]https://bosh.io/docs/cloud-config.html#vm-extensions

## How to review

Code review.
This change should be a no-op at deployment. Deploy against a current master deploy to verify that this is the case.

## Who can review

Anyone but myself.


mtekel:
Superb work. Naming of the extensions makes it clear how the VM differs and I don't need to look into other files any more to figure out what the actual extensions/speciality of the VM were. Even better, upon merging bosh figures all this out and doesn't re-deploy and VMs or apply any other changes. Merging.

PR #732, opened 2017-01-19
[#135224045] Stagger CF RDS maintenance windows in prod/staging/CI

alext:
## What

This means that if there's a problem with an upgrade, we are likely to notice
and have a chance to rectify the problem before it reaches production.

We want this to be deployed separately ahead of upcoming changes to DB versions etc so that it's in place beforehand and therefore we can be confident that these updated maintenance windows will apply to the version upgrades.

## How to review

Code review.
Deploy to a dev environment, and verify that the maintenance window for the CF RDS instance gets updated to be on Tuesday, and that this change takes effect immediately.

## Who can review

Anyone but myself.


mtekel:
Why not do it like in https://github.com/alphagov/paas-bootstrap/pull/40? Meaning, have default, that common for ci and dev, and only override staging and prod?

alext:
> Why not do it like in alphagov/paas-bootstrap#40? Meaning, have default, that common for ci and dev, and only override staging and prod?

I chose to follow the pattern that had been used for the other cf db parameters for consistency.

terraform/dev.tfvars
```diff
@@ -4,4 +4,5 @@ apps_dns_zone_id = "Z3R6XFWUT4YZHB"
 cf_db_multi_az = "false"
 cf_db_backup_retention_period = "0"
 cf_db_skip_final_snapshot = "true"
+cf_db_maintenance_window = "Tue:04:00-Tue:05:00"
```
saliceti: Can we move it to Monday?

PR #737, opened 2017-01-24
[#135224045] Cleanup FIXMEs from CF RDS upgrade.

alext:
🚨  **DO NOT MERGE** until the upgrade has applied to all environments - production is scheduled to be applied between 4-5am on Thursday 26th, and any pending dev environments will be applied the following Monday (30th)

## What

The parameter group had to be left in place from the upgrade (#734)
because the upgrade wasn't applied immediately, instead it was applied
in the defined maintenance window for the instance.

Now that the upgrade has been applied, the old parameter group can be
removed.

Similarly, we no longer need to set allow_major_version_upgrade because
the upgrade has completed.

## How to review

Code review.
Verify these changes apply cleanly to a deployed environment.

## Before merging

Verify the upgrade has applied to all environments.

## Who can review

Anyone but myself.

### DIFF
--- a/terraform/cloudfoundry/cf_rds.tf
+++ b/terraform/cloudfoundry/cf_rds.tf
@@ -29,19 +29,6 @@ resource "aws_security_group" "cf_rds" {
   }
 }
 
-# FIXME: Remove this once the upgrade to 9.5 has applied everywhere.
-resource "aws_db_parameter_group" "cf" {
-  name        = "${var.env}-cf"
-  family      = "postgres9.4"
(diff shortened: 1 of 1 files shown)


alext:
The RDS upgrade has now applied to all environments, so this is good to merge.

mtekel:
This is good to merge. Unfortunately due to [small issue in TF](https://github.com/hashicorp/terraform/issues/11519), 1st run of terraform job will fail and it needs re-running to fix the pipeline flow. I'll do this to help the PR through...

PR #738, opened 2017-01-24
[#137701781] Set up smoke test alerts to PagerDuty using datadog

mtekel:
## What

[Set up smoke test alerts to PagerDuty using datadog](https://www.pivotaltracker.com/n/projects/1275640/stories/137701781)

Create a monitor that notifies via pagerduty if >= 3 continuous smoke tests failures in 15m or no_data for 15m.

## How to review

Deploy with datadog enabled. Put yourself on [schedule](https://gds-paas.pagerduty.com/schedules#PEQKA3V) in pagerduty. Run continuous smoke tests. Make them fail 3 times in 15 min window (you can e.g. stop cloud controllers). You should get notification based on the preferences you set-up in PagerDuty. Stop running tests for 15 mins - you should get no_data notification. Note that if you test out of "office hours (9 - 17oo), then this will be marked as low priority and based on your notification settings you might not be notified. Before merging, remove temp commit 57346d2.

## Who can review

not @mtekel or @paroxp or @combor

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1507,53 +1507,89 @@ jobs:
 
                   bosh cleanup --all
 
-        - task: datadog-terraform-apply
-          config:
-            platform: linux
-            image_resource:
-              type: docker-image
-              source:
-                repository: governmentpaas/terraform
-            inputs:
-              - name: datadog-tfstate
-              - name: paas-cf
-              - name: config
-            outputs:
-              - name: updated-tfstate
-            params:
-              TF_VAR_env: {{deploy_env}}
-              TF_VAR_datadog_api_key: {{datadog_api_key}}
-              TF_VAR_datadog_app_key: {{datadog_app_key}}
-              TF_VAR_aws_account: {{aws_account}}
-              ENABLE_DATADOG: {{enable_datadog}}
-            run:
-              path: sh
-              args:
-                - -e
-                - -c
-                - |
-                  if [ -n "${TF_VAR_datadog_api_key}" ] && [ -n "${TF_VAR_datadog_app_key}" ] && [ "${ENABLE_DATADOG}" = "false" ]; then
-                    echo "Datadog disabled but keys present, running check cleanup..."
-                    terraform destroy -force \
-                       -state=datadog-tfstate/datadog.tfstate \
-                       -state-out=updated-tfstate/datadog.tfstate \
-                       paas-cf/terraform/datadog
-                    exit 0
-                  fi
+        - do:
+          - task: datadog-terraform-apply
+            config:
+              platform: linux
+              image_resource:
+                type: docker-image
+                source:
+                  repository: governmentpaas/terraform
+              inputs:
+                - name: datadog-tfstate
+                - name: paas-cf
+                - name: config
+              outputs:
+                - name: updated-tfstate
+              params:
+                TF_VAR_env: {{deploy_env}}
+                TF_VAR_datadog_api_key: {{datadog_api_key}}
+                TF_VAR_datadog_app_key: {{datadog_app_key}}
+                TF_VAR_aws_account: {{aws_account}}
+                ENABLE_DATADOG: {{enable_datadog}}
+              run:
+                path: sh
+                args:
+                  - -e
+                  - -c
+                  - |
+                    if [ -n "${TF_VAR_datadog_api_key}" ] && [ -n "${TF_VAR_datadog_app_key}" ] && [ "${ENABLE_DATADOG}" = "false" ]; then
+                      echo "Datadog disabled but keys present, running check cleanup..."
+                      terraform destroy -force \
+                         -state=datadog-tfstate/datadog.tfstate \
+                         -state-out=updated-tfstate/datadog.tfstate \
+                         paas-cf/terraform/datadog
+                      exit 0
+                    fi
 
-                  if [ "${ENABLE_DATADOG}" = "true" ]; then
-                    terraform apply \
-                      -var-file=paas-cf/terraform/{{aws_account}}.tfvars \
-                      -state=datadog-tfstate/datadog.tfstate -state-out=updated-tfstate/datadog.tfstate \
-                      -var-file=config/job_instances.tfvars paas-cf/terraform/datadog
-                  else
-                    echo "Datadog disabled, skipping terraform run..."
-                    cp datadog-tfstate/datadog.tfstate updated-tfstate/datadog.tfstate
-                  fi
-          ensure:
-            put: datadog-tfstate
-            params:
-              file: updated-tfstate/datadog.tfstate
+                    if [ "${ENABLE_DATADOG}" = "true" ]; then
+                      terraform apply \
+                        -var-file=paas-cf/terraform/{{aws_account}}.tfvars \
+                        -state=datadog-tfstate/datadog.tfstate -state-out=updated-tfstate/datadog.tfstate \
+                        -var-file=config/job_instances.tfvars paas-cf/terraform/datadog
+                    else
+                      echo "Datadog disabled, skipping terraform run..."
+                      cp datadog-tfstate/datadog.tfstate updated-tfstate/datadog.tfstate
+                    fi
+            ensure:
+              put: datadog-tfstate
+              params:
+                file: updated-tfstate/datadog.tfstate
+
+          - task: datadog-apply-attributes
+            config:
+              platform: linux
+              image_resource:
+                type: docker-image
+                source:
+                  repository: ruby
+                  tag: 2.2-alpine
+              inputs:
+                - name: datadog-tfstate
+                - name: paas-cf
+              params:
+                TF_VAR_env: {{deploy_env}}
+                TF_VAR_datadog_api_key: {{datadog_api_key}}
+                TF_VAR_datadog_app_key: {{datadog_app_key}}
+                TF_VAR_aws_account: {{aws_account}}
+                ENABLE_DATADOG: {{enable_datadog}}
+              run:
+                path: sh
+                args:
+                  - -e
+                  - -c
+                  - |
+                    # FIXME: Remove this task when https://github.com/zorkian/go-datadog-api/issues/56 is fixed
+                    if [ "${ENABLE_DATADOG}" = "true" ]; then
+                      BUNDLE_GEMFILE=paas-cf/Gemfile bundle install --without=default
+                      echo "Applying require_full_window atributes for monitors that set it to false..."
+                      echo
+                      paas-cf/concourse/scripts/update_monitor_options.rb < datadog-tfstate/datadog.tfstate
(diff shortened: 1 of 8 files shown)


terraform/datadog/concourse.tf
```diff
@@ -42,6 +42,45 @@ resource "datadog_monitor" "continuous-smoketests" {
...
+    The `continuous-smoke-tests` have not reported any metrics for a while.
+
+    We need to investigate.
+
+    Notify: @pagerduty-Datadog_no_data
+  {{/is_no_data}}
+EOF
+
+  require_full_window = false
+  notify_no_data      = true
```
saliceti: I'm getting no data alerts (and pagerduty escalation) even when the build finishes successfully. I believe the "build failed" is not sent, and it triggers this alert?
If it's the case, do we need this `notify_no_data`?

terraform/datadog/concourse.tf
```diff
@@ -42,6 +42,45 @@ resource "datadog_monitor" "continuous-smoketests" {
...
+  {{#is_no_data}}
+    The `continuous-smoke-tests` have not reported any metrics for a while.
+
+    We need to investigate.
+
+    Notify: @pagerduty-Datadog_no_data
+  {{/is_no_data}}
+EOF
+
+  require_full_window = false
```
saliceti: The UI shows "Require"

mtekel: I've seen that before. Looks like Datadog API problem. The actual setting is ignored and default is picked (which is based on kind of aggregation you are doing, e.g. avg, in_total etc.). See [this bit](https://www.terraform.io/docs/providers/datadog/r/monitor.html#require_full_window) of TF documentation.

But @keymon mentioned it doesn't happen to him, so maybe he can describe how he worked around that?

terraform/datadog/concourse.tf
mtekel: Good spot. I have also noticed that we have a monitor for slow continuous tests and no-data is one case of very slow test... so perhaps modify that monitor...

terraform/datadog/concourse.tf
keymon: Not sure anymore about that :(

I think there is something that is not working 100% there.

saliceti:
I've tested getting and setting `require_full_window` via [the API](http://docs.datadoghq.com/api/?lang=console#monitors) and it works fine. There may be an issue in the terraform provider or the library?

combor:
We've hit the problem described here: https://github.com/zorkian/go-datadog-api/issues/56

combor:
This can be re-reviewed now.

dcarley:
> re-reviewed

the_artful_dodger_ _re-rewind

Makefile
```diff
@@ -30,7 +30,7 @@ spec:
 	cd scripts &&\
 		BUNDLE_GEMFILE=Gemfile bundle exec rspec
 	cd concourse/scripts &&\
-		go test
+		BUNDLE_GEMFILE=../../scripts/Gemfile bundle exec go test
```
dcarley: We tried this in the past and it slowed the test suite down considerably. There's some context in:

- with Bundler: 8b403df1a1f5b422d6466ce3d2f3a1ccc79b8b5f
- without Bundler: ebca6530aa1917dcac8f17c9fd716ce4891b2587

Can you add it to the `exec.Command()` within the individual test instead?

combor: Ok, changed it as suggested.

concourse/scripts/scripts_suite_test.go
```diff
@@ -1,13 +1,117 @@
...
+	})
+	err := http.ListenAndServe("127.0.0.1:8080", nil)
+	Expect(err).ToNot(HaveOccurred())
+}
+
 func TestScripts(t *testing.T) {
 	RegisterFailHandler(Fail)
+
+	BeforeSuite(func() {
+		go mockDatadogApi()
```
dcarley: I've pushed the commit 2069f06892580727b571c2061b8e519c6d19566e to a branch which refactors this part of the tests and explains why.

Could you cherry-pick+rebase if you agree with it?

combor: I agree with most of the explanation - it was actually my design choice at the beginning ( apart from ghttp) but I'm not sure if I fully understand "- server is started and torndown between each test because it's fast enough". Can you explain?

concourse/scripts/update_monitor_options.rb
```diff
@@ -0,0 +1,54 @@
...
+end
+
+api_key = ENV['TF_VAR_datadog_api_key']
+app_key = ENV['TF_VAR_datadog_app_key']
+api_host = ARGV[0]
+dog = Dogapi::Client.new(api_key, app_key, nil, nil, nil, nil, api_host)
+monitor_ids = find_monitors(tfstate)
+err = update_monitors(monitor_ids, dog)
+if err != nil
+  printf("Error: %s", err)
```
dcarley: I think this should call [`abort`](https://ruby-doc.org/core-2.2.3/Kernel.html#method-i-abort) with the error string so that the script returns a non-zero exit code if it's unable to update the monitor and the failure will be made visible in Concourse. It's worth a test too - reset the handler to return 500 and assert the exit code.

dcarley:
Sorry about the stream of comments. I should have started a review and submitted them as a batch. I've only code reviewed so far and will test it tomorrow, so feel free to address all comments after I've done that.

concourse/scripts/scripts_suite_test.go
dcarley: I was explaining the change from `BeforeSuite` to `BeforeEach` which starts and stops the mock server before each test (in that file) rather than before all tests (in all files). It's natural to think that it would make the tests slower, but starting a server is so fast that it makes no appreciable difference to the time the tests take to run.

combor: Ok, that looks good. I've cherry picked your commit and squashed it with https://github.com/alphagov/paas-cf/commit/785f6260eeb69770f3f63c30561509930a3b0da3 preserving comment.

concourse/scripts/update_monitor_options.rb
combor: This is corrected and new tests are implemented as well. I've moved a lot of stuff so it would be good to re-check the entire thing.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1507,53 +1507,90 @@ jobs:
...
+                ENABLE_DATADOG: {{enable_datadog}}
+              run:
+                path: sh
+                args:
+                  - -e
+                  - -c
+                  - |
+                    # FIXME: Remove this task when https://github.com/zorkian/go-datadog-api/issues/56 is fixed
+                    if [ "${ENABLE_DATADOG}" = "true" ]; then
+                      gem install dogapi -v 1.24.0
```
dcarley: Can you use Bundler here so that we get the same version of the gem that's used by the tests?

As mentioned in https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/84#issuecomment-276138891, we'll need to use groups to prevent all of the test dependencies being installed.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1507,53 +1507,90 @@ jobs:
...
-              file: updated-tfstate/datadog.tfstate
+        - do:
+          - task: datadog-terraform-apply
+            config:
+              platform: linux
+              image_resource:
+                type: docker-image
+                source:
+                  repository: governmentpaas/terraform
+                  tag: 161376949_TF_ruby
```
dcarley: I think this `tag` was left in by mistake.

combor:
> I think we should squash together all of the commits related to update_monitor_options; adding the script, tests, and using in the pipeline.

IMHO changes are much more readable now when commits are separate - especially when you read git diff/log to figure out why the change was made.  Putting everything in one commit will make it huge and not so easy to read. We could group update script, tests and pipeline changes in separate 3 commits. What do you think ?

mtekel:
I have re-based on master and fixed monitor tags to support TF 0.8.5. I have also added use of bundler groups and re-arranged commits. Hopefully things look more clear now.

dcarley:
> IMHO changes are much more readable now when commits are separate - especially when you read git diff/log to figure out why the change was made. Putting everything in one commit will make it huge and not so easy to read. We could group update script, tests and pipeline changes in separate 3 commits. What do you think ?

That's a reasonable compromise.

My concern was that they're not currently in usable chunks (e.g. the tests can't run without the Gemfile and the mock API results aren't relevant until the tests are added) and some of the refactoring would be better described as "we do this because" a single commit message (e.g. non-zero exit code for bad responses, which is also tangled with test refactoring).

combor:
I have made requested changes. We've got less commits now in a sane order ( not on github though).

dcarley:
Commits look good and it tests out OK. Thanks for including tests, btw.

One last question: will this alert the out-of-hours person if the smoke tests fail in CI or staging? If so, how would we restrict those to in-hours only?

mtekel:
There are 2 alerts that send notifications to pagerduty here:
1. conti smoke tests too slow or no_data
2. failures of conti smoke tests

2nd monitor has a env. specific variable that only enables notifications in prod. This works via controling creation of that monitor. That is, if notify, then monitor 2 gets created, if not then not. There is a tmp commit to enable this in dev so that you can test this PR. This opens a question, if we should handle the 1st monitor the same way as 2nd, that means not create it in dev/ci/staging - because we don't want to be notifified, or instead - do we want to be notified about that? 

Note that 1st monitor only sends no_data notifications to pagerdurty - and to a different service destination than 2. more important test. We have configured this 1st pagerduty service to set incidents to low during office hours and high otherwise - but all the config is in pagerduty.

We should decide now how we want to organise this. Perhaps document our alerting policies. Think if we want to have 2 different "services" for pagerduty, so that we can send alerts with different priority/configuration...

dcarley:
Ahhh, I see. Thanks for explaining. I think we would want to use the same mechanism to alert the in-hours support person about CI and staging because the dashboard alone isn't sufficient - but that's beyond the scope of this story and we can do it in follow-up work.

Could you write a story? (I'd offer but my list is already pretty deep from the incident)

mtekel:
If we want to alert for CI and Staging, we can easily do it in this PR. I'll create story to have better look/thinking about alerting policies (Which environments, which severities, what things to alert on etc.).

dcarley:
> If we want to alert for CI and Staging, we can easily do it in this PR.

Would it be only in-hours for those environments? I don't think we should do it now if it might result in someone being woken up about CI and staging.

> I'll create story to have better look/thinking about alerting policies (Which environments, which severities, what things to alert on etc.).

👍

mtekel:
Ok, so now it will only send alerts in prod. In CI/Staging it will be unchanged. We can decide what to do for those environments later.

dcarley:
Rebased to remove TMP commit and then rebased again to fix Travis build from #755.

PR #741, opened 2017-01-25
Bump bbs database max_open_connections to 40

alext:
## What

If this is too low we risk triggering a deadlock in bbs due to a bug in
bbs itself. We believe the bug is the on fixed in
https://github.com/cloudfoundry/bbs/commit/b33de3b07ed47077b6afbb3ad1a8331cee4c8b85.
At the time of writing, this has not yet been released.

## How to review

Code review.

## Who can review

Anyone but myself.


combor:
LGTM

PR #745, opened 2017-01-31
[docs] Change self-update instructions for DataDog in dev

dcarley:
## What

Replace the sentence, which says what to do if you have self-update
pipelines disabled but not what to do if you don't, with a command to
also disable self-update pipelines.

Because I copied the command without reading the preceding sentence, so
self-update pipeline overwrote my changes and I was left momentarily
confused.

## How to review

Text review only should be sufficient.

## Who can review

Not @dcarley

### DIFF
--- a/doc/datadog.md
+++ b/doc/datadog.md
@@ -21,9 +21,9 @@ make <ENV> upload-datadog-secrets
 make dev upload-datadog-secrets
 ```
 
-2. If you have auto-update pipelines disabled
+2. Deploy the pipeline with self-update disabled
 ```
-ENABLE_DATADOG=true make dev pipelines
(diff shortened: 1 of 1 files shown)


dcarley:
Travis build failed due to a problem in Travis:
```
The command "sudo -E apt-add-repository -y "ppa:git-core/ppa"" failed and exited with 1 during .
```

I've restarted the build and it now passes.

mtekel:
LG, would be also nice to fix numbering, because on github, all steps display as 1.

dcarley:
> LG, would be also nice to fix numbering, because on github, all steps display as 1.

Good point. Pushed a commit to fix it.

mtekel:
thanks, merging

PR #746, opened 2017-02-01
[#130660899] Add monitors for diego single-instance service locks and for etcd

keymon:
[#130660899 Monitor that single instance services are single instance (forever alone)](https://www.pivotaltracker.com/story/show/130660899)

What?
----

We want to monitor all the single-instance services from Diego as we do with route-emitter in #662.

We will add monitors only for the services: bbs, nsync, tps and auctioneer, for the lock flag in consul.

We also add monitor the etcd flag `cf.etc.IsLeader`.

We finally improve the query a little bit by using max/min depending if we are checkign the upper or lower value. See commits.

How to review?
-------------

Code review

You can deploy the changes only in CI by doing:

```
cd paas-cf/terraform/datadog
mkdir tmp
cd tmp

ln -s ../auctioneer.tf ../bbs.tf ../datadog.tf ../etcd.tf ../nsync.tf ../tps.tf ../variables.tf .
terraform apply -var aws_account=ci  -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var env=master

```

Remember to destroy later:

```
terraform destroy -var aws_account=ci  -var datadog_api_key=$(paas-pass datadog/dev/datadog_api_key) -var datadog_app_key=$(paas-pass datadog/dev/datadog_app_key) -var env=master
```

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/datadog/etcd.tf
+++ b/terraform/datadog/etcd.tf
@@ -1,0 +1,23 @@
+resource "datadog_monitor" "etcd_one_leader" {
+  name                = "${format("%s etcd IsLeader count is exactly one (%s)", var.env, element(list("upper", "lower"), count.index))}"
+  type                = "query alert"
+  message             = "There is not exactly one etcd reporting IsLeader."
+  escalation_message  = "There is still not exactly one etcd reporting IsLeader."
+  notify_no_data      = false
+  require_full_window = true
+
+  count = 2
+
+  query = "${
+    format(
+      "min(last_5m):sum:cf.etcd.IsLeader{deployment:%s}.fill(last, 60) %s",
+      var.env,
+      element(list("> 1", "< 1"), count.index)
+    )}"
+
+  thresholds {
+    critical = "1"
+  }
+
+  tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:etcd"]
+}
--- a/terraform/datadog/etcd.tf
+++ b/terraform/datadog/etcd.tf
@@ -10,7 +10,8 @@ resource "datadog_monitor" "etcd_one_leader" {
 
   query = "${
     format(
-      "min(last_5m):sum:cf.etcd.IsLeader{deployment:%s}.fill(last, 60) %s",
+      "%s(last_5m):sum:cf.etcd.IsLeader{deployment:%s}.fill(last, 60) %s",
+      element(list("max", "min"), count.index),
       var.env,
       element(list("> 1", "< 1"), count.index)
     )}"
--- a/terraform/datadog/auctioneer.tf
+++ b/terraform/datadog/auctioneer.tf
@@ -1,0 +1,23 @@
+resource "datadog_monitor" "auctioneer_lock_held_once" {
+  name                = "${format("%s auctioneer lock held exactly once (%s)", var.env, element(list("upper", "lower"), count.index))}"
+  type                = "query alert"
+  message             = "There is not exactly one auctioneer holding the lock."
+  escalation_message  = "There is still not exactly one auctioneer holding the lock."
+  notify_no_data      = false
+  require_full_window = true
+
+  count = 2
+
+  query = "${
+    format(
+      "min(last_5m):sum:cf.auctioneer.LockHeld.v1_locks_auctioneer_lock{deployment:%s}.fill(last, 60) %s",
+      var.env,
+      element(list("> 1", "< 1"), count.index)
+    )}"
+
+  thresholds {
+    critical = "1" # This value must match the threshold set in the query. Otherwise datadog API would fail obscurely
+  }
+
+  tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:auctioneer"]
+}
--- a/terraform/datadog/auctioneer.tf
+++ b/terraform/datadog/auctioneer.tf
@@ -10,7 +10,8 @@ resource "datadog_monitor" "auctioneer_lock_held_once" {
 
   query = "${
     format(
-      "min(last_5m):sum:cf.auctioneer.LockHeld.v1_locks_auctioneer_lock{deployment:%s}.fill(last, 60) %s",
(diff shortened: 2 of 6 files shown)


terraform/datadog/etcd.tf
```diff
@@ -0,0 +1,23 @@
...
+
+  query = "${
+    format(
+      "min(last_5m):sum:cf.etcd.IsLeader{deployment:%s}.fill(last, 60) %s",
+      var.env,
+      element(list("> 1", "< 1"), count.index)
+    )}"
+
+  thresholds {
+    critical = "${element(list("1", "1"), count.index)}"
```
mtekel: why not simply 1?

mtekel: Same for all other monitors...

keymon: Because this way is obvious how to define an lower/upper limit and that it needs to be updated. In this case the upper and lower limits are the same, but I think keeping it will ease reuse this pattern later. 

It also makes obvious that this critical value is related to the query expression.

BTW, if the `critical` value mismatches the value in the query for `<` or `>`, the datadog api call fails in a really obscure way.

mtekel: What's really not obvious is why you want to do it this way for a thing that needs to monitor _exactly_ 1. That is, any other threshold for these monitors is not OK. I though it was some TF limitation. We can use the pattern when it makes sense - when there are different items/values for different monitors (like `> 1`  and `< 1 `).

dcarley: I also found it quite confusing to read. Can you change it to a single value please?

keymon: OK, I changed it and I put a warning.

But funny thing, we are already thinking to set different limits, as using "min" might be too sensitive. as commented in slack:

https://govuk.slack.com/archives/the-government-paas/p1486042172009010
> During a rolling deployment of CF, the metric might be reported twice or we might miss metrics, so the sum across the members might get 2 or 0 for one value.

> Using min, the monitor is sensible to restarts and can alert for 5m during a rolling deployment. 

> one way to avoid this could be use "avg" instead of "min" and a lower/upper limit of 0.80 and 1.20, for instance. As the metric is reported every 1 minute, the average in 5 minutes would be within 0.9-1.10

keymon: e.g:
```
diff --git a/terraform/datadog/tps.tf b/terraform/datadog/tps.tf
index eef05b6..6eaefb8 100644
--- a/terraform/datadog/tps.tf
+++ b/terraform/datadog/tps.tf
@@ -10,13 +10,13 @@ resource "datadog_monitor" "tps_watcher_lock_held_once" {
 
   query = "${
     format(
-      "min(last_5m):sum:cf.tps_watcher.LockHeld.v1_locks_tps_watcher_lock{deployment:%s}.fill(last, 60) %s",
+      "avg(last_5m):sum:cf.tps_watcher.LockHeld.v1_locks_tps_watcher_lock{deployment:%s}.fill(last, 60) %s",
       var.env,
-      element(list("> 1", "< 1"), count.index)
+      element(list("> 0.8", "< 1.2"), count.index)
     )}"
 
   thresholds {
-    critical = "1" # This value must match the threshold set in the query. Otherwise datadog API would fail obscurely
+    critical = "${element(list("0.8", "1.2"), count.index)}"
   }
 
   tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:tps_watcher"]
```

keymon: Or maybe the solution is use `last(zero)` :-m

```diff
diff --git a/terraform/datadog/tps.tf b/terraform/datadog/tps.tf
index eef05b6..66450ae 100644
--- a/terraform/datadog/tps.tf
+++ b/terraform/datadog/tps.tf
@@ -10,7 +10,7 @@ resource "datadog_monitor" "tps_watcher_lock_held_once" {
 
   query = "${
     format(
-      "min(last_5m):sum:cf.tps_watcher.LockHeld.v1_locks_tps_watcher_lock{deployment:%s}.fill(last, 60) %s",
+      "min(last_5m):sum:cf.tps_watcher.LockHeld.v1_locks_tps_watcher_lock{deployment:%s}.fill(zero, 60) %s",
       var.env,
       element(list("> 1", "< 1"), count.index)
     )}"
```

mtekel:
maybe you should also use min for upper and max for lower?

keymon:
> maybe you should also use min for upper and max for lower?

+1, I will do so

dcarley:
Out of interest, in order to make sure the values are the same in the query and threshold, could you store them in an intermediate variable? e.g.

```
variable lock_threshold {
  default = "1"
}

…
  query = "${
    format(
      "min(last_5m):sum:cf.auctioneer.LockHeld.v1_locks_auctioneer_lock{deployment:%s}.fill(last, 60) %s",
      var.env,
      element(list("> ${var.lock_threshold}", "< ${var.lock_threshold}"), count.index)
    )}"

  thresholds {
    critical = "${var.lock_threshold}"
  }
…
```

mtekel:
LG now. I'll check the tests once again (with min, max being created this time) and merge then. Hopefully we can fix travis build by that time...

mtekel:
There seems to be some problem with DataDog. The intention with min/max was to create robust alerts:
* take max for the lower check to ensure that less than 1 locks held is not a random glitch situation, but something that is persistent
* take min for the upper check to ensure that more than 1 locks held is a persistent situation

Nevertheless, this doesn't work so well, as data stops being sent when you have 0 locks held (both instances of component not active). The fill function in DD doesn't seem to work correctly:

image

We should raise this with DD. Until then, we can swap min/max criteria for the upper and lower checks. This will alert on every glitch, e.g. during deploy, but I guess that's better than no alerts...

keymon:
@mtekel I changed it as you suggested.

mtekel:
This works now. I will re-base on master to get travis fix and then merge.

keymon:
Thx!! 👍

PR #747, opened 2017-02-01
[#135948443] Fix name of java buildpack.

alext:
## What

The package was renamed in CF v246[1]. This omission wasn't an obvious
problem for existing deployments because the buildpack had already been
uploaded, however for a clean deployment the java buildpack would be
missing. For existing deployments it means that the buildpack wouldn't have
been updated from the previous version.

[1]https://github.com/cloudfoundry/cf-release/commit/72222a8

## How to review

Do a fresh deploy from master and verify that the java buildpack is present.

## Who can review

Anyone but myself or @combor

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -316,7 +316,7 @@ properties:
       - name: staticfile_buildpack
         package: staticfile-buildpack
       - name: java_buildpack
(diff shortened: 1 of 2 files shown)


keymon:
The change makes sense and the test ensures that it works. I will merge.

dcarley:
👍  for test.

> For existing deployments it means that the buildpack wouldn't have been updated from the previous version.

Should we warn tenants about this?

PR #749, opened 2017-02-01
[#138486833] CF v251 upgrade

alext:
## What

Upgrade to CloudFoundry v251 and Diego 1.5.3

This includes fixes for how we pin the version of cf-release in the pipeline, which is currently broken for new deployments on master. See the "Pin cf-release version in tag_filter" commit for details.

## How to review

Code review - may be easiest to review commit-by-commit.

For this sort of upgrade we should test both an upgrade to an existing deployment, and a clean deploy.

## Who can review

Anyone but myself or @combor.


keymon:
> Configure SSL for UAA internal endpoint.
>  Enabling SSL on UAA also give us the option to point the ELB at the endpoint in future to encrypt the ELB -> UAA traffic.

Is there going to be a separate PR?

alext:
I think that's out of scope for this story.  I'll add it as a follow-up in pivotal.

keymon:
In v247 they added a serial: true to serialise the update of the traffic controller. We missed that in the v250 upgrade, and @alext suggested that could be the cause why enable TLS broke the traffic controller communication to doppler. 

Was added in this commit:

https://github.com/cloudfoundry/cf-release/commit/96727123#diff-f4669efacfe0e8bebab22559af6edea4R865

Do we need it?

alext:
Note: while testing this, I had an app deployed that was scaled to 150 instances, and the upgrade completed without issue.

keymon:
I tested scaling to 200, then restarting several times. Works.

PR #750, opened 2017-02-01
[#138486833] Revert "TMP - test terraform update"

keymon:
This reverts commit d7501a5d42c67562582962f930f12089c6da7f8a. added in #744

Strangely the commit was included because github did not detect the
changes from the `git push --force` and the `make merge_pr` got the
previous commit from the API.

We noticed that in https://status.github.com/ mean hook was too high
in that moment, 7.8s

We should wait before running merge_pr after a git push -f


alext:
:+1: Looks good to me as soon as Travis is green.

PR #752, opened 2017-02-01
[#136435555] Add user to mailing list notification

combor:
## What

As a part of improve incident communications story we've decided to
notify operator that every user should be a member of GOV.UK PaaS
Announce. As a first step before implementing automation for this task
we will would like to add notification that will be displayed after user
account was successfully created.

Also it includes a small fix to correctly detect if user account already exists. 

## How to review

- run it on your dev env and check if it notification is displayed and makes sense 

## Who can review

Not @combor

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -139,7 +139,7 @@ create_user() {
 
   if cf create-user "${EMAIL}" "${PASSWORD}" 2>&1 | tee "${TMP_OUTPUT}"; then
     if ! grep -q "already exists" "${TMP_OUTPUT}"; then
-      SEND_EMAIL=true
+      USER_CREATED=true
     fi
   else
     abort "Error creating user ${EMAIL}"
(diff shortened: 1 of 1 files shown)


mtekel:
Works fine. I have noticed other issue though: If you do --no-email and the user already exists, the script prints a message about new password being set. But it is not set, because user already exists...

mtekel:
Looks like we should check if RESET_USER is true here: https://github.com/alphagov/paas-cf/blob/a8861510f746f79c2603723eebbc7dbdce781f3c/scripts/create-user.sh#L205

mtekel:
```
@@ -202,7 +202,7 @@ print_password() {
 }
 
 emit_password() {
-  if [ "${NO_EMAIL:-}" = "true" ]; then
+  if [ "${NO_EMAIL:-}" = "true" ] && [ "${RESET_USER:-}" = "true" ]; then
     print_password
   else
     send_mail
```

alext:
@mtekel I disagree with your fix. I think a better approach would be to move the `SEND_EMAIL=true` check ([here](https://github.com/alphagov/paas-cf/blob/a8861510f746f79c2603723eebbc7dbdce781f3c/scripts/create-user.sh#L170)), up to the `emit_password` function. That way the same logic is applied no matter what method is used to emit the password.

mtekel:
How about this?
```
@@ -167,34 +167,30 @@ get_body() {
 }
 
 send_mail() {
-  if [[ "${SEND_EMAIL}" == "true" ]]; then
-    MESSAGE_JSON="
-    {
-      \"Subject\": {
-        \"Data\": \"$(get_subject)\",
+  MESSAGE_JSON="
+  {
+    \"Subject\": {
+      \"Data\": \"$(get_subject)\",
+      \"Charset\": \"utf8\"
+    },
+    \"Body\": {
+      \"Text\": {
+        \"Data\": \"$(get_body)\",
         \"Charset\": \"utf8\"
-      },
-      \"Body\": {
-        \"Text\": {
-          \"Data\": \"$(get_body)\",
-          \"Charset\": \"utf8\"
-        }
       }
     }
-    "
-
-    aws ses send-email \
-      --destination "{ \"ToAddresses\": [\"${EMAIL}\"] }" \
-      --message "${MESSAGE_JSON}"\
-      --from "${FROM_ADDRESS}"  \
-      --region eu-west-1 \
-      --output text > /dev/null
-
-    echo "An email has been sent to ${EMAIL} with their new credentials."
-    show_notification
-  else
-    echo "User was already present and has not been recreated. No mail sent."
-  fi
+  }
+  "
+
+  aws ses send-email \
+    --destination "{ \"ToAddresses\": [\"${EMAIL}\"] }" \
+    --message "${MESSAGE_JSON}"\
+    --from "${FROM_ADDRESS}"  \
+    --region eu-west-1 \
+    --output text > /dev/null
+
+  echo "An email has been sent to ${EMAIL} with their new credentials."
+  show_notification
 }
 
 print_password() {
@@ -202,7 +198,9 @@ print_password() {
 }
 
 emit_password() {
-  if [ "${NO_EMAIL:-}" = "true" ]; then
+  if [[ "${SEND_EMAIL}" == "false" ]]; then
+    echo "User was already present and password was not reset. Use -r to re-create users, which forces new password."
+  elif [[ "${RESET_USER:-}" == "true" || "${NO_EMAIL:-}" == "true" ]]; then
     print_password
   else
     send_mail
```

combor:
Thanks for the commit but now someone else has to review it.

dcarley:
> Thanks for the commit but now someone else has to review it.

It's sufficient for you to review and test the additional commit and for @mtekel to merge.

mtekel:
So @combor are you fine with the commit?

combor:
LGTM

PR #753, opened 2017-02-02
[#136353673] Add ability to create CDNs programmatically

paroxp:
## What

We would like to create a CDN for the PaaS Documentation and the Product Page and be able to access it, from the new subdomain.

This change, should allow any user with an access, create CloudFront Distributions. The CNAME record, will be created as part of the system domain name, for each distribution.

## How to review

1. Run `create-cloudfoundry` pipeline from this branch
1. Run `ACTION=plan make dev setup_cdn_instances`
1. Experience no errors from the above
1. Run `ACTION=apply make dev setup_cdn_instances`
1. Experience no errors from the above
1. Wait... This may take a while. I suggest logging into AWS Console and monitoring CloudFront Distributions
1. Visiting `docs.${DEPLOY_ENV}.dev.cloudpipeline.digital` should result in `paas-tech-docs`
1. Visiting `www.${DEPLOY_ENV}.dev.cloudpipeline.digital` should result in `paas-product-page`

## Who can review

Not @paroxp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint_yaml:
 lint_terraform: dev
 	$(eval export TF_VAR_system_dns_zone_name=$SYSTEM_DNS_ZONE_NAME)
 	$(eval export TF_VAR_apps_dns_zone_name=$APPS_DNS_ZONE_NAME)
-	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t terraform graph > /dev/null
+	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t sh -c 'terraform get $$1 && terraform graph $$1' -- > /dev/null
 	@if [ "$$(terraform fmt -write=false terraform)" != "" ] ; then \
 		echo "Use 'terraform fmt' to fix HCL formatting:"; \
 		terraform fmt -write=false -diff=true terraform ; \
@@ -197,6 +197,11 @@ pingdom: check-tf-version ## Use custom Terraform provider to set up Pingdom che
 	$(eval export PASSWORD_STORE_DIR?=~/.paas-pass)
 	$(eval export PINGDOM_CONTACT_IDS=11089310)
 	@terraform/scripts/set-up-pingdom.sh ${ACTION}
+
+.PHONY: setup_cdn_instances
+setup_cdn_instances: check-tf-version check-aws-credentials ## Setup the CloudFront Distribution instances, by reading their config from terraform/cloudfront/instances.tf.
+	$(if ${ACTION},,$(error Must pass ACTION=<plan|apply|...>))
+	@terraform/scripts/set-up-cdn-instances.sh ${ACTION}
 
 merge_pr: ## Merge a PR. Must specify number in a PR=<number> form.
 	$(if ${PR},,$(error Must pass PR=<number>))
--- a/Makefile
+++ b/Makefile
@@ -43,15 +43,11 @@ spec:
 lint_yaml:
 	find . -name '*.yml' -not -path '*/vendor/*' | xargs $(YAMLLINT) -c yamllint.yml
 
-lint_terraform: dev
+.PHONY: lint_terraform
+lint_terraform: dev ## Lint the terraform files.
 	$(eval export TF_VAR_system_dns_zone_name=$SYSTEM_DNS_ZONE_NAME)
 	$(eval export TF_VAR_apps_dns_zone_name=$APPS_DNS_ZONE_NAME)
-	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t sh -c 'terraform get $$1 && terraform graph $$1' -- > /dev/null
-	@if [ "$$(terraform fmt -write=false terraform)" != "" ] ; then \
-		echo "Use 'terraform fmt' to fix HCL formatting:"; \
-		terraform fmt -write=false -diff=true terraform ; \
-		exit 1; \
-	fi
+	@terraform/scripts/lint.sh
 
 lint_shellcheck:
 	find . -name '*.sh' -not -path '*/vendor/*' | xargs $(SHELLCHECK)
--- a/terraform/cf-certs/upload.tf
+++ b/terraform/cf-certs/upload.tf
@@ -7,6 +7,7 @@ resource "aws_iam_server_certificate" "system" {
   certificate_body  = "${var.system_domain_crt}"
   private_key       = "${var.system_domain_key}"
   certificate_chain = "${var.system_domain_intermediate_crt}"
+  path              = "/cloudfront/${var.env}-system-domain/"
 
   lifecycle {
     create_before_destroy = true
@@ -18,6 +19,7 @@ resource "aws_iam_server_certificate" "apps" {
   certificate_body  = "${var.apps_domain_crt}"
   private_key       = "${var.apps_domain_key}"
   certificate_chain = "${var.apps_domain_intermediate_crt}"
+  path              = "/cloudfront/${var.env}-apps-domain/"
 
   lifecycle {
     create_before_destroy = true
--- a/terraform/cloudfront/cloudfront_distribution/distribution.tf
+++ b/terraform/cloudfront/cloudfront_distribution/distribution.tf
@@ -45,6 +45,8 @@ resource "aws_cloudfront_distribution" "cdn_instance" {
     max_ttl                = 86400
   }
 
+  price_class = "PriceClass_100"
+
   restrictions {
     geo_restriction {
       restriction_type = "none"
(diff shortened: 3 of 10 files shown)


terraform/cloudfront/cloudfront_distribution/distribution.tf
```diff
@@ -0,0 +1,61 @@
...
+      restriction_type = "none"
+    }
+  }
+
+  tags {
+    Environment = "${var.env}"
+  }
+
+  viewer_certificate {
+    cloudfront_default_certificate = true
```
dcarley: Do we need to use the wildcard certs that are uploaded by [`make <env> manually_upload_certs`](https://github.com/alphagov/paas-cf/blob/3e3ba8b18d2c9ccb6928713ad4e9e0859b0e92e5/Makefile#L185-L192) in order for the client browser to get a valid cert when you visit `www.cloud.service.gov.uk` or `docs.cloud.service.gov.uk`?

Not sure how you'd test it in dev, but I think we should mention it in the review section of the PR.

terraform/cloudfront/cloudfront_distribution/distribution.tf
```diff
@@ -0,0 +1,61 @@
...
+resource "aws_cloudfront_distribution" "cdn_instance" {
+  origin {
+    domain_name = "${var.origin}"
+    origin_id   = "${var.origin}"
+
+    custom_origin_config {
+      origin_protocol_policy = "https-only"
+      http_port              = 80
+      https_port             = 443
+      origin_ssl_protocols   = ["TLSv1", "TLSv1.1", "TLSv1.2"]
```
dcarley: Not a blocker, but out of curiosity, I wonder if we need to support TLSv1 and TLSv1.1

There was some discussion about it recently, do you know the outcome @bleach @davbo @davidillsley ?

Sidenote: our ELBs do because the default AWS policies do: https://github.com/alphagov/paas-cf/blob/3e3ba8b18d2c9ccb6928713ad4e9e0859b0e92e5/terraform/globals.tf#L125

terraform/cloudfront/cloudfront_distribution/distribution.tf
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  aliases         = "${var.aliases}"
+  comment         = "${var.comment}"
+  enabled         = true
+  is_ipv6_enabled = true
```
dcarley: The future! 🐳

terraform/cloudfront/cloudfront_distribution/distribution.tf
davbo: So far we are prioritising work to measure the impact of dropping TLSv1[.1] from GOV.UK. I'm not sure where that has got to yet though.

terraform/cloudfront/cloudfront_distribution/distribution.tf
paroxp: Good point. I had a wrong assumption, it would just apply straight away. 

Pushed new commit. Thanks.

Makefile
```diff
@@ -46,7 +46,7 @@ lint_yaml:
 lint_terraform: dev
 	$(eval export TF_VAR_system_dns_zone_name=$SYSTEM_DNS_ZONE_NAME)
 	$(eval export TF_VAR_apps_dns_zone_name=$APPS_DNS_ZONE_NAME)
-	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t terraform graph > /dev/null
+	find terraform -mindepth 1 -maxdepth 1 -type d -not -path 'terraform/providers' -not -path 'terraform/scripts' -print0 | xargs -0 -n 1 -t sh -c 'terraform get $$1 && terraform graph $$1' -- > /dev/null
```
alext: This addition of `terraform get` causes the creation of a `.terraform` directory in the repo root when running it locally.

Also, this is getting quite long, and not particularly easy to read. I'd suggest extracting the body of this task into a script. That script could then use a temporary working dir to avoid polluting the repo (in a similar manner to some of the other scripts in `terraform/scripts`).

terraform/cloudfront/cloudfront_distribution/distribution.tf
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  tags {
+    Environment = "${var.env}"
+  }
+
+  viewer_certificate {
+    acm_certificate_arn            = "${var.system_domain_cert_arn}"
+    cloudfront_default_certificate = false
+  }
```
alext: I think this is missing a `ssl_support_method` param, which is required when specifying a cert ([see docs](https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#ssl_support_method)). This need to be set to one of `vip` or `sni-only`. I think we can probably use `sni-only`.

At present, I get the following error:
```
2 error(s) occurred:

* aws_cloudfront_distribution.cdn_instance: MalformedXML: 1 validation error detected: Value '' at 'distributionConfigWithTags.distributionConfig.viewerCertificate.sSLSupportMethod' failed to satisfy constraint: Member must satisfy enum value set: [sni-only, vip]
        status code: 400, request id: be2f4fc6-ea00-11e6-b6c6-29d0a4c8b7f3
* aws_cloudfront_distribution.cdn_instance: MalformedXML: 1 validation error detected: Value '' at 'distributionConfigWithTags.distributionConfig.viewerCertificate.sSLSupportMethod' failed to satisfy constraint: Member must satisfy enum value set: [sni-only, vip]
        status code: 400, request id: be2f9e4f-ea00-11e6-b089-c378731fc466
```

terraform/cloudfront/cloudfront_distribution/distribution.tf
alext: Looking more closely at this, I think this is only specifying the policies for connecting to origin, not for end-user requests. End-user SSL details are configured in the `viewer_certificate` section lower down in this file.

Makefile
paroxp: This has now been extracted into a separate file.

I would like to point out, that this command will fail locally, as it did before with two `provider pingdom couldn't be found` errors.

terraform/cloudfront/cloudfront_distribution/variables.tf
```diff
@@ -0,0 +1,17 @@
...
+
+variable "name" {}
+
+variable "origin" {}
+
+variable "system_dns_zone_id" {}
+
+variable "system_dns_zone_name" {}
+
+variable "system_domain_cert_arn" {}
```
paroxp: Sure! I don't think we've been doing that in many of the other files, but you're right.

terraform/scripts/set-up-cdn-instances.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -eu
+TERRAFORM_ACTION=${1}
+STATEFILE=cloudfront-distribution-${MAKEFILE_ENV_TARGET}.tfstate
```
paroxp: Good point. Thanks.

terraform/scripts/set-up-cdn-instances.sh
```diff
@@ -0,0 +1,40 @@
...
+terraform remote config \
+  -backend=s3 \
+  -backend-config="bucket=gds-paas-${DEPLOY_ENV}-state" \
+  -backend-config="key=${STATEFILE}" \
+  -backend-config="region=${AWS_DEFAULT_REGION}"
+
+# Run the terraform action on the instances.
+terraform "${TERRAFORM_ACTION}" \
+  -var-file="${PAAS_CF_DIR}/terraform/${AWS_ACCOUNT}.tfvars" \
+  -var "env=${MAKEFILE_ENV_TARGET}" \
```
paroxp: Agree. It did confuse me a little when I've looked at it in the Pingdom script.

Makefile
```diff
@@ -198,6 +198,11 @@ pingdom: check-tf-version ## Use custom Terraform provider to set up Pingdom che
 	$(eval export PINGDOM_CONTACT_IDS=11089310)
 	@terraform/scripts/set-up-pingdom.sh ${ACTION}
 
+.PHONY: setup_cdn_instances
+setup_cdn_instances: check-tf-version check-aws-credentials
```
paroxp: Sorry, added.

terraform/cloudfront/cloudfront_distribution/distribution.tf
paroxp: Sorry, I did have it in, however must have lost it in rebasing process.

terraform/cf-certs/upload.tf
```diff
@@ -7,6 +7,7 @@ resource "aws_iam_server_certificate" "system" {
   certificate_body  = "${var.system_domain_crt}"
   private_key       = "${var.system_domain_key}"
   certificate_chain = "${var.system_domain_intermediate_crt}"
+  path              = "/cloudfront/${var.env}-system-domain"
```
paroxp: As suggested in the IRL conversation, I have created a new commit in this PR. Please, let me know, if that still prevents the pipeline from progression.

alext:
This looks good.

While reviewing this, I ran into a [terraform bug](https://github.com/hashicorp/terraform/issues/11716) that means that the cf-certs terraform needs to be run twice before the CDN can be deployed.

alext:
I'd also note that this will need some manually deploying in staging and prod due because the cert properties need to be updated.

PR #756, opened 2017-02-03
[SUPPORT] Change parser VM type to medium

combor:
## What
Due to huge volume of logs to parse we went out of cpu credits on
parser VMs. This slowed down logstash considerably and resulted in
almost no log ingestion to elasticsearch.  To avoid such situation in
the future we would like to use a VM type that is not burstable.

medium type is EC2 m3.medium VM. 

## How to review

Deploy and check if you are getting logs in Elasticsearch ( by checking in Kibana )

## Who can review

Not @combor


dcarley:
Per conversation in Slack: merged without testing in dev because we understand the change relatively well and it doesn't have any tenant impact.

PR #757, opened 2017-02-03
Fix generate-git-keys task

mtekel:
## What

Fix generate-git-keys task

ssh-keygen is present in the git-ssh container. We use that in other tasks
that also need ssh-keygen. Use this image instead of awscli, which was
missing ssh-keygen.

## How to review

Task generate-git-keys should work now. This is in 'operator' tab. Without this PR, it fails with `sh: ssh-keygen: not found`

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2267,7 +2267,7 @@ jobs:
           image_resource:
             type: docker-image
             source:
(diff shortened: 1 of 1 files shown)


dcarley:
Works. It was originally using `governmentpaas/git-ssh` in d255f02f975d5f9d853931c0fa386c0ce49ccc9c. Not sure why or when it changed, but I don't think it matters.

PR #759, opened 2017-02-06
Update tf version check for 0.8.5

alext:
## What

We've upgraded Terraform in the pipelines, so we should ensure we're
using the same version when running the local tasks for consistency.

## How to review

Try to run one of the local terraform tasks with an older terraform version and verify it errors. eg:

`ACTION=plan make dev pingdom`

## Who can review

Anyone but myself.


mtekel:
works, merging

PR #760, opened 2017-02-06
[#137700569] Set up Pingdom alerts to PagerDuty

mtekel:
## What

[Set up Pingdom alerts to PagerDuty](https://www.pivotaltracker.com/n/projects/1275640/stories/137700569)

New contacts in pingdom have been created which point to new services created in pagerduty for pingdom alerts - one for in-hours notifications only and one for 24/7. This PR configures pagerduty notifications for CI/Staging (in-hours) and prod (24/7).

## How to review

Locally edit dev tfvars file. Copy pingdom contact ids from ci/staging tfvars files. This will enable in-hours pager duty notifications for dev when you create pingdom checks.

Override in-hours schedule to your name. `ACTION=apply make dev pingdom`. Stop your healthcheck app. Wait. Get notified by pagerduty - the way you configured your notification options in PD. 

Don't forget to clean up the override in PD (it's in the right column of schedules interface) and testing pingdom checks: `ACTION=destroy make dev pingdom`

## Who can review

not @mtekel or @paroxp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,6 @@ manually_upload_certs: check-tf-version ## Manually upload to AWS the SSL certif
 pingdom: check-tf-version ## Use custom Terraform provider to set up Pingdom check
 	$(if ${ACTION},,$(error Must pass ACTION=<plan|apply|...>))
 	$(eval export PASSWORD_STORE_DIR?=~/.paas-pass)
-	$(eval export PINGDOM_CONTACT_IDS=11089310)
 	@terraform/scripts/set-up-pingdom.sh ${ACTION}
 
 .PHONY: setup_cdn_instances
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,6 @@ globals:
 
 .PHONY: dev
 dev: globals check-env-vars ## Set Environment to DEV
-	$(eval export MAKEFILE_ENV_TARGET=dev)
 	$(eval export AWS_ACCOUNT=dev)
 	$(eval export ENABLE_DESTROY=true)
 	$(eval export ENABLE_AUTODELETE=true)
@@ -102,7 +101,6 @@ dev: globals check-env-vars ## Set Environment to DEV
 
 .PHONY: ci
 ci: globals check-env-vars ## Set Environment to CI
-	$(eval export MAKEFILE_ENV_TARGET=ci)
 	$(eval export AWS_ACCOUNT=ci)
 	$(eval export ENABLE_AUTO_DEPLOY=true)
 	$(eval export TAG_PREFIX=staging-)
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -191,7 +191,7 @@ manually_upload_certs: check-tf-version ## Manually upload to AWS the SSL certif
 pingdom: check-tf-version ## Use custom Terraform provider to set up Pingdom check
 	$(if ${ACTION},,$(error Must pass ACTION=<plan|apply|...>))
 	$(eval export PASSWORD_STORE_DIR?=~/.paas-pass)
-	$(eval export PINGDOM_CONTACT_IDS=11089310)
+	$(eval export PINGDOM_CONTACT_IDS=11089310,11189971)
```
dcarley: @saliceti is planning to reconfigure the PagerDuty services to the 4 proposed by @mtekel on the whiteboard this morning. Should we take this opportunity now to use separate contact IDs (and thus escalation policies) for CI and staging?

saliceti: 11189971 is the id for 24x7 notification
I've also created 11190300 for in hours notifications
Could you add logic to have different ids for different environments?

mtekel:
Updated. This now also needs testing that correct (different) targets are alerted in ci/staging and prod.

terraform/scripts/set-up-pingdom.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 TERRAFORM_ACTION=${1}
 VERSION=0.2.2
 BINARY=terraform-provider-pingdom-tf-0.8.5-$(uname -s)-$(uname -m)
-STATEFILE=pingdom-${MAKEFILE_ENV_TARGET}.tfstate
```
dcarley: Context for why we originally used `MAKEFILE_ENV_TARGET` is described in b8158b24b29ef93db67c6bcd5872434a4b8996e1. The use in this script contradicts that though 😸 

So I think it's fine to remove it. We have two places where configuration can change depending on the environment - `Makefile` or Terraform `tfvars` (as used in this script).

combor:
Just a quick update. This is in progress but I've got problems with my env to it may take a little bit longer.

terraform/ci.tfvars
```diff
@@ -10,3 +10,4 @@ support_email="govpaas-alerting-ci@digital.cabinet-office.gov.uk"
 # Enabled/disabled resources
 # Disable datadog_monitor.total_routes_drop resource
 datadog_monitor_total_routes_drop_enabled = 0
+pingdom_contact_ids=11089310,11190300
```
combor: After I copy this to `dev.tfvars` my terraform (v0.8.5) complains about:
```
invalid value "/go/src/github.com/alphagov/paas-cf/terraform/dev.tfvars" for flag -var-file: Error parsing /go/src/github.com/alphagov/paas-cf/terraform/dev.tfvars: At 9:30: expecte
d: IDENT | STRING | ASSIGN | LBRACE got: NUMBER
```

alext: It probably needs to be quoted.

Having said that, this looks like a good opportunity to update this to use an array variable.

mtekel:
Changed this to use list variables now.

combor:
Tested this and it works. LGTM.

Merging.

PR #761, opened 2017-02-06
[WIP] Cache ruby and python deps in travis builds

Jonty:
This PR is to discuss/poke [caching dependencies in travis](https://docs.travis-ci.com/user/caching/) builds to both speed them up and make them more reliable.

Currently it (might):
* Cache python dependencies using travis' built in support for pip
* Cache ruby dependencies using travis' built in support for bundler

It could also:
* Cache things installed in the ~/bin directory if we added some simple checks to the `before_install` sections: https://github.com/alphagov/paas-cf/blob/master/.travis.yml#L19
* Cache all the things built in the `GOPATH`

Thoughts? Criticisms? Prophecies?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ addons:
 cache:
   bundler: true
   pip: true
-  apt: true
   directories:
     - $HOME/.cache/pip
     - $HOME/.rvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ cache:
   directories:
     - $HOME/.cache/pip
     - $HOME/.rvm
+    - $HOME/gopath
+    - $HOME/vendor/bundle
 
 before_install:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ cache:
   apt: true
   directories:
     - $HOME/.cache/pip
+    - $HOME/.rvm
 
 before_install:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ cache:
   bundler: true
   pip: true
   apt: true
+  directories:
+    - $HOME/.cache/pip
 
 before_install:
   - |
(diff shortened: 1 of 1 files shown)


dcarley:
Can you provide some information about what difference this has made to the build time? That would help us gauge whether it's worth doing.

Are we affected by travis-ci/travis-ci#2518 because we also install gems from `scripts/Gemfile`?

> Cache things installed in the ~/bin directory if we added some simple checks to the before_install sections: https://github.com/alphagov/paas-cf/blob/master/.travis.yml#L19

I don't think it would help. Travis docs say:

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source:

mtekel:
I would like to mention that I had to wait (what felt like unnecessarily too long) for tests to pass for https://github.com/alphagov/paas-cf/pull/759. If caching would have made that faster then it's a welcome improvement.

Jonty:
I'll run some tests. One thing to note is that installing git takes quite a long time, mostly due to the ppa adding process. I'm not sure what we're using git for other than installing gomega/ginkgo, so it might be better to pull in binary versions of them?

dcarley:
> I'm not sure what we're using git for other than installing gomega/ginkgo

We pull a later version to test `tag-release.sh` (ref: 00f84122e3a797fdbe829ce56692d897e24a69ef)

> so it might be better to pull in binary versions of them?

Is that possible? I don't think it's distributed as a static OS-agnostic binary.

alext:
> I'm not sure what we're using git for

00f8412 Suggests that we need a newer git version for the `tag-releases.sh` tests.

dcarley:
Changed the subject to "WIP" so that it get collapsed on our PR dashboard while we're not actively working on it.

Jonty:
I am concluding that no matter how much I poke this I can't make Travis do what it should. The savings I've achieved are not worth the extra complexity.

Closing.

PR #764, opened 2017-02-07
[#136353673] Reduce the default TTL for CloudFront instances

paroxp:
## What

We have decided, it may be unreasonable to have a 60 minute delay, before CloudFront is going to check for the changes.

This has been justified with the scenario: During user testing lab, we may want to quickly run an update on the docs and expect it to appear fairly quickly online, for the users to look up.

This has been changed to 5s as agreed with Ben, @dcarley, @alext and @mtekel. It may be worth, keeping an eye on this change.

## How to review

Code review, should be sufficient test.

After the change is applied, the manual `ACTION=apply make prod setup_cdn_instances` will need to be ran.

## Who can review

Anyone but @paroxp.


alext:
:+1: Looks good to merge as soon as Travis is green.

PR #769, opened 2017-02-09
Prevent failures by pinning autodelete to master

Jonty:
## What

Autodelete will often fail if the environment was deployed from a branch that has been merged, meaning the remote branch no longer exists. This pins the autodelete branch to master to avoid this.

90% of the time the branch is now part of master and the operations carried out are quite simple, so this should cause no problems.

## How to review

Code review.

## Who can review

Anyone!

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -15,7 +15,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-cf.git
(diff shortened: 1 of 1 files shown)


dcarley:
LGTM. Reviewed code only.

alext:
> 90% of the time the branch is now part of master

That's an interesting statistic, do you mind sharing how you arrived at it? :trollface:

Jonty:
@alext Sure, here's a direct quote:
> <@alext> I reckon that’s true 80-90% of the time.

PR #765, opened 2017-02-08
[#135953741] Datadog pagerduty integration

saliceti:
## What
Stories:

* [Configure schedules and escalation policies for PagerDuty](https://www.pivotaltracker.com/story/show/135953741)
* [Send smoke test and Pingdom alerts from CI and staging to the in-hours support person using PagerDuty](https://www.pivotaltracker.com/story/show/139005729)

Create explicit variables for alerting in and out of hours. They can be set differently for each environment. We currently set email notification for dev, in hours pagerduty only in ci and staging, in hours and out of hours pagerduty in production.

## How to review
* There is a "[TEMP]" commit to create datadog monitors even if you don't have datadog enabled
* Apply to your environment, the monitors should be created
* Change values in `dev.tfvars`, the notifications should change
* You can double check the pagerduty service notifications correspond to real services configured in the integration

## Before merging
* Remove the "[TEMP]" commit

## Who can review
Not me

PR #770, opened 2017-02-09
[#135953741] Fix terraform destroy bug

saliceti:
## What

Story: [Configure schedules and escalation policies for PagerDuty](https://www.pivotaltracker.com/story/show/135953741)

In https://github.com/alphagov/paas-cf/pull/765 new variables were added to env.tfvars files and used in datadog terraform. The destroy task wasn't reading the env.tfvars files so the task was hanging.
The env.tfvars are now read in the destroy task.

## How to review
Run the destroy-cloudfoundry pipeline, it shouldn't hang at task datadog-TF-destroy
Should you want to keep your environment, you can comment the delete-deployment and pause the the terraform-desrtoy job.

## Who can review
Not me


Jonty:
Works as described.

PR #773, opened 2017-02-10
Increase assertion timeout

combor:
## What

We noticed that Travis can be pretty slow to run tests so we have to
increase session timeout to avoid occasional failures.

## How to review

Review code changes

## Who can review

Not @combor


alext:
:+1: Looks good. Will merge when Travis is green.

PR #776, opened 2017-02-10
[#134350583] Fix RDS Broker and syslog trusted advisor warnings

paroxp:
## What

In order for our Trusted Advisor to report when and what we desire, we would like to fix some issues we're currently facing.

Two issues targeted in this PR are: 
- the RDS Broker could use HTTPS connection
- the logsearch-ingestor could use ssl connection

In order to prevent downtime, we're adding an extra listener, which then will be removed in another PR.

## How to review

This changes need to be applied in strict order to avoid downtime for log forwarding:

1. Review and apply this PR. It will create new secure ELB listeners for rds-broker and ingestor 
2. Review and apply ( on bootstrap ) https://github.com/alphagov/paas-bootstrap/pull/46. It will change bosh runtime config to point rsyslog to new listener and use SSL
3. Review and apply https://github.com/alphagov/paas-cf/pull/782. It will remove old rds-broker listenr and **implicitly trigger deployment of rsyslog with TLS**
4. Review and apply https://github.com/alphagov/paas-cf/pull/783. It will remove now unused ingestor ELB 

- Run `create-cloudfoundry` pipeline from this branch
- Expect it to go through without errors
- Check if you're receiving logs to your Kabana
- Visit [Trusted Advisor Security](https://console.aws.amazon.com/trustedadvisor/home?region=eu-west-1#/category/security) and make sure,  
`{DEPLOY_ENV}-logsearch-ingestor` and`{DEPLOY_ENV}-rds-broker` are not listed under ELB Listener Security

## Who can review

Neither, @combor nor @paroxp

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1275,9 +1275,9 @@ jobs:
                   echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
 
                   if cf service-brokers | grep "rds-broker\s"; then
-                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS http://$RDS_BROKER_SERVER
+                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS https://$RDS_BROKER_SERVER
                   else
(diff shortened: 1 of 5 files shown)


keymon:
what a pita :stuck_out_tongue:

keymon:
This looks good, merging. I will later merge paas-bootstrap and apply it to all environments.

PR #778, opened 2017-02-10
Switch documentation URL's to the new docs

Jonty:
## What

We've moved our documentation off Readthedocs and onto the PaaS, this updates the documentation links in the API and in emails we send to tenants.

## How to review

Check the new links refer to the same documentation as the old links, but on the new server.

## Who can review

Anyone!


mtekel:
links are correct, cf still deploys and mails get sent, merging

PR #779, opened 2017-02-10
[#121934253] Use blue-green-deploy plugin for healthcheck app

dcarley:
## What

This will ensure that there is no downtime when we deploy the healthcheck
app - which happens during every run of the pipeline. It will enable us to
merge the `post-deploy` job into `cf-deploy` because it will now be safe to
run in parallel with `availability-tests`.

Although it hasn't been a problem to date, it will also prevent us from
getting false-positive alerts from Pingdom (which monitors this app) and may
allow us to reduce the threshold for alerts.

## How to review

- see the individual commit messages for more information
- test without the healthcheck database
  - deploy from `master` to ensure that you have a running environment (not deleted overnight)
  - deploy the pipeline from a branch
  - run it all the way through
  - confirm that the availability-tests pass
- test with the healthcheck database
  - deploy from `master` to ensure that you have a running environment (not deleted overnight)
  - delete `DISABLE_HEALTHCHECK_DB` from your local `Makefile`
  - deploy the pipeline from a branch with `SELF_UPDATE_PIPELINE=false`
  - run it all the way through
  - confirm that the availability-tests pass
- **important before merging**
  - merge alphagov/paas-docker-cloudfoundry-tools#89 first
  - remove the "TMP" commit

## Who can review

Not @dcarley

Unfortunately I'm on holiday after today, so someone else may need to address any comments that arise.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1228,30 +1228,59 @@ jobs:
 
                 ls -l config/*
 
-      - task: create-admin-org-space
-        config:
-          platform: linux
-          image_resource:
-            type: docker-image
-            source:
-              repository: governmentpaas/cf-cli
-          inputs:
-            - name: paas-cf
-            - name: config
-            - name: bosh-CA
-          run:
-            path: sh
-            args:
-              - -e
-              - -c
-              - |
-                ./paas-cf/concourse/scripts/import_bosh_ca.sh
-                . ./config/config.sh
-                echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
-                echo | cf create-org admin
-                cf create-space admin -o admin
-
       - aggregate:
+        - task: create-admin-org-space
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
+            inputs:
+              - name: paas-cf
+              - name: config
+              - name: bosh-CA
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1695,120 +1751,25 @@ jobs:
...
-          params:
-            file: deployed-healthcheck/healthcheck-deployed
-
   - name: smoke-tests
     serial_groups: [smoke-tests]
     plan:
       - aggregate:
           - get: pipeline-trigger
-            passed: ['post-deploy','availability-tests']
+            passed: ['availability-tests']
```
alext: This needs to trigger from both `cf-deploy` and `availability-tests`. Otherwise if the deploy fails, but the availability tests are green, the pipeline will continue, which we don't want. (similarly for `paas-cf` below)

dcarley: ACK, will update. I think I did this in a few places.

dcarley: Rebased - I've basically replaced all previous occurrences of `post-deploy` with `cf-deploy`.

alext:
I've tested this, and it works correctly in both cases (with and without the healthcheck DB).

I've spotted one potential issue with the ordering of the post-deploy jobs, which could cause issues for an initial deploy of an environment. I've added a commit to address this.

Because of this, it would be good for someone else to review this additional change.

Jonty:
@alext LGTM

PR #782, opened 2017-02-14
[#134350583] Cleanup of rds listener 

combor:
## What

This needs to be rebased after https://github.com/alphagov/paas-cf/pull/776 is merged.

As we have moved to https for cloud controller to rds broker
communication we can now remove old listener and related security rule.

## How to review

This needs to be applied after https://github.com/alphagov/paas-cf/pull/776. Check the sequence in the master PR.

Run deployment from this branch and make sure you still can interact with rds-broker. For instance execute `cf create-service` command.

## Who can review
Not @paroxp or @combor

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1275,9 +1275,9 @@ jobs:
                   echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
 
                   if cf service-brokers | grep "rds-broker\s"; then
-                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS http://$RDS_BROKER_SERVER
+                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS https://$RDS_BROKER_SERVER
                   else
-                    cf create-service-broker rds-broker rds-broker $RDS_BROKER_PASS http://$RDS_BROKER_SERVER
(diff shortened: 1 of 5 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1363,9 +1363,9 @@ jobs:
                   echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
 
                   if cf service-brokers | grep "rds-broker\s"; then
-                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS http://$RDS_BROKER_SERVER
+                    cf update-service-broker rds-broker rds-broker $RDS_BROKER_PASS https://$RDS_BROKER_SERVER
```
keymon: This is wrong, because these commands will run AFTER you remove the HTTP listener! so we will have downtime because the broker is registered with the wrong old url for a while.

keymon: This should actually be moved to the previous PR #776

manifests/cf-manifest/manifest/800-logsearch.yml
```diff
@@ -131,8 +131,6 @@ jobs:
   templates:
   - name: ingestor_syslog
     release: logsearch
-  - name: ingestor_relp
```
keymon: this could also be moved to #776, but it is fine.

PR #785, opened 2017-02-14
[#139706089] Enable route services

combor:
## What

This will enable tenants to add an authentication layer in front of an app without having to modify the app itself.

## How to review

- Deploy from this branch
- Push the [cf_basic_auth_route_service](https://github.com/alext/cf_basic_auth_route_service) app to CF (setting the AUTH_USERNAME,  AUTH_PASSWORD and SKIP_SSL_VALIDATION env vars).
- Define a user-provided-service for it:
```
cf create-user-provided-service my-auth -r https://basic-auth-route-service.${DEPLOY_ENV}.dev.cloudpipelineapps.digital
```
- Bind it to the route for an app:
```
cf bind-route-service ${DEPLOY_ENV}.dev.cloudpipelineapps.digital my-auth --hostname my-app-route
```
- Check if your app has basic auth now
## Who can review

Not @combor

### DIFF
--- a/manifests/cf-manifest/scripts/generate-cf-secrets.rb
+++ b/manifests/cf-manifest/scripts/generate-cf-secrets.rb
@@ -14,6 +14,7 @@ generator = SecretGenerator.new(
   "bulk_api_password" => :simple,
   "nats_password" => :simple,
   "router_password" => :simple,
+  "route_services_secret" => :simple,
   "uaa_batch_password" => :simple,
   "uaa_admin_password" => :simple,
   "test_user_password" => :simple,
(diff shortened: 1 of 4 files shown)


mtekel:
How about enabling route services tests? 
https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/manifest/900-cf-tests.yml#L23

combor:
Good idea. Enabled.

manifests/cf-manifest/scripts/generate-cf-secrets.rb
```diff
@@ -14,6 +14,7 @@
   "bulk_api_password" => :simple,
   "nats_password" => :simple,
   "router_password" => :simple,
+  "route_services_secret" => :simple,
```
alext: The docs mention that this needs to be a "strong passphrase". We should investigate what that really means, and if necessary introduce a stronger scheme for the password generator.

alext: We've done some digging, and I think this password scheme is fine. I've added details of this to the commit message.

mtekel:
works

dcarley:
👏

PR #786, opened 2017-02-14
Enable cf tasks acceptance tests

mtekel:
## What
As of our upgrade to v251, we have announced support of cf tasks.
Do run acceptance tests to keep verifying that the functionality
still works correctly over time.

Note that the tests can still pass even if task_creation feature
flag is disabled as tests run with user that has admin privileges,
which can run tasks even if they are disabled.

## How to review

Run. Acceptance tests should pass.

## Who can review
not @mtekel


keymon:
LGTM

PR #789, opened 2017-02-20
[#130020741] Password rotation pipeline 

combor:
## What

This is a first batch of changes to implement password rotation pipelines.

There are two of them in the `operator` group. 

- reset cloudfoundry passwords - rotates all but grafana password
- reset bosh passwords - rotates all but NATS password

Grafana changes will come in a next PR as we decided to improve grafana release to implement password rotation via deployment manifest.

We found rotation of NATS particularly difficult so we agreed to put it out of scope. See https://www.pivotaltracker.com/n/projects/1275640/stories/127339659 for more details.   

## How to review
- deploy your `create` pipeline
- run `reset-cloudfoundry-passwords` and again `create` pipeline to test if creds are rotated
- run `reset-bosh-passwords` and next `bootstrap` pipeline to test bosh creds rotation 

## Who can review

Not @combor

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -41,6 +41,7 @@ groups:
       - bump-patch-version
       - pipeline-check-lock
       - pipeline-release-lock
+      - reset-cloudfoundry-passwords
   - name: tests
     jobs:
       - availability-tests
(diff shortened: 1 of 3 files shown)


mtekel:
LG, but I don't think concourse ssh keys should be rotated in `reset-cloudfoundry-passwords` job. I think it should be separated into its own job which can be later extended with resetting other concourse credentials....

combor:
Might be...it's a naming convention...we can also do `rotate-passwords` pipeline and merge bosh bosh and cf pipelines.

mtekel:
I prefer these be separate tasks in operator panel, as you can more easily decide which kind of secrets to rotate. You don't always need to rotate everything at once...

combor:
Ok, suggested improvements are implemented now.

mtekel:
As a part 1 LGTM. Part 2 could add more things apart from grafana. E.g. concourse passwords, perhaps bosh CA or deployment ssh key. There's also IPSec CA and CF certs as well that we could rotate.

PR #790, opened 2017-02-21
[#140308141] Whitelist skyscape IPs for #140308141

keymon:
[#140308141 Create org and users for new tenant prototypes](https://www.pivotaltracker.com/n/projects/1275640/stories/140308141)

What?
-----

We want to allow the new tenant prototypes from a skyscape list of ips

We whitelist the skyscape IPs for the ticket[1]

[1] https://www.pivotaltracker.com/n/projects/1275640/stories/140308141

How to review?
-------------

code review

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/prod.tfvars
+++ b/terraform/prod.tfvars
@@ -52,6 +52,9 @@ tenant_cidrs = [
   "194.73.212.3/32",
   # MOJ Digital and Technology
   "81.134.202.29/32",
+  # Tenant egress for story #140308141
+  "185.40.8.212/32",
+  "86.188.177.234/32",
 ]
 bosh_db_backup_retention_period = "35"
 bosh_db_skip_final_snapshot = "false"


terraform/prod.tfvars
```diff
@@ -52,6 +52,9 @@ tenant_cidrs = [
   "194.73.212.3/32",
   # MOJ Digital and Technology
   "81.134.202.29/32",
+  # UK Cloud / Skyscape egress for #140308141
+  185.40.8.212/32,
+  86.188.177.234/32,
```
alext: Is this all of UKCloud/Skyscape, or a subset specific for a single tenant? I think it would be useful to capture who they're for in the comment here.

dcarley: Bit of an odd situation, but there's some context in the referenced story: https://www.pivotaltracker.com/n/projects/1275640/stories/140308141

alext: Ah, now that I've read it more carefully that makes sense.

alext: Another point. Do these not need to be quoted like the others?

dcarley: Actually, having thought about it for a minute, I think we should remove the provider from the description and just reference the story ID/URL.

keymon: I fixed that

alext:

PR #791, opened 2017-02-21
[#139296467] Add bare domain of product page to cloudfront

keymon:
[#139296467 Redirect from cloud.service.gov.uk (no www prefix)](https://www.pivotaltracker.com/story/show/139296467)

What?
-----

Set it up so that if users try to navigate to `cloud.service.gov.uk` in their browsers, they will be redirected to `https://www.cloud.service.gov.uk`

We added logic to the product page to redirect to the page with the www prefix in [1]. Here we map the domain in cloudfront.

[1] https://github.com/alphagov/paas-product-page/pull/2

How to review?
------------

 * Code review
 * Deploy from master with `make dev setup_cdn_instances ACTION=apply`, then check the difference checking out this branch and running `make dev setup_cdn_instances ACTION=plan`

You can deploy it by `make dev setup_cdn_instances ACTION=apply` again, and check that you are reaching the govuk product page when running `curl -v -I -k https://${DEPLOY_ENV}.dev.cloudpipeline.digital`

Dependencies
-----------

https://github.com/alphagov/paas-product-page/pull/2 should be merged first and deployed to prod.

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/cloudfront/cloudfront_distribution/distribution.tf
+++ b/terraform/cloudfront/cloudfront_distribution/distribution.tf
@@ -3,9 +3,13 @@ resource "aws_route53_record" "cdn_domain" {
 
   zone_id = "${var.system_dns_zone_id}"
   name    = "${element(var.aliases, count.index)}."
-  type    = "CNAME"
-  ttl     = "60"
-  records = ["${aws_cloudfront_distribution.cdn_instance.domain_name}"]
+  type    = "A"
+
+  alias {
+    name                   = "${aws_cloudfront_distribution.cdn_instance.domain_name}"
+    zone_id                = "${aws_cloudfront_distribution.cdn_instance.hosted_zone_id}"
+    evaluate_target_health = true
+  }
 }
 
 resource "aws_cloudfront_distribution" "cdn_instance" {
--- a/terraform/cloudfront/instances.tf
+++ b/terraform/cloudfront/instances.tf
@@ -8,7 +8,7 @@ module "cloudfront_paas_product_page" {
   source = "./cloudfront_distribution"
 
   name    = "PaaS Product Page"
-  aliases = ["www.${var.system_dns_zone_name}"]
+  aliases = ["www.${var.system_dns_zone_name}", "${var.system_dns_zone_name}"]
   origin  = "govuk-paas.cloudapps.digital"
   comment = "Serve the govuk-paas under the gov.uk domain."
 


alext:
How is DNS for this going to work? (note `cloud.service.gov.uk` is the system domain, not the apps domain, so there's no wildcard entry). Given this is an apex domain, we can't use a CNAME to point to the Cloudfront endpoint.

dcarley:
> How is DNS for this going to work? (note cloud.service.gov.uk is the system domain, not the apps domain, so there's no wildcard entry). Given this is an apex domain, we can't use a CNAME to point to the Cloudfront endpoint.

[`ALIAS` record](http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html)? Does CloudFront create DNS records for you automatically?

keymon:
Cloudfront does not create domains we do: https://github.com/alphagov/paas-cf/blob/9b2a18cdbbbc0eaec8afb1ad2d64ff6cb363fe54/terraform/cloudfront/cloudfront_distribution/distribution.tf#L1-L10

But `cloud.service.gov.uk` is  is not an APEX.  APEX is `service.gov.uk` 

Or am I getting confused here. 

Anyway, if it does not work, I would change that with ALIAS as @dcarley suggested.

keymon:
I mean, APEX or ROOT domains are the ones that are configured the ROOT DNS servers. but in this case `cloud.service.gov.uk` is a delegation, isn't it? Maybe I am wrong...

alext:
> Does CloudFront create DNS records for you automatically?

I don't believe it does. Our terraform modules sets them up - https://github.com/alphagov/paas-cf/blob/master/terraform/cloudfront/cloudfront_distribution/distribution.tf#L1-L9  This is going to need to be modified ebcause it will currently attempt to create a CNAME at the APEX.

> But cloud.service.gov.uk is is not an APEX. APEX is service.gov.uk

service.gov.uk delegates to cloud.service.gov.uk, but cloud.service.gov.uk is still the apex of the hosted zone in Route53.

keymon:
> but cloud.service.gov.uk is still the apex of the hosted zone in Route53.

I think that is not a problem and you can use a CNAME in that case... but I need to test it...

alext:
The thing is that you can't have a CNAME and any other records at a given level in the DNS hierarchy. In our case we have all the delegation records (NS etc) at that level in DNS, so we can't additionally have a CNAME.

keymon:
OK, you are right. More info: https://en.wikipedia.org/wiki/CNAME_record#Restrictions

Then I need to change it to ALIAS as @dcarley suggested.

alext:
I've run into problems with this:

CloudFront doesn't set the `X-Forwarded-Host` header, which alphagov/paas-product-page#2 relies on. Additionally it's not presently configured to cache differently based on the incoming host header, so it won't differentiate between the redirect from the bare hostname, and the actual content from the www hostname.

It is possible to configure cloudfront to pass through (and cache based on) the Host header, but then this Host header is presented to the CF router, which returns a 404 because it doesn't know about that domain.

I think we need to rethink how we implement this, and maybe implement the bare domain redirect independently from the product page hosting. Is this something that could be done with a lambda function or something like that?

keymon:
OK, I will propose something else. I have the original implementation of a cloudfront+S3 bucket, but I am not sure if it does set the HSTS headers. 

I will move this to doing.

keymon:
We will use this approach, we will use https://github.com/alphagov/paas-cf/pull/794 instead.

PR #792, opened 2017-02-21
[#138702349] TO DISCUSS: Fix flapping route53 alias with hardcoded zone id

keymon:
What?
----

It seems to be an issue in terraform [1] which makes the route53 alias
to flap changes. The solution for the time being is:

 1. Add a `.` in the domain reported by ELB
 2. Hardcode the zone id of the ELBs. (Z32O12XQLNTSW2 for eu-west-1)

We get the zone IDs from http://docs.aws.amazon.com/general/latest/gr/rande.html

How to test?
----------

This problem does not happen in dev :(

You can test it in CI, though. Or simply do a code review.

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/cloudfoundry/dns-records.tf
+++ b/terraform/cloudfoundry/dns-records.tf
@@ -60,8 +60,15 @@ resource "aws_route53_record" "apps_apex" {
   type    = "A"
 
   alias {
-    name                   = "${aws_elb.cf_router.dns_name}"
-    zone_id                = "${aws_elb.cf_router.zone_id}"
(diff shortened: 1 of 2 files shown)


dcarley:
Hector and I talked about this after standup. We were both uncomfortable with hardcoding the zone IDs, in case they change in the future. We've provided some more information to the upstream issue hashicorp/terraform#10007 and created a follow-up story [#140398377](https://www.pivotaltracker.com/story/show/140398377) to see if it's fixed or consider recreating the ELBs.

PR #793, opened 2017-02-23
[#130020741] Update grafana to v 4.1.2

combor:
## What

This is a follow up to https://github.com/alphagov/paas-cf/pull/789 that updates Grafana to v 4.1.2 to make use of newly provided `grafana-cli` admin command.
The `grafana-cli admin` allows to change admin user password via cli.
This will let us to set new password in the manifest and rotate it during deployment.

## How to review

**This is master PR for set of PRs related to grafana update. PRs need to be merged in the sequence listed below**:
- https://github.gds/government-paas/aws-account-wide-terraform/pull/83
- https://github.com/alphagov/paas-release-ci/pull/12
- https://github.com/alphagov/paas-grafana-boshrelease/pull/1
- this PR

## Who can review

Not @combor


mtekel:
Works. There is one catch though: pw reset doesn't break existing sessions - it does not force logout - and you can still keep browsing stats if you were logged in before pw reset.

PR #794, opened 2017-02-23
[#139296467] Add redirection to add www to the system domain

keymon:
[#139296467 Redirect from cloud.service.gov.uk (no www prefix)](https://www.pivotaltracker.com/story/show/139296467)

What?
-----

Set it up so that if users try to navigate to `cloud.service.gov.uk` in their browsers, they will be redirected to `https://www.cloud.service.gov.uk`

We use a configuration based on Cloudfront + API Gateway to redirect to the www.${apex} domain from the APEX system domain.

Limitation: Any other path will return a 403 error. In order to make API Gateway work with any path, we must use proxy mode and send the request to a Lambda function or similar.

Dependencies
------------

This supersedes the PRs https://github.com/alphagov/paas-cf/pull/791 and https://github.com/alphagov/paas-product-page/pull/2

How to review?
------------

 * Code review
 * Deploy from master with `make dev setup_cdn_instances ACTION=apply`, then check the difference checking out this branch and running `make dev setup_cdn_instances ACTION=plan`

You can deploy it by `make dev setup_cdn_instances ACTION=apply` again, and check that you are reaching the govuk product page when running `curl -v -I -k https://${DEPLOY_ENV}.dev.cloudpipeline.digital`

Note that cloudfront creation takes some minutes.

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/cloudfront/instances.tf
+++ b/terraform/cloudfront/instances.tf
@@ -31,3 +31,16 @@ module "cloudfront_paas_docs" {
   system_dns_zone_id    = "${var.system_dns_zone_id}"
   system_domain_cert_id = "${var.system_domain_cert_id}"
 }
+
+module "redirect_paas_product_page" {
+  source = "./cloudfront_redirect"
+
+  name = "${var.env}-paas-product-page"
+
+  aliases         = ["${var.system_dns_zone_name}"]
+  redirect_target = "https://www.${var.system_dns_zone_name}"
+
+  env            = "${var.env}"
+  dns_zone_id    = "${var.system_dns_zone_id}"
+  domain_cert_id = "${var.system_domain_cert_id}"
+}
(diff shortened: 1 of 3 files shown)


alext:
Travis failures are due to terraform lint.

keymon:
Fixed that

terraform/cloudfront/instances.tf
```diff
@@ -31,3 +31,16 @@ module "cloudfront_paas_docs" {
   system_dns_zone_id    = "${var.system_dns_zone_id}"
   system_domain_cert_id = "${var.system_domain_cert_id}"
 }
+
+module "redirect_paas_product_page" {
+  source = "./cloudfront_redirect"
+
+  name = "paas-product-page"
```
alext: It's probably worth including the `DEPLOY_ENV` in here so we can distinguish them in the AWS console.

alext:
This looks to be working correctly now with the changes in the WIP PR.

Happy to merge once those are squashed in, and the `DEPLOY_ENV` thing is adderessed.

PR #796, opened 2017-02-23
Make Terraform version check more robust

alext:
## What

This awk monstrosity (that I introduced in 50bb83f) was not working with
gawk 3.1.8 or mawk 1.3.3 (as available on Ubuntu Precise). It turns out
that the regex interval expressions (`{1,4}` etc) only work with gawk 3
if you specify the `--posix` flag. They're enabled by default with gawk
4+.

Instead of attempting to resolve this in a portable way, I've instead
reimplemented this check in ruby, which is arguably much more readable.

## How to review

Test that `make check-tf-version` works as expected.

## Who can review

Anyone but myself.


dcarley:
Terraform have added this a feature since we first did it:

- https://www.terraform.io/docs/configuration/terraform.html

Could we use that instead?

alext:
Oooo interesting. I hadn't seen that. I'll give it a go.

combor:
Alex, please let me know when you are done with improvements.

alext:
@combor This is now ready for review.

Note: this is currently based on #797, so includes the commit from that. We can either merge this one, or I can rebase after #797 is merged.

combor:
I'll review the other one first.

combor:
You an rebase now.

alext:
rebased.

terraform/version_constraint.tf
```diff
@@ -0,0 +1,3 @@
+terraform {
+  required_version = ">= 0.8.5"
```
combor: TBH I'm not sure what's the binary compatibility of the compiled pingdom provider. I've got local terraform 0.8.7 and the linting tests pass but can we guarantee this ?
Shall we be more strict here?

alext: This is currently replicating the existing constraint that was in place. I think if we want to alter the versions we constrain this to, that's a separate change, and should be a separate PR.

combor:
OK, so this LGTM. Merging.

PR #797, opened 2017-02-23
Revert "Makefile: refactor environment check"

alext:
## What

This is currently breaking the build because the AWS cred check doesn't work when
run in the pipeline. Reverting this while we work out a fix for this.

This reverts commit e1a9a229ea543a369136d4553b41b7ebd1cf4d6c (part of #730).

## How to review

Run the pipeline from this branch and verify that the self-update-pipelines task runs correctly.

## Who can review

Anyone but myself


combor:
It works. Merging.

PR #798, opened 2017-02-23
Skip AWS cred. validation in self-update pipelines

mtekel:
## What

We are using IAM profile on concourse for AWS credentials. We thus want to
skip STS token and user credential validation that we normally want to run
when running make locally.

## How to review

Deploy this branch to your concourse. Run self update pipelines task (in the 1st pipeline-lock job). It should pass.

## Who can review

not @mtekel

### DIFF
--- a/scripts/validate_aws_credentials.sh
+++ b/scripts/validate_aws_credentials.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 
+if [ "${SKIP_AWS_CREDENTIAL_VALIDATION:-}" == "true" ]  ; then
+  exit 0
+fi
+
 if [ -z "$AWS_SESSION_TOKEN" ]; then
   echo "No temporary AWS credentials found, please run create_sts_token.sh"
   exit 255;
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,7 @@ pingdom: check-env check-tf-version ## Use custom Terraform provider to set up P
 	@terraform/scripts/set-up-pingdom.sh ${ACTION}
 
 .PHONY: setup_cdn_instances
-setup_cdn_instances: check-tf-version check-aws-credentials ## Setup the CloudFront Distribution instances, by reading their config from terraform/cloudfront/instances.tf.
+setup_cdn_instances: check-tf-version check-env ## Setup the CloudFront Distribution instances, by reading their config from terraform/cloudfront/instances.tf.
 	$(if ${ACTION},,$(error Must pass ACTION=<plan|apply|...>))
 	@terraform/scripts/set-up-cdn-instances.sh ${ACTION}
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
-.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck lint_concourse check-env-vars
-
+.PHONY: help test spec lint_yaml lint_terraform lint_shellcheck lint_concourse check-env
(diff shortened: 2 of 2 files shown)


scripts/validate_aws_credentials.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 
+if ${SKIP_AWS_CREDENTIAL_VALIDATION:-false} ; then
```
alext: This will actually execute the contents of the variable as a command. eg:

```
$ cat test.sh
#!/bin/bash

if ${FOO:-false} ; then
  echo "true"
  exit 0
fi

echo "false"

$ FOO=wibble ./test.sh 
./test.sh: line 3: wibble: command not found
false
```

It would be better to test for string values instead like we do elsewhere.

mtekel:
updated

dcarley:
@mtekel should this PR re-apply the change that was reverted in #797?

mtekel:
well, I need to check what everything was added/changed in makefile since I made the original PR as well...

mtekel:
updated

alext:
This doesn't successfully deploy without manual intervention. On the first run of self-update-pipelines, the pipeline hasn't been updated to include the `SKIP_AWS_CREDENTIAL_VALIDATION` parameter, so it attempts to check the AWS creds and fails.

mtekel:
We could do 2 PRs then. Alternatively, temporarily check if `PIPELINES_TO_UPDATE` is set, indicating that this is run inside pipeline. But cleanup would require 2nd PR...

alext:
I can't think of a way of doing this without needing 2 PRs. I'm not sure I feel strongly about which way to go.

dcarley:
`¯\_(ツ)_/¯`

Let's go with the first option of one PR to set `SKIP_AWS_CREDENTIAL_VALIDATION` and 2nd to implement the check because I think it's slightly more explicit.

mtekel:
I have created https://github.com/alphagov/paas-cf/pull/803

alext:
I've merged #803. This can now be rebased on master to pick up that change.

mtekel:
rebased

alext:
This now works for the create pipeline, but we need the same flag added for all the other pipelines.

mtekel:
Good point: https://github.com/alphagov/paas-cf/pull/804

mtekel:
ok, so to apply this, we need to run those pipelines :| not sure we want to....

mtekel:
```
diff --git a/scripts/validate_aws_credentials.sh b/scripts/validate_aws_credentials.sh
index 941e1d59..d55f2759 100755
--- a/scripts/validate_aws_credentials.sh
+++ b/scripts/validate_aws_credentials.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 
+if [ ! -z "${PIPELINES_TO_UPDATE+x}" ] ; then
+  echo "PIPELINES_TO_UPDATE set to ${PIPELINES_TO_UPDATE}. Assuming run in pipeline, skipping STS credentials check."
+  exit 0
+fi
+
 if [ "${SKIP_AWS_CREDENTIAL_VALIDATION:-}" == "true" ]  ; then
   exit 0
 fi
```

alext:
@mtekel I don't think we need to worry about those pipelines. They're not deployed anywhere other than dev.

alext:
I think this is good to merge. I'm going to leave it until Monday to merge so that #803 and #804 have applied everywhere.

PR #800, opened 2017-02-23
[#139003315] Test subsequent deployments while under big load

paroxp:
## What

Prior to our BBS incident, we'd like to test against a large number of instances running to establish if our deployment is unaffected.

We would like to test that the subsequent deployments still work when there are many app instances running.

## How to review

- Review the code
- Run `make dev pipelines` from this branch
- Run `create-cloudfoundry` pipeline
- Expect it to succeed

**Optionally**
- Create separate branch
- Modify the `platform-tests/example-apps/healthcheck/manifest.yml` and remove the `disk_quota` or set it to `1G`
- Run `make dev pipelines` from new branch
- Run `create-cloudfoundry` pipeline
- Expect it to fail

## Who can review

Not @paroxp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export ENABLE_DATADOG ?= false)
 	$(eval export CONCOURSE_AUTH_DURATION=48h)
 	$(eval export DISABLE_PIPELINE_LOCKING=true)
+	$(eval export TEST_HEAVY_LOAD=true)
 	@true
 
 .PHONY: ci
@@ -112,6 +113,7 @@ ci: globals check-env-vars ## Set Environment to CI
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DEPLOY_ENV=master)
+	$(eval export TEST_HEAVY_LOAD=true)
 	@true
 
 .PHONY: staging
@@ -129,6 +131,7 @@ staging: globals check-env-vars ## Set Environment to Staging
 	$(eval export DISABLE_CF_ACCEPTANCE_TESTS=true)
 	$(eval export ENABLE_DATADOG=true)
 	$(eval export DEPLOY_ENV=staging)
+	$(eval export TEST_HEAVY_LOAD=true)
 	@true
 
 .PHONY: prod
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1338,6 +1338,72 @@ jobs:
                   ./paas-cf/concourse/scripts/set_quotas_from_manifest.rb cf-manifest/cf-manifest.yml
 
       - aggregate:
+        - task: deploy-simulated-load-application
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
+            params:
+              TEST_HEAVY_LOAD: {{test_heavy_load}}
+            inputs:
+              - name: paas-cf
+              - name: config
+              - name: bosh-CA
+            run:
+              path: sh
+              args:
+                - -e
+                - -u
+                - -c
+                - |
+                  if [ "${TEST_HEAVY_LOAD:-}" != "true" ] ; then
+                    echo Heavy load test has been disabled. Skipping...
+                    exit 0
+                  fi
+
+                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
+
+                  . ./config/config.sh
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf create-org heavy-testers
+                  cf set-quota heavy-testers medium
+                  cf create-space test-application -o heavy-testers
+                  cf target -o heavy-testers -s test-application
+
+                  cd paas-cf/platform-tests/example-apps/healthcheck
+                  cf push simulated-load -i 180
+
+                  timeout=300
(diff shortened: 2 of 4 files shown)


alext:
I'm not sure this is testing the correct thing. We want an application constantly running on the platform with c 180 instances. Specifically we need it to be present at that scale while the bosh deployment is running so that we catch any issues with rotating the cell VMs etc.

alext:
I'd also question whether the healthcheck endpoint is the right one to use for this. It feels like overloading this apps purpose.

I'd suggest that we use a separate deployment (could be the same codebase) to be the representative load.

paroxp:
@alext 

> I'm not sure this is testing the correct thing. We want an application constantly running on the platform with c 180 instances. Specifically we need it to be present at that scale while the bosh deployment is running so that we catch any issues with rotating the cell VMs etc.

Could you tell me, why do the changes from the PR not satisfy this requirement? Maybe you could recommend some changes?

> I'd also question whether the healthcheck endpoint is the right one to use for this. It feels like overloading this apps purpose.
> 
> I'd suggest that we use a separate deployment (could be the same codebase) to be the representative load.

Yes, you're right. I had a wrong understanding from kick-off. I'll change that.

alext:
> Could you tell me, why do the changes from the PR not satisfy this requirement?

Because this scales the app up to 180, and then back down (using the exit trap) to 2 afterwards. Which means that it'll only have 2 instances most of the time (including while the deployment is happening)

> Maybe you could recommend some changes?

Instead we should look to deploy a separate app instance that permanently has 180 instances.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1338,6 +1338,116 @@ jobs:
...
+                    . ./config/config.sh
+                    echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                    cf create-org testers
+                    cf set-quota testers test_apps
+                    cf create-space test-application -o testers
+                    cf target -o testers -s test-application
+
+                    cd paas-cf/platform-tests/example-apps/healthcheck
+
+                    if  [ "${DISABLE_HEALTHCHECK_DB:-}" != "true" ] ; then
```
alext: This instance never needs a database. We can omit this block completely.

paroxp: I thought that was the case. But then I was considering, if we'd like to put more load on. Thanks. Removing.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1338,6 +1338,73 @@ jobs:
...
+                  . ./config/config.sh
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf create-org heavy-testers
+                  cf set-quota heavy-testers medium
+                  cf create-space test-application -o heavy-testers
+                  cf target -o heavy-testers -s test-application
+
+                  cd paas-cf/platform-tests/example-apps/healthcheck
+                  cf push test-application
+                  cf scale test-application -i 180
```
alext: This doesn't need to be a separate call, you can do `cf push -i 180`

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1338,6 +1338,73 @@ jobs:
...
+
+                  . ./config/config.sh
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf create-org heavy-testers
+                  cf set-quota heavy-testers medium
+                  cf create-space test-application -o heavy-testers
+                  cf target -o heavy-testers -s test-application
+
+                  cd paas-cf/platform-tests/example-apps/healthcheck
+                  cf push test-application
```
alext: I think we should give this app a more descriptive name. How about simulated-load or something like that.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1338,6 +1338,73 @@ jobs:
...
+              - name: config
+              - name: bosh-CA
+            run:
+              path: sh
+              args:
+                - -e
+                - -u
+                - -c
+                - |
+                  if [ "${TEST_HEAVY_LOAD:-}" -eq "false" ] ; then
```
alext: `-eq` is meant for comparing integers. For strings `=` or `!=` should be used. See `man test` for full details.

Makefile
```diff
@@ -148,6 +148,7 @@ prod: globals check-env-vars ## Set Environment to Production
 	$(eval export ENABLE_PAAS_DASHBOARD=true)
 	$(eval export DEPLOY_RUBBERNECKER=true)
 	$(eval export DEPLOY_ENV=prod)
+	$(eval export TEST_HEAVY_LOAD=false)
```
alext: I think we should invert the sense of this variable - ie have it default to false, and then set it to true for dev/ci/staging. That will make it consistent with other variables, and make it less likely that this gets deployed to prod by mistake.

concourse/pipelines/create-cloudfoundry.yml
alext: This is also why Travis failed.

Makefile
paroxp: Yeah, I had it like that at the beginning, but thought it's less code to do otherwise. Changing now.

concourse/pipelines/create-cloudfoundry.yml
paroxp: Good to know. Thanks.

concourse/pipelines/create-cloudfoundry.yml
paroxp: Bah... Silly "quick" change. Thanks.

PR #802, opened 2017-02-23
[#130020741] Rotate more credentials 

mtekel:
## What

[Implement automation of credential rotation (continued)](https://www.pivotaltracker.com/story/show/130020741)

Add more secrets and credentials into rotation.

## How to review

Run all clear credentials tasks. Then run boostrap and main pipelines. Observe new credentials being generated and applied to bosh, concourse and CF deployment. All should work fine and all tests should pass.

## Who can review

not @mtekel or @combor


mtekel:
I thought it would be good to include tasks that need to run bootstrap to generate new creds in bootstrap pipeline as well, but then I thought about that story and realised that soon we'll have bootstrap pipeline available in the deployer concourse, so that work would have very short lived benefits.

mtekel:
renamed name and moved the tasks to 'credentials' tab

dcarley:
Having some problems redeploying BOSH. It looks like the keypair might not have been updated in EC2:

```
  Compiling package 'collectd/e2d748928525612e8bc7f99173dfa9ce96367acb'... Finished (00:00:48)
  Updating instance 'bosh/0'... Finished (00:00:17)
  Waiting for instance 'bosh/0' to be running... Failed (00:05:05)
Failed deploying (00:24:06)

Stopping registry... Finished (00:00:00)
Cleaning up rendered CPI jobs... Finished (00:00:00)

Command 'deploy' failed:
  Deploying:
    Received non-running job state: 'starting'
```

Going to investigate further on Monday.

> I thought it would be good to include tasks that need to run bootstrap to generate new creds in bootstrap pipeline as well, but then I thought about that story and realised that soon we'll have bootstrap pipeline available in the deployer concourse, so that work would have very short lived benefits.

I think we might need to do something in the meantime about making it clearer what you need to do after running the clear credentials tasks. I'm not sure about the implementation though; can talk about it in-person.

alext:
> I thought it would be good to include tasks that need to run bootstrap to generate new creds in bootstrap pipeline as well, but then I thought about that story and realised that soon we'll have bootstrap pipeline available in the deployer concourse, so that work would have very short lived benefits.

Even when we have the bootstrap pipelines in the deployer concourse, the pipeline definitions (and secret generation) will still live in the paas-bootstrap repo, so it probably makes sense to have the tasks that reset the creds in the same repo as the tasks that generate them.

dcarley:
> Even when we have the bootstrap pipelines in the deployer concourse, the pipeline definitions (and secret generation) will still live in the paas-bootstrap repo, so it probably makes sense to have the tasks that reset the creds in the same repo as the tasks that generate them.

Also, I agree with this. What's the amount of effort to move them?

concourse/tasks/delete-ssh-keys.yml
```diff
@@ -0,0 +1,21 @@
...
+  type: docker-image
+  source:
+    repository: governmentpaas/awscli
+inputs:
+  - name: paas-cf
+run:
+  path: sh
+  args:
+    - -e
+    - -c
```
mtekel: re: -u
https://github.com/alphagov/paas-cf/pull/802/files#diff-8f7658bb02a08dfb771f38694a7b075fR2272
I actually need it. I can change it do full "prefix" - that is, the whole name of key.

mtekel:
regarding moving tasks: how about moving _all_ of them to bootstrap (we have 3, bosh, concourse and CF secrets. Bosh and concourse belong to bootstrap) ? This way they are not present in the main prod pipeline...

concourse/tasks/delete-ssh-keys.yml
dcarley: It'll work because `-u` checks if the variable is set, not if it's contents are empty:
```sh
➜  paas-cf git:(master) ✗ env ./test.sh
./test.sh: line 3: SSH_KEY_PREFIX: unbound variable
➜  paas-cf git:(master) ✗ env SSH_KEY_PREFIX="" ./test.sh
SSH_KEY_PREFIX:
➜  paas-cf git:(master) ✗ env SSH_KEY_PREFIX="foo" ./test.sh
SSH_KEY_PREFIX: foo
```

dcarley:
> regarding moving tasks: how about moving all of them to bootstrap (we have 3, bosh, concourse and CF secrets. Bosh and concourse belong to bootstrap) ? This way they are not present in the main prod pipeline...

I'm not sure about moving the CF task. Can you describe the pros and cons? I'm thinking..

- pro:
  - keep all related code in one place
  - less likely to run the task by mistake in production
- con:
  - not necessarily where you'd expect to find it
  - you'd need to provision a Bootstrap Concourse even if you were only rotating CF credentials

mtekel:
your 1st pro and con contradict themselves
re 2nd con: provisioning is now quite fast. Rotating credentials is IMO such a task that you want to ensure you _really_ don't accidentally trigger. Having to deploy a server (even if it only takes 5 mins) seems like a good safeguard, but also not too difficult/complicated to disable running it when necessary

dcarley:
> your 1st pro and con contradict themselves

I think they're different things. The pro is about duplication of code (the task and spruce templates). The con is about making it easy for team members to use the functionality.

I don't feel really strongly either way. Happy for you to choose or someone else pitch in with their opinion. Let's move it for BOSH and Concourse either way.

mtekel:
I would say the main problem is code duplication. E.g. it would not make sense to have key deletion as a task in paas-cf, as there's nothing else using it. If we end up having bootstrap pipeline uploaded to deployer concourse, the credentials deletion task would be on the server any way - but at least in a separate pipeline.

mtekel:
It seems all comments but moving of tasks were addressed. We should think if we want to do it - as that will take a bit more time. Moving functionality to paas-bootrap shoudl be more or less OK, unless we encounter any issues/tech debt on the way. Which is quite possible.

dcarley:
> Let's move it for BOSH and Concourse either way.

We definitely want to do this. Can you go ahead?

Let's not move the CF task, even if it results in code duplication.

dcarley:
Works 👏  👏  👏

PR #807, opened 2017-02-28
Include ADR017 threshold in show-cf-memory-usage task

alext:
## What

In order to help decide whether we need to increase capacity when adding an org, this adds a line to the `show-cf-memory-usage` task to show the total cell memory required to meet the [ADR017](https://government-paas-team-manual.readthedocs.io/en/latest/architecture_decision_records/ADR017-cell-capacity-assignment/) requirement.

While working on this, I've removed the "Memory actually used by apps:" line because it was incorrect, and wasn't reporting what we thought it was. See the commit message for details.

## How to review

Code review
Run the task from this branch. Verify the new output makes sense and agrees with the ADR.

## Who can review

Anyone but myself.


mtekel:
LG, merging

PR #808, opened 2017-02-28
[#140768615] IP whitelist for mod-dbs-trial

timmow:
## What

Add an IP address for mod-dbs-trial to the whitelist

https://gaap.deskpro.com/agent/#app.tickets,t:3858
[#140768615](https://www.pivotaltracker.com/n/projects/1275640)
## How to review

Check the ip matches that in the deskpro ticket and the changes look like they will work

## Who can review

Anyone but me

### DIFF
--- a/terraform/prod.tfvars
+++ b/terraform/prod.tfvars
@@ -58,6 +58,8 @@ tenant_cidrs = [
   "86.188.177.234/32",
   # Our own build Concourse in CI
   "52.213.245.135/32",
+  # mod-dbs-trial org
+  "82.6.141.121/32",
 ]
 bosh_db_backup_retention_period = "35"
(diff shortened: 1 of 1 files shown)


terraform/prod.tfvars
```diff
@@ -58,6 +58,8 @@ tenant_cidrs = [
   "86.188.177.234/32",
   # Our own build Concourse in CI
   "52.213.245.135/32",
+  # mod-dbs-trial org
+  "82.6.141.121",
```
alext: Needs a `/32` on the end (AWS requires that these all be CIDRs).

alext:
:+1: Looks good to me.  Will merge when Travis is green.

PR #809, opened 2017-03-01
[#138938495] Move CF RDS maintenance window to 7-8am

alext:
## What

RDS maintenance windows can cause short and in exceptional cases longer
periods of downtime.

We have more people available to deal with incidents when it's closer
working hours, but we still want to have the maintenance window before
the working day when tenants are likely to be using the PaaS more.

## How to review

Run the deploy pipeline against an existing environment. Observe the maintenance window for the Cf RDS instance being updated. This change should apply immediately.

## Who can review

Anyone but myself.


keymon:
1kkasz

As commented, already tested by @alext and the code looks sane

PR #811, opened 2017-03-01
[#140853603] Fix CONCOURSE_IP in ssh.sh script

saliceti:
## What
Story: [Fix ssh_concourse](https://www.pivotaltracker.com/story/show/140853603)

There was a typo that was not a problem before, but now we check AWS credentials and this was failing.

## How to review
Try `ssh_concourse` on a non DEV environment

## Who can review
Not me


mtekel:
works, thanks for the fix

PR #817, opened 2017-03-07
Revert "Increase number of CC workers to 3 per VM"

saliceti:
## What

This reverts commit c292aad01133c3cb79117e1aca4bbd11e22bcf1d.

The workers are constantly polling the RDS CC DB. This causes significant CPU usage and we have run out of credits. The impact is that services connecting to the databases in that instance are
slower (api, uaa, bbs). This has already been reported by some users.

We now revert to the default. This is safe as this commit hadn't made an improvement.

## How to review
Code review

## Who can review
Not me


mtekel:
applies fine and does what we expect... merging

PR #818, opened 2017-03-07
[#139196683] Monitor BBS health

mtekel:
## What

This is set of PRs to deploy BBS monitoring solution. It adds new `cf.bbs.Healthy` and `cf.bbs.ResponseTime` metrics. 

It consists of:
- https://github.com/alphagov/paas-bootstrap/pull/55 - WHICH MUST BE DEPLOYED FIRST
- https://github.com/alphagov/paas-datadog-bbs-integration/pull/1 - Provides BBS check and tests.
- https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/13 - Integrates BBS check into BOSH release and provides default config
- https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/13 - Implements new mechanism to copy custom checks into datadog agent directory
- This PR

## How to review

- deploy from this branch and check if `cf.bbs.Healthy` and `cf.bbs.ResponseTime` are available in DD for your env. Check if monitor works
- Review and merge https://github.com/alphagov/paas-datadog-bbs-integration/pull/1
- update git submodule in  https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/13 to point to master branch of `alphagov/paas-datadog-bbs-integration`
- Merge https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/13 - this will trigger final build for `paas-datadog-for-cloudfoundry-boshrelease`
- Once https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/13 is merged, replace TMP commit with final build of the release.

## Who can review

not @mtekel or @combor

### DIFF
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -207,6 +207,8 @@ jobs:
         release: cf
       - name: datadog-consul
         release: datadog-for-cloudfoundry
+      - name: datadog-bbs
+        release: datadog-for-cloudfoundry
       - name: bbs
         release: diego
       - name: cfdot
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -224,6 +224,9 @@ jobs:
       - name: cf
     update:
       serial: true
+    properties:
+      bbs:
+          url: (( concat "https://" meta.bbs.api_location "/v1/actual_lrp_groups/list" ))
 
   - name: uaa
     azs: [z1, z2]
--- a/terraform/datadog/bbs.tf
+++ b/terraform/datadog/bbs.tf
@@ -22,3 +22,20 @@ resource "datadog_monitor" "bbs_lock_held_once" {
 
   tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:bbs"]
 }
+
+resource "datadog_monitor" "bbs_healthy" {
+  name                = "${format("%s bbs healthy", var.env)}"
+  type                = "query alert"
+  message             = "${format("BBS health check failed. Check BBS status immediately. @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
+  escalation_message  = "BBS health check still failing. Check BBS status immediately."
+  notify_no_data      = true
+  require_full_window = false
+
+  query = "${format("min(last_1m):min:cf.bbs.Healthy{bosh-deployment:%s} < 1", var.env)}"
+
+  thresholds {
+    critical = "1"
+  }
+
+  tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:bbs"]
+}
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -26,9 +26,9 @@ releases:
     url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/paas-haproxy-0.1.3.tgz
     sha1: 732ceb817afe33ee117b85a202d87f6f5c3dd760
   - name: datadog-for-cloudfoundry
-    version: 0.1.7
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/datadog-for-cloudfoundry-0.1.7.tgz
-    sha1: 3bdfdbd98cac6e9452b80788229e83cddf9549d3
+    version: 0.1.13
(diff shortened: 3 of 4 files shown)


manifests/cf-manifest/manifest/010-cf-jobs.yml
```diff
@@ -222,6 +224,9 @@ jobs:
       - name: cf
     update:
       serial: true
+    properties:
+      bbs:
+        url: (( grab meta.bbs.api_location  ))
```
dcarley: What's happening when this runs on both database VMs? Do they both submit the same metric with the same name?

terraform/datadog/bbs.tf
```diff
@@ -22,3 +22,20 @@ resource "datadog_monitor" "bbs_lock_held_once" {
...
+
+resource "datadog_monitor" "bbs_healthy" {
+  name                = "${format("%s bbs healthy", var.env)}"
+  type                = "query alert"
+  message             = "${format("BBS health check failed. Check BBS status immediately. @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
+  escalation_message  = "BBS health check still failing. Check BBS status immediately."
+  notify_no_data      = true
+  require_full_window = false
+
+  query = "${format("min(last_1m):min:cf.bbs.Healthy{bosh-deployment:%s} < 1", var.env)}"
```
dcarley: Can you explain to me why we use `bosh-deployment` here but `deployment` in `tags` below?

Why do other monitors use `deploy_env` instead?

alext: IMO we should use `deploy_env` everywhere we can. At present, the CF bosh deployment name happens to be the same as `deploy_env`, but if we decide to split things in future, this will no longer necessarily be the case.

manifests/cf-manifest/manifest/010-cf-jobs.yml
combor: So we have two checks running on database VMs both submitting `cf.bbs.Health` and `cf.bbs.ResonseTime`. That's fine as check is using `bbs.service.cf.internal:8889` to connect.

combor: This also prevents any weird behaviour during re-deployment.

dcarley: After some testing, I'm not confident that `database/0` is submitting metrics. The monitor reported `0` when I stopped `bbs` on `database/0` and left it running on `database/1`. But reports `no data` when I stopped it on `database/1` and left it running on `database/0`. Worth noting that `cfdot` still works.

mtekel: BBS only has one process active at a time. The check goes to official bbs url. So both machines are checking active bbs. When you stop one, BBS fails over and updates consul, which in turn updates `bbs.service.cf.internal` record to point to correct VM. To make check fail, you need to stop both bbs instances

dcarley: > So we have two checks running on database VMs both submitting cf.bbs.Health and cf.bbs.ResonseTime

What happens if one can connect and the other one can't?

> To make check fail, you need to stop both bbs instances

I'm getting `no data` when stopping just one of them.

mtekel: If only one connects, you should still get data. I think polling interval is 10 or 15s.

mtekel: btw., don't look at the monitor, look at the metric. Datadog still has this bug with no_data on monitors

Also, you can try deleting the monitor and creating it again by running cf deploy again, which would run TF task to create monitors

dcarley: The metric explorer showed an absence of datapoints for that period of time too. I haven't been able to reproduce it since, but I now understand that it's not related to both machines submitting.

Piotr and I used `iptables` to prevent one machine from connecting to the active BBS. The monitor triggered successfully, because `min()` selected the `0` metric from the non-active machine rather than the `1` metric from the active machine.

manifests/cf-manifest/manifest/010-cf-jobs.yml
```diff
@@ -222,6 +224,9 @@ jobs:
       - name: cf
     update:
       serial: true
+    properties:
+      bbs:
+          url: (( concat "https://" meta.bbs.api_location "/v1/actual_lrp_groups/list" ))
```
dcarley: Note (partly to self): I've tested this change from alphagov/paas-datadog-for-cloudfoundry-boshrelease#16 and it works. We'll need to update this PR with a new final release when merged.

dcarley:
Merging, but won't take effect until I deploy alphagov/paas-bootstrap#55 in all environments.

PR #820, opened 2017-03-07
[#141300765] Increase CF DB instance type 

keymon:
What?
------

We ran out of CPU credits in the CF RDS instance because of unexpectedly
high CPU usage. This caused CPU usage to be capped at 20% and it made
the system slow. We noticed most CPU was consumed by cloud controller
workers. They constantly poll the DB for new jobs to execute. We reduced
the number by reverting [1]

We change the instance type to a type that is not credit based.

[1]  https://github.com/alphagov/paas-cf/pull/817

How to review?
--------------

Code review

After merge
-----------

We suggest that the change is done manually with "apply immediatelly" in prod. We have already tested that the upgrade with no impact for the end user, as RDS will be upgraded in a HA way, but we want to babysit the upgrade anyway.

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/cloudfoundry/cf_rds.tf
+++ b/terraform/cloudfoundry/cf_rds.tf
@@ -40,7 +40,7 @@ resource "aws_db_instance" "cf" {
   allocated_storage    = 10
   engine               = "postgres"
   engine_version       = "9.5.4"
-  instance_class       = "db.t2.small"
+  instance_class       = "db.m3.medium"
   username             = "dbadmin"
   password             = "${var.secrets_cf_db_master_password}"
   db_subnet_group_name = "${aws_db_subnet_group.cf_rds.name}"


terraform/cloudfoundry/cf_rds.tf
```diff
@@ -40,7 +40,7 @@ resource "aws_db_instance" "cf" {
   allocated_storage    = 10
   engine               = "postgres"
   engine_version       = "9.5.4"
-  instance_class       = "db.t2.small"
+  instance_class       = "db.m4.large"
```
alext: Any reason to use a m4.large over an m3.medium? The m3.medium is roughly half the cost of the m4.large.

alext:
Note: This change should be merged on a Thursday, Friday or Monday so that the change gets applied in the correct sequence across CI, staging and prod due to the staggered maintenance windows (#732).

terraform/cloudfoundry/cf_rds.tf
keymon: >  Any reason to use a m4.large over an m3.medium? The m3.medium is roughly half the cost of the m4.large.

Not sure what is the point of trying to save costs with this DB.

saliceti: The cost is significant and `t2.medium` is sufficient.
As a reference, this developer has a similar configuration with all 3 DBs in the same RDS instance and they use `t2.medium`: https://github.com/cloudfoundry/bbs/issues/13#issuecomment-263539941

keymon: I wouldn't use a credit based instance type.

dcarley: > Not sure what is the point of trying to save costs with this DB.

Could you explain why?

> I wouldn't use a credit based instance type. If there is any issue related to the DB it will eventually run out of credits and we will be in the same situation.

I *think* @saliceti might have meant that the CPU and RAM specifications of `t2.medium` were sufficient, were it not throttled. So `m3.medium` would meet the same needs without being throttled.

keymon: > Could you explain why?

There are a lot of places were we do not worry about the costs. We have plenty of VMs running that could be merged and save costs. There are tenant DB instances running that are either oversized or not being used. 

But this DB can have direct impact on the platform performance and even availability. Why worry so much about the costs when we do not worry so much in other places.

Said that, it is fine, `m3.medium` looks good.

saliceti: Even throttled `t2.medium` is sufficient.

But I agree with @keymon, this is bikeshedding as there are many other places where we should look to save costs.

mtekel:
t2.medium:
 + 2 cpus, can handle 2 concurrent queries not blocking each other = lower latency
 + fast cpu when running on 100% burst (which should be 100% of time - we want to _never_ run out of credits)
 + higher base CPU performance when out of credits (40%)
 + cost ($112/month for HA pair)
 + large credit bucket (if we have monitoring and we trigger e.g. at 20% credits left, it will be more time than smaller t2s)

m3.medium:
  + not credit based
  - slower than t2.medium in normal use cases (when t2 is not throttled)
  ~ cost ($150/m for HA pair)

m4.large:
  + 2 cpus. Same advantages as t2.medium
  + 8G ram = more disk cache, can handle more connections. Gives us much bigger buffer in case we need it (e.g. emergency, attacks etc.)
  + faster cpus than m3
  ~ cost ($290/m for HA pair) - almost 2x more than m3, but you get more than twice ram and CPUs...

Given how complicated is this to upgrade, I'd go with m4.large. Or instead split uaa, bbs and api into separate instances. Running those at m3.medium would be more expensive though than having one shared m4.large.

IMO we should investigate why this is happening. It is possible that the instance is for some reason running out of memory (not enough for disk cache) - swap then increases and all this in turn increases CPU consumption over the t2.small long term limit (20%).

If costs are issue, then t2.medium is the best. It also has more memory, so if that is the problem it fixes that and double CPU performance when out of credits - which means the slowdown would be much less severe. On the other side, if we don't want to fiddle with instance types too much, I'd go m4.large, as it's the latest instance type with best performance, connectivity and capacity.

dcarley:
> We suggest that the change is done manually with "apply immediatelly" in prod.

Could you explain the steps to do this?

> We have already tested that the upgrade with no impact for the end user, as RDS will be upgraded in a HA way, but we want to babysit the upgrade anyway.

This should be clarified that there is no impact to applications running on the PaaS but there will be a downtime for API operations against the PaaS. Based on that, when would you suggest we do it?

keymon:
> Could you explain the steps to do this?

Go to the AWS console, CF DB instance, modify instance type, check "apply immediately"  and apply.

> This should be clarified that there is no impact to applications running on the PaaS but there will be a downtime for API operations against the PaaS. Based on that, when would you suggest we do it?

There will be no impact on the applications. Some API requests might fail during deployment. The API is not guaranteed to be online during deployments of the platform, and for other changes we did not send notifications.

bleach:
Cost calculations for t2.small to m4.large

Where | hourly cost increase per instance | monthly increase per env | monthly total increase
------- | --------- | --------| ---------
staging and prod (multi-AZ) | $0.324 (from $0.078 to $0.402) | $233.28 | $466.56
CI and dev (single AZ, ~ 15 envs) | $0.162 (from $0.039 to $0.201) | $116.64 | $1749.60
TOTAL |  |  |  $2216.16

dcarley:
@alext is going to do this while on support. We had a short discussion about:

- instance type: we're going to stick with `m3.medium`, but the commit message needs updating to reflect the change
- deployment: whether we could wait for this to deployed in the maintenance window now that we've mitigated the issue in #817, but we decided against it because we don't really understand the problem well enough and the maint window is over a week away. Alex is going to test the procedure for applying immediately - he's not sure whether it's possible to a) merge first, let it apply, and then change manually, or whether we need to b) pause the pipelines, change manually, merge and then unpause.

alext:
From testing, it works fine to let this apply first, and then make the same change in the console with "apply immediately" set, so I'm going to merge this and let it deploy, I'll then upgrade the instances with "apply immediately" manually once it's deployed.

PR #821, opened 2017-03-09
Add script and pipeline task for cf-psql

saliceti:
## What

I started writing this when I was on support. This could be useful to debug RDS.

Allows to open a psql shell on the CF RDS instance.
    
For security reasons, the connection is made with a specially crafted readonly user.
    
A funcion `pg_stat_allusers()` is created as a replacement for view `pg_stat_activity`. It is the same, with elevated privilege.

## How to review

* Deploy
* Run: `DEPLOY_ENV=XXX make dev cf-psql`
* Check you can connect to any database
* Check you can run select on any table
* Connect to `postgres` database (default). Check that `select * from pg_stat_activity;` doesn't show you the queries, but `select * from pg_stat_allusers();` does
* Check you can't delete, insert, update the data, delete database, etc

## Who can review
Not me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,9 @@ prod: globals ## Set Environment to Production
 bosh-cli: check-env ## Create interactive connnection to BOSH container
 	concourse/scripts/bosh-cli.sh
 
+cf-psql: check-env ## Create interactive connnection to the CF RDS instance
+	concourse/scripts/cf-psql.sh
+
 .PHONY: pipelines
 pipelines: check-env ## Upload pipelines to Concourse
 	concourse/scripts/pipelines-cloudfoundry.sh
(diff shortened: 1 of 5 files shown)


dcarley:
I'm not opposed to this, but..

I wonder whether this is an anti-pattern that we shouldn't encourage by making it easy. It seems like having a shell as the admin user leaves a lot of room for mistakes.

Could you explain the use cases where we've needed this and Datadog metrics haven't been sufficient?

How about using a read-only user instead?

saliceti:
I guess a readonly account would be sufficient as long as it has permissions on all databases, statistics, can do explain, etc.

This was particularly useful to have access when I was on support last time. Something was consuming CPU in RDS and only with psql access I was able to identify there were a lot of queries from the api workers, vi pg_stat_activity.

For the story [Investigate API workers DB CPU usage](https://www.pivotaltracker.com/story/show/141313847) this will be useful too because I would run EXPLAIN on the queries to get an even better understanding.

I used it too during the BBS outage to check if there was any locks, and to check activity in general.

It's been useful at times to check the content of the tables in the databases: events, routes, users, groups, LRPs, etc. So we can see the data without depending on an API to access it.

I can rewrite the PR to use a readonly user instead if it's useful.

saliceti:
Closing as it has not been very popular

PR #826, opened 2017-03-14
Follow pagination in show-cf-memory-usage

alext:
## What

We were previously only fetching the first page of organisations and
quotas. This hasn't been an issue up to now because we haven't exceeded
a single page (50 items).

We hit an issue testing previous updates to this on a dev environment
that had more than 50 quotas (left from acceptance tests) that meant it
couldn't find the quota for an org.

This adds a method to follow the `next_url` in paginated responses
until it's nil so that we get all data for these endpoints.

## How to review

* In a def environment create more than 50 orgs:
`for i in $(seq 1 55); do cf create-org test-$i; done`
* On master run the memory report: `make dev show-cf-memory-usage`
* Observe that not all orgs are returned.
* Run the report on this branch and verify that all orgs are now returned.

## Who can review

Anyone but myself.

### DIFF
--- a/scripts/show-cf-memory-usage.rb
+++ b/scripts/show-cf-memory-usage.rb
@@ -24,12 +24,23 @@ if File.exist?(config_path)
   end
 end
 
+# Fetch all resources from the path, iterating over all pages of responses.
+def cf_curl_all_resources(path)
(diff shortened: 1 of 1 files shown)


mtekel:
works, merging

PR #830, opened 2017-03-15
[#136027405] - Use two-factor authentication for CF admin accounts

henrytk:
## What

[#136027405](https://www.pivotaltracker.com/n/projects/1275640/stories/136027405)

We want to use two-factor authentication for our Cloud Foundry admin accounts, to enable us to eventually remove the API whitelist. UAA doesn't support this natively, so we have to use an external provider. This pull request provides the necessary code to configure Google as an OpenID Connect provider and migrate our admin users to ones without passwords.

## How to review

This review will be fun. Ultimately what we need to review is that all admin accounts (except the main admin account) are forced to use two-factor authentication. Other things to consider are:

* whether we can still log in using the main admin account if our 2FA has issues.
* that old accounts are not still active.

### Before merging

* You will need to wait until [credentials#61](https://github.gds/government-paas/credentials/pull/61) is merged and then update your local paas-pass. This will give you the credentials you need to upload with:

  ```
  DEPLOY_ENV="$DEPLOY_ENV" make dev upload-google-oauth-secrets
  ```
  
  We will need to do this for all of the persistent environments before merging.

* If you are testing in a dev environment you will need to authorise the redirect URI in the [Google developer console](https://console.developers.google.com/apis/credentials?project=govuk-paas). While you are in there you can take a look at the redirect URIs configured for the ci, staging, and production credentials to make sure they look correct, as there is no easy way to test these.

The following is an example of how you could test this with a deployment:

* deploy from master and manually create a regular user using your email address. Check you can log in with the password.
* edit the temporary commit to include your user details. Make sure you use a real email address. Make sure you use the same `origin` as the other users (google).
* re-deploy using this branch. You should get an email notifying you of the new account and your old account should be deleted (it will use the same email address as a username).
* make sure you cannot login using the password anymore.
* check you can log in using Google. There is [a bug with UAA we have raised an issue for](https://github.com/cloudfoundry/uaa/issues/562), so for now you will have to follow the horrible workaround (see below).
* **IMPORTANT:** remove the temporary commit

### The horrible workaround

```
cf login -a https://api.<system_domain> -u your.email@example.com --sso [--skip-ssl-validation ]
```

You will be presented with a command prompt to enter a one-time password:

```
One Time Code ( Get one at https://login.<system_domain>/passcode )> 
```

Open a browser and visit the URL above. Click on the 'Login with Google' link. You will see an error. This is because there is a bug in UAA resulting in the incorrect `redirect_uri` query parameter. It is missing the domain of the login page. Edit the parameter to add the missing `login.<system_domain>%2F`, so this:

```
https://accounts.google.com/o/oauth2/v2/auth?client_id=REDACTED&response_type=code&redirect_uri=https%3A%2F%2Flogin%2Fcallback%2Fgoogle&scope=openid+email&nonce=REDACTED
```

becomes something like this:

```
https://accounts.google.com/o/oauth2/v2/auth?client_id=REDACTED&response_type=code&redirect_uri=https%3A%2F%2Flogin.henry.dev.cloudpipeline.digital%2Flogin%2Fcallback%2Fgoogle&scope=openid+email&nonce=REDACTED
```

It should then do the normal login process and give you the one-time password to enter on the command line.

## Who can review

Anyone with access to the Google developer console, which at the moment I think is just me and @bleach - so we will have to change that.


dcarley:
Should we remove the password and encryption functionality from `sync-admin-users` (and underlying libs) now that we don't need it?

henrytk:
@dcarley yes, we should. I originally left it in because it supports the review process, but I can remove it and add it back in as a temporary commit.

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -361,6 +361,25 @@ properties:
     links:
       passwd: (( concat "https://login." properties.system_domain "/forgot_password" ))
       signup: (( concat "https://login." properties.system_domain "/create_account" ))
+    oauth:
+      providers:
+        google:
```
paroxp: I believe, we'd need to edit all entries in the database and rename their provider name from `google` to `google_gds`.

Agree, it should be a more related alias name.

scripts/upload-google-oauth-secrets.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
paroxp: Totally agree. As hector would say, we could solve it with KV Store :D

But, I don't think it should be all combined at the current state. Let's say, I'd like to test DataDog on my Dev environment. Do I need to upload all possible credentials, like Google Console API Keys, to my "private" bucket, if I don't need them?

manifests/cf-manifest/manifest/020-cf-properties.yml
henrytk: The script will delete the users and recreate them if it detects a change in origin, which is determined by what we call it. As Rafal points out we will then change the name (in our UAA configuration). However, UAA puts the name in the redirect URIs, and the redirect URIs have to be whitelisted in the Google developer console. This means manually updating in the console, but you can have multiple whitelistings for each set of OAuth credentials, so you can dual run the old and new names.

bleach: I think you're saying that it's not that hard to change later and I agree now :)

bleach:
I've tested this now and although the workaround is horrid, we've agreed to merge now.

@paroxp @henrytk I've removed the temporary commits. Can you have a quick look over it and check that all is as expected please?

henrytk:
@bleach looks as expected. The only unknown is the CF upgrade, which the rebase against master has thrown into the mix. Exciting.

PR #833, opened 2017-03-16
[#141394313] Upgrade ipsec release to disable stats logging

dcarley:
## What

Reduce the amount of logs generated by racoon which are filling up our
logsearch cluster. Depends on alphagov/paas-ipsec-release#2

## How to review

1. Observe the vast amount of logs from racoon on the `cell` VMs, either:
    - in logsearch with query `@source.component:racoon`, but I had problems in my dev environment
    - in `/var/log/syslog` on the VM itself
1. Deploy from this branch
1. Observe the vast reduction in logs from racoon in the same places as previous

## Who can review

Not @dcarley


dcarley:
I've rebased with the final build.

PR #836, opened 2017-03-20
[#139196683] Use deploy_env tag in bbs monitor definition

mtekel:
## What

Before the tag didn't seem to be available on the metric (probably because of issues fixed by https://github.com/alphagov/paas-datadog-agent-boshrelease/pull/10).

The tag is now present, so we can use like we do in all other monitors.

## How to review

Deploy. See TF apply change to monitor. Check the monitor - the state should not change. Stop your dev bbs servers to trigger the monitor to verify it still works.

## Who can review

not @mtekel


combor:
LGTM

PR #837, opened 2017-03-20
[support] Remove "router state" messages from cell monitors

dcarley:
## What

These descriptions are misleading because the presence of the `rep` or
`guardian` processes shouldn't be affected by the routers. I think they may
have been copy/pastad from another monitor by accident.

## How to review

Code review only.

## Who can review

Not @dcarley


mtekel:
works, merging

PR #839, opened 2017-03-21
[#136027405] Fix dev oauth property

combor:
## What

This implements a condition that checks if creation of admin users is
enabled and adds one more stub to the spruce command.

Oauth config has been moved to separate file to include it only when we
want to create admins.

## How to review

Deploy on dev and check if it finishes OK

## Who can review

Not @combor


mtekel:
works as expected, merging

PR #840, opened 2017-03-21
[#141948375] Split Consul client and server checks 

henrytk:
## What

Currently we run the Consul server checks on all VMs, because the datadog-consul job in the datadog-for-cloudfoundry release couples the two together. They have been split into datadog-consul-agent-client and datadog-consul-agent-server jobs in the release as part of https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/18. We leave the client checks on all VMs, but only include the server check for the VMs which are part of the Consul cluster.

## How to review

First, this pull request must be merged: https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/18 is being reviewed. Once that is done the temporary commit should be replaced with one that points to the final release. The commits may or may not be squashed together, I am not fussy.

The other pull request should involve reviewing that the new release actually puts the right configuration files in the right places, so I will leave it up to the reviewer to decide whether code review is sufficient here.

## Who can review

Anyone but me.


mtekel:
works fine, merging

PR #842, opened 2017-03-22
[#142128979] Add DWP Carer's Allowance IP to the whitelist

paroxp:
## What

As part of the #142128979, we'd like to add the DWP Carer's Allowance IP to our whitelist. This should be followed by creation of the organisation and users, for them to try out the PaaS.

## How to review

Make sure the correct IP address has been added.

## Who can review

Not @paroxp


combor:
LGTM

PR #843, opened 2017-03-22
[#142059793] Remove Hector from admin users and commit signers

paroxp:
## What

Hector has left the team. His accounts in staging and production will be manually cleaned up.

## How to review

Check if correct data has been removed.

## Who can review

Not @paroxp


dcarley:
👍  confirmed key. Will merge when Travis has finished.

😢  👋  @keymon

PR #844, opened 2017-03-22
[support] show-cf-memory-usage: Link to memory graph

dcarley:
## What

Make it easier for people to know what they should compare the "memory
required" number against. I've previously run the script but not verified
whether we are meeting ADR017 because it was too much effort to look.

The link shows a sum of all cells by environment. I was going to hardcode it
to `bosh-deployment:prod` since that's the environment that we most often
add orgs and cells to, but the script can be used against any environment.

## How to review

1. Checkout the branch
1. Run `make show-cf-memory-usage`
1. Click the link and verify that it shows useful information

## Who can review

Not @dcarley

### DIFF
--- a/scripts/show-cf-memory-usage.rb
+++ b/scripts/show-cf-memory-usage.rb
@@ -1,6 +1,14 @@
 #!/usr/bin/env ruby
 
 require 'json'
+require 'uri'
+
+datadog_url = URI("https://app.datadoghq.com/metric/explorer")
+datadog_url.query = URI.encode_www_form(
+  exp_agg: "sum",
+  exp_metric: "system.mem.total",
+  exp_scope: "bosh-job:cell,bosh-deployment:#{ENV.fetch('DEPLOY_ENV')}",
+)
 
 config_path = File.expand_path('~/.cf/config.json')
 if File.exist?(config_path)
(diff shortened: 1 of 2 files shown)


scripts/show-cf-memory-usage.rb
```diff
@@ -76,3 +76,4 @@ def format_memory(amount)
 puts "Memory reserved by apps: #{format_memory(apps_reserved_memory)}"
 puts
 puts "Total cell memory required to meet ADR017: #{format_memory(required_cell_memory)}"
+puts "Compare this against: https://app.datadoghq.com/metric/explorer?tile_size=s&exp_agg=sum&exp_metric=system.mem.total&exp_scope=bosh-job%3Acell&exp_group=bosh-deployment"
```
alext: Any reason to use this instead of https://metrics.cloud.service.gov.uk/dashboard/file/cell-memory-usage.json ? I find that Datadog page pretty confusing because it's got graphs for all the deployments.

dcarley: Did you see the PR description about hardcoding the environment? We could do that for DataDog.

I originally chose DataDog over Grafana because I'm normally already logged into DataDog and my individual account credentials are easy to access than the shared Grafana login.

alext: Given this is run from the Makefile, it should be possible to parameterise it using `$DEPLOY_ENV` or `$SYSTEM_DNS_ZONE_NAME`.

dcarley: I'll have a look at how many 🐂  (closest I could find) are involved.

dcarley:
Closing while I rework.

PR #845, opened 2017-03-23
[#142128979] Add additional IP addresses for DWP CA

paroxp:
## What

After the communication on DeskPro, Muhammad requested two additional IP
addresses to be registered on our whitelist.

Please, follow the conversation on DeskPro with ID 4328.

## How to review

Define if these IPs belong to DWP CA.

## Who can review

Not @paroxp


mtekel:
Interestingly, only 1st IP belongs to Healthcare Commission, other two are Atos. Which is OK, given I might know that Atos is working on this project as well. Merging.

PR #847, opened 2017-03-23
[#142128979] Increase number of cells

paroxp:
## What

In order to support more load and ensure enough cells are up, we'd like
to add aditional cell to each zone, while adding DWP Carer's Allowance.

Change requested by Dan in #142128979.

## How to review

Assess the need and sanity of increase.

## Who can review

Not @paroxp


mtekel:
LG, merging

PR #848, opened 2017-03-23
[#142042903] Fix no_data in failed job and error count monitors

mtekel:
## What
Apply rollup function, because, as responded by DataDog support:

>The issue on the initial monitors was that during evaluation, the timeshift
that necessarily operates on a change monitor, couldn't find datapoints that
were there if the datapoints were not submitted at the exact timeshift
timestamp of the monitor:

>    Let's say the timeshift was looking for a datapoint at 8:34:25 am , and
you had one at 8:34:23 and one at 8:34:29, you'd artificially get a no data
result.

>Appending the rollup(avg,30) function is correcting that in the way that it
will average the points buckets over a 30 sec timeframe, so you'd always
have data.

## How to review

Apply to your DD enabled environment. Observe monitors being updated. Observe monitors not having no_data issues any more (give it few hours - but change should be visible very soon given how much no_data there is on the monitors before fix).

## Who can review

not @mtekel

### DIFF
--- a/terraform/datadog/cloud_controller.tf
+++ b/terraform/datadog/cloud_controller.tf
@@ -60,7 +60,7 @@ resource "datadog_monitor" "cc_failed_job_count_total_increase" {
   escalation_message  = "Amount of failed jobs in Cloud Controller API still growing considerably, check the API health."
   require_full_window = false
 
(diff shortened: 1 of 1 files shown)


saliceti:
LGTM

PR #849, opened 2017-03-23
[#141785699] Fix ssh upload

saliceti:
## What
Story: [Resolve issues with large database imports](https://www.pivotaltracker.com/story/show/141785699)

Fix the issue reported by Notify that prevents them to upload more than 1GB data via ssh. The issue was reported here: https://github.com/cloudfoundry/diego-ssh/issues/31
A PR was raised upstream to update the crypto library globally in diego-release: https://github.com/cloudfoundry/diego-release/pull/288
For our issue we only need to update the ssh_proxy and file_server jobs.
An acceptance test covers uploads 1.9GB to cover the fix we've made. Another issue prevents us from uploading more than 2GB via `cf` command line: https://github.com/cloudfoundry/cli/issues/1098

We should increase the limit when it's fixed.

## How to review
* Install [pv](https://linux.die.net/man/1/pv)
* Test ssh against a vanilla environment. This should fail:

```
$ dd if=/dev/zero count=1040 bs=1024k | pv | cf ssh healthcheck -c "cat > /dev/null"
```
* Deploy
* Restage the healthcheck app
* Test again, it should work now
* Check it fails above 1GB

## Who can review

Not me

### DIFF
--- a/platform-tests/src/acceptance/cf_ssh_test.go
+++ b/platform-tests/src/acceptance/cf_ssh_test.go
@@ -1,12 +1,25 @@
 package acceptance_test
 
 import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gbytes"
 	. "github.com/onsi/gomega/gexec"
+)
+
+const (
+	BYTE     = int64(1)
+	KILOBYTE = 1024 * BYTE
+	MEGABYTE = 1024 * KILOBYTE
+	GIGABYTE = 1024 * MEGABYTE
+	TERABYTE = 1024 * GIGABYTE
 )
 
 var _ = Describe("CF SSH", func() {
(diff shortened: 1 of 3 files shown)


platform-tests/src/acceptance/cf_ssh_test.go
```diff
@@ -24,4 +37,41 @@ var _ = Describe("CF SSH", func() {
 		Expect(cfSSH).To(Exit(0))
 		Expect(cfSSH).To(Say("load average:"))
 	})
+
+	It("allows uploading a large payload via standard ssh client", func() {
+		// FIXME: Increase to 10GB once the following issue is solved:
+		// https://github.com/cloudfoundry/cli/issues/1098
+		const payload = 1*GIGABYTE + 900*MEGABYTE
```
alext: `payloadSize` would be a better name for this constant.

platform-tests/src/acceptance/cf_ssh_test.go
```diff
@@ -24,4 +37,41 @@ var _ = Describe("CF SSH", func() {
...
+
+		file, err := os.Open("/dev/zero")
+		Expect(err).NotTo(HaveOccurred())
+
+		go func() {
+			defer sshStdin.Close()
+			defer GinkgoRecover()
+			copied, error := io.CopyN(sshStdin, file, payload)
+			finalString := fmt.Sprintf("Successfully copied %d bytes", copied)
+			io.WriteString(GinkgoWriter, finalString)
```
alext: These 2 lines could be replaced with one using `fmt.Fprintf` directly to the `GinkgoWriter`.

platform-tests/src/acceptance/cf_ssh_test.go
```diff
@@ -24,4 +37,41 @@ var _ = Describe("CF SSH", func() {
...
+		sshStdin, err := cfSSHCommand.StdinPipe()
+		Expect(err).NotTo(HaveOccurred())
+
+		file, err := os.Open("/dev/zero")
+		Expect(err).NotTo(HaveOccurred())
+
+		go func() {
+			defer sshStdin.Close()
+			defer GinkgoRecover()
+			copied, error := io.CopyN(sshStdin, file, payload)
```
alext: `error` is a built-in type, so we should probably avoid using a variable with the same name as that can get confusing.

platform-tests/src/acceptance/cf_ssh_test.go
```diff
@@ -24,4 +37,41 @@ var _ = Describe("CF SSH", func() {
...
+			finalString := fmt.Sprintf("Successfully copied %d bytes", copied)
+			io.WriteString(GinkgoWriter, finalString)
+			Expect(error).NotTo(HaveOccurred())
+			Expect(copied).To(Equal(payload))
+		}()
```
alext: Is this separate Goroutine necessary? Could this not be done between the `Start`, and the `Wait` below? I think that would make the flow easier to follow.

saliceti:
This should be ok now

PR #852, opened 2017-03-24
Fix the counters on window with certain width

paroxp:
## What

Now that we have a third counter available, we may meet the case where all three will need to be displayed on the dashboard. This will cause some layout disturbances and could be fixed by making the counters smaller on a window below 544 pixels wide.

## How to review

1. Run the dashboard locally
2. ~Cause many DataDog monitors to fail on one of our environments~ or remove `data-hideif="X | equals 0"` from `widgets/health/health.html` lines `9`, `15` and `21`
3. Check the behaviour on different window widths - You need to refresh every time you set a new window width

## Who can review

Not @paroxp

### DIFF
--- a/tools/paas_dashboard/widgets/health/health.scss
+++ b/tools/paas_dashboard/widgets/health/health.scss
@@ -6,6 +6,8 @@ $warning: #F25F00;
 $failure: #AF0900;
 $error: #5B0274;
 
+$small: 34em;
+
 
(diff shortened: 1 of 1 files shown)


mtekel:
It's a small step for this dashboard, but big step for overall esthetic impression of tv display we have in the office. Now if we can only match the color shades and fonts with fourth wall...

PR #851, opened 2017-03-23
Fix admin users login with OpenID connect 

bleach:
## What

Fix some problems that we found after merging the original PR

### Unable to validate google JWT signatures

CF v253 introduced a UAA change that appears to require a new `tokenKeyUrl` property to be set to allow Google's signatures to be validated. A [bug](https://github.com/cloudfoundry/uaa/issues/565) exists for this upstream.

### Passcode issuance requires sticky sessions

The user visits login.domain/passcode to get a one-time code to use in the CLI, they then get redirected to /login, visit google to login and then return to access the passcode page.

This requires sticky sessions to be enabled on the UAA load balancer because the the URL that a user was intending to visit is saved in a server side session and there does not seem to be a way of sharing the sessions between UAA instances.

The expectation from the UAA team seems be to that it will be deployed behind gorouter, which automatically uses sticky sessions when it sees a `JSESSIONID` cookie. https://docs.cloudfoundry.org/concepts/http-routing.html 

We switched away from gorouter load balancing of UAA (and all CF endpoints) in #196 

In order to fix this, an ELB stickiness policy has been defined, based on the UAA JSESSIONID cookie. I don't believe this will have any effect on most of our requests through the ELB, as most of them will be from the CLI.

## How to review

- Ensure that your environment creates admin users:
```
export NEW_ACCOUNT_EMAIL_ADDRESS='the-multi-cloud-paas-team+dev@digital.cabinet-office.gov.uk'
make dev pipelines
```
- [optional] Out of politeness you may add a temporary commit to remove your colleagues from `config/admin_users.yml` to ensure they don't receive confusing emails
- Ensure you have pulled `paas-pass` / credentials recently and run `make dev upload-google-oauth-secrets`
- Add `https://login.$DEPLOY_ENV.dev.cloudpipeline.digital/login/callback/google` to the list of _Authorised redirect URIs_ for dev in the [Google API developer console](https://console.developers.google.com/apis/credentials?project=govuk-paas)
- Deploy this branch
- Check your email, including your spam folder for an email saying an account has been set up
- Follow the workaround instructions [here](https://github.com/alphagov/paas-team-manual/blob/d551300451c4c6cc43106ab5248006b07e80b2c2/docs/guides/cf_admin_users.md) to login, working around a UAA bug

## Who can review

Anyone but me.

PR #853, opened 2017-03-27
WIP: [#136027405] Script to workaround SSO redirect_uri bug

dcarley:
## What

Paste the broken URL and click on the corrected URL:

    ➜  paas-cf git:(bugfix/136027405-fix_sso_url_script) ✗ make dev fix_sso_url
    Paste URL:
    https://accounts.google.com/o/oauth2/v2/auth?client_id=REDACTED.apps.googleusercontent.com&response_type=code&redirect_uri=https%3A%2F%2Flogin%2Fcallback%2Fgoogle&scope=openid+email&nonce=REDACTED

    Corrected URL:
    https://accounts.google.com/o/oauth2/v2/auth?client_id=REDACTED.apps.googleusercontent.com&response_type=code&redirect_uri=https%3A%2F%2Flogin.dcarley.dev.cloudpipeline.digital%2Flogin%2Fcallback%2Fgoogle&scope=openid+email&nonce=REDACTED

## How to review

I used this myself while reviewing #851 because I found modifying the URL by hand was cumbersome. If it's useful to anyone else when I'll write tests and remove the WIP label. If it's not then I'll close the PR after a week or so.

1. Run it as shown above
1. Feedback whether it's useful

## Who can review

Not @dcarley

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,3 +234,6 @@ shake_concourse_volumes: check-env ## Restarts concourse services and workers an
 show-cf-memory-usage: ## Show the memory usage of the current CF cluster
 	$(eval export API_ENDPOINT=https://api.${SYSTEM_DNS_ZONE_NAME})
 	@./scripts/show-cf-memory-usage.rb
+
+fix_sso_url: check-env
+	@./scripts/fix_sso_url.rb
--- a/scripts/fix_sso_url.rb
+++ b/scripts/fix_sso_url.rb
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+# Temporary workaround for:
+# - https://github.com/cloudfoundry/uaa/issues/562
+# - https://github.com/cloudfoundry/uaa/pull/575
+
+require 'uri'
+
+REDIRECT_HOSTNAME = "login.#{ENV.fetch('SYSTEM_DNS_ZONE_NAME')}".freeze
+
+puts "Paste URL:"
+url = URI.parse(gets.chomp)
+
+query = Hash[URI.decode_www_form(url.query)]
+redirect_url = URI.parse(query.fetch('redirect_uri'))
+redirect_url.path = "/#{redirect_url.host}#{redirect_url.path}"
+redirect_url.host = REDIRECT_HOSTNAME
+query['redirect_uri'] = redirect_url
+url.query = URI.encode_www_form(query)
+
+puts
+puts "Corrected URL:"
+puts url


Makefile
```diff
@@ -234,3 +234,6 @@ shake_concourse_volumes: check-env ## Restarts concourse services and workers an
 show-cf-memory-usage: ## Show the memory usage of the current CF cluster
 	$(eval export API_ENDPOINT=https://api.${SYSTEM_DNS_ZONE_NAME})
 	@./scripts/show-cf-memory-usage.rb
+
+fix_sso_url: check-env
```
dcarley: This `check-env` makes it a bit annoying to use because of the AWS credentials check - you don't need AWS credentials to run this and it slows down the execution by a second or tow.

saliceti:
It works fine.
However to be able to run this script I also needed a second terminal, DEPLOY_ENV and AWS credentials so it's a bit cumbersome too. So I'm not sure of the value right now.

scripts/fix_sso_url.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'uri'
+
+REDIRECT_HOSTNAME = "login.#{ENV.fetch('SYSTEM_DNS_ZONE_NAME')}".freeze
+
+puts "Paste URL:"
+url = URI.parse(gets.chomp)
+
+query = Hash[URI.decode_www_form(url.query)]
+redirect_url = URI.parse(query.fetch('redirect_uri'))
+redirect_url.path = "/#{redirect_url.host}#{redirect_url.path}"
```
timmow: Its a little bit confusing why we are using the `redirect_url.host` part here - probably doesn't matter that much but anyone trying to follow the code could get a bit confused.

Makefile
timmow: It works without the `check-env` - unless i'm missing something can we remove it?

timmow:
I thought it was useful while approving the story

Makefile
dcarley: It needs `SYSTEM_DNS_ZONE_NAME` and `check-env` is currently our only way of (indirectly) checking whether one of the environment targets have been run.

scripts/fix_sso_url.rb
dcarley: The redirect URI gets munged to `https://login/callback/google` so we have to extract part of the path from the hostname.

dcarley:
I've found this more cumbersome than useful. Closing.

PR #854, opened 2017-03-28
[#142302879] Increase default number of cells

combor:
## What

Due to performance problems during acceptance tests execution we decided
to increase number of cells on CI and STAGING environments to cope with
the load and allow tests to finish undisrupted. During
investigation with Pivotal we have discovered that metron is dropping
messages and UDP packed are being dropped by kernel as the load on cells
was extremely high.

## How to review

Sanity check

## Who can review

Not @combor

### DIFF
--- a/manifests/cf-manifest/manifest/env-specific/cf-default.yml
+++ b/manifests/cf-manifest/manifest/env-specific/cf-default.yml
@@ -1,4 +1,4 @@
 ---
 meta:
   cell:
-    instances: 2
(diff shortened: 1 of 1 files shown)


dcarley:
I think we should change the default and apply it to dev environments as well. Rationale being that we only run cf-acceptance-tests in dev and CI, and we've had trouble trying to reproduce this problem in dev, which would have been harder still if dev and CI diverge.

PR #857, opened 2017-03-30
[#142302879] Provision 6 cells

combor:
## What

As we still experience intermittent acceptance tests failures and one of
the reasons might be load related we have decided to try to use more
cels and spread the load.

## How to review

Sanity check

## Who can review

Not @combor


dcarley:
Could you amend the commit message subject to clarify that it's only for CI?

combor:
Done.

PR #858, opened 2017-03-30
[#142165547] Increase elasticsearch disk space

henrytk:
## What

⚠️ Make sure you remove the temporary commit before merging. No, seriously.

[#142165547](https://www.pivotaltracker.com/n/projects/1275640/stories/142165547)

This is based on the original work done in https://github.com/alphagov/paas-cf/pull/658, but since then the manifest test layout has changed (cloud config and manifest were split), so I have re-implemented it on this branch.

There's currently an open alert for elasticsearch disk space, which is ~78% used. Usage is not increasing significantly. This introduces extra disk capacity in prod only. See commit messages for full details.

## How to review

If you want to deploy:

1. deploy from master
1. there is a temporary commit on this branch. Deploy from this branch and it should increase the disk sizes in your elasticsearch_master VMs. The cell instance count is reduced because we are currently hitting our limit of r4.xlarge instances in our dev account. The temporary commit will make the Travis tests fail.
1. Remove the temporary commit and make sure Travis tests pass.

If you want to be thorough you can first deploy without the temporary commit to make sure we are not impacting non-production environments, but I will leave that in the hands of the reviewer.

### Before merging

* ⚠️ Make sure you remove the temporary commit

## Who can review

Anyone but me.


henrytk:
I've marked this as WIP while I rebase against master to include the changes in https://github.com/alphagov/paas-cf/pull/857

manifests/cf-manifest/spec/cloud-config/env_spec.rb
```diff
@@ -0,0 +1,16 @@
+
+RSpec.describe "Environment specific configuration" do
+  let(:default_cloud_config) { cloud_config_with_defaults }
+  let(:prod_cloud_config) { render_cloud_config("prod") }
+
+  def get_disk_size(cloud_config, disk_name)
+    cloud_config["disk_types"].select { |j| j["name"] == disk_name }.first["disk_size"]
+  end
+
+  it "should specify a larger elasticseach disk size in production" do
```
paroxp: /s/elasticseach/elasticsearch

alext: Syntax error in regex. I think you meant `s/elasticseach/elasticsearch/` :trollface:

henrytk: Fixed

PR #859, opened 2017-03-30
[#142066597] VPC peering

Jonty:
## What

This introduces VPC peering infrastructure to allow tenants access to their own VPC so they can expose services we don't have as brokers to their applications.

This also introduces the live configuration for DIT's VPC peering. Once this has been merged and deployed they should be notified that they need to accept the peering request and add the route.

It does this by having a per-environment json definition of the VPCs to peer with, along with who they belong to and what ranges to route. This json definition is then parsed during the pipeline and both a `tfvars` file is generated so that the VPC peering can be established & routing configured, and a CF manifest file is generated to configure a CF security group that can be bound to an organisation.

The workflow is as follows:

* We give a tenant a CIDR in the 172 range
* A tenant creates a VPC using this range, then gives us their account ID & VPC ID
* We add this information to our `prod.vpc_peering.json` file and deploy to production
* We bind the newly generated security group to the tenants organisations
* The tenant accepts the VPC peering request
* The tenant creates a route directing all 10.0.0.0/16 traffic to the new vpc peering
* Tenants now have access to their VPC in the orgs that are bound

## How to review

Peering requests between VPC's in the same account work the same way as peering requests between accounts, so for simplicity you can simply test in dev.

I have left a test peering VPC configured in dev, along with a deployed elasticache node in it. Please delete these when you are done testing. The VPC is `vpc-911a03f5`, and the redis instance is `peering-test.p9lva7.0001.euw1.cache.amazonaws.com`.

Create & commit `paas-cf/terraform/dev.vpc_peering.json` with the following contents in your dev env:
```
[
    {
        "peer_name": "peering_test",
        "account_id": "595665891067",
        "vpc_id": "vpc-911a03f5",
        "subnet_cidr": "172.16.1.0/24"
    }
]
```

Deploy.

Either deploy an application, or `cf ssh` to a container in the org you'll be testing with and `nc peering-test.p9lva7.0001.euw1.cache.amazonaws.com 6379`. It should fail to connect.

Bind the newly created security group `vpc_peer_peering_test` to an org, and `cf restart` the application you're testing with to have the group take effect.

Repeat the previous test, and `nc` should now connect.

## After deploy

Once this has been merged and deployed DIT should be notified that they need to accept the peering request and add the route. We also need to bind the group to their org.

## Who can review

Not @jonty.

### DIFF
--- a/terraform/cloudfoundry/vpc_peering.tf
+++ b/terraform/cloudfoundry/vpc_peering.tf
@@ -1,0 +1,27 @@
+resource "aws_vpc_peering_connection" "vpc_peer" {
+  count         = "${length(var.peer_vpc_ids)}"
+  peer_owner_id = "${element(var.peer_account_ids, count.index)}"
+  peer_vpc_id   = "${element(var.peer_vpc_ids, count.index)}"
+  vpc_id        = "${var.vpc_id}"
+}
+
+resource "aws_route" "vpc_peer_route_0" {
+  count                     = "${length(var.peer_vpc_ids)}"
+  route_table_id            = "${aws_route_table.internet.0.id}"
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
+  vpc_peering_connection_id = "${element(aws_vpc_peering_connection.vpc_peer.*.id, count.index)}"
+}
+
+resource "aws_route" "vpc_peer_route_1" {
+  count                     = "${length(var.peer_vpc_ids)}"
+  route_table_id            = "${aws_route_table.internet.1.id}"
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
+  vpc_peering_connection_id = "${element(aws_vpc_peering_connection.vpc_peer.*.id, count.index)}"
+}
+
+resource "aws_route" "vpc_peer_route_2" {
+  count                     = "${length(var.peer_vpc_ids)}"
+  route_table_id            = "${aws_route_table.internet.2.id}"
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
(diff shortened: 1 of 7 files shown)


terraform/prod.vpc_peering.json
```diff
@@ -0,0 +1,9 @@
+[
+    {
+        "peer_name": "dit",
+        "security_group": "dit",
```
alext: I don't see this value used anywhere. Is it still needed?

Jonty: No! My error, forgot to remove it.

terraform/cloudfoundry/vpc_peering.tf
```diff
@@ -0,0 +1,27 @@
...
+  peer_vpc_id   = "${element(var.peer_vpc_ids, count.index)}"
+  vpc_id        = "${var.vpc_id}"
+}
+
+resource "aws_route" "vpc_peer_route_0" {
+  count                     = "${length(var.peer_vpc_ids)}"
+  route_table_id            = "${aws_route_table.internet.0.id}"
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
+  vpc_peering_connection_id = "${element(aws_vpc_peering_connection.vpc_peer.*.id, count.index)}"
+}
```
alext: These are currently flapping (being added and removed on alternate terraform runs). This looks to be because we're combining `aws_route` resources with inline route definitions ([here](https://github.com/alphagov/paas-cf/blob/master/terraform/cloudfoundry/nat.tf#L16-L19)). This is documented to cause problems - https://www.terraform.io/docs/providers/aws/r/route.html

alext:
This can be rebased onto #861 which will fix the route flapping issues.

PR #861, opened 2017-03-31
[#142066597] Extract aws_route terraform resources

alext:
## What

Use `aws_route` resources for the `internet` route table 

Terraform doesn't allow mixing routes defined inline in an
`aws_route_table` resource with separate `aws_route` resources[1]. We
need to start using the latter to add routes for the VPC peering, so we
need to migrate this.

This is non-straightforward because if we remove the inline `route`
statement, terraform won't remove the routes, which means that when it
attempts to create the replacement `aws_route` resource, it gets a
conflict. This is solved by modifying the tfstate file before
running terraform.

[1]https://www.terraform.io/docs/providers/aws/r/route_table.html

**Note:** depending on timings, we may include this in the VPC peering PR on Monday, but this would still benefit from a separate review.

## How to review

Run this branch, and verify that the cf-terraform change is a no-op.
Run it a second time to verify that the migration step only happens once, and that it's still a no-op.

## Who can review

Anyone but myself or @Jonty

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -811,8 +811,13 @@ jobs:
                 mkdir generated-certificates
                 tar xzf cf-certs/cf-certs.tar.gz -C generated-certificates
 
+                # FIXME: remove this once it's run everywhere.
+                wget -O tf_aws_route_migrator https://github.com/alext/tf_aws_route_migrator/releases/download/0.0.1/tf_aws_route_migrator_linux_amd64
+                chmod +x tf_aws_route_migrator
+                ./tf_aws_route_migrator < cf-tfstate/cf.tfstate > migrated-cf.tfstate
+
                 terraform apply -var-file=paas-cf/terraform/{{aws_account}}.tfvars \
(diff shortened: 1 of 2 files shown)


combor:
Can I see source for https://github.com/alext/tf_aws_route_migrator/releases/download/0.0.1/tf_aws_route_migrator_linux_amd64 please?

alext:
> Can I see source for...

To capture conversation IRL:

The source for that is in the same repo that the release is hosted in - https://github.com/alext/tf_aws_route_migrator

combor:
We don't want to migrate those routes https://github.com/alphagov/paas-bootstrap/blob/master/terraform/vpc/subnet.tf#L8 ?

alext:
> We don't want to migrate those routes

Not at the moment. This is presently focussed on the necessary changes to get the VPC peering working. Generalising this would add significant complexity to both the implementation and deployment.

Once the VPC peering is rolled out, we could do some follow-up work to update those, but only if that's deemed to be a priority.

combor:
That looks good. Migration works ok and bosh runs of Terraform were no-op.

PR #865, opened 2017-04-03
[#141777063] Tests for VPC peering config generation

Jonty:
## What

These tests are mostly designed to spot incorrect .json VPC peer configurations, but they also minimally validate that the generators are working as expected.

## How to review

* Code review.
* Run the tests.
* Twiddle the `prod.vpc_peering.json` until each of them breaks, or create a new one that is interestingly broken.

## Who can review

Not @jonty.

### DIFF
--- a/terraform/scripts/generate_vpc_peering_test.go
+++ b/terraform/scripts/generate_vpc_peering_test.go
@@ -1,0 +1,188 @@
+package scripts_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+var _ = Describe("VPC peering", func() {
+	var examplePeerConfig = map[string]string{
+		"peer_name":   "cheese",
+		"account_id":  "123cheese",
+		"vpc_id":      "vpc123cheese",
+		"subnet_cidr": "0.0.0.0/32",
+	}
+
+	var files []string
+
+	BeforeEach(func() {
+		files, _ = filepath.Glob("../*.vpc_peering.json")
+	})
+
+	Context("config files", func() {
+		It("have all required values", func() {
+			for _, filename := range files {
+				file, _ := ioutil.ReadFile(filename)
+
+				var peers []map[string]string
+				err := json.Unmarshal([]byte(file), &peers)
+				Expect(err).To(BeNil(), "Config file: %s", filename)
+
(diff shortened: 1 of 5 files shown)


terraform/scripts/generate_vpc_peering_test.go
```diff
@@ -0,0 +1,138 @@
...
+		files, _ = filepath.Glob("../*.vpc_peering.json")
+	})
+
+	Context("config files", func() {
+		It("have all required values", func() {
+			type PeerConfig struct {
+				Peer_name   string
+				Account_id  string
+				Vpc_id      string
+				Subnet_cidr string
```
alext: It might be better to use idiomatic Go naming conventions here (eg `SubnetCIDR` etc), and to add json field tags to make it explicit which fields in the JSON doc are being read. Otherwise we may miss some case typos or similar things.

Jonty: Fair, I'll add field tags.

Jonty:
@alext I was hesitant to add a test like that as we'd said the tests were mostly to catch errors in the configuration rather than testing the scripts themselves, but if you think they're necessary I'm happy to add it.

Jonty:
@alext As discussed, I've reworked the json parsing and added tests that check the generators crash in every situation where a field is missing.

terraform/scripts/generate_vpc_peering_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"os/exec"
+	"path/filepath"
+)
+
+var _ = Describe("VPC peering", func() {
+	type PeerConfig struct {
+		PeerName   string `json:"peer_name"`
+		AccountID  string `json:"account_id"`
+		VpcID      string `json:"vpc_id"`
+		SubnetCIDR string `json:"subnet_cidr"`
```
alext: Gah! This is still not validating the case of the fields in the input. From [the docs](https://golang.org/pkg/encoding/json/#Unmarshal):

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match. Unmarshal will only set exported fields of the struct.

I think we'll need to use a `map[string]string` to achieve case validation. At least we can use [Gomega's collection assertions](http://onsi.github.io/gomega/#working-with-collections) for this (`Expect(foo).To(HaveKey("bar"))`, or even `Expect(foo).To(HaveKeyWithValue("bar", Not(BeEmpty()))`)

Jonty:
@alext And reworked so we're case sensitive, and a little cleaner.

PR #868, opened 2017-04-04
[#134175741] create-user.sh: Replace passwords with invites

dcarley:
## What

Use UAA's builtin mechanism for generating (but not sending) invites for new
accounts. A link with a one-time use code allows the user to set their own
password. The code is invalidated after first use. This is preferable to
generating passwords because we don't need to send them in plaintext email
and we don't need to recommend that people change their password after the
first login.

## How to review

Code:

1. Read the individual commits in order for more details about the implementation.
1. Forgive me for shell scripting atrocities.

Testing:

1. Deploy the branch to your dev environment
1. Run `./scripts/create_user.sh` to create yourself an account in your dev environment
1. Follow the link in the email that you receive
1. Confirm that you see a link to the documentation after setting your password
1. Confirm that you can login with the CLI using your new account
1. Run `./scripts/create_user.sh` with the `-r` option to send a new invite
1. Confirm that you can no longer login with the CLI
1. Confirm that the new invite works
1. Review alphagov/paas-tech-docs#32 also

Optional:

1. Confirm that it works with an SSO account in dev (I've done this once, but not recently)
1. Confirm that normal users don't have tokens with `scim.invite` scope, you can use:

       cf oauth-token | awk -F. '{print $2}' | (base64 -D; echo -n '}') | jq

## Who can review

@jollery or @tomskerous may want to review the email content.


scripts/create-user.sh
```diff
@@ -11,22 +11,24 @@ source "${SCRIPT_DIR}/common.sh"
...
+ - username: ${EMAIL}
+ - organisation: ${ORG}
 
- - login: ${EMAIL}
- - password: ${PASSWORD}
+Please use this link to activate your account and set a password. The link will only work once:
+${INVITE_URL}
+
+You can find advice about choosing a password:
+https://docs.cloud.service.gov.uk/#choosing-passwords
```
tomskerous: Suggest adding:

You can get support here:
https://www.cloud.service.gov.uk/support.html

scripts/create-user.sh
```diff
@@ -11,22 +11,24 @@ source "${SCRIPT_DIR}/common.sh"
...
 
 To get started, look at our Quick Setup Guide:
 https://docs.cloud.service.gov.uk/#quick-setup-guide
 
-You should make sure to change your password, as explained in the Quick Setup Guide.
-
 You can find our privacy policy here:
 https://docs.cloud.service.gov.uk/#privacy-policy
```
tomskerous: Suggest adding

You can get support here:
https://www.cloud.service.gov.uk/support.html

tomskerous: After "Regards" suggest

GOV.UK PaaS team.

PS Some departmental email systems will check links in inbound emails as part of their virus protection. This may have invalidated your one-time link. If this is the case please contact support to set your password another way.

dcarley: I'll do it in a separate PR because it's not strictly related to this change and I don't want it to drag out much longer.

dcarley: Done 👍

scripts/create-user.sh
```diff
@@ -138,19 +140,67 @@ create_org_space() {
...
+        -H "If-Match: *" \
+        -d "{\"verified\": false}" \
+        "${uaa_endpoint}/Users/${uaa_uuid}" >"${TMP_OUTPUT}"
     fi
   fi
 
-  if cf create-user "${EMAIL}" "${PASSWORD}" 2>&1 | tee "${TMP_OUTPUT}"; then
-    if ! grep -q "already exists" "${TMP_OUTPUT}"; then
-      USER_CREATED=true
+  if [[ -z "${uaa_uuid}" ]] && [[ "${RESET_USER}" == "true" ]]; then
```
henrytk: Under what circumstances would a user's ID be zero length?

scripts/create-user.sh
```diff
@@ -138,19 +140,67 @@ create_org_space() {
...
+  if jq -e '.resources | length == 1' "${TMP_OUTPUT}" >/dev/null; then
+    uaa_uuid=$(jq -er '.resources[0].id' "${TMP_OUTPUT}")
+
+    if [[ "${RESET_USER}" == "true" ]]; then
+      curl -sf "${ssl_arg}" \
+        -X PATCH \
+        -H "Authorization: ${auth_token}" \
+        -H "Accept: application/json" -H "Content-Type: application/json" \
+        -H "If-Match: *" \
+        -d "{\"verified\": false}" \
```
henrytk: The UAA API documentation says:

> True, if this user has verified her/his email address.

So the email verification and invite mechanism are coupled together?

henrytk: Had a chat with @dcarley offline about this. Setting it to `false` allows the invite mechanism to work for existing users.

henrytk:
I've rebased this against master.

henrytk:
I really don't think the [Get started](https://docs.cloud.service.gov.uk/#get-started) page is an appropriate landing page for somebody that has just finished a sign up journey, even though the title makes it sound like it is. I would expect a user to want to actually login at this point, so [Setting up the command line](https://docs.cloud.service.gov.uk/#setting-up-the-command-line) seems more appropriate, especially since the email doesn't give you the information required to log in using the newly created account.

However, this story has been going for a while, so I won't go on about it.

PR #872, opened 2017-04-06
Raise CC error alert threshold

Jonty:
## What

This is triggered during almost every deploy due to what appears to be a bug in cloud controller. This raises it to require around 5% of the requests-per-minute to be errors.

## How to review

Code review.


alext:
Is this still needed now that we've updated the metrics collector app?

Jonty:
@alext After checking the logs probably not, will close.

PR #877, opened 2017-04-12
Remove the API whitelist

Jonty:
## What

🐝 🐝 **THIS CAN BE REVIEWED, BUT DO NOT MERGE IT UNTIL AFTER EASTER WEEKEND**🐝🐝 

This duplicates the way we have set up access for `web_access_cidrs` for the API, removing the current set of whitelisted addresses and replacing it with 0.0.0.0/0. It also renames the configuration variable from `tenant_cidrs` to `api_access_cidrs` in order to bring it in line with `web_access_cidrs` and make the new purpose clearer.

I have not completely removed the whitelisting code because we still require the ability to toggle access between dev/prod, and I think that renaming things to make them clearer makes the .tfvars configuration easier to understand.

## How to review

🐝 🐝**THIS CAN BE REVIEWED, BUT DO NOT MERGE IT UNTIL AFTER EASTER WEEKEND**🐝 🐝  

Copy the `api_access_cidrs` block from `prod.tfvars` to your `dev.tfvars`. Deploy. Check you can hit the API when you're routed via an IP address that was previously not on the whitelist.

Please do not merge this until after the easter weekend, but you can review it at will.

### DIFF
--- a/terraform/cloudfoundry/security-groups.tf
+++ b/terraform/cloudfoundry/security-groups.tf
@@ -27,7 +27,7 @@ resource "aws_security_group" "cf_api_elb" {
 
     cidr_blocks = [
       "${compact(var.admin_cidrs)}",
-      "${compact(var.tenant_cidrs)}",
+      "${compact(var.api_access_cidrs)}",
       "${var.concourse_elastic_ip}/32",
       "${formatlist("%s/32", aws_eip.cf.*.public_ip)}",
     ]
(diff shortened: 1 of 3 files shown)


terraform/cloudfoundry/security-groups.tf
```diff
@@ -94,7 +94,6 @@ resource "aws_security_group" "web" {
 
     cidr_blocks = [
       "${compact(var.admin_cidrs)}",
-      "${compact(var.tenant_cidrs)}",
```
paroxp: Could you clarify, Is this one not getting replaced with the new `api_access_cidrs ` like the other two? Why?

Jonty: This is because this is about web control panel access and has `web_access_cidrs` on the line below it - previously when this was for `tenant_cidrs` it made sense to give tenants access to both simultaneously, but as this commit aims to make it clearer what the setting is for it didn't make much sense to have `api_access_cidrs` also govern access to web control panels when `web_access_cidrs` already exists for that purpose.

PR #878, opened 2017-04-12
[Support] Fix literal not terminated

combor:
## What

Lack of quotation made Terraform to fail miserably crying about
```
invalid value "paas-cf/terraform/prod.tfvars" for flag -var-file: Error
parsing paas-cf/terraform/prod.tfvars: At 82:20: literal not terminated
```

## How to review

Sanity check

## Who can review

Not me


dcarley:
Travis 😴

PR #880, opened 2017-04-13
[#support] Replace rotate-user-password with create-user.sh -r

combor:
## What

As `rotate-user-password.sh` has been deprecated in https://github.com/alphagov/paas-cf/commit/ab676558e78bffe476b555dbf20623e4be109904 we need to use `create-user.sh -r` instead.

## How to review

Sanity check.

## Who can review

Not me

### DIFF
--- a/scripts/reset-org.py
+++ b/scripts/reset-org.py
@@ -92,7 +92,7 @@ class OrgReset(object):
                 raise SubprocessException('Aborting: \'%s\' failed with exit code %d\n%s' % (err.cmd, err.returncode, err.output))
 
     def parse_current_user(self, target_output):
-        match = re.search(r'User:\s+(\S+)', target_output)
+        match = re.search(r'[Uu]ser:\s+(\S+)', target_output)
         return match.group(1)
 
 
--- a/scripts/reset-user-testing.sh
+++ b/scripts/reset-user-testing.sh
@@ -29,9 +29,9 @@ check_logged_in_cf "${TARGET_ENVIRONMENT_API}"
 	--space-developers \
 	--quota small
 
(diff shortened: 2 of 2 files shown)


dcarley:
This didn't work for me when the users don't already exist:
```
Running 'cf set-org-role  paas_user_research OrgManager'
Traceback (most recent call last):
  File "/Users/dcarley/projects/paas/paas-cf/scripts/reset-org.py", line 118, in <module>
    org_reset.set_roles()
  File "/Users/dcarley/projects/paas/paas-cf/scripts/reset-org.py", line 60, in set_roles
    self.run(['cf', 'set-org-role', user, self.org, 'OrgManager'])
  File "/Users/dcarley/projects/paas/paas-cf/scripts/reset-org.py", line 73, in run
    raise SubprocessException('Aborting: \'%s\' failed with exit code %d\n%s' % (err.cmd, err.returncode, err.output))
__main__.SubprocessException: Aborting: '['cf', 'set-org-role', '', 'paas_user_research', 'OrgManager']' failed with exit code 1
Assigning role OrgManager to user  in org paas_user_research as admin...
FAILED
Server error, status code: 404, error code: 20003, message: The user could not be found:
```

Rather than invest any more time in making it partially or fully support this use case, I think we should go back to how it was before with the multiple lines/commands and let it fail explicitly when the user doesn't exist.

combor:
Ok, so I simply replaced `rotate-user-password.sh` with `create-user.sh -r`. The later will fail here: https://github.com/alphagov/paas-cf/blob/master/scripts/create-user.sh#L176 if user doesn't exist and `reset-org.py` already passed CF error: `Server error, status code: 404, error code: 20003, message: The user could not be found:`

dcarley:
Worked 👍

PR #883, opened 2017-04-21
[#141315433] Monitor RDS instances for CPU credits

dcarley:
## What

Similar to 1f33b77 for EC2 instances. I omitted this from the first PR
because I thought we didn't have any RDS t2 instances, but BOSH has been
since alphagov/paas-bootstrap@043f3dd.

The use of the `deploy_env` tag depends on alphagov/paas-bootstrap#59

I've used `dbinstanceidentifier` as the group `by` because it's a mandatory
attribute for RDS instances, unlike `dbname` (Terraform's `name` attribute)
or `name` (which comes from a standard AWS tag).

The `!created_by:aws_rds_service_broker` will prevent us from ever being
warned about the performance of tenant database instances which aren't
relevant to the health of our platform. The filter isn't strictly necessary
at the moment because they aren't tagged with `deploy_env`, but it's
possible that they could be in the future.

## How to review

This is even more difficult to test than the last PR 🐼  

1. Review, merge, and deploy alphagov/paas-bootstrap#59 first
1. Upload secrets: `make dev upload-datadog-secrets`
1. Deploy this branch to your dev environment. 
1. Have a look in DataDog for a monitor called "RDS CPU credits" which is tagged with your dev environment but uses data from CI. Confirm that the query is correct by observing a single metric line for BOSH.
1. I **have not** tested triggering the monitor by exhausting the RDS instance credits. It's going to be tricky to do and disruptive to CI. Based on the previous PR I'm reasonably confident that it will work, but please test it or talk to me if you're unsure.
1. Manually delete the monitor in DataDog
1. ⚠️  Remove the WIP commit before merging ⚠️ 
1. Merge

## Who can review

Not @dcarley

### DIFF
--- a/terraform/datadog/aws.tf
+++ b/terraform/datadog/aws.tf
@@ -20,7 +20,7 @@ resource "datadog_monitor" "rds-cpu-credits" {
   type           = "query alert"
   message        = "${format("Instance is {{#is_warning}}low on{{/is_warning}}{{#is_alert}}out of{{/is_alert}} CPU credits and may perform badly. See: %s#cpu-credits @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.datadog_documentation_url, var.aws_account)}"
   notify_no_data = false
-  query          = "${format("avg(last_30m):avg:aws.rds.cpucredit_balance{deploy_env:%s,!created_by:aws_rds_service_broker} by {dbinstanceidentifier} <= 1", var.env)}"
+  query          = "${format("avg(last_30m):avg:aws.rds.cpucredit_balance{deploy_env:%s,!created_by:aws_rds_service_broker} by {dbinstanceidentifier} <= 1", "master")}"
 
   thresholds {
     warning  = "20"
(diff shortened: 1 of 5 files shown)


terraform/datadog/aws.tf
```diff
@@ -1,7 +1,7 @@
 resource "datadog_monitor" "ec2-cpu-credits" {
   name           = "${format("%s EC2 CPU credits", var.env)}"
   type           = "query alert"
-  message        = "${format("Instance is {{#is_warning}}low on{{/is_warning}}{{#is_alert}}out of{{/is_alert}} on CPU credits and may perform badly. See: https://government-paas-team-manual.readthedocs.io/en/latest/support/responding_to_alerts/#cpu-credits @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
+  message        = "${format("Instance is {{#is_warning}}low on{{/is_warning}}{{#is_alert}}out of{{/is_alert}} CPU credits and may perform badly. See: %s#cpu-credits @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.datadog_documentation_url, var.aws_account)}"
```
saliceti: There is a terraform variable `support_email` that you could use instead of `@govpaas-alerting-%s@digital.cabinet-office.gov.uk`

dcarley: I looked at that, but in production it points at DeskPro as is only used by the CVE monitor. I considered renaming it and adding another variable to replace all of the other address references but I'd already changed quite a few things in this PR.

PR #895, opened 2017-05-02
[#144443475] Simplify how we run tests

saliceti:
## What

Story: [Review how we run upstream acceptance tests](https://www.pivotaltracker.com/story/show/144443475)

* Stop faking errand environment to run acceptance and smoke tests and use the same approach as upstream
* Fix usage of temporary credentials
* Remove obsolete properties

## How to review

* Deploy and let all tests run, including continuous smoke tests
* Check `config.json` doesn't include the main admin credentials, it must be the temporary ones

## Who can review
Not me


platform-tests/upstream/run_acceptance_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+ln -s "$(pwd)/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests" "${CF_GOPATH}/cf-acceptance-tests"
+
+echo "Linking test artifacts directory"
+# FIXME: We would like to avoid writing in /var/vcap the artifacts directory
+# path is hardcoded. See: https://github.com/cloudfoundry/cf-release/issues/1197
+mkdir -p /var/vcap/sys/log
+ln -s "$(pwd)/artifacts" /var/vcap/sys/log/acceptance_tests
+
+echo "Sleeping for ${SLEEPTIME} seconds..."
+for i in $(seq $SLEEPTIME); do echo -ne "$i"'\r'; sleep 1; done; echo
```
dcarley: Small request:

Could you change this to `$(seq $SLEEPTIME 1)` so that it counts down instead of up?

It confuses me every time I see the output.

saliceti: Done 👍

dcarley: Fanks!!

saliceti:
Refactored how we generate the properties to allow reuse and remove dependency on cf-release templates.

dcarley:
Really good, thanks 👍 👍 👍

PR #898, opened 2017-05-03
[Support] Fix create-user to work with cf-cli 6.26

alext:
## What

With version 6.26 of the cli, the output from `cf target` has changed
slightly. The fieldname is now "user" (lowercase) instead of "User".
This was breaking the script.

Instead of just changing this, I've used the `tolower` awk function (as
described [here](https://unix.stackexchange.com/questions/273624/case-insensitive-search-in-awk)) so that it remains compatible with older versions of
the CLI.

## How to review

Run the script against a dev environment with both version 6.25 and 6.26 of the cf cli (the `--no-email` flag is probably useful for this).

## Who can review

Anyone but myself.


dcarley:
Works 👍

PR #899, opened 2017-05-04
[#144763021] Bump stemcell and cflinux2 rootfs to fix USNs

alext:
## What

This updates the stemcell to the latest in the series to pick up fixes for some USNs (see commit for details).

It also updates cflinux2-rootfs for the same reason (again see commit for details)

I've taken this opportunity to switch to the renamed boshrelease for cflinux2. This was [announced on cf-dev](https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/thread/XEX6Z2YMTYSPK3I6RQE6TQILKQESUDCB/), and the current release is deprecated, and will be discontinued at some point.

## How to review

Deploy from this branch and verify all works.

## Who can review

Not me.

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -17,10 +17,10 @@ releases:
     version: 1.4.0
     url: https://bosh.io/d/github.com/cloudfoundry/garden-runc-release?v=1.4.0
     sha1: 1d6020e761806d7f355ceda06c889c582b47dc32
-  - name: cflinuxfs2-rootfs
+  - name: cflinuxfs2
     version: 1.117.0
-    url: https://bosh.io/d/github.com/cloudfoundry/cflinuxfs2-rootfs-release?v=1.117.0
-    sha1: 0b2c0b5c34bd9b17ac18aaba13ff9ce275849e22
+    url: https://bosh.io/d/github.com/cloudfoundry/cflinuxfs2-release?v=1.117.0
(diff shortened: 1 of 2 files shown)


dcarley:
Testing in my dev environment ⏳ 

Do we need to send an announcement out to tenants so that their apps can make use of the new rootfs?

alext:
I don't think so. Because this is a stemcell bump, all the cells will be re-created, so all the apps will pick up the new rootfs automatically (unless I'm misunderstanding when the rootfs is applied).

dcarley:
> I don't think so. Because this is a stemcell bump, all the cells will be re-created, so all the apps will pick up the new rootfs automatically (unless I'm misunderstanding when the rootfs is applied).

I thought it was applied during application staging? I may have also misunderstood though.

alext:
From investigation, I think the app droplet does not include the rootfs. I found [this thread](https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/thread/HF3BNHQ7OHMBE6W6BJJGW4Y5OWMN2XSF/) on cf-dev which goes into lots of detail. [This post](https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/message/AMERCG7CMDBBMAF3PIEZNAUAHYLDUD5C/) in particular states:
> Just to clarify, applications intentionally do NOT require restaging to be
placed onto a new rootfs; a droplet is compatible with all future versions
of a specific rootfs.

Based on that, I think we're good.

dcarley:
👍  good to know.

It passed tests. I'll merge in the morning, so that we're in the office for the prod deployment.

PR #900, opened 2017-05-05
[#142400337] Extend cf ssh tests to 10GB

alext:
## What

Now that the CF CLI issue has been fixed (with version 6.26) this test
can be increased.

## How to review

Deploy from this branch and verify that the tests pass.

Note: alphagov/paas-docker-cloudfoundry-tools#92 must not be merged until this has been deployed to production.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -566,6 +566,7 @@ jobs:
               type: docker-image
               source:
                 repository: governmentpaas/cf-cli
+                tag: cf_cli_6.26
             run:
               path: sh
               args:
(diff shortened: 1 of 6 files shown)


alext:
I've had to add a workaround to this because the smoke and acceptance tests don't work with cf-cli 6.26 until CF release 257 (we're currently on 256). I'm in 2 minds whether this is worth doing, or whether it would be better to leave this change until after we've upgraded to cf 257.

dcarley:
> I'm in 2 minds whether this is worth doing, or whether it would be better to leave this change until after we've upgraded to cf 257.

I think it's OK to continue, since we've now started the work and workaround isn't terrible.

alext:
The pain point is that the smoke tests will start failing in prod as soon as the container update is merged. I'll have to rework this so that the `[TMP]` commit is less temporary, and the container is only merged once this PR has been deployed to prod.

alext:
I've updated the instructions so that this can be merged pointing at the dockerhub branch build. Once this has reached prod, the container PR can be merged, and the pinning in here reverted.

PR #902, opened 2017-05-08
[#141231851] Open web access security group

saliceti:
## What

Story: [CDN (big) story following spike](https://www.pivotaltracker.com/story/show/141231851)

Web access to the platform (via router ELB then haproxy/gorouter) was open in staging but closed in other environments. This does not protect from a security risk, it requires keeping a list of IPs up-to-date and makes testing cumbersome. For example to test cloudfront, the security group has to be manually edited to open to the world.

We now open it to the world in all environments and remove the web_access_cidrs list.

## How to review
* Code review
* Check you can access your environment via an external IP (ex from your phone)

## Who can review
Anyone but @LeePorte or me


dcarley:
With `aws_security_group.web` open, there's no need to have the `retrieve-pingdom-probes-ips` job in the pipeline and separate `aws_security_group.pingdom-probes-N` security groups - should we remove this in this PR?

saliceti:
@dcarley Good point, it's removed now

dcarley:
Tests pass.

So happy to see this code go 👏  👏  👏

PR #905, opened 2017-05-09
[SUPPORT] Allow re-inviting user without assigning to org/space

alext:
## What

When re-inviting a user (eg because they've forgotten their password),
we don't need to specify an org because they'll already be assigned to
the orgs/spaces then need access to. In this case it's desirable to not
touch the orgs/spaces.

This therefore allows this script to run without an org specified when
re-inviting a user.

## How to review

Test that the script works against a dev environment (the `--no-email` flag is useful here) for the following:
* Creating a user assigned to an org
* Re-inviting a user without an org specified

## Who can review

Not me.

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -253,7 +253,12 @@ send_mail() {
 }
 
 print_invite() {
-  success "Created invite ${INVITE_URL} for ${EMAIL}"
+  success "Created invite for ${EMAIL}"
+
+  echo "Invitation email:"
+  echo "Subject: $(get_subject)"
+  # shellcheck disable=SC2059
+  printf "$(get_body)"
 }
 
 emit_invite() {
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -17,8 +17,8 @@ MESSAGE='Hello,
 
 Your account for GOV.UK PaaS is ready:
 
- - username: ${EMAIL}
- - organisation: ${ORG}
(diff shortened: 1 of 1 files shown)


alext:
Closing because this breaks when it comes to email the user. The email template includes the org at present.

alext:
I've addressed the issue with the email template. I've additionally modified the behaviour when `--no-email` is specified so that the email contents are printed, which makes testing easier.

scripts/create-user.sh
```diff
@@ -18,7 +18,7 @@ MESSAGE='Hello,
 Your account for GOV.UK PaaS is ready:
 
  - username: ${EMAIL}
- - organisation: ${ORG}
+${ORG:+ - organisation: ${ORG}}
```
dcarley: Minor: is there a not-terrible way to omit the extra newline that this generates?

alext: It's possible. I'm not sure about the non-terrible part. Some examples:

```sh
 - username: ${EMAIL}${ORG:+\n - organisation: ${ORG}}
```

```sh
 - username: ${EMAIL}${ORG:+
 - organisation: ${ORG}}
```

What do you think?

dcarley: The latter seems the best of a bad bunch.

alext: I've updated it with this version now.

dcarley:
Used and works 👍

PR #906, opened 2017-05-09
[#144515851] Better concourse thrashing monitoring

Jonty:
## What

This introduces a new monitor tracking the number of btrfs processes running at any one time. When concourse starts to fail the number of processes spikes drastically, so this should allow us to quickly spot when concourse needs shaking before the load average catches up.

🐝 https://github.com/alphagov/paas-bootstrap/pull/60 should be merged and deployed before this, although really it doesn't matter much 🐝

At @alext's suggestion this also tweaks the existing CPU load monitor so that it will trigger within a minute of the load average spiking over the limit.

## How to review

* Deploy https://github.com/alphagov/paas-bootstrap/pull/60 to your concourse
* Deploy this PR to your cluster (with `ENABLE_DATADOG=true`)
* Check that a `$DEPLOY_ENV concourse excess btrfs processes running` monitor has been created in datadog, and that it goes green
* If you'd like to test the monitor, `make dev ssh_concourse` and run the following script to trigger it:
```
#!/bin/bash
list=(/var/vcap/data/baggageclaim/volumes/live/*)
while true; do
        /var/vcap/packages/btrfs_tools/sbin/btrfs qgroup show -F --raw ${list[-1]}/volume > /dev/null &
done
```

### DIFF
--- a/terraform/datadog/concourse.tf
+++ b/terraform/datadog/concourse.tf
@@ -4,7 +4,7 @@ resource "datadog_monitor" "concourse-load" {
   message            = "${format("Concourse load is too high: {{value}}. Check VM health. @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.aws_account)}"
   escalation_message = "Concourse load still too high: {{value}}."
   notify_no_data     = false
-  query              = "${format("avg(last_5m):avg:system.load.5{bosh-job:concourse,deploy_env:%s} > 200", var.env)}"
+  query              = "${format("max(last_1m):max:system.load.1{bosh-job:concourse,deploy_env:%s} > 200", var.env)}"
 
   thresholds {
     warning  = "150.0"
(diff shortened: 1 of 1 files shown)


terraform/datadog/concourse.tf
```diff
@@ -52,6 +52,24 @@ resource "datadog_monitor" "continuous-smoketests-failures" {
   tags = ["deployment:${var.env}", "service:${var.env}_monitors"]
 }
 
+resource "datadog_monitor" "concourse-btrfs" {
+  name                = "${format("%s concourse excess btrfs processes running", var.env)}"
+  type                = "query alert"
+  message             = "Large number of btrfs processes running. Concourse probably needs shaking."
```
Jonty: Good idea, will implement that.

terraform/datadog/concourse.tf
```diff
@@ -52,6 +52,24 @@ resource "datadog_monitor" "continuous-smoketests-failures" {
...
+  escalation_message  = "Large number of btrfs processes still running. Concourse probably needs shaking."
```
Jonty: I have no idea why we do it, and wondered the same - I was just keeping in line with other monitors.

[`escalation_message`](https://www.terraform.io/docs/providers/datadog/r/monitor.html#escalation_message) is optional, and I would happily remove it from both this and all the other cases where it seems to be needless.

dcarley: Let's remove it from this monitor so as not to further the pattern.

Up to you about whether we remove it from all the others. It should probably be a separate PR if we do.

dcarley:
Confirmed that it works 👍

terraform/datadog/concourse.tf
Jonty: @dcarley Done, and a PR for the team manual opened: https://github.com/alphagov/paas-team-manual/pull/113

dcarley:
@combor is kindly deploying alphagov/paas-bootstrap#60 and then this can be merged.

I've amended the commit to fix the `terraform fmt` failure in Travis:
```patch
diff --git a/terraform/datadog/concourse.tf b/terraform/datadog/concourse.tf
index 2eaf670..f356ae4 100644
--- a/terraform/datadog/concourse.tf
+++ b/terraform/datadog/concourse.tf
@@ -53,9 +53,9 @@ resource "datadog_monitor" "continuous-smoketests-failures" {
 }

 resource "datadog_monitor" "concourse-btrfs" {
-  name                = "${format("%s concourse excess btrfs processes running", var.env)}"
-  type                = "query alert"
-  message             = "${format("Large number of btrfs processes running on the concourse host. It may perform badly. See: %s#btrfs-processes @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.datadog_documentation_url, var.aws_account)}"
+  name    = "${format("%s concourse excess btrfs processes running", var.env)}"
+  type    = "query alert"
+  message = "${format("Large number of btrfs processes running on the concourse host. It may perform badly. See: %s#btrfs-processes @govpaas-alerting-%s@digital.cabinet-office.gov.uk", var.datadog_documentation_url, var.aws_account)}"

   notify_no_data      = false
   require_full_window = false
```

PR #907, opened 2017-05-09
Revert "Use updated version of cf-cli containers."

timmow:
## What

This reverts commit bf4bb32f70dfd2534440279813a602e9374839cc.

See #900 for more details on why - we had to pin the acceptance tests container temporarily

## How to review

Code review
Check the build has completed in dockerhub and the latest container uses the correct version of cf cli
I don't think you need to re run the acceptance tests as I think we can trust the tagged image and image built from master are roughly equivalent

## Who can review

Anyone but me


alext:
:+1: Looks good to me.  I'll merge when Travis is green.

PR #909, opened 2017-05-10
[#143868179] Deploy walking skeleton of Compose service broker

henrytk:
## What

https://www.pivotaltracker.com/story/show/143868179

We want to deploy the Compose service broker as an application on the platform.

## How to review

Preferably review in conjunction with https://github.com/alphagov/paas-compose-broker/pull/1

Deploy from this branch and make sure all of the tests passed. The tests haven't been changed, but the ordering of tasks in the CF deploy has.

The broker is a 'walking skeleton', so doesn't need to actually provision anything. Its 'correct' behaviour is to error when actually performing all operations except the _Services_ endpoint for returning the catalog.

Example of checking it serves the catalog:

```
curl -k -u compose-broker:$PASSWORD_FROM_CF_SECRETS https://compose-broker.${DEPLOY_ENV}.dev.cloudpipelineapps.digital/v2/catalog
```

### Before merging ⚠️ 

Review and merge https://github.com/alphagov/paas-compose-broker/pull/1 and then remove the temporary commit on this branch.

## Who can review

Anyone but me or @combor

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1485,41 +1485,14 @@ jobs:
 
                   cp ./paas-cf/config/service-brokers/compose/catalog.json ./paas-compose-broker/
                   cd paas-compose-broker/
-                  cf push --no-start
-                  cf set-env compose-broker LOG_LEVEL DEBUG
-                  cf set-env compose-broker USERNAME compose-broker
-                  cf set-env compose-broker PASSWORD "$COMPOSE_BROKER_PASS"
+                  ruby -ryaml -e "
+                    manifest = YAML.load_file('manifest.yml')
+                    env = { 'env' => {'LOG_LEVEL' => 'DEBUG' , 'USERNAME' => 'compose-broker', 'PASSWORD' => '${COMPOSE_BROKER_PASS}'} }
+                    manifest['applications'].each { |app| app.merge!(env) }
+                    File.write('manifest.yml', manifest.to_yaml)
+                  "
                   cf blue-green-deploy compose-broker
-
-        - task: register-compose-broker
-          config:
-            platform: linux
-            image_resource:
-              type: docker-image
-              source:
-                repository: governmentpaas/cf-cli
-            inputs:
-              - name: paas-cf
-              - name: config
-              - name: bosh-CA
-            params:
-              APPS_DNS_ZONE_NAME: {{apps_dns_zone_name}}
-            run:
-              path: sh
-              args:
-                - -e
-                - -u
-                - -c
-                - |
-                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
-                  . ./config/config.sh
-                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
-
-                  if cf service-brokers | grep "compose-broker\s"; then
-                    cf update-service-broker compose-broker compose-broker "$COMPOSE_BROKER_PASS" "https://compose-broker.${APPS_DNS_ZONE_NAME}"
-                  else
-                    cf create-service-broker compose-broker compose-broker "$COMPOSE_BROKER_PASS" "https://compose-broker.${APPS_DNS_ZONE_NAME}"
-                  fi
+                  cf delete -f compose-broker-old
 
         - task: deploy-simulated-load-application
           config:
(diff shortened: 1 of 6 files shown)


henrytk:
I have marked this as WIP because we hit this issue: https://github.com/cloudfoundry/cli/issues/939, which means if you deploy twice the deployment fails with:

```
Enabling access to all plans of service mongo for the org service-brokers as admin...
FAILED
Server error, status code: 400, error code: 260002, message: This combination of ServicePlan and Organization is already taken: organization_id and service_plan_id unique
```

Discussed with @alext and @combor and agreed to not enable the service at all and add a custom acceptance test which adds it (which should be removed automatically when the test org is deleted). Also, it was suggested to move the catalog.json file from paas-compose-broker to paas-cf, as it is deploy/runtime configuration. A copy of it can be kept in the paas-compose-broker as a useful example.

henrytk:
The pipeline no longer enables the service access. I've added a custom acceptance test which just checks the marketplace. See commit message for details. Ready for review again.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1391,6 +1399,43 @@ jobs:
...
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf create-org service-brokers
+                  cf set-quota service-brokers small
+                  cf create-space compose -o service-brokers
+                  cf target -o service-brokers -s compose
+
+                  cp ./paas-cf/config/service-brokers/compose/catalog.json ./paas-compose-broker/
+                  cd paas-compose-broker/
+                  cf push --no-start
+                  cf set-env compose-broker BROKER_API_HOST 0.0.0.0
```
alext: As commented on the broker PR, it would be good for this to be the default behaviour making it unnecessary to specify this here.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1391,6 +1399,43 @@ jobs:
                   cf enable-feature-flag diego_docker
 
       - aggregate:
+        - task: deploy-compose-service-broker
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
```
alext: This needs the version pinning to match the current changes in master from #910.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1391,6 +1399,43 @@ jobs:
...
+                - -e
+                - -u
+                - -c
+                - |
+                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
+
+                  . ./config/config.sh
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf create-org service-brokers
+                  cf set-quota service-brokers small
```
alext: I think we should move the creation of this org out of this task. It might be better to make the existing `create-admin-org-space` task more generic to cover this org as well.

combor:
I believe I have addressed all comments. This is probably ready for re-review.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1447,6 +1458,40 @@ jobs:
...
+                  echo | cf login -a ${API_ENDPOINT} -u ${CF_ADMIN} -p ${CF_PASS}
+                  cf target -o service-brokers -s compose
+
+                  cp ./paas-cf/config/service-brokers/compose/catalog.json ./paas-compose-broker/
+                  cd paas-compose-broker/
+                  cf push --no-start
+                  cf set-env compose-broker LOG_LEVEL DEBUG
+                  cf set-env compose-broker USERNAME compose-broker
+                  cf set-env compose-broker PASSWORD "$COMPOSE_BROKER_PASS"
+                  cf blue-green-deploy compose-broker
```
alext: This sequence won't achieve a proper blue-green deploy. The `cf push --no-start` will stop the app. Additionally, the `cf set-env` commands will only apply to the old app. The new app that the blue-green plugin creates won't inherit them.

To solve this, we should have this script add the env vars to the `manifest.yml` before doing the `cf blue-green-deploy`.

We should also add a step to delete the old app after a successful push

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1746,6 +1735,92 @@ jobs:
...
+                export ES_PORT="9200"
+                ruby "${SCRIPT_DIR}/kibana_set_utc.rb"
+
+        - task: enable-elasticsearch-shard-reallocation
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/curl-ssl
```
alext: The version pinning here has been lost in the rebase.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1746,6 +1735,92 @@ jobs:
...
                   bosh cleanup --all
 
+      - aggregate:
+        - task: register-compose-broker
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
```
alext: This container needs pinning as well.

config/service-brokers/compose/catalog.json
```diff
@@ -0,0 +1,43 @@
...
+            "displayName":"mongo",
+            "imageUrl":"https://webassets.mongodb.com/_com_assets/cms/MongoDB-Logo-5c3a7405a85675366beb3a5ec4c032348c390b3f142f5e6dddf1d78e2df5cb5c.png",
+            "longDescription":"Compose MongoDB instance",
+            "providerDisplayName":"GOV.UK PaaS",
+            "documentationUrl":"https://compose.com/mongodb",
+            "supportUrl":"https://www.cloud.service.gov.uk/support.html"
+         },
+         "plans":[
+            {
+               "id":"1",
```
alext: This ID needs to be a UUID as well.

concourse/pipelines/create-cloudfoundry.yml
combor: This is done. We have short ruby script to add required env variables to the manifest.

combor:
This is ready to re-re-check.

PR #910, opened 2017-05-11
[#145211537] Pin all governmentpaas containers to a version

dcarley:
## What

So that we can control how containers are updated between environments and
prevent forwards/backwards incompatibilities with code in the pipeline.

This will also prevent Concourse from re-downloading all of the containers
each time we merge a change to alphagov/paas-docker-cloudfoundry-tools
because Docker Hub rebuilds everything.

Using the functionality added in:

- alphagov/paas-docker-cloudfoundry-tools#93

## How to review

1. Eyeball the diff and make sure that it looks sensible
1. Run at least one pipeline and confirm that it still works

## Who can review

Not @dcarley

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -4,11 +4,13 @@ resource_types:
   type: docker-image
   source:
     repository: governmentpaas/s3-resource
+    tag: 9e7cf1dd32699c091e22e46b349e050443d9ac1a
 
 - name: semver-iam
   type: docker-image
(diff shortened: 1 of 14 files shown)


dcarley:
Closing while I double check a problem with resources.

dcarley:
Fixed, depends on:

- alphagov/paas-semver-resource#5
- alphagov/paas-s3-resource#5

dcarley:
I noticed that the `run-bosh-cli` task was using the old format and was missed. I've updated it now and tested, but it's still using the old format - details are in the commit message.

PR #912, opened 2017-05-12
[SUPPORT] Use European keyservers in git-resource GPG check

dcarley:
## What

We have seen intermittent problems with requests to keyservers timing out
and failing the `pipeline-lock` job in our pipeline:

    gpg: requesting key 6DCAB6FA from hkp server keys.gnupg.net
    gpg: keyserver timed out
    gpg: keyserver receive failed: Keyserver error

The default of `keys.gnupg.net` is a CNAME for `pool.sks-keyservers.net`
which is described at:

- https://sks-keyservers.net/overview-of-pools.php

Try using the European pool, instead of the default pool, in the hope that
the lower latency will make them more reliable. If this doesn't work and we
think it's still a problem them we could consider inline all of the public
key contents in the future.

## How to review

1. Temporarily enable commit verification in your dev environment:

    ```patch
    index 1e2e2c4..5fe3444 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -84,7 +84,6 @@ dev: globals ## Set Environment to DEV
            $(eval export ENABLE_AUTODELETE=true)
    -       $(eval export SKIP_COMMIT_VERIFICATION=true)
            $(eval export DISABLE_HEALTHCHECK_DB=true)
            $(eval export ENABLE_DATADOG ?= false)
    ```
1. Checkout this branch:

    ```
    git fetch && git checkout bugfix/support-use_eu_keyservers
    ```
1. Deploy the pipelines from this branch but don't change the branch used by the pipelines (because the commit isn't signed):

    ```
    SELF_UPDATE_PIPELINE=false make dev pipelines
    ```
1. Run the `pipeline-lock` job and confirm that it downloads the keys and verifies the commit signature

## Who can review

Not @dcarley

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -18,6 +18,7 @@ resources:
       branch: master
       tag_filter: {{paas_cf_tag_filter}}
       commit_verification_key_ids: {{gpg_ids}}
+      gpg_keyserver: hkp://eu.pool.sks-keyservers.net/
 
   - name: cf-secrets
     type: s3-iam
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -81,6 +81,7 @@ resources:
       branch: {{branch_name}}
       tag_filter: {{paas_cf_tag_filter}}
       commit_verification_key_ids: {{gpg_ids}}
+      gpg_keyserver: hkp://eu.pool.sks-keyservers.net/
 
   - name: graphite-nozzle
     type: git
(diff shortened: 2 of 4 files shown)


dcarley:
Co-incidentally this has happened twice today - once in CI and once in staging.

timmow:
And once in prod!

saliceti:
I've tested with the EU pool but it was slow and I had a few of these errors:

```
gpg: requesting key XXX from hkp server eu.pool.sks-keyservers.net
gpg: keyserver timed out
gpg: keyserver receive failed: Keyserver error
```

In one occasion the job passed as the missing key was not the one used to commit. In another occasion it failed after 2 of these errors.

Reverting to normal pool worked straightaway, quickly and without any error.

I suggest to investigate more to understand this behaviour better.

dcarley:
It worked for me, but the testing doesn't bode well. I'm going to close this because it's not an improvement.

PR #915, opened 2017-05-12
[SUPPORT] Revert "Remove *_monitors tag from CC monitors"

dcarley:
## What

This reverts commit 00cfcd466b94cf066616a9e19f5bd0b679b79ace.

I believe that this was fixed by 7bf2b74f58b69f203e32f9c606b2b749754f64ef so
we should display these on the dashboard again. I noticed this because I was
called out by PagerDuty in-hours and the alert wasn't on the dashboard.

## How to review

Code review.

## Who can review

Anyone.


saliceti:
LGTM

PR #922, opened 2017-05-17
[SUPPORT] Increase the ES disk size

paroxp:
## What

We're getting alerts on running low on the disk space in production. For
the time being, we'd like to increase the disk space size, as we don't
have any archiving in place.

This change will add additional 256G to the current disk, which
currently is 512G.

## How to review

- Sanity review over the change
- Agreement upon additional space

## Who can review

Preferably @alext

### DIFF
--- a/manifests/cf-manifest/env-specific/cf-prod.yml
+++ b/manifests/cf-manifest/env-specific/cf-prod.yml
@@ -3,4 +3,4 @@ meta:
   cell:
     instances: 15
   elasticsearch_master:
-    disk_size: 512000
+    disk_size: 768000


dcarley:
Could you include "Elasticsearch" in the subject of the commit message so that someone looking at the commit log has context about the scope of the change (i.e. not all disks)?

paroxp:
I knew you'd say that. Sure :D

dcarley:
Difference in cost by USD/mo:
```
➜  paas-cf git:(master) ✗ bc -q
(0.11 * 512) * 3
168.96
(0.11 * 768) * 3
253.44
```

Discussed with Alex on Slack. This is a reasonable thing to do since we expect logs to grow as we have more tenants and traffic. It's not worth the effort of finding out exactly what logs they are at this stage. Hopefully we'll move ELB to another provider in the future.

PR #924, opened 2017-05-18
Raise concourse btrfs monitor levels

Jonty:
## What

The concourse btrfs monitor levels appear to be too low for production as we're getting spurious warnings. This raises the levels and also ensures that we only track btrfs processes on the concourse host rather than the wider cluster.

## How to review

* Code review.
* Check that the levels are higher than what we've seen in production: https://app.datadoghq.com/monitors#2083141?group=*&from_ts=1495098656256&to_ts=1495102256543

## Who can review

Anybody, probably @paroxp.


paroxp:
Looks good :)

PR #925, opened 2017-05-18
Add user search functionality

paroxp:
## What

Previously, it would be useful, for the [Finding activity](https://government-paas-team-manual.readthedocs.io/en/latest/support/finding_activity/) story. Now
we have had a case, where it would be useful tool, while checking if the
user still exists in the database.

## How to review

- Change to this branch
- Attempt to find any user

### DIFF
--- a/scripts/find-user.sh
+++ b/scripts/find-user.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+set -euo pipefail
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+SCRIPT=$0
+
+# shellcheck disable=SC1090
+source "${SCRIPT_DIR}/common.sh"
+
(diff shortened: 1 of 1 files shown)


dcarley:
This seems quite complex (and potentially taxing on the API). Whereas [uaac](https://github.com/cloudfoundry/cf-uaac) provides a single command to do the same thing:
```
uaac user get <email>
```

What we should do instead is document how and make it easier to use `uaac`. I'm happy to do that, since I have the unfortunate experience of doing a lot of work with it in the past, if you like?

paroxp:
Sure. I was unaware of the uaac function. It didn't come out from the
previous story I've played.

I agree, this could have been making 7 requests each use, which could
affect the API somewhat.

Thanks for catching that up.

PR #926, opened 2017-05-19
[SUPPORT] Print the error but do not stop the tests

paroxp:
## What

The job [availability-tests/470](https://deployer.cloud.service.gov.uk/teams/main/pipelines/create-cloudfoundry/jobs/availability-tests/builds/470) has failed. We think it actually
shouldn't matter to us that much, if there is a one of connection issue
with the concourse API. We don't mean to get rid of the error
completely, therefore we're still printing it, allowing the script to
continue it's loop and hoping that the issue has resolved it self in the
mean time.

## How to review

- Code review
- Run the deployment from this branch and expect the
`availability-tests` to pass

### DIFF
--- a/platform-tests/src/availability/helpers/concourse.go
+++ b/platform-tests/src/availability/helpers/concourse.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"crypto/tls"
+	"log"
 	"net"
 	"net/http"
 	"os"
@@ -116,7 +117,10 @@ func buildsWithVersion(team concourse.Team, pipelineName, resourceName, resource
 	}
 
 	resourceVersions, _, resourceExists, err := team.ResourceVersions(pipelineName, resourceName, page)
-	ExpectWithOffset(2, err).NotTo(HaveOccurred())
+	if err != nil {
+		log.Println(err)
+		return make([]atc.Build, 0)
+	}
 	ExpectWithOffset(2, resourceExists).To(BeTrue())
 
 	for _, version := range resourceVersions {


platform-tests/src/availability/helpers/concourse.go
```diff
@@ -116,7 +117,9 @@ func buildsWithVersion(team concourse.Team, pipelineName, resourceName, resource
 	}
 
 	resourceVersions, _, resourceExists, err := team.ResourceVersions(pipelineName, resourceName, page)
-	ExpectWithOffset(2, err).NotTo(HaveOccurred())
+	if err != nil {
+		log.Println(err.Error())
+	}
 	ExpectWithOffset(2, resourceExists).To(BeTrue())
```
alext: Do we know what state `resourceExists` will be in when this returns an error? I'd expect the value to be undefined (and probably the zero value - false). I think this expectation (and probably the following code) should only be run if err is nil.

paroxp: Overlooked that. Thanks.

platform-tests/src/availability/helpers/concourse.go
```diff
@@ -116,7 +117,10 @@ func buildsWithVersion(team concourse.Team, pipelineName, resourceName, resource
 	}
 
 	resourceVersions, _, resourceExists, err := team.ResourceVersions(pipelineName, resourceName, page)
-	ExpectWithOffset(2, err).NotTo(HaveOccurred())
+	if err != nil {
+		log.Println(err.Error())
```
paroxp: Thanks! Noted.

dcarley:
Thanks 👍

PR #927, opened 2017-05-19
Invert the order of displayed boxes

paroxp:
## What

When you put the dashboard right next to grid of three concourse
pipelines, the order is inverted. It is not a bug. It is not a feature.
But it bugs me a lot :D

## How to review

- Code review
- Run the ruby app locally and see if everything is as expected


timmow:
I'd never noticed this, but now you've mentioned it It is going to annoy me :)

timmow:
When travis finally finishes 👍

PR #930, opened 2017-05-22
[SUPPORT] Improve the failure in bosh-tests

paroxp:
## What

Currently, we're not getting any error message which could confuse
someone who hasn't previously encounter this particular error. With this
change, we're reporting on the status of VMs, while the `bosh-tests`
run.

## How to review

- Code review
- Run the pipeline and allow `bosh-tests` to pass
- Optionally, stop (change state) one of the VMs and re-run `bosh-tests` expecting them to fail

## Who can review

Neither @dcarley nor @paroxp

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2115,7 +2115,13 @@ jobs:
               ./paas-cf/concourse/scripts/bosh_login.sh {{bosh_fqdn}} bosh-secrets/bosh-secrets.yml
               bosh deployment cf-manifest/cf-manifest.yml
               bosh vms {{deploy_env}} | tee vms.txt
-              [ $(grep '|' vms.txt | grep -cEv "(\-\-|VM|running)") -eq 0 ]
+              vms_not_running=$(grep '|' vms.txt | grep -cEv "(\-\-|VM|running)" || true)
+              if [ ${vms_not_running} -gt "0" ]; then
+                echo "Error: Number of not running VMs: ${vms_not_running}."
+                exit 1
+              else
+                echo "Success: All VMs are up and running."
+              fi
 
   - name: performance-tests
     plan:


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2109,13 +2109,18 @@ jobs:
           run:
             path: sh
             args:
-            - -e
```
dcarley: I'm uncomfortable about removing this because it's the best kind of error checking you get in shell scripting.

Without it the script will continue if any of the previous `bosh` commands fail and we may end up in an unknown state. I can't tell for sure what the `grep` will do if `vms.txt` is empty?

Can you explain why it needed to be removed? We might be able to find another solution.

paroxp: Not knowing much about shellscript, I've removed this as it has caused the code to stop at [line 2117](https://github.com/alphagov/paas-cf/pull/930/files/a115a365b2540f6d5cf216565880b5de7ca26a81#diff-8f7658bb02a08dfb771f38694a7b075fR2117) and not continuing to the following if statement.

Regarding the `grep` and empty `vms.txt`:

```sh
→ mv vms.txt _vms.txt
→ touch vms.txt
→ ./test.sh
Success: All VMs are up and running.
```

I'd be more than happy, to use different solution, but I may need some more guidance.

dcarley: Let's pair on it. I'll arrange a time either later this afternoon or tomorrow.

paroxp: 👍

paroxp:
We had a look at the code and found a better yet somewhat ugly approach to the problem I've been having. It does however solve the issue I've been facing.

Would be good, to have yet someone else review the latest changes.

dcarley:
Happy with this, but taking my name off because Rafal and I worked together on it.

PR #931, opened 2017-05-24
[#145024711] Custom acceptance tests for HTTP(s) edge cases

combor:
## What
We want our acceptance tests to tell us about regressions similar to the
MTU response size incident. This PR implements a set of HTTP tests
for edge cases that will try to replicate weird requests to apps we host
on our platform. It also adds a HTTP tester app with multiple endpoints
that can return headers or query strings in the body so tests can make
assertions.

This set of tests replaces response_size_test and print_request_headers
app that we used before.

List of tests:
- large request bodies
- large request headers
- large response headers
- egress connectivity to a service outside the PaaS
- app-to-app connectivity between two apps on the PaaS
- slow response
- GET with many parameters
- GET with UTF8 parameters?
- GET with up to 2000 characters

## How to review

Run a deployment from this branch and make sure the custom acceptance tests pass. 

## Who can review
Not @combor or @LeePorte


platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+				returnedBody, err := ioutil.ReadAll(response.Body)
+				Expect(err).NotTo(HaveOccurred())
+				returnedBodySize, err := strconv.Atoi(string(returnedBody))
+				Expect(err).NotTo(HaveOccurred())
+				Expect(returnedBodySize).Should(Equal(requestkB * int(KILOBYTE)))
+			}
+		})
+
+		It("should accept large request header sizes", func() {
+			for headerkB := 1; headerkB <= 7; headerkB += 1 {
```
LeePorte: It's not a limit in the HTTP spec rather a limit imposed by the webserver 8KB is the generally accepted max with IIS upping it to 16KB. This was verified during developing the test.

https://stackoverflow.com/questions/686217/maximum-on-http-header-values

alext:
It would be good to update the Go version of the `http_tester` test app while we're here. It's currently set to 1.6 which is marked as deprecated in the current Go buildpack.

platform-tests/example-apps/http_tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+
+func bigHeader(w http.ResponseWriter, r *http.Request) {
+
+	var headerSize = r.URL.Query().Get("size")
+	headerInt, err := strconv.Atoi(headerSize)
+	if err != nil {
+		http.Error(w, fmt.Sprint(err), http.StatusInternalServerError)
+		return
+	}
+	rand.Seed(time.Now().UnixNano())
```
combor: I don't think so. It looks like leftover from the code that we replaced with `randStringBytes`. Removed.

platform-tests/example-apps/http_tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+	"math/rand"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	delay   = 250
```
combor: Done.

platform-tests/example-apps/http_tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		http.Error(w, fmt.Sprint(err), http.StatusInternalServerError)
+		return
+	}
+	egressResponse, err := httpClient.Do(req)
+
+	if err != nil {
+		responseMessage = err.Error()
+		responseCode = 502
+	} else {
```
combor: True...fixed.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+	utfchars    = "スタック・オーバーフロー はプログラマ"
+	paramLength = 8
+	queryLength = 4096
+	requesttext = "slow response"
+)
+
+var (
+	appName        string
+	numberOfParams = (queryLength / paramLength)
+	queryString    string
```
combor: Leftover...deleted.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
+)
+
+const (
+	letters     = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	utfchars    = "スタック・オーバーフロー はプログラマ"
+	paramLength = 8
+	queryLength = 4096
+	requesttext = "slow response"
```
combor: Moved to the specific test.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+var (
+	appName        string
+	numberOfParams = (queryLength / paramLength)
+	queryString    string
+	parameters     url.Values
```
combor: Moved to the specific tests.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+				"push", appName,
+				"-b", config.RubyBuildpackName,
+				"-p", "../../../../cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/assets/dora",
+				"-d", config.AppsDomain,
+				"-i", "1",
+				"-m", "256M",
+			).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+		})
+
+		It("should serve request and response bodies of increasing sizes", func() {
```
combor: Corrected.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+		BeforeEach(func() {
+			appName = generator.PrefixedRandomName("CATS-APP-HTTP-TESTER-")
+			Expect(cf.Cf(
+				"push", appName,
+				"-b", config.GoBuildpackName,
+				"-p", "../../example-apps/http_tester",
+				"-d", config.AppsDomain,
+				"-c", "./bin/debug_app; sleep 1; echo 'done'",
+			).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+			rand.Seed(time.Now().UnixNano())
```
combor: Leftover.  Removed.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			}
+			httpClient = http.Client{Transport: tr}
```
combor: I have removed this bit to use suite provided http client.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,205 @@
...
+			parameters = url.Values{}
+			for i := 1; i <= numberOfParams; i++ {
+				keyName := "p" + strconv.Itoa(i)
+				valueLength := paramLength - len("&"+keyName+"=")
+				parameters.Add(keyName, randStringBytes(valueLength))
+			}
+			requestURL.RawQuery = parameters.Encode()
+			response := helpers.CurlApp(appName, requestURL.RequestURI())
+			Expect(response).To(BeEquivalentTo(requestURL.RawQuery))
+			Expect(len(response)).To(Equal(queryLength - len("?")))
```
combor: Fixed.

combor:
> It would be good to update the Go version...

Updated. The import path was kinda weird as well.
```
"ImportPath": "debug_app",
```

platform-tests/example-apps/http_tester/main.go
dcarley: From offline..

We discovered that the results are deterministic without this:

https://play.golang.org/p/r9L_b6GgT2

So we're going to add it back, within the function. It's more computationally expensive that seeding once but it shouldn't matter for our use case.

platform-tests/src/acceptance/http_edge_cases_test.go
dcarley: Thanks 👌

dcarley:
> The import path was kinda weird as well.
> 
> "ImportPath": "debug_app",

Does this have an effect on the start command?
```
"-c", "./bin/debug_app; sleep 1; echo 'done'",
```
I'll test it anyway, but if you know the answer before the tests finish..

platform-tests/example-apps/http_tester/main.go
combor: It's back now in the randStringBytes to avid deterministic 'random' strings :)

platform-tests/src/acceptance/http_forwarded_for_test.go
```diff
@@ -44,13 +44,13 @@ var _ = Describe("X-Forwarded headers", func() {
 		Expect(cf.Cf(
 			"push", appName,
 			"-b", config.GoBuildpackName,
-			"-p", "../../example-apps/print_request_headers",
+			"-p", "../../example-apps/http_tester",
 			"-d", config.AppsDomain,
 			"-c", "./bin/debug_app; sleep 1; echo 'done'",
```
dcarley: Needs to be changed to `./bin/http_tester` after `ImportPath` changed.

platform-tests/src/acceptance/http_forwarded_proto_test.go
```diff
@@ -22,13 +22,13 @@ var _ = Describe("X-Forwarded headers", func() {
 		Expect(cf.Cf(
 			"push", appName,
 			"-b", config.GoBuildpackName,
-			"-p", "../../example-apps/print_request_headers",
+			"-p", "../../example-apps/http_tester",
 			"-d", config.AppsDomain,
 			"-c", "./bin/debug_app; sleep 1; echo 'done'",
```
dcarley: Needs to be changed to `./bin/http_tester` after `ImportPath` changed.

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -0,0 +1,197 @@
...
+			appName2 := generator.PrefixedRandomName("CATS-APP-HTTP-TESTER-")
+			Expect(cf.Cf(
+				"push", appName2,
+				"-b", config.GoBuildpackName,
+				"-p", "../../example-apps/http_tester",
+				"-d", config.AppsDomain,
+				"-c", "./bin/http_tester; sleep 1; echo 'done'",
+			).Wait(CF_PUSH_TIMEOUT)).To(Exit(0))
+
+			response := helpers.CurlApp(appName, fmt.Sprintf("/egress?domain=%s.%s", appName2, config.AppsDomain))
```
dcarley: This test fails since we removed `InsecureSkipVerify` from the `/egress` handler in the app:
```
HTTP edge cases
/tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_edge_cases_test.go:188
  Using http_tester
  /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_edge_cases_test.go:187
    can connect to other app in the paas [It]
    /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_edge_cases_test.go:145

    Expected
        <string>: Get https://CATS-APP-HTTP-TESTER-3e568d99-f5c0-42a8-7d92-2b3c86db4ea4.dcarley.dev.cloudpipelineapps.digital/: x509: certificate signed by unknown authority
        
    to be equivalent to
        <string>: OK

    /tmp/build/3151afb2/paas-cf/platform-tests/src/acceptance/http_edge_cases_test.go:144
```

I think the options are either:

1. Add a query parameter to disable cert verification for this one test based on `config.SkipSSLValidation`
1. Add our CA to the rootfs for environments that are self-signed

platform-tests/example-apps/http_tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func egress(w http.ResponseWriter, r *http.Request) {
+
+	domain := r.URL.Query().Get("domain")
+
+	var responseMessage string
+	var responseCode int
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
dcarley: This was removed and added back without comment in:

https://github.com/alphagov/paas-cf/pull/931/files/09e8efbcaa31d8aae76cde7f66d7fe3aa0ac613f#diff-2c8d4ffb16bb71b88fd7b89690145eed

I don't think the solutions are easy though because `config.SkipSSLValidation` is set to false in all environments. Let's leave it for now.

PR #936, opened 2017-05-30
[#146152015] Remove Jonty from commit signers and admins

henrytk:
## What

Jonty has left the team :(

## How to review

Check it's his key that's been removed and there is nothing else that needs to be removed.

## Who can review

Anyone but me.


.gpg-id
```diff
@@ -1,4 +1,3 @@
-B1A5BC141F6BEEE5
```
dcarley: > `BEEE5`

🐝  🐝  🐝

PR #939, opened 2017-05-31
[#145024711 ] Reduce http_tester instance memory

combor:
## What

As we run tests in parallel we deploy multiple instances of http_tester
app at once. This may have an impact on the platform so we decided to
reduce available instance memory to 64m

## How to review

Run the pipeline and check if tests pass. 

## Who can review
Not @combor

### DIFF
--- a/platform-tests/src/acceptance/http_edge_cases_test.go
+++ b/platform-tests/src/acceptance/http_edge_cases_test.go
@@ -26,12 +26,9 @@ const (
 	utfchars = "スタック・オーバーフロー はプログラマ"
 )
 
-var (
-	appName string
-)
-
 var _ = Describe("HTTP edge cases", func() {
 	Describe("Using dora", func() {
+		var appName string
 
 		BeforeEach(func() {
 			appName = generator.PrefixedRandomName("CATS-APP-DORA-")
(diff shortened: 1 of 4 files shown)


dcarley:
Because we're attempting to fix a specific problem, could you tag this PR with the story ID [#145024711](https://www.pivotaltracker.com/story/show/145024711) and include the error message that we've been seeing in the commit message?

combor:
Done. BTW note this is still WIP.

dcarley:
Just getting my requests in early so that it doesn't hold up the review 😀

platform-tests/src/acceptance/http_edge_cases_test.go
```diff
@@ -65,10 +62,8 @@ var _ = Describe("HTTP edge cases", func() {
 			appName = generator.PrefixedRandomName("CATS-APP-HTTP-TESTER-")
 			Expect(cf.Cf(
 				"push", appName,
-				"-b", config.GoBuildpackName,
 				"-p", "../../example-apps/http_tester",
 				"-d", config.AppsDomain,
-				"-c", "./bin/http_tester; sleep 1; echo 'done'",
```
alext: I don't think this is picking up the `manifest.yml`. I think the cf CLI always looks in the current directory for that (even when `-p` is specified), so you'll need to add a `"-f", "../../example-apps/http_tester/manifest.yml",` line.

PR #940, opened 2017-06-02
[#145024711] Skip HTTP edge cases

combor:
## What

As HTTP edge cases tests are misbehaving we have decided to temporarily
disable them to investigate the root cause.

This PR should be reverted once we know how to fix underlying
problem.

## How to review

Sanity check.

## Who can review

Not @acombor


dcarley:
Haven't tested, but looks 👌

PR #943, opened 2017-06-06
Mysql support 146511067

LeePorte:
## What

As Part of enabling MySQL within the RDS broker, we need to add the MySQL ingress port to the RDS Broker security group and to the Cloud Foundry security group. Without this the integration tests for the RDS broker with MySQL fail.

## How to review

1. Code Review
2. Deploy from this branch
3. Verify appropriate ingress rules are in the security groups

## Who can review

Not me

### DIFF
--- a/terraform/cloudfoundry/cf_rds.tf
+++ b/terraform/cloudfoundry/cf_rds.tf
@@ -24,6 +24,16 @@ resource "aws_security_group" "cf_rds" {
     ]
   }
 
+  ingress {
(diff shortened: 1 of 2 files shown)


alext:
To capture a conversation: The RDS broker integration tests should not depend on anything in paas-cf. It seems that the changes in this branch are needed while developing as a temporary measure until the work to run the tests on the build concourse is complete.

We're therefore going to close this PR, but keep the branch around for now.

PR #944, opened 2017-06-07
#[146759947]  Bump the UAA release version to v30.4

timmow:
## What

update  the UAA release, to protect us
against the CVE-2017-4994[1].

[1] https://www.cloudfoundry.org/cve-2017-4994/

## How to review

Code review. I've applied this in a dev env and the tests pass.

## Who can review

Anyone but me


alext:
I don't think you meant to commit the `spec/examples.txt` files. These should probably be in the .gitignore file.

PR #948, opened 2017-06-09
[SUPPORT] Fix the script to upload Tracker Token

paroxp:
## What

We have changed the bucket name a while ago. This script currently
errors out, for a good reason. We'd like to correct that, before we
rotate the token properly.

## How to review

- Attempt to `make dev upload-tracker-token` from the `master` branch - experience an error
- Checkout this branch and try the above command again - experience **no error**
- After merging, it's advised to run `make prod upload-tracker-token` so we can comeback to working rubbernecker next working day


alext:
Looks good. I've updated the token in prod after merging.

PR #951, opened 2017-06-13
[#140849465] Remove some rds-broker tests

paroxp:
## What

Our new RDS broker functionality will include the set of CI tests, which
essentially gives us the opportunity to remove some slow and not
strictly related tests.

We'd like to remove the final snapshot related set of tests along with
their functionality as there isn't enough relation between our CF
deployment and the broker functioning properly at the time. We think, it
could be well covered by the blackbox ci testing.

## How to review

Review in conjunction with https://github.com/alphagov/paas-rds-broker/pull/46, because the idea of the story is to have all of the same tests, but split across the two repositories. Therefore, we want to ensure no tests have been lost in the process.

- Code review
- Deploy from the branch and witness the tests passing

## Who can review

Neither @henrytk nor @paroxp


platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -106,87 +103,8 @@ var _ = Describe("RDS broker", func() {
...
-	Context("final snapshots", func() {
-
-		var (
-			dbInstanceName  string
-			rdsInstanceName string
-		)
-
-		BeforeEach(func() {
-			dbInstanceName = generator.PrefixedRandomName("test-db-")
-			Expect(cf.Cf("create-service", serviceName, testPlanNameWithSnapshot, dbInstanceName).Wait(DEFAULT_TIMEOUT)).To(Exit(0))
```
alext: I think you can remove the `testPlanNameWithSnapshot` const as well.

paroxp: Also removed `region` which happens not to be used.

PR #954, opened 2017-06-14
[#141705365] Add MySQL plans

paroxp:
## What

We're slowly enabling the MySQL support in our RDS broker. In preparation, we need to add plans, tests and company of changes to our security groups.

## How to review

- Somewhat dependant on alphagov/paas-rds-broker#50
- Code review
- See travis happy [...]
- Deploy from this branch
- Experience the tests go green
- Run `cf marketplace -s mysql` and see variety of plans
- **Optionally:** Create an instance for your self - _It's covered by the tests_
- Check the plan descriptions and pricing


platform-tests/example-apps/healthcheck/db.go
```diff
@@ -35,7 +44,7 @@ func testDBConnection(ssl bool) error {
 		dbURL.RawQuery = dbURL.RawQuery + "&sslmode=disable"
 	}
 
-	db, err := sql.Open("postgres", dbURL.String())
+	db, err := sql.Open(service, dbURL.String())
```
alext: I'm not sure how this can work for mysql without the mysql driver being imported.

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
+	_ "github.com/go-sql-driver/mysql"
```
alext: This shouldn't be needed here (only in the healthcheck app itself).

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -105,6 +106,93 @@ var _ = Describe("RDS broker", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode).To(Equal(200), "Got %d response from healthcheck app. Response body:\n%s\n", resp.StatusCode, string(body))
+
+			By("disallowing connections from the Healthcheck app without TLS")
+			resp, err = httpClient.Get(helpers.AppUri(appName, "/db?service=mysql&ssl=false"))
+			Expect(err).NotTo(HaveOccurred())
+			body, err = ioutil.ReadAll(resp.Body)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode).NotTo(Equal(200), "Got %d response from healthcheck app. Response body:\n%s\n", resp.StatusCode, string(body))
+			Expect(body).To(MatchRegexp("no pg_hba.conf entry for .* SSL off"), "Connection without TLS did not report a TLS error")
```
alext: This doesn't look right. It looks like a Postgres specific error message that I wouldn't expect from a MySQL instance.

platform-tests/src/acceptance/rds_broker_test.go
```diff
@@ -105,6 +106,93 @@ var _ = Describe("RDS broker", func() {
 	})
 })
 
+var _ = Describe("RDS broker - MySQL", func() {
+	const (
+		serviceName              = "mysql"
+		testPlanName             = "Free"
+		testPlanNameWithSnapshot = "S-dedicated-5.7"
+		region                   = "eu-west-1"
```
alext: Are these last 2 constants used?

platform-tests/example-apps/healthcheck/db.go
paroxp: Aaagh! Not sure, why it has imported the mysql driver in platform-tests/src/acceptance/rds_broker_test.go - which you already commented on.

Sorry about that. Fixing.

platform-tests/src/acceptance/rds_broker_test.go
paroxp: Leftover from postgres tests before the [recent merge](https://github.com/alphagov/paas-cf/commit/431ad38ffdd67e1f3b3296e0eecf998b2edf188b#diff-ebbb4adbeb7f9ea59f6a7015a3a3afbaL30).

Fixing. Thanks.

bleach:
What's the rationale behind not running `cf enable-service-access mysql` in the pipeline? We already do for [postgres](https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-cloudfoundry.yml#L1353) and even [have code to ensure that the tests can access the not-yet-ready mongo service](https://github.com/alphagov/paas-cf/blob/master/platform-tests/src/acceptance/init_test.go#L67).

The story description seems pretty clear that mysql should be available for tenants, so I think we should turn it on for everyone in the pipeline.

paroxp:
It does. It hasn't been mentioned in the kick-off and I wasn't sure if there were no blockers from different aspects.

bleach:
Pulled in & rebased against master in order to fix a problem with compose broker deployment that was fixed in master.

platform-tests/example-apps/healthcheck/db.go
```diff
@@ -24,18 +26,44 @@ func dbHandler(w http.ResponseWriter, r *http.Request) {
...
-func testDBConnection(ssl bool) error {
-	dbURL, err := url.Parse(os.Getenv("DATABASE_URL"))
+func testDBConnection(ssl bool, service string) error {
+	var sslTrigger = "sslmode"
+	var sslOn = "verify-full"
+	var sslOff = "disable"
+
+	dbu := os.Getenv("DATABASE_URL")
+
+	switch service {
```
bleach: I find this hard to read because it combines:
- setting the default service to be postgres
- setting the various db specific parameter names for MySQL, whereas the postgres ones are set as defaults

How about:
- first setting postgres as the default service (above this switch statement
- moving the postgres settings into the switch statement
?

PR #956, opened 2017-06-15
[Support] Setup new VPC peering connection for DIT services

alext:
## What

Details in Deskpro ticket 5766

This is a new VPC they've setup to provision things needed (initially
Redis) by various services they're putting onto the PaaS.

## How to review

* Code review
* Read Deskpro ticket and verify this matches what's requested.

## Who can review

Not me

### DIFF
--- a/terraform/prod.vpc_peering.json
+++ b/terraform/prod.vpc_peering.json
@@ -4,5 +4,11 @@
         "account_id": "513071381340",
         "vpc_id": "vpc-c1f83da6",
         "subnet_cidr": "172.16.1.0/24"
+    },
+    {
+        "peer_name": "dit_services",
+        "account_id": "650575079077",
+        "vpc_id": "vpc-82c913e5",
+        "subnet_cidr": "172.17.0.0/22"
     }
 ]


terraform/prod.vpc_peering.json
```diff
@@ -4,5 +4,11 @@
         "account_id": "513071381340",
         "vpc_id": "vpc-c1f83da6",
         "subnet_cidr": "172.16.1.0/24"
+    },
+    {
+        "peer_name": "dit_services",
+        "account_id": "650575079077",
+        "vpc_id": "vpc-6775a900",
+        "subnet_cidr": "10.30.0.0/16"
```
dcarley: Discussed offline:

This differs from the ranges previously described in the [ADR](https://government-paas-team-manual.readthedocs.io/en/latest/architecture_decision_records/ADR019-accessing-user-provided-services/#using-private-address-space-vpc-peering) and [docs](https://github.com/alphagov/paas-cf/blob/master/doc/vpc_peering.md#how), which makes it a bit harder for us to manage what subnets we can and can't use. Although they should never overlap with `10.0/16`.

@alext is going to ask if it's much work for them to change. If they're not able to change it, then we should consider:

- updating the ADR and docs
- adding tests to prevent overlaps between our infrastructure and peerings

alext:
They've agreed to change the IP range, and have created a new VPC. I've updated the PR now.

This is ready to review again.

dcarley:
👍

PR #957, opened 2017-06-15
[#139001857] Allow platform RDS instances to auto_minor_version_upgrade

alext:
## What

So that we automatically get upgraded to newer point releases when there
are security issues etc.
    
This includes updating the engine_version to only specify the major
version so that terraform won't attempt to downgrade instances that have
been upgraded. It also means that newly created instances will get the
latest point release.

## How to review

* Run the bootstrap pipeline.
* Observe that the CF and CDN RDS instance get auto_minot_version_upgrade set to true.
* Observe that terraform doesn't attempt to change the engine versions.

## Who can review

Not me.


combor:
I don't think that this change needs a bootstrap testing. IMHO code review is enough and it LGTM.

combor:
BTW. I think it's a copy/paste bug :)

PR #960, opened 2017-06-20
[#143876117] Pin compose-broker version

combor:
## What

This pins version of compose-broker. Also it deploys few fixes to prefix broker managed DBs.

## How to review

- Deploy and check if it works
- merge https://github.com/alphagov/paas-compose-broker/pull/3
- tag `master` from https://github.com/alphagov/paas-compose-broker to `v0.1.0`
- remove the temporary commit 

## Who can review

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1047,7 +1047,6 @@ jobs:
               DISABLE_USER_CREATION: {{disable_user_creation}}
               OAUTH_CLIENT_ID: {{oauth_client_id}}
               OAUTH_CLIENT_SECRET: {{oauth_client_secret}}
-              COMPOSE_ACCOUNT_ID: {{compose_account_id}}
               COMPOSE_ACCESS_TOKEN: {{compose_access_token}}
             run:
               path: sh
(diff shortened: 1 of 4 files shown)


dcarley:
> Pin compose broker version
> As we want to have control over versions we deploy and prevent
> auto-updates we need to pin the broker version.

Could you include some details in the commit message of a4c81a5 to describe why the previous pinning solution wasn't suitable?

> Remove compose account ID
> As new version of compose broker doesn't need to read this from
> environment we can safely remove it from our pipeline.

Should I re-run `upload-compose-secrets` to remove the account ID from `compose-secrets.yml` in persistent environments?

combor:
> Should I re-run upload-compose-secrets to remove the account ID from compose-secrets.yml in persistent environments?

It's not required but it would be a good practice to remove unused credentials from the store. We can actually delete it from `paas-pass` as well.

combor:
> Could you include some details in the commit message of a4c81a5 to describe why the previous pinning solution wasn't suitable?

Done.

dcarley:
Removed TMP commit and retested deployment 👍 

I'll re-run `upload-compose-secrets` after it has been deployed to all environments so that I don't break any existing pipelines.

dcarley:
> I'll re-run upload-compose-secrets after it has been deployed to all environments so that I don't break any existing pipelines.

:white_check_mark: Done

PR #962, opened 2017-06-20
[#147124125] Add deployment kick-off functionality

paroxp:
## What

We'd like to kick our environments every morning if possible. To achieve
that, we can create a trigger based on the `time-resource` provided by
concourse, which has pretty neat configuration.

The intention is to speed up our development process.

## How to review

- Code review
- Either:
  - Deploy from branch right before leaving work and expect it to run the next morning
  - Branch off from here and amend the `start` value; `git push`; `make dev pipelines` from a new branch

## Related

- alphagov/paas-bootstrap#71

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ dev: globals ## Set Environment to DEV
 	$(eval export CONCOURSE_AUTH_DURATION=48h)
 	$(eval export DISABLE_PIPELINE_LOCKING=true)
 	$(eval export TEST_HEAVY_LOAD=true)
+	$(eval export ENABLE_MORNING_DEPLOYMENT=true)
 	@true
 
 .PHONY: ci
@@ -151,6 +152,10 @@ bosh-cli: check-env ## Create interactive connnection to BOSH container
 .PHONY: pipelines
 pipelines: check-env ## Upload pipelines to Concourse
 	concourse/scripts/pipelines-cloudfoundry.sh
+
+.PHONY: trigger-deploy
+trigger-deploy: check-env ## Trigger a run of the create-cloudfoundry pipeline.
+	concourse/scripts/trigger-deploy.sh
 
 .PHONY: showenv
 showenv: check-env ## Display environment information
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -102,6 +102,12 @@ jobs:
           params:
             file: updated-datadog-tfstate/datadog.tfstate
 
+
+      - task: await-turn
+        file: paas-cf/concourse/tasks/splay.yml
+        params:
+          DEPLOY_ENV: {{deploy_env}}
+
       - task: delete-deployment
         config:
           platform: linux
@@ -121,17 +127,7 @@ jobs:
             args:
               - -e
               - -c
-              # The following sleep monstrosity deterministically sleeps for a
-              # period of time between 0-20mins in order to prevent all our
-              # deletion jobs running at the same time. See the commit message for
-              # how it works.
               - |
-                sum=$(echo {{deploy_env}} | md5sum);
-                short=$(echo $sum | cut -b 1-15)
-                decimal=$((0x$short));
-                sleeptime=$((${decimal##-} % 60*20));
-                echo "Sleeping for $sleeptime seconds before deletion.."
-                sleep $sleeptime
                 ./paas-cf/concourse/scripts/bosh_login.sh {{bosh_fqdn}} bosh-secrets/bosh-secrets.yml
                 bosh -n delete deployment {{deploy_env}} --force
 
--- a/concourse/pipelines/deployment-kick-off.yml
+++ b/concourse/pipelines/deployment-kick-off.yml
@@ -1,0 +1,56 @@
+---
+resources:
+  - name: paas-cf
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-cf.git
+      branch: {{branch_name}}
+      tag_filter: {{paas_cf_tag_filter}}
+      commit_verification_key_ids: {{gpg_ids}}
+
+  - name: deployment-timer
+    type: time
+    source:
+      days: [Monday, Tuesday, Wednesday, Thursday, Friday]
+      location: Europe/London
+      start: 7:45 AM
+      stop: 8:15 AM
+
+jobs:
+  - name: kick-off
+    serial: true
+    plan:
+      - get: deployment-timer
+        trigger: true
+      - get: paas-cf
+
+      - task: await-turn
+        file: paas-cf/concourse/tasks/splay.yml
+        inputs:
+          - name: deployment-timer
+        params:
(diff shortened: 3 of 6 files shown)


dcarley:
Discussed offline with @paroxp:

We think we'll need to increase the number of retries the CPI attempts, like we did previously in [#112411703](https://www.pivotaltracker.com/n/projects/1275640/stories/112411703), otherwise all environments may fail to start. Example from my deployment this morning:
```
   Failed creating missing vms > router/79647a44-5f98-48b4-957b-268552854b46 (0): Unknown CPI error 'Unknown' with message 'Request limit exceeded.' in 'set_vm_metadata' CPI method (00:07:20)
   Failed creating missing vms > doppler/477a1883-8c6b-48af-8467-c0346a96d247 (2): Unknown CPI error 'Unknown' with message 'Request limit exceeded.' in 'set_vm_metadata' CPI method (00:07:36)
   Failed creating missing vms (00:07:36)

Error 100: Unknown CPI error 'Unknown' with message 'Request limit exceeded.' in 'set_vm_metadata' CPI method
```

alext:
I think this will need some sort of splay applied so that we don't have all dev environments launching at the same time. See https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/autodelete-cloudfoundry.yml#L124-L134 for prior art.

This was added to the autodelete because we were always hitting the rate limit when all the dev environments autodeleted at once.

Makefile
```diff
@@ -73,6 +73,7 @@ list_merge_keys: ## List all GPG keys allowed to sign merge commits.
 .PHONY: globals
 PASSWORD_STORE_DIR?=${HOME}/.paas-pass
 globals:
+	$(eval export ENABLE_MORNING_DEPLOYMENT=false)
```
dcarley: I don't think we should set this when https://github.com/alphagov/paas-cf/pull/962/files#diff-ae4438819742e1613315377720ffa024R153 has a default of `false` because it would follow the pattern of other feature flags and prevent the two settings getting out-of-sync.

dcarley:
We talked offline about the problem of not being able to manually trigger the pipeline in new environments until the timer resource had emitted at least one version overnight. Introducing the additional splay job wouldn't help with this because we still needed to pass a resource over that triggers.

So we're going to use the approach of having a separate pipeline which triggers `create-cloudfoundry` using the Concourse API. We'll also add splaying, to further help prevent them failing due to AWS rate limits, and take the opportunity to move the splay code into a separate task file that can be included in two places.

concourse/pipelines/deployment-kick-off.yml
```diff
@@ -0,0 +1,59 @@
...
+      location: Europe/London
+      start: 7:45 AM
+      stop: 8:15 AM
+
+jobs:
+  - name: kick-off
+    serial: true
+    plan:
+      - get: deployment-timer
+        trigger: {{morning_deployment_trigger}}
```
alext: Given this pipeline is only uploaded to dev environments, do we need this to be parameterised? (It's not parameterised in the autodelete pipeline for example).

concourse/pipelines/deployment-kick-off.yml
```diff
@@ -0,0 +1,59 @@
+---
+resource_types:
+- name: semver-iam
+  type: docker-image
+  source:
+    repository: governmentpaas/semver-resource
+    tag: dc33fee5c9061263e83159a893340f46728d93b5
```
alext: I don't see this resource type used anywhere in this pipeline. Do we need it?

concourse/pipelines/deployment-kick-off.yml
paroxp: I left it as I didn't think we want to enforce everyone to use it... There may be a case when someone would prefer to create an alias `ENABLE_MORNING_DEPLOYMENT=false make dev pipelines`.

If you feel strongly about it, I suppose we can take it out :)

concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -102,6 +102,10 @@ jobs:
           params:
             file: updated-datadog-tfstate/datadog.tfstate
 
+
+      - task: await-turn
+        file: paas-cf/concourse/tasks/splay.yml
+
```
alext: Does this not need a DEPLOY_ENV param like the other pipeline?

concourse/pipelines/deployment-kick-off.yml
```diff
@@ -0,0 +1,59 @@
...
+    serial: true
+    plan:
+      - get: deployment-timer
+        trigger: {{morning_deployment_trigger}}
+      - get: paas-cf
+
+      - task: await-turn
+        file: paas-cf/concourse/tasks/splay.yml
+        inputs:
+          - name: deployment-timer
```
alext: What's this input used for?

concourse/pipelines/deployment-kick-off.yml
paroxp: Missed during cleanup. Thanks.

concourse/scripts/kick-off-pipeline.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "Resource paas-cf must be checkout"
+  exit 1
+fi
+
+if [ "${ENABLE_MORNING_DEPLOYMENT}" != "true" ]; then
+  echo "Pipeline kick-off is disabled. Skipping. (set ENABLE_MORNING_DEPLOYMENT=true to enable)"
+else
+  echo "Pipeline kick-off is enabled. Updating. (set ENABLE_MORNING_DEPLOYMENT=false to disable)"
+
+  make -C ./paas-cf "${AWS_ACCOUNT}" run_job JOB="pipeline-lock"
```
alext: I don't think this is the right thing to use. The `run_job` make task does 2 things. It unpins the `paas-cf` dependency, and then triggers the job. We don't want this to unpin anything.

concourse/scripts/kick-off-pipeline.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -u
+set -e
+
+if [ ! -d "./paas-cf" ]; then
+  echo "Resource paas-cf must be checkout"
+  exit 1
+fi
+
+if [ "${ENABLE_MORNING_DEPLOYMENT}" != "true" ]; then
```
alext: Again, do we need this conditional in here given the pipeline that calls this will only be deployed when this flag is enabled?

concourse/pipelines/deployment-kick-off.yml
alext: But in that case, the whole pipeline would be removed, so there's still no need for this conditional.

paroxp:
Thanks Alex, addressed all.

concourse/pipelines/deployment-kick-off.yml
paroxp: Established, it is not used but there to satisfy the set of tests we're running.

If the lines above would be taken away, travis would say, the source is defined, however not used.

PR #968, opened 2017-06-29
[#148059799] Upgrade ipsec boshrelease

paroxp:
## What

We've applied some logrotate fixes to the boshrelease, which will no
longer prevent logrotate from rotating the logs on ipsec.

## How to review

- Code review
- Run the pipeline from this branch - inb4 Dan: I have run that myself :)
- Experience no issues
- For bonus points: Check if logrotate does warn about permission

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -30,9 +30,9 @@ releases:
     url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/datadog-for-cloudfoundry-0.1.16.tgz
     sha1: 268944eb92dacbdcc8e5920eb36e824dbdb715c0
   - name: ipsec
-    version: 0.1.2
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/ipsec-0.1.2.tgz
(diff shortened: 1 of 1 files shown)


alext:
> For bonus points: Check if logrotate does warn about permission

It would be good to include details on how to check this.

alext:
I'm struggling to reproduce this problem. Even with the directory permissions set to 0775, logrotate appears to rotate the logs without issue.

I tested by creating an extra log file with `dd if=/dev/zero of=/var/vcap/data/sys/log/racoon/test.log bs=1M count=60`, and then running `logrotate -v /etc/logrotate.conf`. It runs without error and successfully rotates the logfile.

paroxp:
Hey!

Sorry, I've described the issue in more detail in the pivotal story comment.

Here's the command to run the logrotator on specific files (ipsec ones):

```
logrotate -dv /etc/logrotate.d/rsyslogs
```

If the folder mentioned in the above configuration is of permission 0775, you should get errors for it.

dcarley:
> Sorry, I've described the issue in more detail in the pivotal story comment.

No need! Ignore your email and be on holiday! :beach_umbrella:

alext:
Given the upstream PR has been merged, I think it still makes sense to merge this so that we remain up to date. I've added to the commit message to capture this point.

PR #976, opened 2017-07-12
[#147829023] Force TLS for MySQL service instances

dcarley:
## What

Use a new version of the RDS broker than forces TLS connections to MySQL service instances. Adjust the acceptance tests to verify this in the same way we do for Postgres.

## How to review

It's best to look at alphagov/paas-rds-broker#52

Before merging:

1. Make sure you've updated alphagov/paas-rds-broker-boshrelease#40 to use a merge commit
1. Make sure you've updated the `wip` commit in this PR to use a final build

## Who can review

Anyone.


chrisfarms:
LGTM

PR #980, opened 2017-07-17
[#145672449] Update link to credentials-high repo

dcarley:
## What

Update link to credentials-high repo

It has been moved to GitHub.com

## How to review

1. Check the new link works.
1. Check that I have't missed any other links.

## Who can review

Anyone.


dcarley:
> The URL text could be different. But it could remain the same as it's pretty generic.

I've updated it to match the new repo name.

PR #982, opened 2017-07-18
[#145682111] Fix Nats Stream Forwarder process check

henrytk:
## What

This Bosh release includes the fix provided on the paas-datadog-for-
cloudfoundry-boshrelease repository[1].

[1] https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/20

## How to review

Check in the build CI this is the correct version and includes the fix mentioned above.

## Who can review

Anyone but me.


dcarley:
> Check in the build CI this is the correct version and includes the fix mentioned above.

👍

PR #984, opened 2017-07-20
[#145517679] Use anomaly detection for cc errors.

combor:
## What

Current alert triggers always if the change if more that 15 errors per half an hour which feels not very suitable for prod. Also we've got false positives when metric is reset after CC redeployment.
The anomaly detection method works for UAA and it seems to be correctly detecting anomalies
while allowing for counter resets.

## How to review

We use exactly same settings for uaa so compare this code with https://github.com/alphagov/paas-cf/blob/master/terraform/datadog/uaa.tf#L6 . Checking on dev may be time consuming as agile algorithm need one day of historical data to learn.  

## Who can review
Not @combor

### DIFF
--- a/terraform/datadog/cloud_controller.tf
+++ b/terraform/datadog/cloud_controller.tf
@@ -71,18 +71,13 @@ resource "datadog_monitor" "cc_failed_job_count_total_increase" {
 }
 
 resource "datadog_monitor" "cc_log_count_error_increase" {
-  name                = "${format("%s Cloud Controller API log error count", var.env)}"
-  type                = "query alert"
-  message             = "${format("Amount of logged errors in Cloud Controller API grew considerably. See logsearch: '@source.component:cloud_controller_ng AND @level:ERROR' {{#is_alert}}%s{{/is_alert}}", var.datadog_notification_in_hours)}"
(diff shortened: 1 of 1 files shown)


combor:
Note: Some lines are changed only due to `terraform fmt`

alext:
> Note: Some lines are changed only due to terraform fmt

Appending `?w=1` to the diff URL makes it much easier to see what's really changed in this case.

combor:
> Appending ?w=1

Nice one!

dcarley:
Looks like a good solution 👍 

I had a to go back to read the commit message for the UAA monitors (attempt) to understand when this would fire. But I guess that's OK, because it's not a simple threshold, we're relying on their algorithm.

I think this is safe to go during our CATS related merge freeze because it doesn't affect the configuration of the platform itself. Team, please shout if you disagree.

> Checking on dev may be time consuming as agile algorithm need one day of historical data to learn.

🙌  could you confirm whether you've tested that it applies cleanly to/from a dev environment? If so, I'll only do code review.

combor:
It applies cleanly ( in term so terraform execution) but I didn't wait one day for agile algorithm to gather enough data.  However I compared metric pattern with UAA and it seems to be very similar ( raising with occasional drops due to restarts.)

dcarley:
Cool, thanks 👍

PR #986, opened 2017-07-21
[#149220549] Patch cf acceptance tests

henrytk:
## What

Until we have a version of capi-release that is compatible with the cf-release we will need to patch the cf-acceptance-tests. See commit message for further details.

## How to review

Deploy on top of master. Make sure the acceptance tests actually run and pass.

The patch of the cf-acceptance-tests is stored on [a branch of our fork of the upstream tests](https://github.com/alphagov/paas-cf-acceptance-tests/tree/fix-v3-app-delete). This is just a branch with no intention to be merged into anything, because it is only useful to us.

## Who can review

Anyone but me or @chrisfarms

### DIFF
--- a/platform-tests/upstream/run_acceptance_tests.sh
+++ b/platform-tests/upstream/run_acceptance_tests.sh
@@ -7,6 +7,23 @@ if  [ "${DISABLE_CF_ACCEPTANCE_TESTS:-}" = "true" ]; then
   exit 0
 fi
 
+# FIXME: Remove this once we are deploying a version of cf-release
+# that includes capi-release >= 1.35.0.
+(
+  cd  "$(pwd)/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/"
+  expected_commit_hash="4a6c59b27bf09aac222ffa02628d1fd47b3a708d"
+  current_commit_hash="$(git log --pretty=format:'%H' -n 1)"
+  if [ "${expected_commit_hash}" != "${current_commit_hash}" ]; then
+    echo "ERROR: Current commit for cf-acceptance-test is different than expected one: ${expected_commit_hash} != ${current_commit_hash}"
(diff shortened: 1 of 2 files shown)


platform-tests/upstream/run_acceptance_tests.sh
```diff
@@ -7,6 +7,23 @@ if  [ "${DISABLE_CF_ACCEPTANCE_TESTS:-}" = "true" ]; then
   exit 0
 fi
 
+# FIXME: Remove this once we are deploying a version of cf-release
+# that includes capi-release >= 1.35.0.
+(
+  cd  "$(pwd)/cf-release/src/github.com/cloudfoundry/cf-acceptance-tests/"
```
combor: Why not simply get this by concourse git resource with a version or branch filter ?

combor: I meant get our fork and point to specific commit.

henrytk: I based this off Hector's prior art: https://github.com/alphagov/paas-cf/commit/716384899c2486af26a1a5a463425adfa6e324da

The advantage is all the code in one place I suppose.

henrytk: Also, another git resource means extra things to poll for and download every single time.

alext: I think this is fine as is. It would be harder to add the check for the expected commit SHA using a resource, and would also require quite a bit more restructuring of how we pull these in (switching from using the submodule to a separate git resource).

combor: TBH I was sure that git resource has `version` filter so it's possible to select specific SHA but I can't find anything like this in docs 🤔 . It's a temp change so I guess I's good if it works.

PR #993, opened 2017-08-04
[#146074949] Add a CF API Pingdom check

henrytk:
## What

We want to monitor the availability of the API in Pingdom, so that
we can report on its availability.

https://www.pivotaltracker.com/n/projects/1275640/stories/146074949

## How to review

* Plan with `make dev pingdom ACTION=plan`. It should want to create all three Pingdom checks. When the same is done for ci, staging, and prod it should only want to create the new check.
* Apply with `make dev pingdom ACTION=apply`. This will test it can create them properly, but won't test the checks actually work because we don't have proper certificates in dev.
* It should alert `the-multi-cloud-paas-team@digital.cabinet-office.gov.uk` after two consecutive failed checks, which should run every minute.
* merge and apply the checks to ci, staging, and prod.

## Who can review

Anyone but me

### DIFF
--- a/doc/pingdom.md
+++ b/doc/pingdom.md
@@ -4,7 +4,7 @@
 
 The checks are defined in terraform configuration file `terraform/pingdom/pingdom.tf`.
 
-The IDs of the contacts to be notified are stored as a comma-delimited string in variable `$PINGDOM_CONTACT_IDS` in the Makefile.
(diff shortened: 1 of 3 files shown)


combor:
This seems to work so LGTM. I also created https://github.com/alphagov/paas-cf/pull/1002 as an addition to this.

PR #997, opened 2017-08-08
[#148289089] Remove leftovers after #991

combor:
## What

We left quite a lot FIXMEs after https://github.com/alphagov/paas-cf/pull/991
This PR cleans them up.

## How to review

Do a sanity check. 

## Who can review

Not @combor


dcarley:
Good catch. Have double checked that they are all applied in prod 👍

PR #998, opened 2017-08-09
Bump the CAPI-release

paroxp:
## What

As CVE-2017-8037 is pointing to the incomplete fix for the CAPI access
to CC VM contents, we need to upgrade out CF-release to version v270 or
the CAPI-release to version 1.38.0.

We went with the CAPI upgrade, as it seems to be a smaller change as
oppose to upgrading the whole platform at the time, which we leave for
the future work.

## How to review

- Code review
- Make sure the version used is the version that fixes the CVE.
- Run your pipeline from this branch - I did not do that, due to lack of dev environment running at present.


dcarley:
I've been working on this too as https://www.pivotaltracker.com/n/projects/1275640/stories/150043715 and have hit some problems with the tests. Let's talk!

dcarley:
Rafal and I talked; I've raised #999 with the test fixes.

PR #999, opened 2017-08-09
[#150043715] Upgrade CAPI and backport CATS

dcarley:
## What

Upgrade CAPI in response to a CVE. Backport CATS to make the tests pass. See the commit messages for more detail.

## How to review

I've run the tests in my dev environment, so I think a code review would be sufficient:

- check that we're using the correct release
- check that I've backported the right things from CATS

## Who can review

Not @dcarley


combor:
This look good. To speed up testing I'll merge this now and check on CI if we are not affected any more.

PR #1000, opened 2017-08-10
[#149558193] Add a basic Elasticsearch plan

LeePorte:
## What

This is a minimal plan which can be used for the private beta testing phase. The word
'Compose' has been removed from the catalog because tenants don't need
to know what the underlying provider is (although they still have a
link to the Compose documentation if needed).

Merging this PR will add the catalogue and use the updated version of compose-broker.

## How to review

Code Review

## Who can review

Not @henrytk or me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1165,7 +1165,6 @@ jobs:
               DISABLE_USER_CREATION: {{disable_user_creation}}
               OAUTH_CLIENT_ID: {{oauth_client_id}}
               OAUTH_CLIENT_SECRET: {{oauth_client_secret}}
-              COMPOSE_ACCESS_TOKEN: {{compose_access_token}}
             run:
               path: sh
               args:
@@ -1603,7 +1602,7 @@ jobs:
               - name: config
               - name: bosh-CA
             params:
-              COMPOSE_ACCESS_TOKEN: {{compose_access_token}}
+              COMPOSE_API_KEY: {{compose_api_key}}
               DB_PREFIX: {{deploy_env}}
               CLUSTER_NAME: "gds-eu-west1-c00"
             run:
@@ -1627,7 +1626,7 @@ jobs:
                       'LOG_LEVEL' => 'DEBUG',
                       'USERNAME' => 'compose-broker',
                       'PASSWORD' => '${COMPOSE_BROKER_PASS}',
-                      'ACCESS_TOKEN' => '${COMPOSE_ACCESS_TOKEN}',
+                      'COMPOSE_API_KEY' => '${COMPOSE_API_KEY}',
                       'DB_PREFIX' => '${DB_PREFIX}',
                       'CLUSTER_NAME' => '${CLUSTER_NAME}',
                     }}
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -256,7 +256,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-compose-broker
-      tag_filter: v0.5.0
+      tag_filter: v0.8.0
 
 jobs:
   - name: pipeline-lock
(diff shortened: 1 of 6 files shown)


henrytk:
1000 baby.

chrisfarms:
I just rebased this against master (no conflicts) to deploy to dev.

However the compose broker now expects a `COMPOSE_API_KEY` env var but the jobs export `COMPOSE_ACCESS_TOKEN` so deploy fails.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1165,7 +1165,7 @@ jobs:
               DISABLE_USER_CREATION: {{disable_user_creation}}
               OAUTH_CLIENT_ID: {{oauth_client_id}}
               OAUTH_CLIENT_SECRET: {{oauth_client_secret}}
-              COMPOSE_ACCESS_TOKEN: {{compose_access_token}}
+              COMPOSE_API_KEY: {{compose_api_key}}
```
henrytk: Unrelated to this story: Why is this passed to this task? I don't see where it is actually used in the manifest.

concourse/scripts/lib/compose.sh
```diff
@@ -5,12 +5,14 @@ set -u
...
-  export compose_access_token
+  export compose_api_key
+  export COMPOSE_API_KEY
   if aws s3 ls "${secrets_uri}" > /dev/null ; then
     secrets_file=$(mktemp -t compose-secrets.XXXXXX)
 
     aws s3 cp "${secrets_uri}" "${secrets_file}"
-    compose_access_token=$("${SCRIPT_DIR}"/val_from_yaml.rb compose_access_token "${secrets_file}")
+    compose_api_key=$("${SCRIPT_DIR}"/val_from_yaml.rb compose_api_key "${secrets_file}")
+    COMPOSE_API_KEY=$("${SCRIPT_DIR}"/val_from_yaml.rb compose_api_key "${secrets_file}")
```
henrytk: This upper case one can come out.

concourse/scripts/lib/compose.sh
```diff
@@ -5,12 +5,14 @@ set -u
 get_compose_secrets() {
   # shellcheck disable=SC2154
   secrets_uri="s3://${state_bucket}/compose-secrets.yml"
-  export compose_access_token
+  export compose_api_key
+  export COMPOSE_API_KEY
```
henrytk: This upper case one can come out.

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -141,7 +141,8 @@ disable_custom_acceptance_tests: ${DISABLE_CUSTOM_ACCEPTANCE_TESTS:-}
 disable_pipeline_locking: ${DISABLE_PIPELINE_LOCKING:-}
 datadog_api_key: ${datadog_api_key:-}
 datadog_app_key: ${datadog_app_key:-}
-compose_access_token: ${compose_access_token:-}
+compose_api_key: ${compose_api_key:-}
+COMPOSE_API_KEY: ${compose_api_key:-}
```
henrytk: This upper case one can come out.

chrisfarms:
There are issues with the `make <env> upload-compose-secrets` script that are currently preventing the new secrets file from being pushed to ci/staging/prod environments.

It appears that there are two issues:

* The script is looking for credentials in the wrong password store location (compose secrets for prod were moved from "credentials" to "credentials-high")
* The script is looking for credentials at pass path `compose/<env>/access_token` but no keys are stored at that path.

config/service-brokers/compose/catalog.json
```diff
@@ -27,5 +27,33 @@
         "units": 1
       }
     }]
+  },{
+    "id": "6e9202f2-c2e1-4de8-8d4a-a8c898fc2d8c",
+    "name": "elastic_search",
```
alext: This should not have an underscore in it. Elasticsearch is not two words - https://www.elastic.co/products/elasticsearch

Spelling it like this will only add friction for users.

PR #1001, opened 2017-08-10
[#146074923] Monitor CF API availability during deployment

chrisfarms:
## What

We want to monitor the availability of the CF API during a deploy.

As a first step we do not want to test _everything_ and this change focuses on monitoring a couple of read-only API calls, but with a view to enabling adding more checks in the future.

To achieve this goal this change:

* Adds a new `api-availability-tests` job to the pipeline designed to start before the `cf-deploy` job begins and end after `cf-deploy` finishes.
* Adds a testing/monitoring system that repeatably tries blocks of CF API calls and monitors the number of successes and failures and reports back any collected errors during the run.
* Adds two relatively simple blocks of CF API calls to the monitor the equivalent of `cf apps` and `cf app myapp`

## Notes

The general idea here is similar to the app-availability-tests that use the vegeta library to put an app under load during deploy and monitor the response, only rather than focusing on single HTTP requests we are interested in sets of CF API calls that simulate users doing actions with the CF CLI. 

## How to review

* [x] Read/Review the individual commits
* [x] Deploy to a dev environment
    * check that the new test job runs during cf-deploy.
    * check that the new test job has some useful output at the end.
    * try and force a failure (by for example removing the `API-AVAILABILITY-*` app deployed to the healthchecks space during a deploy).
* [ ] Check logic surrounding when/how pre-deploy blocks the pipeline
* [ ] Check that the monitor task roughly emulates a "cf app myapp" type API call
* [ ] Check that the kind of stats collected are useful enough
* [ ] Check that the conditions that cause the test to fail are suitable

## Who can review

Not @chrisfarms
Not @46bit 
Not @combor (although it changed so much since he was on it that it's probably ok)


platform-tests/src/api/api_test.go
```diff
@@ -0,0 +1,130 @@
...
+		Expect(cfg.Password).ToNot(BeEmpty(), "CF_PASS environment variable must be set")
+
+		deployment := &Deployment{
+			AtcAddress: os.Getenv("CONCOURSE_ATC_URL"),
+			Password:   os.Getenv("CONCOURSE_ATC_PASSWORD"),
+			Username:   os.Getenv("CONCOURSE_ATC_USERNAME"),
+			Version:    os.Getenv("PIPELINE_TRIGGER_VERSION"),
+			Team:       "main",
+		}
+		Expect(deployment.AtcAddress).ToNot(Equal(""), "CONCOURSE_ATC_URL environment variable must be set")
```
paroxp: Being interested, is there a reason you used `Equal("")` whilst above you had `BeEmpty()`?

platform-tests/src/api/api_test.go
```diff
@@ -0,0 +1,130 @@
...
+			return nil
+		})
+
+		monitor.Add("Fetching detailed app information", func(cf *cfclient.Client) error {
+			org, err := cf.GetOrgByName("admin")
+			if err != nil {
+				return fmt.Errorf("Failed to fetch 'admin' org")
+			}
+
+			apps, err := cf.ListAppsByQuery(url.Values{"q": []string{
```
paroxp: Interested again, you seem to have changed the style of how you build `url.Values`. Any reason for that?

platform-tests/src/api/api_test.go
```diff
@@ -0,0 +1,130 @@
...
+				"name:" + appName,
+				"organization_guid:" + org.Guid,
+			}})
+			if err != nil {
+				return err
+			} else if len(apps) != 1 {
+				return fmt.Errorf("expected a single app named '%s' got %d", appName, len(apps))
```
paroxp: Not really sure if the returned error message makes sense. - This could be just me.

platform-tests/src/api/api_test.go
```diff
@@ -0,0 +1,130 @@
...
+					}
+					lg("failed to get status from concourse [", concourseConnectionAttemptsRemaining, " attempts remaining]", err)
+				} else if done {
+					concourseConnectionAttemptsRemaining = 5
+					lg("detected deployment job completed, stopping monitor")
+					monitor.Stop()
+					return
+				}
+			}
+		}(5)
```
paroxp: Could we state that `5` and the one from line 115 in a constant for easier management in the future?

platform-tests/src/api/concourse.go
```diff
@@ -0,0 +1,129 @@
...
+	if resourceVersionID == 0 {
+		return nil, fmt.Errorf("Resource: %s with version: %s did not exist in Concourse", resourceName, resourceVersion)
+	}
+
+	builds, _, err := team.BuildsWithVersionAsInput(pipelineName, resourceName, resourceVersionID)
+	if err != nil {
+		return nil, err
+	}
+
+	return filterBuildsByNameAndSortByTime(builds, jobName), nil
```
paroxp: Did you consider separating that functionality? I wonder if it'd be perhaps beneficial, in case we'd like to change something in the future.

```
filtered := filterBuilds("name", builds, jobName)
sorted := sortBy("time", filtered, "asc")
```

- NOT A REQUIREMENT - We could well cross that bridge when we get to it.

paroxp:
Could you also rebase of master?

platform-tests/src/api/api_test.go
chrisfarms: The reason was that I wrote them on different days 😇  ... fixup en route `BeEmpty` should semantically be used for checking if lists are empty really according to ginkgo docs.

platform-tests/src/api/api_test.go
chrisfarms: Nope no reason... more consistant in latest.

platform-tests/src/api/api_test.go
chrisfarms: sure. Will update to "Failed to find any apps in the 'healthchecks' space, expected at least one to be returned" and alter the other similar error to match.

platform-tests/src/api/api_test.go
chrisfarms: Yep

PR #1002, opened 2017-08-10
Improve pingdom script

combor:
## What

This was crafted out of my frustration while testing https://github.com/alphagov/paas-cf/pull/993

Now we make sure we pin and require terraform version. The version was unpinned and newer terraform doesn’t have `remote` command. Also the state file doesn’t match pingdom plugin version and complains that it was created with newer terraform while trying to use 0.8.5

Also, we try to not download terraform plugin every time. This was supper annoying while working via VPN and trying to find out while this script fails as the download was very slow.

## How to review

Run `make ci pingdom ACTION=plan` and check if it works

## Who can review
Not me

### DIFF
--- a/terraform/scripts/set-up-pingdom.sh
+++ b/terraform/scripts/set-up-pingdom.sh
@@ -18,12 +18,18 @@ PAAS_CF_DIR=$(pwd)
 WORKING_DIR=$(mktemp -d terraform-pingdom.XXXXXX)
 trap 'rm -r "${PAAS_CF_DIR}/${WORKING_DIR}"' EXIT
 
-wget "https://github.com/alphagov/paas-terraform-provider-pingdom/releases/download/${VERSION}/${BINARY}" \
-  -O "${WORKING_DIR}"/terraform-provider-pingdom
-chmod +x "${WORKING_DIR}"/terraform-provider-pingdom
+if [ ! -d bin/ ]; then
+  mkdir bin/
+fi
+
+#wget can only check timestamp on a file in work dir
+cd bin/
+wget -N "https://github.com/alphagov/paas-terraform-provider-pingdom/releases/download/${VERSION}/${BINARY}" 
+cp ./"${BINARY}" "${PAAS_CF_DIR}"/"${WORKING_DIR}"/terraform-provider-pingdom
+chmod +x "${PAAS_CF_DIR}"/"${WORKING_DIR}"/terraform-provider-pingdom
 
 # Work in tmp dir to ensure there's no local state before we kick off terraform, it prioritises it
-cd "${WORKING_DIR}"
+cd "${PAAS_CF_DIR}"/"${WORKING_DIR}"
 
 # Configure Terraform remote state
 terraform remote config \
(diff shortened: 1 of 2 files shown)


terraform/scripts/set-up-pingdom.sh
```diff
@@ -2,10 +2,18 @@
...
 TERRAFORM_ACTION=${1}
+TERRAFORM_VERSION="0.8.7"
 VERSION=0.2.2
 BINARY=terraform-provider-pingdom-tf-0.8.5-$(uname -s)-$(uname -m)
 STATEFILE=pingdom-${AWS_ACCOUNT}.tfstate
 
+#Check terraform version
+
+if [[ $(terraform version 2>&1 | head -1| sed -e 's/Terraform v//') != "${TERRAFORM_VERSION}" ]]; then
+  echo "Required Terrform version ${TERRAFORM_VERSION}"
```
keymon: typo

combor: Fixed :)

terraform/scripts/set-up-pingdom.sh
```diff
@@ -2,10 +2,18 @@
...
 VERSION=0.2.2
 BINARY=terraform-provider-pingdom-tf-0.8.5-$(uname -s)-$(uname -m)
 STATEFILE=pingdom-${AWS_ACCOUNT}.tfstate
 
+#Check terraform version
+
+if [[ $(terraform version 2>&1 | head -1| sed -e 's/Terraform v//') != "${TERRAFORM_VERSION}" ]]; then
+  echo "Required Terraform version ${TERRAFORM_VERSION}"
+  exit 1
+fi
```
alext: Instead of manually checking the version like this, I think it would be better to use Terraform's built-in version constraint feature - https://www.terraform.io/docs/configuration/terraform.html

combor: This is a good idea but doesn't help if internal command is missing. I left the check in the script but added built-in in `global.tf` . Have a look.

terraform/globals.tf
```diff
@@ -1,3 +1,7 @@
+terraform {
+  required_version = "0.8.7"
+}
```
alext: This isn't the correct place to add this - the pingdom terraform doesn't include this file. There's also a separate `version_constraint.tf` file that's included in all the terraform configs that already specifies the version, although it's currently more relaxed and only specifies `>= 0.8.5`.

I've thought that we should pin the version we're using globally more strictly - we should make it exactly match the version in `paas-docker-cloudfoundry-tools`, but that's currently 0.8.5, which poses problems for terraform configs that have moved beyond that.

From here, I think there are 2 options. We can either pin the version used by the pingdom config separately from the rest (which isn't great for consistency), or we can defer the changes to the version pinning to a later story, and reduce this PR to the second commit only (to save unnecessary downloads).

combor: `>= 0.8.5` referenced in `version_constraint.tf` isn't correct as state file was created with version `0.8.7`. As a simplest thing we can change the version in this file to `=0.8.7` that will solve all the problems but still will be a bit confusing about plugin version. That can be refactored during terraform upgrade.

PR #1011, opened 2017-08-16
[WIP] Resilient rebuild of the Buildpack Release Notes Generator

46bit:
## What

We've been upgrading our Cloud Foundry deployment from v264 to v269. Part of this is obtaining a combined buildpack changelog to give our tenants. [An existing shell script](https://github.com/alphagov/paas-cf/blob/2d8b055477e1697f9e8fe4a84c5051c2480e09f3/scripts/generate_buildpack_release_notes.sh) gave incorrect and inconsistent output. It wasn't reasonable to maintain and we rebuilt it in Python with many improvements.

The new script:

1. Finds submodules matching `src/*-buildpack-release` which have changed between the two releases.
2. Obtains a list of the interceding versions of the buildpack by comparing the new and old `releases/.../index.yml`.
3. Generates a summary markdown file.
4. Generates a detailed summary markdown file, containing buildpack changenotes fetched from Github's API.

Newly added and updated buildpacks are handled automatically, but we can't easily detect renamed or removed buildpacks. The script should not crash in such situations. We warn to check manually for them but neither have yet occurred in `cf-release`.

## Status

**While reviewing the output we have identified two problems:**

1. [ ] Some buildpack versions do not have corresponding buildpack releases. For example the [staticfile-buildpack v1.4.7](https://github.com/cloudfoundry/staticfile-buildpack/releases/tag/v1.4.7). Our present logic does not find these. We'll need to look directly at the buildpack tags rather than buildpack-release tags, which should be doable via the API.
2. [ ] The summary's URLs of buildpack releases are broken links. They miss a `v` before the number.

This doesn't block the release of [#149400231]. I intend to do this work in spare time as it needs a little more thought.

## How to review

1. [ ] Follow the instructions from running `./scripts/generate_buildpack_release_notes.py`.
2. [ ] Choose two [versions of cf-release](https://github.com/cloudfoundry/cf-release/releases) and try to verify the script's output is correct.

## Who can review

Not @46bit @LeePorte.


dcarley:
This looks promising 👍 

It would be good to have tests if we progress it further.

matthewford:
As an aside is there a place where the current build pack versions that the paas is running published anywhere - while the update emails are useful, it would be great to be able to view this info in the technical docs for example.

bleach:
@matthewford you can see them by running `cf buildpacks`

alext:
I'm going to close this for now as it's not being actively worked on at the moment. We can reopen the PR when it's ready to be reviewed.

PR #1014, opened 2017-08-17
Increase smoke test timeout

henrytk:
## What

No associated Pivotal story

We are having performance issues with certain CF API operations, such
as deleting orgs during the smoke tests. This means it is taking over
the timeout of 30 seconds and therefore causing continuous smoke test
failures.

We will patch the tests temporarily until the performance issue is investigated.

## How to review

This has not been tested yet. It is unlikely you will experience the same performance issues as production, so I recommend:
* deploying from this branch
* hijacking the smoke tests container while they are running
* verifying the imported cf-release contains a smoke test module with this patch: https://github.com/cloudfoundry/cf-smoke-tests/compare/master...alphagov:increase-timeout
* doing a code review on https://github.com/cloudfoundry/cf-smoke-tests/compare/master...alphagov:increase-timeout to make sure it really does increase the timeout.

## Who can review

Anyone but me.

### DIFF
--- a/platform-tests/upstream/run_smoke_tests.sh
+++ b/platform-tests/upstream/run_smoke_tests.sh
@@ -2,6 +2,22 @@
 
 set -eu
 
+# FIXME: Remove this once we have resolved the performance issues resulting
+# in the SMOKE_TESTS org taking longer than 30 seconds to delete.
+(
+  cd  "$(pwd)/cf-release/src/smoke-tests/"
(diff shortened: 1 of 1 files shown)


henrytk:
I've deployed this and hijacked the container: it really does apply the patch. All we really need is code review on the branch (it doesn't need to be merged).

chrisfarms:
LGTM

PR #1015, opened 2017-08-18
Revert "Increase smoke test timeout"

henrytk:
## What

No associated Pivotal story.

This reverts commit 93a24f38af8fe5beec52bf0c5249e51a91429ad6.

We now know the smoke test failures were due to degraded API performance
caused by the CDN broker taking a long time to process the last operation
requests. This timeout should be restored so we can detect similar performance
degradations in future.

## How to review

In my opinion deployment of this is unnecessary and a code review alone is sufficient.

## Who can review

Anyone but me.

### DIFF
--- a/platform-tests/upstream/run_smoke_tests.sh
+++ b/platform-tests/upstream/run_smoke_tests.sh
@@ -2,22 +2,6 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


46bit:
Exactly reverts #1014. LGTM.

PR #1017, opened 2017-08-23
[#148289089] Fix healthcheck-db destroy.

alext:
## What

This was missed when we moved the healthcheck app into a different
org(#991). This therefore updates this task to look in the correct
org/space for the app.

I've removed the test for the org existing because the admin org always
exists.

## How to review

Code review is probably enough. Alternatively, you could run the destroy pipeline, and verify that the `remove-healthcheck-db` task runs, and deletes the healthcheck app. (The DB isn't enabled in dev by default).

## Who can review

Not me.


dcarley:
LGTM. Code reviewed only.

PR #1018, opened 2017-08-23
[#150288087] Operational/Usage Metrics Collection

chrisfarms:
## What

We want to have more insight into usage of the platform.

This change adds a new cf application `paas-metrics` that is deployed by the pipeline, assigned the `cloudfoundry.global_auditor` role and will then periodically collects various usage statistics and passes them along to datadog.

The metrics currently collected and some more details can be found in the [README](https://github.com/alphagov/paas-cf/blob/150288087-op-metrics/tools/metrics/README.md).

## How to review

* [ ] [Optional] Sign up for a new datadog account and update your dev datadog secrets (just in case we don't want to add unwanted noise to live datadog until ready)
* [x] Deploy to dev
* [x] Leave it running for a little while, maybe create a few apps/users/orgs.
* [x] Check the `op.*` stats in datadog
* [x] Tell @chrisfarms to increase the polling intervals (I've intentionally set the polling interval of many of the metrics quite low so we can see them come in - most should be raised higher to something like every 5 or 10 mins before merge)

## Who can review

Not @chrisfarms or @LeePorte


tools/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+
+func NewMetricBuffer(n int) *MetricBuffer {
+	events := make(chan Metric, n)
+	ctx, cancel := context.WithCancel(context.Background())
+	return &MetricBuffer{events, ctx, cancel}
+}
+
+func NewMultiMetricReader(rs ...MetricReader) MetricReadCloser {
+	if len(rs) < 1 {
+		panic("you need at least one reader")
```
paroxp: Would you not prefer to return an error here as oppose to halting the app?

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2127,6 +2221,67 @@ jobs:
...
+                    }
+                    manifest = YAML.load_file('manifest.yml')
+                    manifest['applications'].each { |app|
+                      app['env'] = {} unless app['env']
+                      app['env'].merge!(env)
+                    }
+                    File.write('manifest.yml', manifest.to_yaml)
+                  "
+
+                  if  [ "${DISABLE_METRICS:-}" != "true" ] ; then
```
paroxp: Could we move this if statement to one of the first lines of this task? We probably don't need all that prep work to be done when we're not going to end up deploying this.

tools/metrics/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+You could then consume it by copying the metrics to a writer using CopyMetrics:
+
+```go
+_ = CopyMetrics(reporter, RandomMetric)
+```
+
+### Updating dependencies
+
+dep is used for dependencies. to update the vendor dir do:
```
paroxp: 🎉  Our first tool using `dep`? 🎉

Makefile
```diff
@@ -80,6 +80,8 @@ globals:
 .PHONY: dev
 dev: globals ## Set Environment to DEV
 	$(eval export AWS_ACCOUNT=dev)
+	$(eval export DISABLE_METRICS=false)
```
chrisfarms: Shall I make it `DISABLE_METRICS ?= true` so you can enable it if you want?

Makefile
```diff
@@ -128,6 +132,7 @@ staging: globals ## Set Environment to Staging
 .PHONY: prod
 prod: globals ## Set Environment to Production
 	$(eval export AWS_ACCOUNT=prod)
+	$(eval export DISABLE_METRICS=false)
```
chrisfarms: 🙌

Makefile
chrisfarms: After reading rest of comments...

I'll make it `ENABLE_METRICS ?= false`

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1243,6 +1242,7 @@ jobs:
           - get: paas-rubbernecker
           - get: paas-compose-broker
 
+
```
chrisfarms: It really brings the code together 👻  ... willfix

manifests/cf-manifest/scripts/generate-cf-secrets.rb
```diff
@@ -45,6 +45,7 @@
   "ssh_proxy_host_key" => :ssh_key,
   "kibana_admin_password" => :simple,
   "bbs_encryption_key" => :simple,
+  "paas_metrics_password" => :simple,
```
chrisfarms: No need for paas-pass ... this pipeline generates everything it needs

paroxp: https://github.com/alphagov/paas-credentials/pull/87

tools/metrics/metrics.go
chrisfarms: That check isn't really needed tbh ... so yeah I'll just remove it.

I like the Readers all being composable and the errors coming out during ReadEvent. :)

concourse/pipelines/create-cloudfoundry.yml
chrisfarms: Yep. Agreed. Done

tools/metrics/gauges.go
```diff
@@ -0,0 +1,202 @@
...
+				return err
+			}
+			numberOfUsers += response.TotalResults
+		}
+		return w.WriteMetrics([]Metric{
+			{
+				Kind:  Gauge,
+				Time:  time.Now(),
+				Name:  "users.count",
+				Value: float64(numberOfUsers),
```
chrisfarms: Yep you're right ... this way of calculating users was a quick fix for the fact the global auditor role could not fetch `/v2/users`... clearly _too_ quick.

I've de-duplicated it now by counting unqiue user guids.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1798,6 +1798,100 @@ jobs:
...
+                    # CF v269, and now just prints the `201` code on its own line
+                    # followed by a space.
+                    if [ $result = 0 ] && echo "${result}" | grep '^201 $'; then
+                      echo "${result}"
+                      echo "Added $2 to $1"
+                    else
+                      echo "${result}"
+                      exit 1
+                    fi
+                  }
```
chrisfarms: Yes I think you're right a standard "create user" script is probably a good idea.

...but I don't feel confident messing with the create_admin script to factor it out as part of this changeset :)

chrisfarms:
All squashed up, and believe I've addressed all comments.

concourse/pipelines/create-cloudfoundry.yml
dcarley: Is it possible/preferable to add the user from the CF manifest instead?

chrisfarms: @dcarley ... Could you expand on that a bit? 

.... I'm not too familiar with these bits, I could probably do with a pair to get oriented before making a change like that 😬

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2127,6 +2220,70 @@ jobs:
...
+        - task: deploy-paas-metrics
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
+                tag: 895cf6752c8ec64af05a3a735186b90acd3db65a
+            params:
+              ENABLE_METRICS: ${ENABLE_METRICS}
```
paroxp: This ins't going to parse. You may need to modify to fit what we're doing for datadog below.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1798,6 +1797,100 @@ jobs:
...
+        - task: create-paas-metrics-user
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-uaac
+                tag: 895cf6752c8ec64af05a3a735186b90acd3db65a
+            params:
+              ENABLE_METRICS: ${ENABLE_METRICS}
```
paroxp: This ins't going to parse. You may need to modify to fit what we're doing for datadog in the other comment.

chrisfarms: Good catch. Resolved.

concourse/pipelines/create-cloudfoundry.yml
chrisfarms: Resolved.

tools/metrics/gauges.go
```diff
@@ -0,0 +1,264 @@
...
+				Kind:  Gauge,
+				Time:  time.Now(),
+				Name:  "quota.routes.reserved", // number of routes reserved
+				Value: float64(reservedRoutes),
+			},
+		})
+	})
+}
+
+func UserCountGauge(c *Client, interval time.Duration) MetricReadCloser {
```
paroxp: Just learned one more thing. If the user is not assigned to the org, it won't count them.

It's an edge case scenario, but something to consider/fix/document.

chrisfarms: Yeah good point... I'm not sure if we care about users not-assigned to orgs since we're interested in tennant usage but you're right it's worth mentioning in the README.

chrisfarms: I've add a footnote to the README to mention this.

paroxp:
We're going to be relying on this app. We should love it, take care of it :D

Do you think it would be beneficial to have some tests done here?

inb4 Dan:
> if you liked it then you shoulda put a test on it

chrisfarms:
💍 Added some tests around the lower-level Reader / Writer bits

 ... tricky to do integration tests against datadog / cloud foundry, probably best to just create an alert in datadog to check that the `op.*` metrics are coming in regularly if we want to check that.

tools/metrics/metrics_test.go
```diff
@@ -0,0 +1,169 @@
...
+		return bang
+	})
+	defer poller.Close()
+	_, err := poller.ReadMetric()
+	if err != bang {
+		t.Fatalf("expected poller to return error 'BANG!' got: %v", err)
+	}
+}
+
+func TestCopyMetrics(t *testing.T) {
```
paroxp: There are times when I get the following error:

```
--- FAIL: TestCopyMetrics (0.00s)
	metrics_test.go:167: expected src == dst, got: [{  test.a 0001-01-01 00:00:00 +0000 UTC 0 []} {  test.b 0001-01-01 00:00:00 +0000 UTC 0 []}]
```

Tested with:
```sh
for i in {1..10}; do go test -v; done
```

paroxp:
Could you add `go test -v` to our `make spec` target?

This will ensure the tests are run on travis.

concourse/pipelines/create-cloudfoundry.yml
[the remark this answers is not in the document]
dcarley: Sorry, delayed response..

I think a client like one of these might work for this use case: https://github.com/alphagov/paas-cf/blob/816d1830cd42eeb8b48bf1d38e563825d18f6113/manifests/cf-manifest/manifest/020-cf-properties.yml#L486-L504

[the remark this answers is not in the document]
chrisfarms: ah ok yes this looks much neater... will give it a go

chrisfarms:
@paroxp ... should be ready for (re)-review now:

* user/client now created via the manifest as per dan's suggestion, which is much more sensible. It does however cause uaa to get restarted during deploy so expect to see a few "api-availability-tests" failures. 
* Travis will now run the tests, but it required a bit of GOPATH jiggery-pokery in the before_install.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1798,6 +1797,27 @@ jobs:
...
+              - name: bosh-secrets
+            run:
+              path: sh
+              args:
+                - -e
+                - -c
+                - |
+                  ./paas-cf/concourse/scripts/bosh_login.sh {{bosh_fqdn}} bosh-secrets/bosh-secrets.yml
+
+                  bosh cleanup --all
```
alext: This addition looks to be erroneous (this is a duplicate of the existing job in this pipeline). I've raised #1062 to remove it.

PR #1021, opened 2017-08-29
[#147668907] Upgrade stemcells to mitigate USN-3392-2

LeePorte:
## What

This commit applies the mitigation to USN3392-2 by upgrading the
stemcells in use from 3421.19 to 3421.20

## How to review

Deploy this branch on top of a dev master.

Some small API downtime may be recorded in the API availability tests.

## Who can review

Anyone but @LeePorte


combor:
I think code review is enough in this case as this is a minor stemcell change and we know that UAA upgrade process will create downtime and we can't currently avoid it. I have confirmed on https://github.com/cloudfoundry/bosh-linux-stemcell-builder/releases/tag/stable-3421.20 that this is a correct stemcell version that patches USN-3392-2

PR #1022, opened 2017-08-30
[#149945364] Provision user provided Compose service

paroxp:
## What

This script will provision a provided type of database for you using
exported Compose token and later will use returned credentials to create
user provided service and make it accessible for a specific org.

## How to review

- Code review

## Who can review

Neither @combor nor @paroxp

### DIFF
--- a/scripts/compose-user-provided-service.sh
+++ b/scripts/compose-user-provided-service.sh
@@ -1,0 +1,171 @@
+#!/bin/bash
+set -euo pipefail
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+SCRIPT=$0
+COMPOSE_API_URL='https://api.compose.io/2016-07'
+
+# shellcheck disable=SC1090
+source "${SCRIPT_DIR}/common.sh"
+
+usage() {
+  cat <<EOF
+  Usage:
+    $SCRIPT [-n <name>] -o <orgname> -s <space[,space,...]> -t <type>
+
+$SCRIPT will manually create an instance
+of defined type with the use of compose API and assign it to the specific
+organisation as a CloudFoundry User Provided Service.
+
+Requirements:
+
+ * You must login with the cf client with an "admin" user.
+ * You must have the compose_api_token environment variable set.
+
+Where:
+  -n <name>    Name of the service that should be created. If flag is not set
+               a default value will be generated.
+
+  -o <orgname> Organisation the service should be assigned to upon creation.
+
+  -s <space>   Comma separated space names the service should be assigned to
+               upon creation.
+
+  -t <type>    Service type to be created. Supported:
+               mongodb, redis, postgresql, elastic_search, rethink, rabbitmq,
+               etcd, scylla, mysql, disque
+EOF
+exit 1
+}
+
+get_cluster_data(){
+  compose_cluster_data=$(curl -s -X GET "${COMPOSE_API_URL}/clusters" \
+    -H "Authorization: Bearer ${COMPOSE_API_TOKEN}" \
+    -H "Content-Type: application/json" \
+    2>/dev/null | jq -r ._embedded.clusters[0])
+
+  compose_cluster_id=$(echo "${compose_cluster_data}" | jq -r .id)
+  compose_account_id=$(echo "${compose_cluster_data}" | jq -r .account_id)
+}
+
+prepare_name() {
+  id=$(date +%s | md5)
+  deployment_name=$(echo "alpha-${TYPE}-${ORG}-${id}" | cut -c1-50)
+}
+
+provision_instance() {
+  local body
+  body=$(printf '{
+    "deployment": {
+      "name": "%s",
+      "account_id": "%s",
+      "cluster_id": "%s",
+      "type": "%s"
+    }
+  }' "${SERVICE_NAME:-$deployment_name}" "${compose_account_id}" "${compose_cluster_id}" "${TYPE}")
+
+  uri=$(curl -s -X POST "${COMPOSE_API_URL}/deployments" \
+      -H "Authorization: Bearer ${COMPOSE_API_TOKEN}" \
+      -H "Content-Type: application/json; charset=utf-8" \
+      -d "${body}" | jq -r .connection_strings.direct[0])
+
+  pattern='^(([a-z]{3,5})://)?((([^:\/]+)(:([^@\/]*))?@)?([^:\/?]+)(:([0-9]+))?)(\/[^?]*)?(\?[^#]*)?(#.*)?$'
+  [[ "$uri" =~ $pattern ]] || return 1;
+
+  uri=${BASH_REMATCH[0]}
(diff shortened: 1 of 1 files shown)


scripts/compose-user-provided-service.sh
```diff
@@ -0,0 +1,122 @@
...
+  local body
+  body=$(printf '{
+    "username": "%s",
+    "password": "%s",
+    "host": "%s",
+    "port": "%s",
+    "uri": "%s"
+  }' "${uri_user}" "${uri_password}" "${uri_host}" "${uri_port}" "${uri}")
+
+  cf create-user-provided-service "${deployment_name}" -p "${body}"
```
alext: This hasn't targeted an org/space, so will create this in the current org/space.

paroxp:
- Added some flags, to support user provided name.
- Added the missed `cf target` which also implements the coma separated value for spaces.
- Added warning message to remind user of whitelist changes has been added.
- Didn't add the `deployment.note` field, as I wasn't sure if generated message would be sufficient and having it as a flag would be an overkill IMO.

paroxp:
- Added some flags, to support user provided name.
- Added the missed `cf target` which also implements the coma separated value for spaces.
- Added warning message to remind user of whitelist changes has been added.
- Didn't add the `deployment.note` field, as I wasn't sure if generated message would be sufficient and having it as a flag would be an overkill IMO.

46bit:
I'm reviewing the script while @alext is busy.

scripts/compose-user-provided-service.sh
```diff
@@ -0,0 +1,164 @@
...
+    info "Targeting organisation '${ORG}' and space '${i}'.\n"
+    cf target -o "${ORG}" -s "${i}"
+    cf create-user-provided-service "${SERVICE_NAME:-$deployment_name}" -p "${body}"
+    printf "\n"
+  done
+}
+
+check_params_and_environment() {
+  if [ -z "${COMPOSE_API_TOKEN:-}" ]; then
+    echo "COMPOSE_API_TOKEN must be exported"
```
46bit: I'm unsure about `COMPOSE_API_TOKEN` despite that Compose describe it as an API Token. It looks like elsewhere in `paas-cf` still names it`COMPOSE_API_KEY`, and it took a lot of work last time we renamed it. But if we're willing to rename it again, this would a start.

scripts/compose-user-provided-service.sh
```diff
@@ -0,0 +1,164 @@
...
+done
+
+load_colors
+check_params_and_environment
+prepare_name
+get_cluster_data
+provision_instance
+create_user_provided_service
+
+success "All done!"
```
46bit: The one thing this script doesn't do is deprovision the instance if `create_user_provided_service` fails. I think we're fine without it if the script is only exploratory.

46bit:
@paroxp @combor I've code reviewed and left a couple of comments, but I don't think either is very important. Let me know what you'd like to do.

scripts/compose-user-provided-service.sh
paroxp: Yeah, we had this discussion at the beginning. My thought is, we shouldn't get deeper into calling things they aren't as it may cause confusion. I know, we've had some bother renaming this before and I know we renamed it from a single wrong thing to another xD

I also think, this script is a stand alone tool aside of `paas-cf`. It's aiming to be a "one-off" use only. If it's not going to be implemented into our pipeline, is it worth keeping our previous naming?

I'm happy to change it if you wish, let me know :)

46bit: Ok, let's leave it. We could/should submit a Pivotal story to rename everything to `COMPOSE_API_TOKEN`. 😄

scripts/compose-user-provided-service.sh
paroxp: You're right. It's nasty that it doesn't deprovision. I'm inclined to change that, but then it may be impossible to do that, as the instance may be still provisioning when we ask to delete it...

We already mention it in the usage, you need to be logged in as admin. Should we implement a `is user cf logged in` check instead?

46bit: I think it's fine to have `cf` error like it did for me, as it's something we're intimately familiar with. I agree automated deprovisioning would be tricky, and it's making for a quite complicated script. 

The simplest solution I can think of is to output an error message if `create_user_provided_service` fails that tells the user to log in and deprovision the instance. 
The notice you have to setup whitelisting is quite visible, so I think something similar would work well?

paroxp: I've addressed this issue, lines 99-101. Please see if that's sufficient.

46bit:
LGTM.

46bit:
I reran the script with a duplicate service name to check it failed nicely, and the reminder to deprovision wasn't output 😞  

```
michaelmokrysz@gds6196:paas-cf ∴ ./scripts/compose-user-provided-service.sh -n 46bits-compose-service -o govuk-paas -s 46bit -t elastic_search
INFO: Targeting organisation 'govuk-paas' and space '46bit'.
api endpoint:   https://api.mmokrysz.dev.cloudpipeline.digital
api version:    2.92.0
user:           admin
org:            govuk-paas
space:          46bit
Creating user provided service 46bits-compose-service in org govuk-paas / space 46bit as admin...
FAILED
Server error, status code: 400, error code: 60002, message: The service instance name is taken: 46bits-compose-service
```

paroxp:
Yes you're right. Good catch. This is due to `set -e` at the top of the file. I took different approach. Thanks.

scripts/compose-user-provided-service.sh
```diff
@@ -0,0 +1,171 @@
...
+  }' "${uri_user}" "${uri_password}" "${uri_host}" "${uri_port}" "${uri}")
+
+  IFS=',';
+  for i in $(printf "%s" "${SPACE}")
+  do
+    info "Targeting organisation '${ORG}' and space '${i}'."
+
+    cf target -o "${ORG}" -s "${i}"
+
+    warning "Please bare in mind, if prior to this message you'll encounter an error, you will need to manually deprovision the instance: '${deployment_name}'"
```
46bit: I'm picking on tiny things, but the latter was a little confusing:
s/bare/bear/
s/prior/post/
After that, time to merge :)

paroxp: English language hard. Thanks! :D

46bit: :) I've been told english is a nightmare!

scripts/compose-user-provided-service.sh
46bit: I filed https://www.pivotaltracker.com/story/show/150728257

alext:
Given we've already started work on the broker support for Redis, I don't think it's worth merging this. We've already spent much more time on this story than anticipated, and at this point it's unlikely we'll need to create any more instances manually.

I'm therefore going to close this PR, and create the instance for trade-tariff manually.

PR #1024, opened 2017-09-01
[#150407935] Add Elasticsearch acceptance test

alext:
## What

This adds an acceptance test for the Elasticsearch service.

As part of this I've taken the opportunity to refactor some of the existing acceptance tests.

## How to review

* Code review - may be easier commit-by-commit.
* Deploy from this branch and verify the tests pass.

## Who can review

Not me.


LeePorte:
Should we split the DB endpoint in the healtcheck app endpoint to be Postgres and MySQL specific endpoints rather than a combined one in order to reflect the updated test pattern?

alext:
> Should we split the DB endpoint in the healtcheck app endpoint to be Postgres and MySQL specific endpoints rather than a combined one in order to reflect the updated test pattern?

The reason I didn't split that one is because there's a lot of commonality between them (they're both SQL services and share the `testSQLConnection` function), so it seemed reasonable to have them combined.

PR #1026, opened 2017-09-05
Order create-cloudfoundry pipeline first in concourse.

alext:
## What

This makes it the default pipeline displayed when you go to the
concourse URL.

The fly command doesn't presently have a command to do this, so this
includes a ruby script to interact with the API using credentials read
from the `.flyrc` file. I've opened https://github.com/concourse/concourse/issues/1561
to request that a command be added to fly.

This is something that's been an irritation since https://www.pivotaltracker.com/story/show/137364635 was played to include
 the `create-bosh-concourse` pipeline. This one was ordered first because it's
the first pipeline pushed to the Concourse.

## How to review

Run the pipeline from this branch and verify that the `self-update-pipelines` step changes the ordering of the pipelines so that the `create-cloudfoundry` pipeline is first.

## Who can review

Not me.

### DIFF
--- a/concourse/scripts/pipelines-cloudfoundry.sh
+++ b/concourse/scripts/pipelines-cloudfoundry.sh
@@ -185,6 +185,8 @@ update_pipeline() {
   case $pipeline_name in
     create-cloudfoundry)
       upload_pipeline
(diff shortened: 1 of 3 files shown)


dcarley:
haha, this was bugging me yesterday too. I thought we might have to use teams but this'll do nicely 👍

PR #1027, opened 2017-09-06
[#150512301] Add Redis service

samcrang:
## What

We have added and enabled support for Redis. 

## How to review

- Deploy to dev.
- Enable the Redis service using `cf enable-service-access`.
- Try to provision a Redis service.

This relies on [changes to `paas-compose-broker`](https://github.com/alphagov/paas-compose-broker/pull/15) on branch `150512301-add-redis-support`, you'll have to drop the commit in this repo that changes `create-cloudfoundry.yml` which does this after reviewing, but before merging.

## Who can review

Anyone who isn't @samcrang or @chrisfarms.


config/service-brokers/compose/catalog.json
```diff
@@ -30,7 +30,38 @@
...
+    "plans": [{
+      "id": "a8574a4b-9c6c-40ea-a0df-e9b7507948c8",
+      "name": "tiny",
+      "description": "256MB Storage / 256MB RAM",
+      "compose": {
+        "units": 1,
+        "databaseType": "redis"
+      },
+      "metadata": {
+        "displayName": "Redis Tiny",
```
dcarley: Should this say anything about it being "storage mode"? If we plan to introduce a "cache mode" plan afterwards and it's not advisable to use this plan for caching. Or are any of the words to describe "non-cache" more confusing than they are helpful? /cc @bleach @alext for thoughts

chrisfarms: heh we had "Redis (storage mode)" in the displayname/description originally but thought it might be confusing without the complementary "Redis (cache mode)" descriptions since this is really a compose thing not a redis thing.

Maybe we could add a note to the story that enables provisioning cache-mode instances to ensure the descriptions are clearly labeled?

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1620,17 +1620,21 @@ jobs:
...
                       'PASSWORD' => '${COMPOSE_BROKER_PASS}',
                       'COMPOSE_API_KEY' => '${COMPOSE_API_KEY}',
                       'DB_PREFIX' => '${DB_PREFIX}',
                       'CLUSTER_NAME' => '${CLUSTER_NAME}',
-                    }}
-                    manifest['applications'].each { |app| app.merge!(env) }
+                    }
+                    manifest = YAML.load_file('manifest.yml')
+                    manifest['applications'].each { |app|
+                      app['env'] ||= app['env']
```
alext: I'm puzzled as to what this line is doing. I suspect you wanted `app['env'] ||= {}`.

chrisfarms: ping @samcrang ... maybe something got lost in translation?

But re this ruby pattern (that is repeated a few times) generally: maybe it would be nicer to have a `make manifest.yml` task in paas-compose-broker that builds the manifest based on env, and remove the yaml/ruby noise?

samcrang: @chrisfarms Yeah @alext is right—it should be `app['env'] ||= {}`.

chrisfarms: Cool ... I'll make a mini PR to fix.

It won't break right now, but if someone removes the `env` section from `manifest.yaml` it will explode.

dcarley: @chrisfarms or @samcrang do you want to raise a new PR and I'll merge?

PR #1030, opened 2017-09-07
[#support] Change parser VM type

combor:
## What

It seems that we have reached a limit what we can achieve with current
VM type for parsers. We are constantly maxing out CPU and it's about a
time to increase processing power for both parsers.

image

## How to review

Sanity check

## Who can review
Not @combor


paroxp:
Looks sane. Is `parser` the best name we could give this configuration?

PR #1034, opened 2017-09-11
[#150847112] Handle errors in MultiMetricReader

dcarley:
## What

Previously the error would be lost and the reader would halt if one of the
readers in a MultiMetricReader returned an error. We think this is the cause
of `ops.users.count` not being reported occasionally.

Handle this by sending the error to the buffer (which will be surfaced by
`ReadMetric()`) and restarting `CopyMetrics()` in a loop. We test this by
sending an error to one of the readers before sending the real metrics,
checking that we get our synthetic error, and then checking that we get all
of the metrics as before.

We had to change the type of `buffers` in order to have access to the
unexported `events` field. This means that we no longer need to do a type
assertion for `WriteMetrics()`. A new slice, of the interface type
`MetricReader`, then had to be made for `NewMultiMetricReader()`. Errors are
collected as a slice to check that we only get one and of the right content.

## How to review

1. Code review
1. Check that the unit tests pass in Travis

We're reasonably confident that this will fix the problem with the `ops.users.count` metric being reported. Especially as we've seen it reported correctly for a short time after the metric app is restarted. However we haven't tested it end-to-end because of the time it takes to setup DataDog. I suggest we merge this if we're happy with the code, wait for it to deploy in prod, and then observe.

## Who can review

Not @dcarley or @LeePorte

### DIFF
--- a/tools/metrics/metrics.go
+++ b/tools/metrics/metrics.go
@@ -116,7 +116,13 @@ func NewMultiMetricReader(rs ...MetricReader) MetricReadCloser {
 		}
 	}()
 	for _, r := range rs {
-		go CopyMetrics(buf, r)
+		go func(r MetricReader) {
+			for {
+				if err := CopyMetrics(buf, r); err != nil {
+					buf.events <- event{err: err}
+				}
+			}
+		}(r)
 	}
 	return buf
 }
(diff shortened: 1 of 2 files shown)


dcarley:
I saw that and then forgot about it 😞 

I've pushed a new commit to fix it.

PR #1035, opened 2017-09-12
[#150512756] IPv4 Keyservers

alext:
## What

Use IPv4-only GPG keyserver pool.

We're experiencing an intermittent problem since we've upgraded
Concourse and garden-runc where gpg will intermittently return an error:
```
gpg: keyserver receive failed: Address family not supported by protocol
```
Switching to an IPv4-only pool works around this, and will unblock the
pipeline while we investigate further.

## How to review

* Push the pipelines using this branch: `BRANCH=ipv4_keyservers make dev pipelines`
* Run the pipeline
* Verify the git fetch in the first job succeeds.

🚨  Remove the `[TMP]` commit before merging.

## Who can review

Not me.


alext:
For future reference, this looks to be caused by https://dev.gnupg.org/T3331.

PR #1037, opened 2017-09-13
[#150512756] Remove Concourse btrfs process Datadog monitor.

alext:
## What

We've upgraded Concourse to 3.4.1, which no longer uses btrfs. This
check is therefore no longer useful.

## How to review

## Who can review

Not me.


dcarley:
(I code reviewed only)

PR #1051, opened 2017-10-02
[#149929104] Add uaa_lock_user.rb script

dcarley:
## What

This will replace the rather cumbersome process for locking (or unlocking)
users that's currently described in the team manual:

- https://government-paas-team-manual.readthedocs.io/en/latest/guides/CF_UAA_user_management/#locking-a-user-account

I've mocked just enough of the UAA responses to make it work and verify that
we're sending PATCH requests when the state needs to change.

Per the original script, I've left TARGET and TOKEN as environments, to call
out the fact that they are mandatory and to help avoid you storing tokens in
your shell history.

## How to review

1. Code review
1. See if the tests make sense and pass
1. Try it in a dev environment:
    1. Create an account using `scripts/create-user.sh`
    1. Verify that you can login
    1. Lock the account and verify that you can no longer login
    1. Unlock the account and verify that you can login again

## Who can review

Anyone.

### DIFF
--- a/scripts/uaa_lock_user.rb
+++ b/scripts/uaa_lock_user.rb
@@ -1,0 +1,74 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'uaa'
+require 'ostruct'
+
+STATE_MAP = {
+  true => "unlocked",
+  false => "locked",
+}.freeze
+
+def parse_args
+  desired_active_state = false
+  settings = OpenStruct.new(
+    target: "",
+    token: "",
+    insecure: false,
+  )
+
+  optparse = OptionParser.new do |opts|
+    opts.banner = <<EOS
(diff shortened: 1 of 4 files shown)


scripts/uaa_lock_user.rb
```diff
@@ -0,0 +1,74 @@
...
+    abort "Username is not unique"
+  end
+  user = users.first
+
+  if user.fetch('active') != active
+    user['active'] = active
+    uaac.patch(:user, user)
+  end
+
+  "#{STATE_MAP.fetch(active)} #{username}"
```
timmow: Not that I think its worth changing now, but i'd be tempted to return 2 values here rather than a string, as it feels more flexible

timmow:
Ready to merge but managed software centre failed to upgrade xcode, and now my dev tools are broken so I cant merge the PR

PR #1054, opened 2017-10-04
[#150926071] Add RDS Instance management

paroxp:
## What

We believe it will decrease the costs for the development environments to
turn off the RDS instances over the night along with the rest of
`autodelete-cloudfoundry` pipeline.

These will start back up in the morning, with the use of
`deployment-kick-off` pipeline.

We've prepared a script which could be also used locally if needed in
order to not duplicate the same functionality before start/stop call.

We need some of the functionality from the newer version of AWS CLI. We
did that in our alphagov/paas-docker-cloudfoundry-tools#100 repo and needed to
include it in our pipelines.

## How to review

- Code review
- Run `autodelete` pipeline
- Check all RDS instances have stopped for your environment
- Run the `kick-off` pipeline
- Check all RDS instances have started again
- Check `create-cloudfoundry` pipeline completes

## Who can review

Neither @dcarley nor myself.

## ~~Before merging~~ 🐝 🐝 🐝 

~~This PR has a [WIP] commit which is tagging a branch tag from docker hub. We need to update that with the proper hash before merging.~~

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -158,7 +158,7 @@ jobs:
             type: docker-image
             source:
               repository: governmentpaas/awscli
-              tag: 465642da06051a55630d39c899697b678f66a7f7
+              tag: 1ab7d6cdac70f2c32563ea173da0cd7791309be5
           run:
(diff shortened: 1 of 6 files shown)


samcrang:
Looks good 👍

PR #1056, opened 2017-10-05
[#150397953] Enable mutual TLS for rep and auctioneer

alext:
## What

In order to migrate away from legacy bridge components, TLS needs to be enabled for rep. This was not done in previous upgrade stories.

This also enables TLS for auctioneer at the same time because it seemed like a good time to do it.

## How to review

Deploy from this branch against an existing deployment. Verify it deploys without downtime, and that all the tests pass.

I've included some refactoring to inline all the job templates, as there is no need to have them separated out since we switched to using Bosh's AZ support.

## Who can review

Not @bandesz or myself.


dcarley:
Waiting for deployment in dev..

> Enable mTLS for communication so auctioneer

Small typo in the commit message, `s/so/to/`?

> I've included some refactoring to inline all the job templates, as there is no need to have them separated out since we switched to using Bosh's AZ support.

Thanks for doing this 👍 👍 👍

alext:
> Small typo in the commit message, s/so/to/?

Fixed and force-pushed.

alext:
I've added an additional commit to this ("Fix cc_uploader server cert") that was previously in #1058. Fixing this in this PR avoids the potential for API downtime while #1058 is deployed.

I've also rebased this on master. This is now ready for review/merge again.

PR #1057, opened 2017-10-06
[#150881142] Update pipeline variables to new format

paroxp:
## What

We've upgraded concourse to the version 3.4.1 not so long ago. Although
the old templating is still supported it will be deprecated in the
future.

The new format supports proper types and variable interpolation.

Pipecleaner had to be modified in order to fit the new syntax.

Some of the functionality has been fixed after the upgrade. Removing it
may benefit us with a clean codebase.

## How to review

- Sanity code check
- See if there aren't variables left out
- See if related `FIXME`s have been resolved

I have tested pushing the pipelines and running the `create-cloudfoundry` pipeline.

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -18,21 +18,21 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-cf.git
       branch: master
-      tag_filter: {{paas_cf_tag_filter}}
-      commit_verification_key_ids: {{gpg_ids}}
+      tag_filter: ((paas_cf_tag_filter))
+      commit_verification_key_ids: ((gpg_ids))
 
   - name: cf-secrets
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
       versioned_file: cf-secrets.yml
 
   - name: cf-manifest
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
       versioned_file: cf-manifest.yml
 
   - name: delete-timer
@@ -45,28 +45,28 @@ resources:
   - name: bosh-secrets
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
       versioned_file: bosh-secrets.yml
 
   - name: bosh-CA
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
       versioned_file: bosh-CA.tar.gz
 
   - name: deployed-healthcheck
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
-      region_name: {{aws_region}}
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
       versioned_file: healthcheck-deployed
 
   - name: datadog-tfstate
     type: s3-iam
     source:
-      bucket: {{state_bucket}}
+      bucket: ((state_bucket))
       versioned_file: datadog.tfstate
       region_name: eu-west-1
 
@@ -92,11 +92,11 @@ jobs:
       - task: datadog-TF-destroy
         file: paas-cf/concourse/tasks/terraform_destroy_datadog.yml
         params:
-          TF_VAR_datadog_api_key: {{datadog_api_key}}
-          TF_VAR_datadog_app_key: {{datadog_app_key}}
-          TF_VAR_env: {{deploy_env}}
-          TF_VAR_aws_account: {{aws_account}}
-          ENABLE_DATADOG: {{enable_datadog}}
+          TF_VAR_datadog_api_key: ((datadog_api_key))
+          TF_VAR_datadog_app_key: ((datadog_app_key))
+          TF_VAR_env: ((deploy_env))
+          TF_VAR_aws_account: ((aws_account))
+          ENABLE_DATADOG: ((enable_datadog))
         ensure:
           put: datadog-tfstate
           params:
@@ -113,7 +113,7 @@ jobs:
           inputs:
             - name: paas-cf
           params:
-            DEPLOY_ENV: {{deploy_env}}
+            DEPLOY_ENV: ((deploy_env))
           run:
             path: ./paas-cf/concourse/scripts/sleep_for_deploy_env.sh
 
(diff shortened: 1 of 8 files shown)


bandesz:
I found the old format in one place: https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-cloudfoundry.yml#L1746. Is there a reason for keeping it?

paroxp:
Good spot. Thanks.

My regex had some problems picking that up for some reason...

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1743,7 +1741,7 @@ jobs:
                   UAA_CLIENT_SECRET=$($VAL_FROM_YAML secrets.uaa_admin_client_secret cf-secrets/cf-secrets.yml)
                   cd paas-cf/scripts
                   bundle
-                  bundle exec sync-admin-users.rb ${API_ENDPOINT} ../config/admin_users.yml "{{NEW_ACCOUNT_EMAIL_ADDRESS}}"
+                  bundle exec sync-admin-users.rb ${API_ENDPOINT} ../config/admin_users.yml "((disable_user_creation))"
```
bandesz: I think we should leave the NEW_ACCOUNT_EMAIL_ADDRESS value here, as the script doesn't handle the "true" value. If the NEW_ACCOUNT_EMAIL_ADDRESS is not set it will throw an error as probably intended.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2171,9 +2169,9 @@ jobs:
                   cd paas-cf/tools/paas_dashboard
                   cf push --no-start paas-dashboard
                   echo Setting DD_API_KEY...
-                  cf set-env paas-dashboard DD_API_KEY {{datadog_api_key}} > /dev/null
+                  cf set-env paas-dashboard DD_API_KEY ((datadog_api_key)) > /dev/null
```
bandesz: we should add quotes here (do we need single quotes in this case?)

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2171,9 +2169,9 @@ jobs:
...
                   echo Setting DD_APP_KEY...
-                  cf set-env paas-dashboard DD_APP_KEY {{datadog_app_key}} > /dev/null
+                  cf set-env paas-dashboard DD_APP_KEY ((datadog_app_key)) > /dev/null
```
bandesz: we should add quotes here (do we need single quotes in this case?)

concourse/scripts/spec/fixtures/pipecleaner_shellcheck.yml
```diff
@@ -12,6 +12,7 @@ jobs:
               - -e
               - -c
               - |
+                echo "((test))"
```
bandesz: what is the reason for adding this line?

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -130,7 +130,6 @@ system_dns_zone_name: ${SYSTEM_DNS_ZONE_NAME}
 apps_dns_zone_name: ${APPS_DNS_ZONE_NAME}
 git_concourse_pool_clone_full_url_ssh: ${git_concourse_pool_clone_full_url_ssh}
 ALERT_EMAIL_ADDRESS: ${ALERT_EMAIL_ADDRESS:-}
-NEW_ACCOUNT_EMAIL_ADDRESS: ${NEW_ACCOUNT_EMAIL_ADDRESS:-}
```
bandesz: If my comment about the script is right we should probably add this back too.

concourse/pipelines/create-cloudfoundry.yml
paroxp: Good spot.

concourse/scripts/spec/fixtures/pipecleaner_shellcheck.yml
paroxp: This file is used by the `concourse/scripts/pipecleaner_test.go`, it is meant to use our pipecleaner python script to run through and see if there are no errors.

Providing the line above, will help us establish if the pipecleaner is running correctly replacing `\(\(.+\)\)` with `DUMY`, and preventing all kinds of shellcheck errors in other tests.

concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -18,21 +18,21 @@ resources:
     source:
       uri: https://github.com/alphagov/paas-cf.git
       branch: master
-      tag_filter: {{paas_cf_tag_filter}}
-      commit_verification_key_ids: {{gpg_ids}}
+      tag_filter: ((paas_cf_tag_filter))
```
bandesz: This now changed from ```""``` to ```null```. I don't know if that would cause any issues.

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -154,19 +154,15 @@ oauth_client_id: ${oauth_client_id:-}
 oauth_client_secret: ${oauth_client_secret:-}
 pagerduty_api_token: ${pagerduty_api_token:-}
 enable_metrics: ${ENABLE_METRICS}
+auto_deploy: ${ENABLE_AUTO_DEPLOY:-false}
+continuous_smoke_tests_trigger: ${ALERT_EMAIL_ADDRESS:-false}
+disable_user_creation: "${NEW_ACCOUNT_EMAIL_ADDRESS:-true}"
```
bandesz: in the code we are explicitly checking for a false value, so this won't work:
```
if [ "${DISABLE_USER_CREATION}" = "false" ] ; then
```

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -154,19 +154,15 @@ oauth_client_id: ${oauth_client_id:-}
 oauth_client_secret: ${oauth_client_secret:-}
 pagerduty_api_token: ${pagerduty_api_token:-}
 enable_metrics: ${ENABLE_METRICS}
+auto_deploy: ${ENABLE_AUTO_DEPLOY:-false}
+continuous_smoke_tests_trigger: ${ALERT_EMAIL_ADDRESS:-false}
```
bandesz: this is used as a trigger variable which probably should be always bool

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -154,19 +154,15 @@ oauth_client_id: ${oauth_client_id:-}
 oauth_client_secret: ${oauth_client_secret:-}
 pagerduty_api_token: ${pagerduty_api_token:-}
 enable_metrics: ${ENABLE_METRICS}
+auto_deploy: ${ENABLE_AUTO_DEPLOY:-false}
```
bandesz: this is used as a trigger variable which probably should be always bool

bandesz:
TF_VAR_datadog_* values also became null instead of ```""```

PR #1063, opened 2017-10-12
[#150881142] Check for "null" string in the terraform_destroy_datadog task.

bandesz:
## What

If the datadog parameters (datadog_api_key, datadog_app_key) are empty they are
passed as "null" string instead of empty strings. It seems this only happens
when we trigger an external task in the pipeline (e.g. in
destroy-cloudfoundry.yml).

## How to review

You should set your datadog keys to empty strings (or remove them completely) and destroy your development environment. The pipeline should succeed.

## Who can review

Not me or @paroxp .

### DIFF
--- a/concourse/tasks/terraform_destroy_datadog.yml
+++ b/concourse/tasks/terraform_destroy_datadog.yml
@@ -17,7 +17,7 @@ run:
     - -c
     - |
       cleanup=false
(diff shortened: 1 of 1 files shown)


bandesz:
I'm closing this PR and moving this change to another one which will contain an other fix.

PR #1065, opened 2017-10-12
[#151808174] Upgrade UAA to v51

bandesz:
## What

In v50 there was a memory leak issue which was fixed in v51. We're
manually bumping the version of UAA to prevent this happening again.

This should be reverted once we upgrade the CF release to >= v276.

This is a critical release as the memory leak can cause another downtime in production soon.

## How to review

* sanity check
* deploy the changes
* all tests should pass

## Who can review

not me or @paroxp

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -37,6 +37,11 @@ releases:
     version: 1.11.7
     url: https://github.com/cloudfoundry-community/nginx-release/releases/download/v1.11.7/nginx-1.11.7.tgz
     sha1: 133bb2260411b197924fff08d4cbd923cc8ec7eb
+    # FIXME: remove once CF is upgraded to >= v276
+  - name: uaa
+    version: "51"
+    url: https://bosh.io/d/github.com/cloudfoundry/uaa-release?v=51
+    sha1: 869b8e6bf58f5431b3579f730f142814aae39d71
 
 stemcells:
   - alias: default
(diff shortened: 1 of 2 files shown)


manifests/cf-manifest/manifest/000-base-cf-deployment.yml
```diff
@@ -37,6 +37,11 @@ releases:
     version: 1.11.7
     url: https://github.com/cloudfoundry-community/nginx-release/releases/download/v1.11.7/nginx-1.11.7.tgz
     sha1: 133bb2260411b197924fff08d4cbd923cc8ec7eb
+    # FIXME: remove once CF is upgraded to > 276
+  - name: uaa
+    version: "52"
+    url: https://bosh.io/d/github.com/cloudfoundry/uaa-release?v=52
+    sha1: e7a1b9f368d30cc5e597be134844280b9887873e
```
alext: This change isn't sufficient. The [job definitions](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/manifest/010-cf-jobs.yml#L123-L136) also need to be updated to point to this release, otherwise they'll continue to use uaa from cf-release.

dcarley:
Do you think we should be concerned that the previous attempt to add metrics introduced this memory leak and that the new release with them enabled has only been out 3 days (ie. not many people will be using it in anger)?

(not a loaded question: I'm unsure myself)

bandesz:
@dcarley it seems they didn't enable the metrics filters in v52 yet, but we eventually agreed with @paroxp that we should only go with v51.

paroxp:
FYI, I have tested the pipeline run with the v51.

dcarley:
Tests passed 👍

PR #1067, opened 2017-10-13
[#151167914] Workaround for bosh-cli timeout

chrisfarms:
# What

I've been using this successfully for ~week now, so it might be of interest to others...

The current bosh-cli make target works by hijacking a concourse
container that has recently run (and succeeded) to use as a temporary
shell to access bosh.

Some recent concourse updates changed the behaviour of how long and when
old containers are garbage collected, and where it used to be the case
that if you were logged in to a hijacked container the container would
be protected from garbage collection (or at least stay around for a very
long time), it is now the case that containers are cleaned up after
around 5mins and we get kicked out of our bosh-cli shell.... which is
frustrating.

To workaround this issue the following changes are made:

* the run-bosh-cli task will now sleep for as long as more than pseudo-terminals are listed in /dev/pts (ie at least one other console is attached)
and only 'succeed' after 5mins of inactivity
slave is detected (ie someone has a terminal wired up to the container)
* the script to launch the job will wait until the container is running,
then connect to it
* the script to launch to shell will abort the job on EXIT to clean up
faster that the /dev/pts based safety net

## How to review

* make dev pipeline
* make dev bosh-cli
* do some boshing for 5mins
* continue doing some boshing for 10mins
* do even more boshing for hours
* exit
* check that the container exits (should be marked "aborted") or if not that it exits after ~5mins of inactivity.

If this seems like an acceptable workaround... then please check that any other tasks that use the bosh-cli script are unaffected. I do not have much context around this.

Also consider the effect of long running jobs/tasks on the prod concourse instance, against I have little context here. 

## Who can review

Not @chrisfarms

### DIFF
--- a/concourse/scripts/bosh-cli.sh
+++ b/concourse/scripts/bosh-cli.sh
@@ -10,12 +10,17 @@ export TARGET_CONCOURSE=deployer
 $("${SCRIPT_DIR}/environment.sh")
 "${SCRIPT_DIR}/fly_sync_and_login.sh"
 
-OUTPUT_FILE=$(mktemp -t bosh-cli.XXXXXX)
-trap 'rm -f "${OUTPUT_FILE}"' EXIT
+BUILD_NUMBER=$($FLY_CMD -t "${FLY_TARGET}" trigger-job -j create-cloudfoundry/bosh-cli | \
+	awk '/started create-cloudfoundry\/bosh-cli/ { print $3 }' | \
+	tr -d '#')
 
-$FLY_CMD -t "${FLY_TARGET}" trigger-job -j create-cloudfoundry/bosh-cli -w | tee "${OUTPUT_FILE}"
+while $FLY_CMD -t "${FLY_TARGET}" builds -j create-cloudfoundry/bosh-cli | awk '{print $3 $4}' | grep "${BUILD_NUMBER}pending" >/dev/null; do
+	echo "waiting for create-cloudfoundry/bosh-cli container (${BUILD_NUMBER}) to start..."
+	sleep 2
+done
+sleep 8 # required since even after status is 'started' it takes a few seconds
 
-BUILD_NUMBER=$(awk '/started create-cloudfoundry\/bosh-cli/ { print $3 }' "${OUTPUT_FILE}" | tr -d '#')
-
+echo "hijacking create-cloudfoundry/bosh-cli container (${BUILD_NUMBER})..."
+trap '$FLY_CMD -t "${FLY_TARGET}" abort-build -j create-cloudfoundry/bosh-cli -b "${BUILD_NUMBER}"' EXIT
 $FLY_CMD -t "${FLY_TARGET}" intercept -j create-cloudfoundry/bosh-cli -b "${BUILD_NUMBER}" \
    -s run-bosh-cli "${@:-ash}"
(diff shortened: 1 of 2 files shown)


concourse/scripts/bosh-cli.sh
```diff
@@ -10,12 +10,17 @@ export TARGET_CONCOURSE=deployer
...
-trap 'rm -f "${OUTPUT_FILE}"' EXIT
+BUILD_NUMBER=$($FLY_CMD -t "${FLY_TARGET}" trigger-job -j create-cloudfoundry/bosh-cli | \
+	awk '/started create-cloudfoundry\/bosh-cli/ { print $3 }' | \
+	tr -d '#')
 
-$FLY_CMD -t "${FLY_TARGET}" trigger-job -j create-cloudfoundry/bosh-cli -w | tee "${OUTPUT_FILE}"
-
-BUILD_NUMBER=$(awk '/started create-cloudfoundry\/bosh-cli/ { print $3 }' "${OUTPUT_FILE}" | tr -d '#')
+while $FLY_CMD -t "${FLY_TARGET}" builds -j create-cloudfoundry/bosh-cli | awk '{print $3 $4}' | grep "${BUILD_NUMBER}pending" >/dev/null; do
+	echo "waiting for bosh-cli container to start..."
```
alext: It would help with debugging to include the the `BUILD_NUMBER` both here and below in the hijacking message.

This task failed the first time for me (I suspect due to office network issues), and debugging which build was involved was a little fiddly.

alext:
I also wonder whether we should include some `FIXME` comments that link to [the concourse issue](https://github.com/concourse/concourse/issues/1269) so that we know when we can revert this.

alext:
This looks to be working well for me, so other than the small comments above, I think this is good to merge.

concourse/scripts/bosh-cli.sh
chrisfarms: 🙌  done

concourse/scripts/bosh-cli.sh
```diff
@@ -10,12 +10,17 @@ export TARGET_CONCOURSE=deployer
...
-
-BUILD_NUMBER=$(awk '/started create-cloudfoundry\/bosh-cli/ { print $3 }' "${OUTPUT_FILE}" | tr -d '#')
+while $FLY_CMD -t "${FLY_TARGET}" builds -j create-cloudfoundry/bosh-cli | awk '{print $3 $4}' | grep "${BUILD_NUMBER}pending" >/dev/null; do
+	echo "waiting for create-cloudfoundry/bosh-cli container (${BUILD_NUMBER}) to start..."
+	sleep 2
+done
+sleep 8 # required since even after status is 'started' it takes a few seconds
 
+echo "hijacking create-cloudfoundry/bosh-cli container (${BUILD_NUMBER})..."
+trap '$FLY_CMD -t "${FLY_TARGET}" abort-build -j create-cloudfoundry/bosh-cli -b "${BUILD_NUMBER}"' EXIT
```
alext: I've discovered that this has a problem on persistent environments. We've configured a 5 minute idle timeout there, so if the bosh-cli session is longer than that, this fails with an authentication error.

chrisfarms: Damn.

This is the auth timeout right? ...not the [concourse step timeout](https://concourse.ci/timeout-step.html) which could be altered?

alext: Correct, this is the auth timeout.

This isn't a massive problem because the while loop in the task itself will cause the task to exit in up to 5 mins. It is a bit of noise though.

PR #1069, opened 2017-10-17
Update readme to include start and end date params

LeePorte:
## What

process_events.go now has start and end flags, these were not
documented when they were added. This commit adds them into the example in the
readme.

## How to review

Code review

## Who can review

Not @LeePorte

### DIFF
--- a/scripts/usage_reporting/README.md
+++ b/scripts/usage_reporting/README.md
@@ -25,6 +25,8 @@ Example:
 ```
 go run event.go process_events.go \
   -org "<ORG_UUID>" \
+  -start "2017-10-01T00:00:00Z" \
+  -finish "2017-11-01T00:00:00Z" \
   < events.json \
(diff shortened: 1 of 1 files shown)


scripts/usage_reporting/README.md
```diff
@@ -25,6 +25,8 @@ Example:
 ```
 go run event.go process_events.go \
   -org "<ORG_UUID>" \
+  -start "2017-10-01T00:00:00Z" \
+  -finish "2017-10-31T23:59:59Z" \
```
dcarley: Should we use 1st of the following month for simplicity? It's easier to read and reduces the chance that we'd make an off-by-one error with the number of days in the current month. I don't think it will have any effect on the results.

LeePorte: I think that would be fine.

PR #1074, opened 2017-10-19
[#139841751] Redirect HTTP for apps

paroxp:
## What

We have upgraded our HAProxy release, to enable redirecting any requests
made to http version of an app to the ssl encrypted protocol https.

We added test for absence of HSTS header when redirecting.

The RFC[1] states:

> An HSTS Host MUST NOT include the STS header field in HTTP responses
> conveyed over non-secure transport.

We therefore want to assert that one isn't set when redirecting.

And at last, we've added HTTP redirect endpoint to router ELB.

This causes all requests for the non-https version of an app to be
redirected to the corresponding https root endpoint. See [ADR444](https://github.com/alphagov/paas-team-manual#120) for details on the rationalle.

## How to review

- Code review
- Deploy `create-cloudfoundry` from this branch
- Expect all acceptance tests to pass
- Visit `healthcheck.${DEPLOY_ENV}.dev.cloudpipelineapps.digital` and witness it being redirected to `https`
- Run `curl -I healthcheck.${DEPLOY_ENV}.dev.cloudpipelineapps.digital` and expect 301

## 🚨  BEFORE MERGING :bee:

Please mind the HAProxy version is currently fixed to the dev build. This will need to be updated once the alphagov/paas-haproxy-release#9 has been merged.

## Who can review

Neither @alext nor myself.


timmow:
👏

bandesz:
I did a search for ```cc.allow_app_ssh_access``` which is still present and set to true. We should probably remove it: https://github.com/alphagov/paas-cf/blob/ee5e987409f5e097ab7776259b1528443edaf0a2/manifests/cf-manifest/manifest/020-cf-properties.yml#L221

bandesz:
Ignore my comment about the cc.allow_app_ssh_access parameter.

platform-tests/src/acceptance/plain_http_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("plain HTTP requests", func() {
```
bandesz: This should be ```Describe("Http client",...```

alext: Why? I think it's clearer like this because it describes the scope of these tests.

platform-tests/src/acceptance/plain_http_test.go
```diff
@@ -0,0 +1,52 @@
...
+	const (
+		CONNECTION_TIMEOUT = 11 * time.Second
+	)
+
+	Describe("to the API", func() {
```
bandesz: This reads better:
```
Context("When it tries to connect to the API"), func() {
  It("should not be able to connect"), func() {
```

platform-tests/src/acceptance/plain_http_test.go
```diff
@@ -0,0 +1,52 @@
...
+		It("has the connection refused", func() {
+			uri := config.ApiEndpoint + ":80"
+			_, err := net.DialTimeout("tcp", uri, CONNECTION_TIMEOUT)
+			Expect(err).To(HaveOccurred(), "should not connect")
+			Expect(err.(net.Error).Timeout()).To(BeTrue(), "should timeout")
+		})
+	})
+
+	Describe("to an app", func() {
```
bandesz: This also:
```
Context("When it tries to connect to the API"), func() {
  It("should be redirected to the https endpoint"), func() {
```

platform-tests/src/acceptance/plain_http_test.go
bandesz: Evantually I changed my mind about the test headers, also I have to re-learn to use the review tool in Github :)

alext:
I've updated this to include the merged haproxy-release versions.

PR #1076, opened 2017-10-23
When starting/stopping the RDS databases we should check against the full database list

bandesz:
## What

The script will only run successfully if all the databases are in the previous
state (e.g. all stopped when we want to start them). If we compare the current
state to the full RDS instance list it will be able to run successfully in any
situation.

This caused an issue in two scenarios:
 - if you're databases are already running the kick-off pipeline will fail
 - if only some of your databases were able to start up the kick-off pipeline will fail

## How to review

It's easier to run the script manually locally: ```./concourse/scripts/rds_instances.sh [start|stop]```

Please test the following scenarios:
 - if you're databases are already running the start command should finish quickly
 - if you stop one database the start command should start it up and finish
 - if you stop a database then the stop command should stop the remaining databases
 - if all databases are stopped the stop command should finish quickly

## Who can review

not me

### DIFF
--- a/concourse/scripts/rds_instances.sh
+++ b/concourse/scripts/rds_instances.sh
@@ -24,6 +24,11 @@ case $ACTION in
   ;;
 esac
 
+get_instance_ids_with_status() {
+  aws rds describe-db-instances \
+    --query "DBInstances[?DBSubnetGroup.VpcId == \`${vpc_ids}\` && DBInstanceStatus == \`${1}\` && MultiAZ == \`false\` && ReadReplicaDBInstanceIdentifiers == \`[]\`].DBInstanceIdentifier" --output text | tr '\t' '\n' | sort
+}
+
 # Obtain the VPC ID from the AWS API and make sure there is only one that we can
 # use.
 vpc_ids=$(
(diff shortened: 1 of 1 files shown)


concourse/scripts/rds_instances.sh
```diff
@@ -42,6 +42,12 @@ fi
 
 # Make a list of RDS Instances that are assigned to the specific VPC ID, are not
 # MultiAZ and are/do not contain any replicas.
+all_instance_ids=$(
```
chrisfarms: since `instance_ids` and `instance_ids_after` are generated from the same query/command, maybe it would make sense to make this into a function so that's clear it's the same query for people editing it later.

dcarley:
> Please test the script manually on your dev.

To be sure I'm testing the right thing:

- stop one RDS instance and trigger the `autodelete-cloudfoundry` pipeline?
- start one RDS instance and trigger the `deployment-kick-off` pipeline?

bandesz:
@dcarley I added extra context and testing scenarios to the PR.

concourse/scripts/rds_instances.sh
bandesz: I added it

dcarley:
> if you're databases are already running the start command should finish quickly
> if you stop one database the start command should start it up and finish
> if you stop a database then the stop command should stop the remaining databases
> if all databases are stopped the stop command should finish quickly

Thanks! All of these scenarios tested true.

Also thanks for implementing @chrisfarms's suggestion.

PR #1077, opened 2017-10-23
[#147674123] Add instance detailing script

paroxp:
## What

As part of #147674123 we would like to have an easy way of establishing
the belongings of an individual instance.

Given we obtain an instance ID from either DataDog, Kibana or AWS
Console, we can run this script to learn which org and space the
instance belong to along with it's name and list of org managers.

## How to review

- Shell code check
- Test the script to obtain some details about random instance (follow alphagov/paas-team-manual#155)

### DIFF
--- a/scripts/get-instance-details.sh
+++ b/scripts/get-instance-details.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+SCRIPT=$0
+RDS_INSTANCE_ID=$1
+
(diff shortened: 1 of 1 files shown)


dcarley:
Works well. Thanks for writing usage information 👍 

As you mentioned in Pivotal, it's quite heavy on API queries, but that's fine at the moment. We can refine it in the future if it becomes a problem.

PR #1078, opened 2017-10-24
[#151014551] Tag app and service `paas-metrics` with their org's quota

46bit:
## What

We have added new tags to app and services metrics reported by the `paas-metrics` application. This allows the product dashboard to be broken down by those tags. The caveat of this work is that minor business logic (the names of trial org quotas and free service plans) is now contained within that app. See the commit message for more details.

## How to review

* I've already shown the Datadog results to Tom and he seems happy
* Thoughts on how we're detecting trial orgs and free services?
* Code review?

## Who can review

Not @46bit @LeePorte.

### DIFF
--- a/tools/metrics/gauges.go
+++ b/tools/metrics/gauges.go
@@ -126,7 +126,7 @@ func AppCountGauge(c *Client, interval time.Duration) MetricReadCloser {
 		if err != nil {
 			return err
 		}
-		org_quotas, err := c.cf.ListOrgQuotas()
+		orgQuotas, err := c.cf.ListOrgQuotas()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


tools/metrics/gauges.go
```diff
@@ -116,32 +117,41 @@ func AppCountGauge(c *Client, interval time.Duration) MetricReadCloser {
 		if err != nil {
 			return err
 		}
-		started := 0
-		stopped := 0
+
+		counters := map[string]map[bool]int{
```
bandesz: It would be nice to have a comment about what this structure means. e.g. counters[STATE][ORG_IS_TRIAL] = number off apps where state is STATE and ORG_IS_TRIAL is true if the app is running in an org which is in trial mode.

tools/metrics/gauges.go
```diff
@@ -155,27 +165,56 @@ func ServiceCountGauge(c *Client, interval time.Duration) MetricReadCloser {
 		if err != nil {
 			return nil
 		}
-		counters := map[string]int{}
+		service_plans, err := c.cf.ListServicePlans()
+		if err != nil {
+			return nil
+		}
+
+		counters := map[bool]map[bool]map[string]int{
```
bandesz: It would be nice to have a comment about what this structure means. e.g. counters[ORG_IS_TRIAL][PLAN_IS_FREE][TYPE] = number of services where ORG_IS_TRIAL is true if the service is running in an org which is in trial mode, PLAN_IS_FREE is true if the service plan is free and TYPE is the service type.

tools/metrics/gauges.go
```diff
@@ -274,3 +312,38 @@ func findService(services []cfclient.Service, guid string) *cfclient.Service {
...
+func findServicePlan(service_plans []cfclient.ServicePlan, guid string) *cfclient.ServicePlan {
+	for _, service_plan := range service_plans {
+		if service_plan.Guid == guid {
+			return &service_plan
+		}
+	}
+	return nil
+}
+
+func findOrgQuotaFromSpaceGUID(c *Client, guid string) (*cfclient.OrgQuota, error) {
```
bandesz: I think it would be a good idea to cache the results of this method to decrease the number of CF api requests.

46bit: I agree, but it might make cleaner code to iterate over spaces than to cache this. What do you think?

bandesz: I'm happy with both, but I think your solution is better as caching would increase the complexity.

tools/metrics/gauges.go
```diff
@@ -170,6 +174,11 @@ func ServiceCountGauge(c *Client, interval time.Duration) MetricReadCloser {
 			return nil
 		}
 
+		// Number of relevant service instances in
+		// - ORG_IS_TRIAL: boolean of whether each app is owned by a trial organisation
```
bandesz: this should say service instead of app

46bit:
I also need to fix my variable names to be `camelCase` https://golang.org/doc/effective_go.html#mixed-caps

tools/metrics/gauges.go
46bit: Unsure how to make this pleasant code in Go, but this was a very good call.

dcarley:
> I've been trying out fixups (to be squashed when merging) instead of editing the original commit, but I think the commits need their own messages.

Using `--fix <SHA> -e` to open an editor and add a short description after the `!fixup` subject might help.

PR #1081, opened 2017-10-27
[#151322324] Move UAA behind gorouter

paroxp:
## What

Whilst investigating UAA downtime during deploys we decided that we should move
UAA to run behind gorouter. This should mean we reduce (and possibly eliminate)
downtime during deploys because connections will be drained by removing the
route whilst UAA is shutting down.

In order to move UAA behind the gorouter, we will need a different way
of registrating our routes. We decided to fallback onto upstream to get
closer to the setup the community is running. We've implemented the
route-registrar configuration, strictly specific to UAA at this point.

Haproxy automatically adds a reason text to its http response statuses. E.g.
"201 Created" instead of plain 201. In our create_admin.yml task we should only
check if the response starts with 201, as the UAA was moved behind gorouter +
haproxy and the tests started to panic.

There is some extra functionality that has been implemented in the
gorouter such as `tls_port` and `server_cert_domain_san`, which we would
be delite to use upon our UAA move behind Gorouter.

This is to ensure the traffic between the two is encrypted.

The release also consists of the custom change to the route_registrar,
simply duplicating the two properties listed above from the gorouter in
order to allow smooth registration.

This should be safe to be reverted when:

- We upgrade cf-release / cf-deployment to the version that contains new
gorouter
- The changes we need have been applied upstream for the route-registrar

## How to review

- Code check
- Deploy your pipeline from this branch
- Make a note of the API failures
- Remove the [TMP] commit
- Re-Deploy your pipeline with the stemcell being set back
- Make a note of the API failures
- Expect all acceptance tests to succeed

## Who can review

Neither @bandesz, @dcarley nor myself

### DIFF
--- a/terraform/cloudfoundry/elbs.tf
+++ b/terraform/cloudfoundry/elbs.tf
@@ -59,7 +59,7 @@ resource "aws_elb" "cf_uaa" {
   }
 
   health_check {
-    target              = "HTTPS:9443/healthz"
+    target              = "HTTP:82/"
     interval            = "${var.health_check_interval}"
     timeout             = "${var.health_check_timeout}"
     healthy_threshold   = "${var.health_check_healthy}"
@@ -67,10 +67,10 @@ resource "aws_elb" "cf_uaa" {
   }
 
   listener {
-    instance_port      = 9443
-    instance_protocol  = "https"
+    instance_port      = 443
+    instance_protocol  = "ssl"
     lb_port            = 443
-    lb_protocol        = "https"
+    lb_protocol        = "ssl"
     ssl_certificate_id = "${var.system_domain_cert_arn}"
   }
 }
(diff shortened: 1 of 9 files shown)


terraform/cloudfoundry/elbs.tf
```diff
@@ -59,18 +59,18 @@ resource "aws_elb" "cf_uaa" {
...
 
   listener {
-    instance_port      = 9443
-    instance_protocol  = "https"
+    instance_port      = 443
+    instance_protocol  = "ssl"
     lb_port            = 443
-    lb_protocol        = "https"
+    lb_protocol        = "ssl"
     ssl_certificate_id = "${var.system_domain_cert_arn}"
```
alext: This is going to cause significant API downtime as the ELB will be reconfigured in the cf-terraform job, several minutes before the UAA nodes get redeployed. to register themselves in gorouter, and the router nodes get added to the ELB.

I think we need to rethink this to do a multi-pass deployment to avoid downtime.

PR #1083, opened 2017-10-30
[#152067205] Improve shellcheck linting for pipelines

alext:
## What

We were previously skipping several checks because it was deemed to much
work to fix them all when shellchecking was introduced to the pipelines.

This re-enables these checks and fixes resulting warnings/errors.

## How to review

Code review is probably enough (may be easiest commit-by-commit)

You could run the pipelines from this branch to test things. I've been running my dev env using this branch for a couple of days now.

## Who can review

Not me.

### DIFF
--- a/concourse/tasks/create_admin.yml
+++ b/concourse/tasks/create_admin.yml
@@ -58,6 +58,7 @@ run:
         # N.B., It stopped printing `201 Created` during the upgrade to 
         # CF v269, and now just prints the `201` code on its own line
         # followed by a space.
+        # shellcheck disable=SC2181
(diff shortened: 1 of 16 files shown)


concourse/tasks/create_admin.yml
```diff
@@ -58,6 +58,7 @@ run:
         # N.B., It stopped printing `201 Created` during the upgrade to 
         # CF v269, and now just prints the `201` code on its own line
         # followed by a space.
+        # shellcheck disable=SC2181
```
camelpunch: Agree that this is a good solution for now, but perhaps in future we could extract the `uaac curl` command so we can name the concept and also appease shellcheck.

PR #1085, opened 2017-11-01
[#145695325] Upgrade buildpacks

alext:
## What

This does 2 things:

* separates the buildpacks from cf-release so that it's easier to upgrade them on a separate cadence.
* upgrades the buildpacks to the latest versions.

## How to review

* Deploy from this branch. Verify that all the tests pass.
* Review the draft notification to tenants (linked from [Pivotal story](https://www.pivotaltracker.com/story/show/145695325)).

## Who can review

Not me.


dcarley:
Do we need to update [`scripts/generate_buildpack_release_notes.sh`](https://github.com/alphagov/paas-cf/blob/master/scripts/generate_buildpack_release_notes.sh) which currently assume that they come from cf-release? Possible yaks ahead.

alext:
> Do we need to update scripts/generate_buildpack_release_notes.sh which currently assume that they come from cf-release? Possible yaks ahead.

That's a good question. I'm not sure it even works in it's current form. Maybe it should be removed. At the moment we're generating sending out a summary of the version changes with links to the relevant release notes in github. This isn't particularly laborious to generate by hand.

paroxp:
I'm in favour of the idea of removing it. I'm not certain it is used nor mentioned anywhere, most of the team doesn't even know of its existence. It seems like an obligation we don't need to maintain this script.

PR #1086, opened 2017-11-01
[#151322324] Move UAA behind Gorouter 1/2

paroxp:
## What

Our previous approach #1081 causes a 30 minute downtime caused by the
modification of terraform resource and not updating the configuration in time.

We're eliminating this problem by splitting the previous approach into two separate
PRs, where at first we register the routes and prepare the use of newly configured 
ELB, causing the healthchecks to fail and not sending any traffic to it, and in the
second PR, we actually reconfigure the ELB which otherwise would cause the 
downtime, until the `cf-deploy` job ran its tasks.

---

Whilst investigating UAA downtime during deploys we decided that we should move
UAA to run behind gorouter. This should mean we reduce (and possibly eliminate)
downtime during deploys because connections will be drained by removing the
route whilst UAA is shutting down.

In order to move UAA behind the gorouter, we will need a different way
of registrating our routes. We decided to fallback onto upstream to get
closer to the setup the community is running. We've implemented the
route-registrar configuration, strictly specific to UAA at this point.

Haproxy automatically adds a reason text to its http response statuses. E.g.
"201 Created" instead of plain 201. In our create_admin.yml task we should only
check if the response starts with 201, as the UAA was moved behind gorouter +
haproxy and the tests started to panic.

There is some extra functionality that has been implemented in the
gorouter such as `tls_port` and `server_cert_domain_san`, which we would
be delite to use upon our UAA move behind Gorouter.

This is to ensure the traffic between the two is encrypted.

The release also consists of the custom change to the route_registrar,
simply duplicating the two properties listed above from the gorouter in
order to allow smooth registration.

This should be safe to be reverted when:

- We upgrade cf-release / cf-deployment to the version that contains new
gorouter
- The changes we need have been applied upstream for the route-registrar

## How to review

- Code check
- Deploy your pipeline from this branch
- Expect all acceptance tests to succeed

To be closely followed by #1087 

## Who can review

Neither @bandesz, @dcarley, @chrisfarms nor myself

### DIFF
--- a/concourse/tasks/create_admin.yml
+++ b/concourse/tasks/create_admin.yml
@@ -55,11 +55,8 @@ run:
             -d "{\"origin\":\"uaa\",\"type\":\"USER\",\"value\":\"${2}\"}"
         )
         # Check that `uaac` output a 201 HTTP status code.
-        # N.B., It stopped printing `201 Created` during the upgrade to 
-        # CF v269, and now just prints the `201` code on its own line
-        # followed by a space.
         # shellcheck disable=SC2181
-        if [ $? = 0 ] && echo "${result}" | grep '^201 $'; then
+        if [ $? = 0 ] && echo "${result}" | grep '^201'; then
           echo "${result}"
           echo "Added $2 to $1"
         else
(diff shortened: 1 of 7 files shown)


alext:
I'm going to review #1087 before I merge this one.

PR #1088, opened 2017-11-02
[#145695325] Update ruby version use in Dashboard

LeePorte:
## What

A recent update of the ruby buildpack has removed the minor version that
was in use by the dashboard. The update now ignores the minor ruby
version.

The dependencies have also been updated as part of this work.

## How to review

On top of a current master in dev do the following, datadog creds are in the password store. I recommend using the prod creds for testing with as it it read only

1. Checkout this branch and cd into `tools/paas_dashboard`
1. `cf push --no-start paas-dashboard`
1. `cf set-env paas-dashboard DD_API_KEY <datadog_api_key>`
1. `cf set-env paas-dashboard DD_APP_KEY <datadog_app_key>`
1. `cf start paas-dashboard`
1. Visit https://pass-dashboard.environmentname.dev.cloudpipelineapps.digital/paas-overview and see the working dashboard.

## Who can review

Not @LeePorte


alext:
Instead of fixing this, we could replace it with https://app.datadoghq.com/screen/238678/triggered-monitors?tv_mode=true

LeePorte:
I'm fine either way :)

dcarley:
I think we should review/merge this because the pipelines are broken and it will take more time to raise and review a PR that removes the dashboard.

I am 👍  on the idea of replacing it afterwards though. I see it supports muting, but wonder how it reports critical and non-critical alerts? Can you write a small task story?

alext:
> I see it supports muting, but wonder how it reports critical and non-critical alerts?

It breaks them out by category. There's an option to hide empty categories whichI've currently enabled. There are categories for critical, warning and no-data.

> Can you write a small task story?

Will do.

PR #1091, opened 2017-11-03
[#152150279] Create API Latency monitor

paroxp:
## What

Following on the UAA downtime incident from October 9, we would like to
monitor our API latency and get alerted if anomaly is detected.

This will create a monitor for each of four components:
- uaa
- cc
- router
- doppler

This will not trigger PagerDuty incident but will trigger our dashboard
to report a single alert.

## How to review

- Code review
- Check if the threshold numbers make sense
- **Optionally:** Deploy from this branch, and manually change from `name:${DEPLOY_ENV}-cf-uaa` to `name:staging-cf-yaa` and perhaps change the thresholds to `c:1`, `w:0` to see the alerts

## Who can review

Neither @camelpunch nor myself

### DIFF
--- a/terraform/datadog/elb.tf
+++ b/terraform/datadog/elb.tf
@@ -1,0 +1,67 @@
+resource "datadog_monitor" "abnormal_api_latency_cc" {
+  name    = "${format("%s Abnormal API Latency - CC", var.env)}"
+  type    = "query alert"
+  message = "${format("{{#is_alert}}We're experiencing >= {{threshold}} change in ELB Latency.{{/is_alert}} \n{{#is_warning}}We're experiencing >= {{warn_threshold}} change in ELB Latency.{{/is_warning}} \n\nVisit the [Team Manual > Responding to alerts > API Latency](%s#api-latency) for more info.", var.datadog_documentation_url)}"
+
+  query = "${format("avg(last_15m):anomalies(avg:aws.elb.latency{name:%s-cf-cc}, 'basic', 2, direction='both') > 0.3", var.env)}"
+
+  thresholds {
+    warning  = 0.15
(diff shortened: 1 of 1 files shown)


terraform/datadog/elb.tf
```diff
@@ -0,0 +1,67 @@
+resource "datadog_monitor" "abnormal_api_latency_cc" {
+  name    = "${format("%s Abnormal API Latency - CC", var.env)}"
+  type    = "query alert"
+  message = "${format("{{#is_alert}}We're experiencing >= {{threshold}} change in ELB Latency.{{/is_alert}} \n{{#is_warning}}We're experiencing >= {{warn_threshold}} change in ELB Latency.{{/is_warning}} \n\nVisit the [Team Manual > Responding to alerts > API Latency](%s#api-latency) for more info.", var.datadog_documentation_url)}"
+
+  query = "${format("avg(last_15m):anomalies(avg:aws.elb.latency{name:%s-cf-cc}, 'basic', 2, direction='both') > 0.3", var.env)}"
+
+  thresholds {
+    warning  = 0.15
```
samcrang: I'm assuming we've not extracted these to variables because we think they're accidentally the same across all monitors? ie we expect that the cc monitor may have different values to the uaa monitor?

samcrang: Also is the plan to deploy it, get some feedback around typical behaviour and then adjust the values?

camelpunch: Yes, we anticipate that each monitor might need different config, and thought it'd be premature to extract these into variables at this time.

samcrang:
This looks good, waiting on staging and prod to go green before merging.

PR #1094, opened 2017-11-08
[#152670116] Bump UAA version to 52.1

paroxp:
## What

In order to protect ourselves from [CVE-2017-8031](https://www.cloudfoundry.org/cve-2017-8031/) exposure, we need to
upgrade the UAA release to at least version 52.1.

## How to review

- Make sure that the version used will cover us against this CVE

~~I haven't tested this yet. It's still deploying in my pipelines.~~


samcrang:
Looks good. From the CVE:

> Releases that have fixed this issue include:
>        UAA: 30.6, 45.4, 52.1

PR #1095, opened 2017-11-08
[#152150279] Tune the Latency monitor

paroxp:
## What

We realised that the CC ELB will create a lot of noise in our production environment. We decided to amend the thresholds as well as the window to for the metrics to be monitored.

I've updated the window on other monitors too and took the decision to change the latency check from "both" to "above" as we probably don't care if it took less time than usual to serve the response.

We're also removing the router monitor as it's not strictly part of the API we mean to monitor and we may get some false alerts if tenant application would take time to serve response.

## How to review

- Code review
- Deploy from this branch with DataDog enabled 
  ```
  BRANCH=$(git rev-parse --abbrev-ref HEAD) ENABLE_DATADOG=true SELF_UPDATE_PIPELINE=false make dev pipelines
  ```
- Go into the DataDog and manually change the monitors (CC in particular), to point to production. For instnace: `name:prod-cf-cc`.
- Have a good long stare at the graph and be sure it wouldn't produce a lot of faulty noise

### DIFF
--- a/terraform/datadog/elb.tf
+++ b/terraform/datadog/elb.tf
@@ -3,11 +3,11 @@ resource "datadog_monitor" "abnormal_api_latency_cc" {
   type    = "query alert"
   message = "${format("{{#is_alert}}We're experiencing >= {{threshold}} change in ELB Latency.{{/is_alert}} \n{{#is_warning}}We're experiencing >= {{warn_threshold}} change in ELB Latency.{{/is_warning}} \n\nVisit the [Team Manual > Responding to alerts > API Latency](%s#api-latency) for more info.", var.datadog_documentation_url)}"
 
-  query = "${format("avg(last_15m):anomalies(avg:aws.elb.latency{name:%s-cf-cc}, 'basic', 2, direction='both') > 0.3", var.env)}"
+  query = "${format("max(last_1h):avg:aws.elb.latency{name:%s-cf-cc} > 2", var.env)}"
 
   thresholds {
-    warning  = 0.15
-    critical = 0.3
+    warning  = 1
+    critical = 2
   }
 
   require_full_window = true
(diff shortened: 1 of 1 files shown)


terraform/datadog/elb.tf
```diff
@@ -3,11 +3,11 @@ resource "datadog_monitor" "abnormal_api_latency_cc" {
   type    = "query alert"
   message = "${format("{{#is_alert}}We're experiencing >= {{threshold}} change in ELB Latency.{{/is_alert}} \n{{#is_warning}}We're experiencing >= {{warn_threshold}} change in ELB Latency.{{/is_warning}} \n\nVisit the [Team Manual > Responding to alerts > API Latency](%s#api-latency) for more info.", var.datadog_documentation_url)}"
 
-  query = "${format("avg(last_15m):anomalies(avg:aws.elb.latency{name:%s-cf-cc}, 'basic', 2, direction='both') > 0.3", var.env)}"
+  query = "${format("avg(last_1d):avg:aws.elb.latency{name:%s-cf-cc} > 1", var.env)}"
```
samcrang: I appreciate setting this aggregation over 1 day smooths out spikes but do we feel this is going to trigger quick enough? For example, if our regular latency is 0.1 seconds it'd take an hour of requests that averaged ~21.7 seconds before this triggered.

samcrang: (at the critical level)

paroxp: I'm _not sure_ if that's right. DataDog works in mysterious ways...

If I attempt to set something below 1h window, I get the warning:

> The selected metric provides data at low frequencies, we strongly recommend alerting on a 1 hour timeframe

It's a difficult one to get right really... Following the guide from DD and taking our concerns into the factor, with the current thresholds the following setup:

> Trigger when the metric is `above` the threshold `at least once` during the last `1 hour`

would trigger 27 warnings and 3 alerts today...  This is too noisy in my opinion and I don't think it's reasonable to increase the thresholds above 1 second.

Your point is still valid, therefore I'd prefer to delay/block merging this PR and have a general tech talk with the team. Maybe someone could clarify things for us or we come up with a different solution.

PR #1099, opened 2017-11-09
[152510403] CF upgrade to v278 

LeePorte:
## What

We have upgraded CF to v278 and upgraded cflinuxfs2 to 1.66.0. There are no buildpack changes in this upgrade.

## How to review

Deploy from this branch on top of current master in dev with `ENABLE_DATADOG=true`
Check that the number of API test failures correlates with the acceptance criteria 

## Who can review
Not @LeePorte or @46bit

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -18,9 +18,9 @@ releases:
     url: https://bosh.io/d/github.com/cloudfoundry/garden-runc-release?v=1.9.6
     sha1: 55fdcc0f2bda526f126d28827fbae7b21c6b0d1c
   - name: cflinuxfs2
-    version: 1.161.0
-    url: https://bosh.io/d/github.com/cloudfoundry/cflinuxfs2-release?v=1.161.0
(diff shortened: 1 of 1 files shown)


manifests/cf-manifest/manifest/000-base-cf-deployment.yml
```diff
@@ -87,7 +87,7 @@ releases:
 stemcells:
   - alias: default
     name: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
-    version: "3445.16"
+    version: "3468"
```
alext: Is there a reason to use this version instead of `3468.1` or `3468.5`?

Given the timelines of [the stemcell releases](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/releases) and the recent crop of USNs that have been fixed, I worry that this version won't include the fixes for the USNs.

In general, when doing CF upgrades, we use the latest version from the recommended stemcell series.

LeePorte: I'm not certain if they hadn't been published on bosh.io at the time (no timestamps on the releases) or were missed on the upgrade story. PR to update is located at https://github.com/alphagov/paas-cf/pull/1101

PR #1100, opened 2017-11-10
[#150711834] Access Bosh with SSH directly

henrytk:
## What

We will be accessing Bosh via SSH directly, so we no longer need to go via Concourse.

## How to review

To be merged after:

* https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/101
* https://github.com/alphagov/paas-bootstrap/pull/100

~~We need to ensure `make ssh_bosh` still works, which is why the above pull requests need to be merged first.~~ The Bosh SSH scripts are now in paas-bootstrap.

Check the new IP whitelist is correct.

## Who can review

Anyone but me or @camelpunch


alext:
We'll also need to remove https://github.com/alphagov/paas-cf/blob/2fb06eb4995e3b0246fe17a81d4ac0f88e21a30d/concourse/pipelines/create-cloudfoundry.yml#L2501-L2522 which tests that the `bosh-cli` script works.

henrytk:
Good spot. I've removed the test-bosh-cli task and squashed the change into the commit that removed the bosh-cli script.

PR #1104, opened 2017-11-15
[#152543946] Eliminate most of the errors during UAA deploys

bandesz:
## What

Sadly we still see some 503 errors. Out of 6 deploys only 2 had 1-1 errors.

We made the following changes:
 * Nginx was removed from UAA as it was causing some of the downtime (the route-registrar health check was running against UAA and not Nginx) - this removed the 502 errors we saw
 * We enabled the connection draining on the UAA ELB (with default draining timeout which is 300s), which removed the remaining EOF errors but introduced new 503 errors (from the haproxy in front of the gorouter).
 * We set the drain_wait to 120s on the gorouter which decreased the number of 503 errors
 * The uaa.self_service_links_enabled parameter now disables the self-service links (like create account and reset password links), not just removes the links therefore there is no need for nginx anymore.

Minor changes:
 * The uaa.self_service_links_enabled parameter was changed to bool true (was a string value)
 * We had to change the tests checking the create account and reset password links as now UAA handles this and returns with a different status code
 * Reorder+rename forgot password tests
 * Changed the UAA ELB drain timeout from the default 300s to 20s (although it doesn't matter much, but it might help when we remove an instance from an ELB)

Remaining errors:
 * We suspect that haproxy should also drain connections at some point (it doesn't have draining atm). As we timeboxed this story we don't go into it any further. Also there is good chance haproxy will be removed soon anyway which might solve this for good.

Note: we intentionally made separate commits to reflect how the story evolved.

## How to review

1. Deploy from this branch. It is expected that the `api-availability-tests` may fail at this stage.
2. Deploy from a temporary branch that includes the following stemcell change for `router` and `uaa`:
    ```patch
    diff --git a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
    index 6583319c..dc0dfc39 100644
    --- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
    +++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
    @@ -84,6 +84,9 @@ stemcells:
       - alias: default
         version: "3468.5"
    +  - alias: test
    +    name: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
    +    version: "3468"

     update:
       canaries: 0
    diff --git a/manifests/cf-manifest/manifest/010-cf-jobs.yml b/manifests/cf-manifest/manifest/010-cf-jobs.yml
    index 0df27b98..ce3fa21a 100644
    --- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
    +++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
    @@ -141,7 +141,7 @@ jobs:
         vm_type: medium
         vm_extensions:
           - cf_rds_client_sg
    -    stemcell: default
    +    stemcell: test
         networks:
           - name: cf
         properties:
    @@ -329,7 +329,7 @@ jobs:
             release: datadog-for-cloudfoundry
         instances: 2
         vm_type: router
    -    stemcell: default
    +    stemcell: test
         networks:
           - name: router
         properties:
    ```

1. Confirm that the `api-availability-tests` passed or only had some 503 errors (we saw less than 5 errors)

Also check the following:
 * the client IP is still recorded correctly in UAA log

## Who can review

Not @dcarley or @bandesz.

### DIFF
--- a/manifests/cf-manifest/manifest/010-cf-jobs.yml
+++ b/manifests/cf-manifest/manifest/010-cf-jobs.yml
@@ -133,8 +133,6 @@ jobs:
         release: (( grab meta.release.name ))
       - name: statsd_injector
         release: (( grab meta.release.name ))
-      - name: nginx
-        release: nginx
       - name: route_registrar
       # FIXME: remove once https://github.com/cloudfoundry/route-registrar/pull/18
       # is merged - It is used ONLY for the route-registrar and gorouter
@@ -147,43 +145,6 @@ jobs:
     networks:
       - name: cf
     properties:
-      nginx_conf: |
-        worker_processes  2;
-        error_log syslog:server=127.0.0.1,severity=error,tag=uaa_nginx_error error;
-        events {
-          worker_connections  1024;
-        }
-        http {
-          include /var/vcap/packages/nginx/conf/mime.types;
-          access_log syslog:server=127.0.0.1,severity=info,tag=uaa_nginx_access combined;
-          sendfile on;
-          keepalive_timeout 20s;
-          upstream uaa {
-            keepalive 10;
-            server 127.0.0.1:8080;
-          }
-          set_real_ip_from  10.0.0.0/16;
-          set_real_ip_from  127.0.0.1/32;
-          real_ip_header    X-Forwarded-For;
-          real_ip_recursive on;
-          server {
-            listen 9443;
-            ssl on;
-            ssl_certificate_key /var/vcap/jobs/nginx/etc/ssl.key.pem;
-            ssl_certificate  /var/vcap/jobs/nginx/etc/ssl_chained.crt.pem;
-
-            location / {
-              rewrite ^/create_account(.do)?$ /create_account_disabled break;
-              rewrite ^/forgot_password(.do)?$ /create_account_disabled break;
-
-              proxy_pass http://uaa;
-              proxy_set_header Host $host;
-              proxy_set_header X-Forwarded-For $remote_addr;
-              proxy_http_version 1.1;
-              proxy_set_header Connection "";
-            }
-          }
-        }
       ssl_key: (( grab secrets.uaa_internal_key ))
       ssl_chained_cert: (( grab secrets.uaa_internal_cert ))
       consul:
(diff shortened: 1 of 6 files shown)


manifests/cf-manifest/manifest/010-cf-jobs.yml
```diff
@@ -147,43 +145,6 @@ jobs:
...
-          server {
-            listen 9443;
-            ssl on;
-            ssl_certificate_key /var/vcap/jobs/nginx/etc/ssl.key.pem;
-            ssl_certificate  /var/vcap/jobs/nginx/etc/ssl_chained.crt.pem;
-
-            location / {
-              rewrite ^/create_account(.do)?$ /create_account_disabled break;
-              rewrite ^/forgot_password(.do)?$ /create_account_disabled break;
-
```
henrytk: How are we now blocking these endpoints?

alext:
The commit message for "Use bool for uaa.login.self_service_links_enabled" links to a non-existent commit in uaa. I suspect you meant to link to uaa-release instead.

alext:
I'm testing this in a dev env now, but after a code review I wonder whether we should make the config for the uaa and cf_router ELBs consistent. Given they both point at the routers it seems odd for them to have different settings for connection draining etc.

dcarley:
> I'm testing this in a dev env now, but after a code review I wonder whether we should make the config for the uaa and cf_router ELBs consistent. Given they both point at the routers it seems odd for them to have different settings for connection draining etc.

I'm not sure. I'm going to create one more follow-up story after the thread on Slack yesterday to improve the app-availability-tests to highlight similar problems with slower or keepalive requests, so perhaps we should tweak it at the same time as that?

dcarley:
> The commit message for "Use bool for uaa.login.self_service_links_enabled" links to a non-existent commit in uaa. I suspect you meant to link to uaa-release instead.

I've rebased it with a correction 👍

alext:
> I'm not sure. I'm going to create one more follow-up story after the thread on Slack yesterday to improve the app-availability-tests to highlight similar problems with slower or keepalive requests, so perhaps we should tweak it at the same time as that?

That makes sense. It's only going to add to the scope of this story otherwise...

alext:
I've successfully run this against dev. Deploying with the temporary stemcell changes resulted in no API downtime reported.

PR #1105, opened 2017-11-16
[#152635692] Deploy new assets to be used by UAA login/accept webpages

46bit:
## What

This updates text and the logo on UAA's webpages (e.g., login.) It also deploys a new app `paas-uaa-assets` in a new space `assets` in the `admin` org. This new app is used in #1106 to serve a custom stylesheet for the UAA webpages. This app must be deployed before merging #1106 in order that the UAA pages never import a stylesheet from an app that does not exist.

Deploying this will update UAA, as the first commit changes its text and logo properties. Thanks to #1104 we expect the downtime to be momentary if at all.

## How to review

Review the manifest changes and whether the new `assets` space is the right decision.

## Who can review

Not @46bit @stephenjoe1 @bleach


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2331,6 +2332,38 @@ jobs:
                   cf blue-green-deploy paas-metrics
                   cf delete -f paas-metrics-old
 
+        - task: deploy-paas-uaa-assets
```
dcarley: In a previous retro we decided that we should try to remove some of the tasks from `cf-deploy` because it had grown too large, but we couldn't refactor all of them in one go so we'd do it incrementally. Do you think this is a good candidate to start with, or not because UAA is dependent on it?

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2331,6 +2332,38 @@ jobs:
...
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
+                tag: 465642da06051a55630d39c899697b678f66a7f7
+            params:
+              DEPLOY_ENV: ((deploy_env))
```
dcarley: Out of curiosity, do you know why we need `DEPLOY_ENV`?

46bit: I think I left this in by mistake!

concourse/pipelines/create-cloudfoundry.yml
46bit: Would that be in the direction of having one task that deploys all the apps? It would make sense to start here.

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -394,11 +394,13 @@ properties:
...
...myLiTa1a31TEKqW+5KNED0fE//ixm4KIU/z1G0VEF/j3UUjcgCrExphdEfH/+ev1gr9+LxHRY4h4pbX2KGPMxhXr/byfr3Q/b7tBe+e86/pn8TCt9a98O0dBYIyQLFEU7Zrub7P1AMA61trDvEB9yr+fX/PziVv8d6FS+rgGbGiMOZqZr/VtLbrPEXEUIv7Fz3VD/MNbQwRw/yxtEsAANWEUw5KRX7NVFNQemj6P/twkDmrXYzWo+cesDLUP8JKVO60nA6FmGrYCAHwcav1eCWr9bpgs/SPIclAbg/8r+2EbMFAzZ1zZt/kxqN07nTIrF5ZMCGrvb4IKfuYeBbVnN4beu2/Qt8e+zSXh3dYX4yA0xsHi+yRk4bSV+0j7drqgdn8uCXOh9D3ZH+Yz3ZsUMfT+czQAer5X+so15qOCgd57H6bb6v37XARw/yxtFMCCIAiCIAiCIAhLByKA+2cRASwIgiAIgiAIglAREcD9s4gAFgRBEARBEARBqIgI4P5ZRAALgiAIgiAIgiBURARw/ywigAVBEARBEARBECoiArh/FhHAgiAIgiAIgiAIFVlz61XgRzfN6nNBJ6W6AH7klV37+vYRBEEQBEEQBEHoP6z6uRVg8PX3wu6XPyWlH5XvXvL00IcmhuZ9FwRBEARBEARBEAAArr322uikf/9bSelfBZaMfO2CIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCIAiCICxl/H9l333eOJxP0AAAAABJRU5ErkJggg==
```
dcarley: I didn't notice until reviewing #1106 but it belongs to this commit..

The image, particularly the "beta" text, looks fuzzy to me. I don't *think* it's my eyes or screen. Am I imagining it? Is there anything that can be done about it?

screen shot 2017-11-17 at 11 21 08

46bit: That *does* look fuzzy, and suggests that the underlying PNG image is too small. I'll look into it and maybe ask Stephen for a larger version. Thanks Dan 💯

concourse/pipelines/create-cloudfoundry.yml
46bit: One task to deploy all the apps has no separation and it'd be scary to change later. We could deploy `rubbernecker`, `paas-dashboard` and *maybe* `paas-metrics` using the CI environment, but I'm not sure what the tradeoffs are.

I feel this extra task starts to make that work necessary, but this task does belong in core `cf-deploy` because it's directly tied to UAA. The refactoring might be best as a maintenance story?

alext: I agree that rubbernecker should move to CI, but the other app you mention form part of the platform itself, so IMO should be deployed by the pipeline.

It's worth noting that when https://www.pivotaltracker.com/story/show/152539815 has been played, `paas-dashboard` won't exist any more,

dcarley: I think it's fine to leave as-is for now because this app isn't a good example of something that's independent from `cf-deploy` - UAA won't function correctly without the app - whereas the other tasks like service brokers and monitoring don't have such hard/circular dependencies. Thought it was worth a discussion though.

manifests/cf-manifest/manifest/020-cf-properties.yml
46bit: It looks okay to me with the default UAA styling, even when blown up. I'll deploy the second PR and see if it looks distorted there.

screenshot 2017-11-17 13 52 53

46bit: @dcarley Oddly, on a similar scale it looks okay for me.

screenshot 2017-11-17 15 48 09

I don't know what to do about this. That it appeared in a screenshot shows it isn't your eyes or your screen. It can't be a simple caching issue as the image is an inline data URI in the HTML file, and you haven't worked on this before. The image is sharp and 960px wide. Thoughts?

46bit: Wait. You're on the invite page…

dcarley: I get the same on the login page too:

screen shot 2017-11-17 at 15 55 57

I checked if it was any different in Safari, but I think it's uncovered a new issue whereby it's not styled at all. Any idea why?

screen shot 2017-11-17 at 15 59 20

46bit: Safari is probably blocking the app's domain because its certificate doesn't validate. If you [visit the `paas-uaa-assets` app](https://paas-assets-app.dcarley.dev.cloudpipelineapps.digital/) and accept its certificate as being invalid, then hopefully refreshing the login page will show the assets.

alext: If this can use external assets, does this need to be an inline data-uri, or could it also point at an external asset?

46bit: As part of making it look good on non-retina screens, we've switched to serving the image from the assets app.

46bit:
@dcarley Both PRs are ready for re-review. The app is now deployed with 2 instances and 64MB of memory--good catch!

dcarley:
Thanks! Sorry it took so much back-and-forth

PR #1106, opened 2017-11-16
[#152635692] Configure UAA to use our custom assets on login/accept webpages

46bit:
🐝🐝🐝 Merge **after** #1105 🐝🐝🐝

## What

Uses an app that #1105 deployed to serve a new custom stylesheet on the UAA webpages. This uses an `asset_base_url` that is marked as deprecated, but we have chosen to use it for now. Tests are added to guard against its removal or our custom CSS becoming out-of-date. See commit messages for details.

Deploying this will update UAA. Thanks to #1104 we expect the downtime to be momentary if at all.

## How to review

* Review #1105 and this PR.
* Merge #1105.
* Merge this PR.

## Who can review

Not @46bit @stephenjoe1 @bleach


platform-tests/src/platform/acceptance/uaa_styling_test.go
```diff
@@ -0,0 +1,71 @@
...
+		response, err := httpClient.Get(uaaLoginURL.String())
+		Expect(err).NotTo(HaveOccurred())
+		uaaLoginDoc, err := goquery.NewDocumentFromResponse(response)
+		Expect(err).NotTo(HaveOccurred())
+
+		stylesheetHrefs := uaaLoginDoc.Find("link[rel=stylesheet]").Map(func(_ int, linkTag *goquery.Selection) string {
+			href, _ := linkTag.Attr("href")
+			return href
+		})
+		Expect(stylesheetHrefs).To(HaveLen(2))
```
dcarley: Very minor, but you might be interested to know that you can simplify these 3 assertions with the [`ConsistOf` matcher](https://godoc.org/github.com/onsi/gomega#ConsistOf) if the order doesn't matter:
```
Expect(stylesheetHrefs).To(ConsistOf([]string{
    "/vendor/font-awesome/css/font-awesome.min.css",
    customStylesheetURL.String(),
}))
```
If the order does matter then `Equal` will do.

platform-tests/src/platform/acceptance/uaa_styling_test.go
```diff
@@ -0,0 +1,71 @@
...
+	var uaaLoginURL url.URL
+	var uaaStylesheetURL url.URL
+	var customStylesheetURL url.URL
+
+	BeforeEach(func() {
+		infoCommand := cf.Cf("curl", "/v2/info")
+		Expect(infoCommand.Wait(testConfig.DefaultTimeoutDuration())).To(Exit(0))
+
+		var infoResp struct {
+			TokenEndpoint string `json:"token_endpoint"`
```
dcarley: It might be a bit of a moot point because they both serve the same content, but I think this should use `authorization_endpoint` (`login.system_domain`) rather than `token_endpoint` (`uaa.system_domain`) to be consistent with some other places:

- The URL returned by `cf login --sso`
- The URL sent to users in invites: https://github.com/alphagov/paas-cf/blob/c17ba3bcd3270158a581524456dba22e4b4f92ee/scripts/create-user.sh#L152

platform-tests/src/platform/acceptance/uaa_styling_test.go
```diff
@@ -0,0 +1,71 @@
...
+			href, _ := linkTag.Attr("href")
+			return href
+		})
+		Expect(stylesheetHrefs).To(HaveLen(2))
+		Expect(stylesheetHrefs[0]).To(Equal("/vendor/font-awesome/css/font-awesome.min.css"))
+		Expect(stylesheetHrefs[1]).To(Equal(customStylesheetURL.String()))
+	})
+
+	It("should have the expected default stylesheet", func() {
+		response, err := httpClient.Get(uaaStylesheetURL.String())
```
dcarley: I would have expected us to test `customStylesheetURL` rather than `uaaStylesheetURL` if that's the one we're now using. Could you describe why?

platform-tests/src/platform/acceptance/uaa_styling_test.go
```diff
@@ -0,0 +1,71 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		returnedBody, err := ioutil.ReadAll(response.Body)
```
dcarley: Should we check the status code too?

46bit:
I've just noticed that this PR doesn't match the `uaa-restyle-pr2-152635692` branch I had locally. It's functionally the same but missing some tidyup I'd done.

dcarley:
> I've just noticed that this PR doesn't match the uaa-restyle-pr2-152635692 branch I had locally. It's functionally the same but missing some tidyup I'd done.

No problem. Leave a comment when you'd like me to re-review.

platform-tests/src/platform/acceptance/uaa_styling_test.go
46bit: This test checks that the upstream stylesheet hasn't changed. If it has then our custom stylesheet may need updating too. I've pushed some changes I left on my work machine on Thursday that should make the purpose of the tests clearer - let me know what you think?

platform-tests/src/platform/acceptance/uaa_styling_test.go
46bit: Awesome. Thanks Dan!

platform-tests/src/platform/acceptance/uaa_styling_test.go
```diff
@@ -0,0 +1,96 @@
...
+	Context("the default stylesheet", func() {
+		It("should not have changed from our saved version", func() {
+			response, err := httpClient.Get(authStylesheetURL.String())
+			Expect(err).NotTo(HaveOccurred())
+			returnedBody, err := ioutil.ReadAll(response.Body)
+			Expect(err).NotTo(HaveOccurred())
+
+			expectedBody, err := ioutil.ReadFile("data/expected_uaa_default_stylesheet.css")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(returnedBody)).To(Equal(string(expectedBody)))
```
dcarley: We talked about adding some test on the assertion to describe what somebody should do if this test fails.

46bit:
Dan had a good idea and we're now testing the checksum of the stylesheet, rather than keeping a copy of it saved. I deployed my pipeline with an incorrect checksum to [see what would happen if the stylesheet changed](https://deployer.miki.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/25):

screen shot 2017-11-21 at 11 21 48

@dcarley This is ready for review once more. 👍

PR #1107, opened 2017-11-16
[#152638275] Restyle the 'Welcome to PaaS' email

bandesz:
## What

This change adds a HTML version to our 'Welcome to PaaS' email which we send
when we create a new user or reset the password for an existing user.

I copied the HTML template from Notify: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/jinja_templates/email_template.jinja2

I extracted the existing plain text template to a Ruby ERB template and saved
the new HTML template next to it. The HTML code would be hard to handle in the
script and this way the templates are easier to manage.

Also we decided to separate the create user and the reset password emails therefore the multiple templates.

I added an email template file so we only have to define the content for the different emails.

## How to review

If your not using your gov.uk email for testing you may have to verify it in SES: https://eu-west-1.console.aws.amazon.com/ses/home?region=eu-west-1#verified-senders-email

1. Run the create user script to create a new user, ORG_NAME can be an existing one or a new:

```
./scripts/create-user.sh -o [ORG_NAME] -e [YOUR_EMAIL]
```

This email should say to create your account and include the organisation name.

2. Run the create user script to reset the password for your user:

```
./scripts/create-user.sh -r -e [YOUR_EMAIL]
```
This email should say to reset your password and not include the organisation name.

3. You should get two emails. Please check if the emails are correct and all links work as expected. Also check the email source (in Gmail: Show original) and verify if the plain text version is correct.

## Who can review

Not @bandesz

### DIFF
--- a/scripts/create-user.sh
+++ b/scripts/create-user.sh
@@ -11,39 +11,8 @@ source "${SCRIPT_DIR}/common.sh"
 DEFAULT_SPACE=sandbox
 FROM_ADDRESS='gov-uk-paas-support@digital.cabinet-office.gov.uk'
 # shellcheck disable=SC2016
-SUBJECT='Welcome to GOV.UK PaaS'
+SUBJECT='Create your GOV.UK PaaS account'
 # shellcheck disable=SC2016,SC1078
-MESSAGE='Hello,
-
-Your account for GOV.UK PaaS is ready:
-
- - username: ${EMAIL}${ORG:+
- - organisation: ${ORG}}
-
-Please use this link to activate your account and set a password. The link will only work once:
-${INVITE_URL}
-
-You can find advice about choosing a password:
-https://docs.cloud.service.gov.uk/#choosing-passwords
-
-To get started, look at our Quick Setup Guide:
-https://docs.cloud.service.gov.uk/#quick-setup-guide
-
-You can find our privacy policy here:
-https://docs.cloud.service.gov.uk/#privacy-policy
-
-To check the status of GOV.UK PaaS, and see the availability of live
-applications and database connectivity, visit
-https://status.cloud.service.gov.uk.  We recommend you sign up to this service
-to get alerts and incident updates.
-
-Regards,
-The GOV.UK PaaS team
-
-PS Some departmental email systems will check links in inbound emails as part of
-their virus protection. This may have invalidated your one-time link. If this is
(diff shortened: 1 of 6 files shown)


scripts/templates/create_user_plain.erb
```diff
@@ -0,0 +1,14 @@
+Hi,
+
+This is your account on GOV.UK PaaS:
+
+Username: <%= email %><% if defined?(org) && org != "" %>
+Organisation: <%= org %><% end %>
```
alext: The original version only output this line if the org was set (it's not set when resetting someone's password for example). We should preserve this behaviour.

scripts/create-user.sh
```diff
@@ -227,7 +196,12 @@ get_subject() {
 
 # Expand variables from message body, escaping new lines and quotes
 get_body() {
-  eval "echo \"${MESSAGE}\"" | \
+  erb email="${EMAIL}" org="${ORG:-}" invite_url="${INVITE_URL}" "${SCRIPT_DIR}/templates/create_user_plain.erb" | \
```
dcarley: Could you add something to the `check_params_and_environment` function to ensure that `erb` is available early on? Otherwise the account could be created/reset but the email would later fail to render and send.

alext: We should add `erb` to the `Gemfile` now that we're using it, and probably run it with `bundle exec` here.

scripts/templates/create_user_plain.erb
alext: Sorry, I misread this. I think now that this is no longer in bash, we can put the conditional on a newline. We should be able to remove the extra whitespace with using `<%-` tag.

scripts/create-user.sh
bandesz: erb is part of the standard library so we don't need to install it

bandesz: Yes, good idea. Added.

scripts/templates/create_user_plain.erb
bandesz: As the templates were separated this if is not needed anymore.

scripts/create-user.sh
alext: Thanks for clarifying that... I hadn't realised this. TIL.

PR #1110, opened 2017-11-21
[#152150403] Suppress update-ca-certificates skipped warning in the logs.

bandesz:
## What

This change will suppress the following warning message:
WARNING: ca-certificates.crt does not contain exactly one certificate or CRL: skipping

It's not actually an indication of a problem but it could distract a team member working on an incident.

## How to review

Run your CF pipeline and check if we're not logging the warnings anymore (e.g. pre-deploy/wait-for-app-availability-tests)

## Who can review

Not @bandesz

### DIFF
--- a/concourse/scripts/import_bosh_ca.sh
+++ b/concourse/scripts/import_bosh_ca.sh
@@ -3,4 +3,8 @@ set -eu
 
 echo "Adding bosh-CA to root certificates"
 tar -xzf bosh-CA/bosh-CA.tar.gz -C /usr/local/share/ca-certificates bosh-CA.crt
-update-ca-certificates
+
+UPDATE_OUTPUT=$(update-ca-certificates 2>&1)
+set +e
+printf "%s" "${UPDATE_OUTPUT}" | grep -v "^WARNING: ca-certificates\.crt does not contain exactly one certificate or CRL: skipping$"
+set -e


concourse/scripts/import_bosh_ca.sh
```diff
@@ -3,4 +3,8 @@ set -eu
 
 echo "Adding bosh-CA to root certificates"
 tar -xzf bosh-CA/bosh-CA.tar.gz -C /usr/local/share/ca-certificates bosh-CA.crt
-update-ca-certificates
+
+UPDATE_OUTPUT=$(update-ca-certificates 2>&1)
+set +e
+printf "%s" "${UPDATE_OUTPUT}" | grep -i -v "skipping$"
```
samcrang: Is it worth specifying the string to match to be something closed to the error message? I don't like the idea that we'd also drop lines that happened to contained "skipping" in them. Also, I don't think we need to be case-insensitive either.

Something like this?:

`grep -v "^WARNING: ca-certificates\.crt does not contain exactly one certificate or CRL: skipping$"`

concourse/scripts/import_bosh_ca.sh
```diff
@@ -3,4 +3,8 @@ set -eu
 
 echo "Adding bosh-CA to root certificates"
 tar -xzf bosh-CA/bosh-CA.tar.gz -C /usr/local/share/ca-certificates bosh-CA.crt
-update-ca-certificates
+
+UPDATE_OUTPUT=$(update-ca-certificates 2>&1)
+set +e
```
samcrang: I'm assuming we're doing this because `update-ca-certificates` can print nothing to `std{out,err}` during normal operation?

bandesz: We're doing this because we can't run it as one command with grep as we need to ignore the grep exit code

bandesz: (Also update-ca-certificates loggs to stderr)

samcrang: I'm saying that `grep` returns `0` if at least one line is printed. If `update-ca-certificates` always prints something normally (which I think it does) then we can just one-line the whole thing?

```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.5 LTS"
$ update-ca-certificates | grep .
Updating certificates in /etc/ssl/certs... 0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d....done.
root@4b342991654d:/# echo $?
0
root@4b342991654d:/# update-ca-certificates | grep z
root@4b342991654d:/# echo $?
1
```

bandesz: Currently it prints one warning, which could go away in the future. In that case we don't want the script to fail.

concourse/scripts/import_bosh_ca.sh
bandesz: Updated

concourse/scripts/import_bosh_ca.sh
bandesz: Ok, I see what you mean, interestingly it doesn't say these things on bosh when the warning is written.

bandesz: So we run this command on Alpine and the command doesn't output anything else there. As we should be prepared that this warning might go away in the future we should keep the commands separate.

samcrang:
Looks good 👍

PR #1112, opened 2017-11-22
[#152144657] Solve some of the credentials rotation issues

bandesz:
## What

We need to make sure we don't have downtime and the create-cloudfoundry pipeline runs successfully when we rotate our credentials.

We identified two major issues:
 - the uaa_admin_password is used before, during and after the cf-deploy job which means e.g. the api-availability-test will have a lot of failures during deployment.
 - the uaa postgres user password change will cause downtime in UAA

This PR contains the following changes:
 * Add concourse task for rotating UAA admin credentials: this task will generate a new uaa_admin_password secret and update it in UAA and the cf-secrets.yml.
 * Do not delete uaa_admin_password when clearing the credentials (now the rotate task will handle it)
 * Trigger clear-cloudfoundry-credentials after rotate-cloudfoundry-credentials: this makes the process more clear
 * Don't rotate CF database user passwords as this causes downtime during deployment.

Minor changes:
 * Order cf-secret keys alphabetically for easier diff

## How to review

1. Update your pipelines from this branch
   ```BRANCH=rotate_credentials_201711_152144657 make dev pipelines```
2. Run the rotate-cloudfoundry-credentials job - it should trigger the clear-cloudfoundry-credentials job
3. Run the create-cloudfoundry pipeline.
4. Check the api-availability pipeline for errors. In our tests we only saw one error like this:
   ```"Failed to fetch app stats: Error requesting app stats: cfclient: error (200002): CF-StatsUnavailable": 1,```
5. All other tests should succeed

## Who can review

Not @bandesz


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2777,37 +2778,99 @@ jobs:
...
+                cf-secrets/cf-secrets.yml \
+                paas-cf/concourse/resources/cf-secrets-keep.yml | \
+                spruce merge - paas-cf/concourse/resources/finalize.yml \
+                > modified-cf-secrets/cf-secrets.yml
+      on_success:
+        put: cf-secrets
+        params:
+          file: modified-cf-secrets/cf-secrets.yml
+
+  - name: rotate-cloudfoundry-credentials
```
alext: Can we order this job before `clear-cloudfoundry-credentials` so that the order in this file matches the order they run in the pipeline. It took me quite a while to figure out that they run in this order with the pipeline as it is now.

bandesz: Changed.

paroxp:
It failed for me too. What does it mean though? With our current setup, do we need to restart the pipeline to get a success? Can you think of anything to avoid that?

bandesz:
@paroxp what do you mean by it failed for you too? Who else? What was the error?

paroxp:
Sorry. I was referring to your initial comment:

> 4. Check the api-availability pipeline for errors. In our tests we only saw one error like this:
`"Failed to fetch app stats: Error requesting app stats: cfclient: error (200002): CF-StatsUnavailable": 1,`

I get exactly the same error message.

bleach:
I received this error message also:

```

  expected no errors
  Expected
      <map[*monitor.Task]map[string]int | len:1>: {
          {
              name: "Fetching detailed app information",
              fn: 0x79d270,
          }: {
              "Failed to fetch app stats: Error requesting app stats: cfclient: error (200002): CF-StatsUnavailable": 2,
          },
      }
  to be empty
```

I don't think it's related to this PR because I got it when deploying `master` before I switched to this branch.

bleach:
Looks like there are conflicts now. Possibly caused by 48f95a9a

bandesz:
@bleach @paroxp Conflicts resolved, also an extra commit to review: https://github.com/alphagov/paas-cf/pull/1112/commits/70f87e0434298e88c44fdeb7ea6149f8a2677325
We forgot to add the ElastiCache broker secrets to the list of keys we don't want to rotate.

bleach:
I've run this a couple of times in my dev env and there were no problems. I've reviewed the code and am happy with it.

I'm not deploying now as I don't have time to babysit by pausing pipelines. I'll merge / deploy on Monday.

PR #1115, opened 2017-11-24
[#152916557] Add custom redirection on user invite

paroxp:
## What

When the user receives an email to complete their registration it will
ask them to provide their password. Once the password is set
successfully, user has accepted an invitation and will be redirected to
the UAA home page.

It isn't very useful page, especially if you're just starting off with
CF. We've created our own landing page we'd like to be providing to our
users upon registration.

## How to review

- Code review
- To be reviewed with alphagov/paas-product-page#41
- Deploy from this branch
- Create a new user in your `dev` env `./scripts/create-user.sh -e user@example.com -o govuk-paas --no-email`
- Click/Copy paste the URL into your browser
- Set user name
- Expect to be redirected into correct page (you may see 404 if the page is not deployed)

## Before merge :bee:

The alphagov/paas-product-page#41 needs to be merged first in order not to redirect people to `404`.

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -550,6 +550,14 @@ properties:
         id: cdn_broker
         scope: cloud_controller.admin_read_only
         secret: (( grab secrets.uaa_clients_cdn_broker_secret ))
+      user_invitation:
+        authorities: oauth.login,scim.write,emails.write,scim.userids
+        authorized-grant-types: password,refresh_token
+        id: user_invitation
+        override: true
+        redirect-uri: "https://www.cloud.service.gov.uk/next-steps?success"
+        scope: openid,password.write,scim.read,scim.write,scim.invite,uaa.user
+        secret: (( grab secrets.uaa_clients_login_secret ))
 
     zones:
       internal:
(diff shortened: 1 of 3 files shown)


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -19,6 +19,7 @@ properties:
     host_key_fingerprint: (( grab secrets.ssh_proxy_host_key.public_fingerprint ))
 
   support_address: "https://docs.cloud.service.gov.uk"
+  start_using_paas_url: "https://www.cloud.service.gov.uk/next-steps?success"
```
alext: I'm not particularly keen on adding this property that's not actually used by any releases. Is there a reason to define it here instead of inline below?

alext:
This could do with some review instructions (the instructions on alphagov/paas-product-page#41 have no details on how to test the changes in this repo...)

manifests/cf-manifest/manifest/020-cf-properties.yml
paroxp: No real reason. I thought We'd like to have an ability to change it quickly without digging into the manifest. I can change it if you like.

paroxp:
You're right. Sorry, I assumed the story context may be enough. Updated.

manifests/cf-manifest/manifest/020-cf-properties.yml
paroxp: I thought it may be OK, as the `support_address` looks similar.

alext:
:+1: Looks good. Will merge once https://github.com/alphagov/paas-product-page/pull/41 has reached prod.

PR #1116, opened 2017-11-28
[#152241494] Enable Billing API Auth

chrisfarms:
## What

* Update the UAA client spec to allow the billing api app to request authorization for a selection of read-only scopes.
* Set paas-usage-events-collector app to pull in the branch at https://github.com/alphagov/paas-usage-events-collector/pull/5

## How to review

Deploy this branch to your dev environment using:

```
ENABLE_USAGE_EVENTS_COLLECTION=true SELF_UPDATE_PIPELINE=false BRANCH=calculate_billing_152241494 make dev pipelines
```

## Before merge

Update the [TMP] commit to point to a tag not a branch.

## Who can review

not @chrisfarms

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -533,12 +533,13 @@ properties:
         redirect-uri: (( concat "https://login." properties.system_domain ))
       paas-usage-events-collector:
         access-token-validity: 1209600
-        authorized-grant-types: client_credentials,refresh_token
+        authorized-grant-types: client_credentials,refresh_token,authorization_code
+        autoapprove: true
         override: true
         secret: (( grab secrets.uaa_clients_paas_usage_events_collector_secret ))
-        scope: openid,oauth.approvals,cloud_controller.admin_read_only
-        authorities: oauth.login,cloud_controller.admin_read_only
-        redirect-uri: (( concat "https://login." properties.system_domain ))
+        scope: openid,oauth.approvals,cloud_controller.admin_read_only,cloud_controller.read,cloud_controller.global_auditor,cloud_controller.admin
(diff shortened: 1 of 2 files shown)


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -533,12 +533,13 @@ properties:
...
         access-token-validity: 1209600
-        authorized-grant-types: client_credentials,refresh_token
+        authorized-grant-types: client_credentials,refresh_token,authorization_code
+        autoapprove: true
         override: true
         secret: (( grab secrets.uaa_clients_paas_usage_events_collector_secret ))
-        scope: openid,oauth.approvals,cloud_controller.admin_read_only
-        authorities: oauth.login,cloud_controller.admin_read_only
-        redirect-uri: (( concat "https://login." properties.system_domain ))
+        scope: openid,oauth.approvals,cloud_controller.admin_read_only,cloud_controller.read,cloud_controller.global_auditor,cloud_controller.admin
```
bandesz: Why do we need the cloud_controller.admin scope here?

chrisfarms: We need to determin if someone has access to edit the `pricing_plans`.

Currently anyone with `admin` `admin_readonly` or `global_auditor` roles can edit pricing.

In the follow up story I have made a suggestion that we could create a new scope in UAA called `billing.admin` for this purpose and assign it to only those who should have access (ie Tom).

bandesz: Ok, because the admin access feels here too much.

chrisfarms: Agreed... In fact I'll make the "invent a billiing.admin role" into it's own story so it doesn't get lost in the list of "nice to haves"

PR #1117, opened 2017-11-28
Prompt before creating orgs when creating users

henrytk:
## What

We have had erroneous orgs generated by typos in the org name passed
to this script. Significant clean up is required once a tenant starts
developing in the wrong org.

## How to review

Code review. Check it still creates users properly and prompts before creating new orgs.

## Who can review

Anyone but me.


samcrang:
This looks good and I'm about to invite some new users so I'll test it out too 👍

PR #1123, opened 2017-11-29
[#152670292] Enable ElastiCache Redis

bandesz:
## What

We introduce the new broker to offer ElastiCache Redis services to our tenants. The new service will be called *redis*.

The existing redis service (Compose Redis) will be renamed to *compose-redis*.

The broker will be installed to two small VMs which will be load balanced with an ELB. The brokers will be able to authenticate to the AWS API with an instance profile.
All the Redis replication groups will be in the same Cache subnet group and will have the same security group.
The cells will have a new security group to be able to connect to the Redis instances.

For now we only introduce one plan which is called tiny and will be essentially the trial plan.

## How to review

Steps:
 1. Review the following PRs:
    * https://github.com/alphagov/paas-aws-account-wide-terraform/pull/106
 1. Review and merge these PRs to make the whole review easier (only small changes):
    * https://github.com/alphagov/paas-release-ci/pull/45
    * https://github.com/alphagov/paas-elasticache-broker/pull/3
 1. Review https://github.com/alphagov/paas-elasticache-broker-boshrelease/pull/1
 1. Change the submodule ref in https://github.com/alphagov/paas-elasticache-broker-boshrelease/pull/1 to the master ref.
 1. Wait for the CI concourse to build the updated BOSH release PR and get the release configuration
 1. Update this PR with the new elasticache-broker release definition (see last TMP commit)
 1. Deploy CF to your dev env
 1. The deploy should be successful and all the tests should pass
 1. Login to CF on your dev: you should not see the new service in the marketplace

## Who can review

Not @samcrang, @LeePorte or @bandesz

## Before merge

❗️ The tenants have to be notified about the Compose redis service name change (only the ones who have access to the Compose services in their marketplace)


platform-tests/example-apps/healthcheck/redis.go
```diff
@@ -30,6 +31,10 @@ func testRedisConnection() error {
 		return errors.Wrap(err, "failed to parse VCAP_SERVICES")
 	}
 
+	if !strings.HasPrefix(credentials.URI, "rediss://") {
```
paroxp: Is that correct? 🤔

platform-tests/src/platform/acceptance/compose_redis_service_test.go
```diff
@@ -0,0 +1,84 @@
...
+		Expect(plans).To(Say(serviceName))
+	})
+
+	It("has the expected plans available", func() {
+		plans := cf.Cf("marketplace", "-s", serviceName).Wait(testConfig.DefaultTimeoutDuration())
+		Expect(plans).To(Exit(0))
+		Expect(plans.Out.Contents()).To(ContainSubstring("tiny"))
+	})
+
+	PContext("creating a database instance", func() {
```
paroxp: Do you need that `PContext`? I see you took it out from compose?

platform-tests/example-apps/healthcheck/redis.go
samcrang: Yup `http` => `https`. `redis` => `rediss`.

platform-tests/src/platform/acceptance/compose_redis_service_test.go
bandesz: In the Compose tests it was originally skipped because we had random 5xx errors from their API. It's just a little bit confusing in the diff as this file was just renamed.

platform-tests/example-apps/healthcheck/redis.go
bandesz: https://www.iana.org/assignments/uri-schemes/prov/rediss

PR #1124, opened 2017-11-29
Fix grammar and typos

samcrang:


bandesz:
Do we want to trigger the full deploy pipeline for a small change like this or should we add this to an other PR?

samcrang:
Is there already a PR with related changes? I'd rather not bundle it up with something else.

It does seem heavy handed to require a deploy to make changes like this though. However, as long as they're infrequent I don't think it matters much?

alext:
I don't think we should be overly concerned with things like this triggering deploys. I think that's better than bundling changes in with unrelated PRs.

PR #1126, opened 2017-11-30
[#131354667] Overcommit cells by 66%

46bit:
## What

This has `rep` advertise cells as having 66% more memory capacity than they really have. This is safe: apps running on the platform use ~34% of their reserved memory, and we calculated this change assuming a more conservative 60% utilisation. See commit messages for details.

## How to review

* Code review
* Deploy in your dev environment and check the `rep` metrics and monitors look good

## Who can review

Not @46bit @camelpunch

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -666,6 +666,9 @@ properties:
       preloaded_rootfses:
         - "cflinuxfs2:/var/vcap/packages/cflinuxfs2/rootfs"
 
+    executor:
+      memory_capacity_mb: (( calc "floor(meta.cell.rep_commit_percentage * meta.cell.instance_memory_mb)" ))
+
     route_emitter:
       dropsonde_port: (( grab properties.metron_agent.dropsonde_incoming_port ))
       bbs:
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1127,6 +1127,7 @@ jobs:
                 ./paas-cf/manifests/cf-manifest/cloud-config/*.yml
                 ./terraform-outputs/*.yml
                 ./cf-secrets/cf-secrets.yml
+                ./paas-cf/manifests/cf-manifest/cell.yml
                 ./paas-cf/manifests/cf-manifest/env-specific/((cf_env_specific_manifest))
             run:
               path: sh
@@ -1170,6 +1171,7 @@ jobs:
                 ./cf-secrets/cf-secrets.yml
                 ./certs/*.yml
                 ./grafana-dashboards-manifest/grafana-dashboards-manifest.yml
+                ./paas-cf/manifests/cf-manifest/cell.yml
                 ./paas-cf/manifests/cf-manifest/env-specific/((cf_env_specific_manifest))
                 ./vpc-peering-manifest/*.yml
               AWS_ACCOUNT: ((aws_account))
--- a/manifests/cf-manifest/cell.yml
+++ b/manifests/cf-manifest/cell.yml
@@ -1,0 +1,6 @@
+---
+meta:
+  cell:
+    instance_type: r4.xlarge
+    instance_memory_mb: 30662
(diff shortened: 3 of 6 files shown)


bleach:
The commit message doesn't have rationale in it. Rationale is this:

- currently apps are occupying ~34% of the memory they have "reserved" on cells and the highest we have ever seen this number is 54%, a long time ago
screen shot 2017-11-30 at 17 22 16
- Seems really conservative to assume the most we will ever see is 60% of memory being used
- That means we can safely advertise  1 / 0.6 = 1.81 of our actual memory
- rounding this down, an 80% overcommit

We have a [monitor set to tell us when actual usage on the cells exceeds a threshold](https://github.com/alphagov/paas-cf/blob/b0f5d3a8971e01ab025391fa1631827d84b228cd/terraform/datadog/cell.tf#L1), so if the amount that apps use of their reservation did increase over time we would expect that monitor to tell us.

46bit:
Thanks Graham! I’d left writing a nice commit message for the morning but
I’ll amend in your explanation :-)

bleach:
D'oh I'd missed that it was still WIP!

bleach:
But anyway, it's the sort of context I *should* be providing.

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -666,6 +666,9 @@ properties:
       preloaded_rootfses:
         - "cflinuxfs2:/var/vcap/packages/cflinuxfs2/rootfs"
 
+    executor:
+      memory_capacity_mb: (( calc "floor(30662 * 1.8)" ))
```
dcarley: Can you (or anyone else) think of a way to make this more robust against us changing the instance type in the future? The old Warden/DEA implementation had an `memory_overcommit_factor` option, but I can't find any equivalent feature or discussion of such a feature for Diego.

https://github.com/alphagov/paas-cf/blob/cd251bdc50994a3bf2f2451db81775e57a3b4c1b/manifests/cf-manifest/cloud-config/020-cf-vm-types.yml#L62

Something that we might be able to change at the same time is that the `30662` number feels a bit magic and like it would benefit from using `(( grab DESCRIPTIVE_NAME ))`.

46bit: You're absolutely right, Dan. `30662` *is* magic. So I'm looking into it.

* From [the use of `1024 * 1024` for `MB`](https://github.com/cloudfoundry/executor/blob/master/initializer/configuration/configuration.go#L49-L59 ), `executor` is using `MiB`.
* `31.5GiB` is `31232MiB`. This is `570MiB` more than what `rep` was advertising.
* `30662` matches the total memory reported by `free -m` on a dev env `cell/0`:
  ```
  free -mcell/7c767cf5-30c0-49d6-8711-803c2fbe425e:~$ free -m
               total       used       free     shared    buffers     cached
  Mem:         30662       9231      21430          6        241       6204
  -/+ buffers/cache:       2786      27875
  Swap:        30663          0      30663
  ```
* Following `garden.Capacity.MemoryInBytes` a long way down it eventually gets the VM's total memory capacity.
* So, `30662` is the VM's total memory capacity. @dcarley any idea if that `570MB` is special?

dcarley: I think it's important to give a name to the number so that someone reading it understands what it represents. I'm not so concerned about the precision of the actual number - saying in the commit message where we got it from seems good enough.

46bit:
I've added a test that the `diego.executor.memory_capacity_mb` value is an integer and at least 30GB (i.e., we aren't *under*provisioning.) Unsure whether to test any more than that.

46bit:
We uncovered that we'd been using a mistaken value of `1 / 0.6`, which I've corrected for.

46bit:
Given that `cloud-config` and `manifest` are separate we're not sure where to move the cell instance values (aws instance type and its memory in MB) to. Hoping to uncover that on Monday.

46bit:
@dcarley Ready for re-review--I needed to check it deployed cleanly first.

PR #1127, opened 2017-12-01
[#152439871] Monitoring ELB nodes

chrisfarms:
## What

We have added a new "gauge" to the `paas-metrics` app that reports the number of healthy / unhealthy ELB nodes that are currently active to datadog.

It does this by fetching all the DNS entries advertised for a given elb address and then attempting to make requests against a known endpoint (healthcheck app) using each IP.

A datadog monitor is configured to alert if there is greater than zero unhealthy responses found. 

## How to review

Code review.

We have tested in dev and triggered the alert by removing the security group from the router vms (which causes connection failures and is the best approx to the issue we could think of).

If you want to test yourself you will need to deploy with:

```
ENABLE_METRICS=true ENABLE_DATADOG=true SELF_UPDATE_PIPELINE=false BRANCH=elb-gauge-152439871 make dev pipelines
```

## Who can review

Not @chrisfarms nor @camelpunch


tools/metrics/pingdumb/pingdumb_test.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, check := range r.Checks {
+			Expect(check.Start).To(BeTemporally("~", time.Now(), 1*time.Hour))
+			Expect(check.Stop).To(BeTemporally("~", time.Now(), 1*time.Hour))
+			Expect(check.Stop).To(BeTemporally(">", check.Start))
+		}
+	})
+
+	It("produces an error when it can't resolve the target", func() {
+		config := ReportConfig{
+			Target:  "https://please.never.register.this.domain.ok/",
```
samcrang: If we're just looking for something that returns `NXDOMAIN` then [RFC 6761](https://tools.ietf.org/html/rfc6761) says we may be able to use the `.invalid` or `.test` TLDs depending on the scenario we're trying to test.

tools/metrics/datadog_reporter.go
```diff
@@ -103,7 +103,7 @@ func (r *DatadogReporter) WriteMetrics(events []Metric) error {
 			continue
 		}
 		m := datadog.Metric{}
-		m.SetMetric("op." + ev.Name)
```
samcrang: I can't see anything in the commit messages that explains why we've put a bunch of metrics under a new namespace(?). I feel like this might be something useful to have a record of.

tools/metrics/fakes/http.go
```diff
@@ -0,0 +1,23 @@
...
+
+type httpHandler struct{}
+
+func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	q := r.URL.Query()
+	if q.Get("hang") == "true" {
+		time.Sleep(1 * time.Hour)
+		panic("giving up after hour of waiting")
+	}
+	w.WriteHeader(http.StatusTeapot)
```
samcrang: 🍵

tools/metrics/pingdumb/pingdumb.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return failures
+}
+
+func (r *Report) OK() bool {
+	failures := r.Failures()
+	return len(failures) == 0
+}
+
+func (r *Report) Addrs() []string {
```
samcrang: I can't see where this is used?

tools/metrics/datadog_reporter.go
camelpunch: This was a refactor - the `op.` namespace used to be added centrally, but we wanted a new namespace, so un-DRYed it first. This is what we meant by "metric name prefixes are now part of names" in https://github.com/alphagov/paas-cf/pull/1127/commits/26d221e626d0a066410943dc8394dc878e8f6823

tools/metrics/pingdumb/pingdumb_test.go
camelpunch: Sounds good!

tools/metrics/pingdumb/pingdumb.go
camelpunch: Good catch :fishing_pole_and_fish:

PR #1130, opened 2017-12-04
[#152144657] Fix cred rotation in envs with valid TLS certs

bleach:
## What

In persistent environments, which have valid TLS certificates, having
CF_SKIP_SSL_VALIDATION unset meant that uaac was passed a quoted empty
string as an argument and failed with `Too many command line parameters given.`

I unquoted the argument to `uaac target` to avoid this happening.

## How to review

Code review. I've run it in my dev env, for what it's worth

## Who can review

Anyone but @bleach


alext:
:+1: Will merge when Travis is green.

dcarley:
Reviewer race condition 😆

PR #1132, opened 2017-12-04
[#152885584] Reduce load on API during availability tests

henrytk:
## What

https://www.pivotaltracker.com/story/show/152885584
    
We want to avoid putting the API and UAA under unnecessary load during availability tests. We only care about the API being continually available, this is not a form of load testing.
    
The tasks performed against the API are not simple, single-request tasks, so we can't use a library such as Vegeta to perform the rate limiting. For simplicity, the rate at which tasks are added to the queue has been throttled.
    
Under the new configuration we will not need as many workers, as only two tasks per second will be added to the queue.

## How to review

If you have a deployment from master you can deploy from this branch and the temporary commit will rotate the `api/*` and `uaa/*` VM stemcells.

We want to know when a deployment happens the API availability tests don't put the API under unnecessary load. The kick-off of the story agreed an API request rate of around 10 per second. Results in my dev environment show they are between 10 and 20. I think that's okay.

### Before merging ⚠️ 

Remove the temporary commit.

## Who can review

Anyone but me

### DIFF
--- a/platform-tests/src/platform/availability/monitor/monitor.go
+++ b/platform-tests/src/platform/availability/monitor/monitor.go
@@ -66,16 +66,18 @@ type result struct {
 
 // Monitor continuously runs each Task given to Add in parallel
 // and gathers basic statistics on
+
 type Monitor struct {
-	tasks           []*Task
-	reporter        chan *Report
-	queue           chan *Task
-	results         chan *result
-	halt            chan bool
-	clientCfg       *cfclient.Config
-	logger          io.Writer
-	numWorkers      int
-	warningMatchers []*regexp.Regexp
+	tasks             []*Task
+	taskRatePerSecond int64
+	reporter          chan *Report
+	queue             chan *Task
+	results           chan *result
+	halt              chan bool
+	clientCfg         *cfclient.Config
+	logger            io.Writer
+	numWorkers        int
+	warningMatchers   []*regexp.Regexp
 }
 
 func (m *Monitor) Add(name string, fn TaskFunc) {
@@ -181,7 +183,7 @@ func (m *Monitor) producer() {
 			select {
 			case <-m.halt:
 				return
-			default:
+			case <-time.After(m.taskRateToDuration()):
 				m.queue <- task
 			}
 		}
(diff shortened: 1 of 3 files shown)


46bit:
LGTM but I'm curious about the 10-20 API requests per second. It appears that task production is over all workers, rather than each worker producing that many tasks. As such I'd expect 2 API requests per second. Is this because each task makes multiple underlying requests?

platform-tests/src/platform/availability/monitor/monitor.go
```diff
@@ -178,6 +180,7 @@ func (m *Monitor) worker() {
 func (m *Monitor) producer() {
 	for {
 		for _, task := range m.tasks {
+			<-time.After(m.taskRateToDuration())
```
chrisfarms: Minor issue:

I believe you want this time.After to replace the `default` case inside the select so that either the timeout passes OR halt is hit.

```
    select {
  			case <-m.halt:
  				return
 			case <-time.After(m.taskRateToDuration()):
 				m.queue <- task
 			}
```

As it is currently it looks like `halt` will have to wait until timeout is hit _before_ deciding to terminate.

platform-tests/src/platform/availability/monitor/monitor.go
```diff
@@ -197,19 +200,26 @@ func (m *Monitor) Stop() {
 	}
 }
 
+func (m *Monitor) taskRateToDuration() time.Duration {
+	nanosecondsInASecond := time.Second.Nanoseconds()
+	return time.Duration(nanosecondsInASecond / m.taskRatePerSecond)
```
chrisfarms: I might just be having trouble mentally parsing this because it's morning, is this the same as just:

```
return time.Second / m.taskRatePerSecond
```

henrytk: Kind of. `time.Second` is of type `time.Duration`, so you can't do maths with it. I suppose I could have done a type cast:

```
return int64(time.Second) / m.taskRatePerSecond
```

henrytk:
@46bit yes, each task has several underlying requests. The [first task](https://github.com/alphagov/paas-cf/blob/df8d8c2262de855c262176e448f1a49923221897/platform-tests/src/platform/availability/api/api_test.go#L45-L58) makes five different requests, and the [second task](https://github.com/alphagov/paas-cf/blob/df8d8c2262de855c262176e448f1a49923221897/platform-tests/src/platform/availability/api/api_test.go#L72-L102) makes even more.

platform-tests/src/platform/availability/monitor/monitor.go
henrytk: I've changed it to:

```
return time.Second / time.Duration(m.taskRatePerSecond)
```

for readability

platform-tests/src/platform/availability/monitor/monitor.go
henrytk: Good catch. Amended.

PR #1133, opened 2017-12-05
[Support] API-availability-tests: Don't assert for >0 successes.

alext:
## What

It's enough for this to assert that no failures occurred.

By asserting for >0 successes, it means that this job can't be
re-triggered in the event of a failure to unblock the pipeline as the
cf-deploy job won't be running, and the re-run of this will therefore
make 0 requests.

## How to review

Code review. Verify that my rationale makes sense.

## Who can review

Not me.

### DIFF
--- a/platform-tests/src/platform/availability/api/api_test.go
+++ b/platform-tests/src/platform/availability/api/api_test.go
@@ -129,7 +129,6 @@ var _ = Describe("API Availability Monitoring", func() {
 		report := monitor.Run()
 		lg(report.String())
 		Expect(report.Errors).To(BeEmpty(), "expected no errors")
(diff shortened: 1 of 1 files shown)


bandesz:
Instead of removing the assertion should we make sure that at least we run the tests for a minimum amount of time or request number? I feel that would be safer.

chrisfarms:
If I understand this change correctly .... this would mean that re-triggering this job will _always_ cause it to pass, so long as the cf-deploy job completed successfully. 

Andras' idea about ensuring a minimum number of requests have occurred before monitor.Stop() gets called sounds like a good move.

alext:
Fair enough, I'm going to close this for now, as other options are more work and should probably be covered by a story.

PR #1135, opened 2017-12-07
[#153072577] Add UAA client for paas-admin app

henrytk:
## What

We need a UAA client configuration for the paas-admin app to use. It currently has limited permissions, because the app only lists orgs at the moment as proof the authentication works.

## How to review

Review in conjunction with https://github.com/alphagov/paas-admin/pull/1

There is little danger with this commit as the client is configured with much less scope than will eventually be required. This is on purpose.

Eyeball the code and deploy with passing tests should be enough.

## Who can review

Anyone but me, @chrisfarms @camelpunch or @carolinegreen

### DIFF
--- a/manifests/cf-manifest/spec/manifest/base_properties_spec.rb
+++ b/manifests/cf-manifest/spec/manifest/base_properties_spec.rb
@@ -126,6 +126,7 @@ RSpec.describe "base properties" do
           "tcp_router",
           "ssh-proxy",
           "graphite-nozzle",
+          "paas-admin",
           "paas-metrics",
           "datadog-nozzle",
           "cc-service-dashboards",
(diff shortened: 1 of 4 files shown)


manifests/cf-manifest/spec/manifest/base_properties_spec.rb
```diff
@@ -136,6 +137,13 @@
         )
       }
 
+      it {
+        clients.each { |_, config|
+          expect(config.has_key?("override")).to be true
```
camelpunch: Because of rspec magic that translates `have_xxx(foo)` to `has_xxx?(foo)`, you could also say `expect(config).to have_key("override")`

henrytk: That reads more gooder. Thanks.

PR #1136, opened 2017-12-07
Metrics test: avoid a race

camelpunch:
## What

This avoids a race between the pingdumb test's custom DNS server spinning up and the test being run.

See e.g. https://travis-ci.org/alphagov/paas-cf/builds/313015567 for an example build failure that should no longer occur.

## How to review

Run `ginkgo -untilItFails tools/metrics/pingdumb` before and after the commit.

## Who can review

Anyone but @camelpunch


chrisfarms:
Yep it appears that it's not trivial to work out if the server is up without some kind of retry either here or after start.

Racey Racey. This does the trick.

PR #1146, opened 2017-12-18
[#153670195] Redirect to /next-steps on product page after UAA login

46bit:
## What

Redirect users to https://www.cloud.service.gov.uk/next-steps when they login to UAA, and add a pipeline task to remove an unused `documentation` UAA client. See commit messages for more details. (The pipeline task will be later removed by another PR #1147.)

## How to review

* Code review
* We've deployed this to dev several times and are confident in the delete task. You can optionally apply to your dev environment.

## Who can review

Not @46bit @paroxp. We'd like @dcarley, as he may have context from the original implementation of the documentation client in #868.


samcrang:
@dcarley is off for Christmas so I'll have to do i'm afraid 🎄.

This looks 👌  though.

PR #1148, opened 2017-12-19
[#142503647] Monitor TLS Certificate Validity

chrisfarms:
## What

We want a monitor that will alert us if any of our public-facing TLS certs are invalid, or due to expire soon.

`paas-metrics` is extended to accept a list of domains for which you want to check the validity of SSL certificates for, and emit metrics for the number of valid days remaining. This is done via environment variable:

```
TLS_DOMAINS=x.example.com,y.example.com
```

A datadog monitor is created to emit WARNINGs at < 7 days and critical alerts at < 1 day.

The metrics also catch some misconfigurations (like incorrect common name, untrusted CA) by marking the cert as invalid (0days validity).

We want these checks in staging too, so have removed the `ENABLE_METRICS` pipeline flag so that deployment is soley based on the `ENABLE_DATADOG` flag.

## How to review

* Code review
* Deploy this branch to dev with `ENABLE_DATADOG=true`
* Setup valid certs for your ELBs in aws console
* Watch the datadog monitor

## Additional

There is a sibling PR for the team-manual: https://github.com/alphagov/paas-team-manual/pull/171

## Who can review

Not @chrisfarms

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2335,7 +2335,13 @@ jobs:
                       'DATADOG_APP_KEY' => '${DATADOG_APP_KEY:-}',
                       'DEPLOY_ENV' => '${DEPLOY_ENV}',
                       'ELB_ADDRESS' => 'https://healthcheck.${APPS_DNS_ZONE_NAME}/',
-                      'TLS_DOMAINS' => 'healthcheck.${APPS_DNS_ZONE_NAME},api.${SYSTEM_DNS_ZONE_NAME},uaa.${SYSTEM_DNS_ZONE_NAME},deployer.${SYSTEM_DNS_ZONE_NAME}',
+                      'TLS_DOMAINS' => [
+                        'healthcheck.${APPS_DNS_ZONE_NAME}',
+                        'api.${SYSTEM_DNS_ZONE_NAME}',
+                        'uaa.${SYSTEM_DNS_ZONE_NAME}',
+                        'deployer.${SYSTEM_DNS_ZONE_NAME}',
+                        '${APPS_DNS_ZONE_NAME}'
+                      ].join(','),
                     }
                     manifest = YAML.load_file('manifest.yml')
                     manifest['applications'].each { |app|
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2293,11 +2293,12 @@ jobs:
                 repository: governmentpaas/cf-cli
                 tag: 465642da06051a55630d39c899697b678f66a7f7
             params:
-              ENABLE_METRICS: ((enable_metrics))
+              ENABLE_DATADOG: ((enable_datadog))
               DATADOG_API_KEY: ((datadog_api_key))
               DATADOG_APP_KEY: ((datadog_app_key))
               DEPLOY_ENV: ((deploy_env))
               APPS_DNS_ZONE_NAME: ((apps_dns_zone_name))
+              SYSTEM_DNS_ZONE_NAME: ((system_dns_zone_name))
             inputs:
               - name: paas-cf
               - name: config
@@ -2310,8 +2311,8 @@ jobs:
                 - -u
                 - -c
                 - |
-                  if  [ "${ENABLE_METRICS:-}" != "true" ] ; then
-                    echo "ENABLE_METRICS=${ENABLE_METRICS} so skipping paas-metrics deploy"
+                  if  [ "${ENABLE_DATADOG:-}" != "true" ] ; then
+                    echo "ENABLE_DATADOG=${ENABLE_DATADOG} so skipping paas-metrics deploy"
                     exit 0
                   fi
 
@@ -2334,6 +2335,7 @@ jobs:
                       'DATADOG_APP_KEY' => '${DATADOG_APP_KEY:-}',
                       'DEPLOY_ENV' => '${DEPLOY_ENV}',
                       'ELB_ADDRESS' => 'https://healthcheck.${APPS_DNS_ZONE_NAME}/',
+                      'TLS_DOMAINS' => 'healthcheck.${APPS_DNS_ZONE_NAME},api.${SYSTEM_DNS_ZONE_NAME}',
                     }
                     manifest = YAML.load_file('manifest.yml')
                     manifest['applications'].each { |app|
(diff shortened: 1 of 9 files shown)


terraform/datadog/certificates.tf
```diff
@@ -0,0 +1,18 @@
...
+  notify_no_data    = true
+  no_data_timeframe = 120
+
+  query = "${format("min(last_1h):min:tls.certificates.validity{deploy_env:%s} by {hostname} <= 0", var.env)}"
+
+  require_full_window = true
+
+  thresholds {
+    warning  = 7
+    critical = 1
```
samcrang: I get this error while trying to deploy to my dev environment:

```
* datadog_monitor.invalid_tls_cert: error updating montor: API error 400 Bad Request: {"errors":["Critical threshold (1) does not match that used in the query (0.0)."]}
```

tools/metrics/README.md
```diff
@@ -21,6 +21,7 @@ The following metrics are currently collected:
 |`op.quotas.memory.allocated` | Gauge | Total amount of memory promised to apps | |
 |`op.quotas.services.reserved` | Gauge | Total number of services promised to orgs | |
 |`op.quotas.services.allocated` | Gauge | Total number of services assigned | |
+|`tls.certificates.validity` | Gauge | Number of days cert will be valid for | `hostname` |
```
samcrang: `s/will be/is/`?

chrisfarms:
Should deploy cleanly now... and resolved the grammer in readme.

henrytk:
I've been caught out again by the need to set `SELF_UPDATE_PIPELINE=false` :(

tools/metrics/gauges.go
```diff
@@ -4,13 +4,48 @@ import (
...
 	"code.cloudfoundry.org/lager"
 	"github.com/alphagov/paas-cf/tools/metrics/pingdumb"
+	"github.com/alphagov/paas-cf/tools/metrics/tlscheck"
 	cfclient "github.com/cloudfoundry-community/go-cfclient"
 )
 
+func TLSValidityGauge(logger lager.Logger, addr string, interval time.Duration) MetricReadCloser {
+	return NewMetricPoller(interval, func(w MetricWriter) error {
+		if !strings.Contains(addr, ":") {
+			addr += ":443"
```
henrytk: What happens if mis-configuration results in the `addr` being `https://example.com:9000`?

chrisfarms: 🙌  I had hoped that by naming the configuration variable `TLS_DOMAINS` rather than `_ENDPOINT` or `_URL` this would be avoided.

... I have added an additional check (by attempting to parse with `net.SplitHostAddr`) to catch this kind of misconfiguration a little better.

tools/metrics/tlscheck/tls.go
```diff
@@ -0,0 +1,45 @@
...
+func GetCertificate(addr string) (*x509.Certificate, error) {
+	conn, err := tls.Dial("tcp", addr, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	for _, chain := range conn.ConnectionState().VerifiedChains {
+		for _, cert := range chain {
+			if !cert.IsCA {
```
henrytk: Doing this makes two assumptions:

* the first `VerifiedChains` element is the one we want (it probably is)
* the certificate we want to test is the first in the slice of certificates. I can't find evidence of this in the docs.

https://golang.org/pkg/crypto/tls/#ConnectionState

henrytk: The above points are probably not a problem, but I would feel better if I had the answers.

terraform/datadog/certificates.tf
```diff
@@ -0,0 +1,18 @@
+resource "datadog_monitor" "invalid_tls_cert" {
+  name              = "${format("%s Invalid TLS/SSL Certificate", var.env)}"
+  type              = "metric alert"
+  message           = "${format("{{hostname.name}} certificate {{#is_alert}}is invalid!{{/is_alert}}{{#is_warning}}expires in {{value}} days{{/is_warning}}\n\nSee [Team Manual > Responding to alerts > Invalid Certificates](%s#invalid-certificates) for more info.", var.datadog_documentation_url)}"
```
henrytk: This doesn't use the [notification addresses](https://github.com/alphagov/paas-cf/blob/b913ccf868c6349a387801eb78cdfef52fc096b7/terraform/datadog/variables.tf#L16-L22) to actually notify us. It will show in Datadog as triggered, but we won't be alerted by email/PagerDuty. Example usage can be found in `terraform/datadog/concourse.tf:22`

terraform/datadog/certificates.tf
```diff
@@ -0,0 +1,18 @@
...
+  notify_no_data    = true
+  no_data_timeframe = 120
+
+  query = "${format("min(last_1h):min:tls.certificates.validity{deploy_env:%s} by {hostname} <= 0", var.env)}"
+
+  require_full_window = true
+
+  thresholds {
+    warning  = 7
```
henrytk: Is seven days enough? Any thoughts @alext ?

terraform/datadog/certificates.tf
henrytk: This doesn't use the [notification addresses](https://github.com/alphagov/paas-cf/blob/b913ccf868c6349a387801eb78cdfef52fc096b7/terraform/datadog/variables.tf#L16-L22) to actually notify us if it is critical. It will show in Datadog as triggered, but we won't be alerted by email/PagerDuty. Example usage can be found in `terraform/datadog/concourse.tf:22`

terraform/datadog/certificates.tf
alext: I agree that this is a little short. I reckon we should trigger warning at 30 days, and critical at 7 days. We can always tune these later (different values may be appropriate if we switch to something with auto-renewing like ACM, but these seems sensible while it's a manual process.).

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2334,6 +2335,7 @@ jobs:
                       'DATADOG_APP_KEY' => '${DATADOG_APP_KEY:-}',
                       'DEPLOY_ENV' => '${DEPLOY_ENV}',
                       'ELB_ADDRESS' => 'https://healthcheck.${APPS_DNS_ZONE_NAME}/',
+                      'TLS_DOMAINS' => 'healthcheck.${APPS_DNS_ZONE_NAME},api.${SYSTEM_DNS_ZONE_NAME}',
```
henrytk: We can't just cover a single subdomain as the certificate should be valid for the parent domain as well. This is because the CDN broker-created CloudFront Distributions use the parent domain as the origin by default and is HTTPS only. We should probably add the parent domain (just `$APPS_DNS_ZONE_NAME`) as well.

terraform/datadog/certificates.tf
chrisfarms: We did not discuss this alert being something that triggers pagerduty.

With the other new monitors/alerts we had decided to run them for a bit without pageduty alerts to ensure we have them tuned nicely before waking people up.

henrytk: I suppose there is very little chance of this turning critical unexpectedly out of hours, given that it would take a misconfiguration or us not seeing the warning for several days/weeks.

PR #1150, opened 2017-12-21
[support] Amend the EC2 CPU monitor

paroxp:
## What

We've decided to amend the monitor in order to prevent it from being in
a warning stage constantly. We thought it would be the best thing to do,
due to our upcoming plans of moving to logit.

## How to review

- Sanity check


chrisfarms:
I'm not sure that's the best move ... the problem is elasticsearch specifically right.

Could we just exclude the elasticsearch job from this monitor if it's causing too much noise?

paroxp:
Good spot. Thanks.

I ignored the ES job in the current monitor and created another which is meant to target the ES job only. I'd propose a thorough check.

PR #1151, opened 2017-12-22
[#147833555] Pass the platform egress IPs to the Compose broker

henrytk:
## What

Pass the NAT gateway IPs along to the compose-broker app environment as a comma-separated list.

The broker consumes these IPs in https://github.com/alphagov/paas-compose-broker/pull/22

paas-bootstrap got https://github.com/alphagov/paas-bootstrap/pull/112 but I think this was for consistency (@henrytk can you confirm?)

## How to review

- Deploy a dev environment with `BRANCH=paas-egres-ips` [sic].
- Once deployed, enable a compose service e.g. `cf enable-service-access compose-redis`
- Create a service instance with e.g. `cf create-service compose-redis tiny myredis`
- Push an app e.g. `cd rails-demo; cf push`
- Bind the service e.g. `cf bind-service rails-blog myredis`
- Restage the app, as the nice message says.
- Get the app's env to retrieve the redis URL: `cf env rails-blog | grep 'uri"'`
- Test that connecting doesn't work from your workstation:
    ```
    $ netcat -v SOME_HOST SOME_PORT
    Connection to SOME_HOST SOME_PORT port [tcp/*] succeeded!
    keys *
    $ # this means you got disconnected    
    ```
- ssh to the app: `cf ssh rails-blog`
- Repeat the previous `netcat` test to ensure that you instead get `-NOAUTH Authentication required.`
- Optional: repeat the above for prod, to ensure that the test was good (but @camelpunch has already checked)

## Before merge

❗️ There is a temporary commit setting the paas-compose-broker repository to the dev branch. After merging paas-compose-broker please replace that commit with the final tag.

## Who can review

Not @henrytk, @camelpunch, or @46bit

### DIFF
--- a/terraform/cloudfoundry/outputs.tf
+++ b/terraform/cloudfoundry/outputs.tf
@@ -173,3 +173,7 @@ output "cell_subnet_cidr_blocks" {
 output "router_subnet_cidr_blocks" {
   value = ["${aws_subnet.router.*.cidr_block}"]
 }
+
+output "nat_public_ips_csv" {
+  value = "${join(",", aws_eip.cf.*.public_ip)}"
+}
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -251,7 +251,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-compose-broker
-      tag_filter: v0.12.0
+      tag_filter: v0.13.0
 
   - name: paas-compose-scraper
     type: git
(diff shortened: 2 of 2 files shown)


concourse/scripts/val_from_yaml.rb
```diff
@@ -51,7 +51,9 @@ def [](key)
   val = property_tree[key]
   abort "Unable to find key: #{key}" if val.nil?
 
-  if val.is_a?(Array) || val.is_a?(Hash)
+  if val.is_a?(Array)
+    puts val.join(',')
+  elsif val.is_a?(Hash)
```
alext: Are you sure that nothing is using the previous behaviour which outputs an Array as YAML?

bandesz: I find this change dangerous too, you should come up with a backward-compatible solution for this.

concourse/scripts/val_from_yaml_test.go
```diff
@@ -132,4 +132,16 @@ val2: b
 			Expect(session.Err.Contents()).To(BeEmpty())
 		})
 	})
+
```
bandesz: we should add a test for a hash value too

bandesz:
There is no temporary commit to change the paas-compose-broker repository to the dev branch

chrisfarms:
Should there be a PR for the [paas-gocomposeapi](https://github.com/alphagov/paas-gocomposeapi) fork too? ... I believe we try to standanise on a `gds_master` branch for forks we make so there is a consistent "master" to work against for any future changes[1]

[1] https://government-paas-team-manual.readthedocs.io/en/latest/team/working_practices/#creating-and-forking-repos

camelpunch:
[PR for our gocomposeapi](https://github.com/alphagov/paas-gocomposeapi/pull/1). Avoided a chicken-egg situation by creating the gds_master branch based on upstream master.

terraform/cloudfoundry/outputs.tf
```diff
@@ -173,3 +173,7 @@ output "cell_subnet_cidr_blocks" {
 output "router_subnet_cidr_blocks" {
   value = ["${aws_subnet.router.*.cidr_block}"]
 }
+
+output "nat_public_ips" {
```
bandesz: should we indicate that this is a comma separated list? e.g. ```nat_public_ips_csv``` or similar

PR #1152, opened 2018-01-02
[support] Add new DiT VPCs for peering.

LeePorte:
## What

This PR adds the newly supplied DiT VPCs to peer with PaaS.

## How to review

Code review and terraform plan.

## Who can review

Not @LeePorte, ideally @bleach as he has the most context.

### DIFF
--- a/terraform/prod.vpc_peering.json
+++ b/terraform/prod.vpc_peering.json
@@ -10,5 +10,17 @@
         "account_id": "650575079077",
         "vpc_id": "vpc-82c913e5",
         "subnet_cidr": "172.17.0.0/22"
+    },
+    {
+        "peer_name": "dit_trade_policy_group_prod",
+        "account_id": "588284153815",
+        "vpc_id": "vpc-57be5f31",
+        "subnet_cidr": "172.16.20.0/22"
+    },
+    {
+        "peer_name": "dit_trade_policy_group_dev",
+        "account_id": "001221602192",
+        "vpc_id": "vpc-b4afb5d3",
+        "subnet_cidr": "172.16.24.0/22"
     }
 ]


terraform/cloudfoundry/vpc_peering.tf
```diff
@@ -25,3 +25,17 @@ resource "aws_route" "vpc_peer_route_2" {
...
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
+  vpc_peering_connection_id = "${element(aws_vpc_peering_connection.vpc_peer.*.id, count.index)}"
+}
+
+resource "aws_route" "vpc_peer_route_4" {
+  count                     = "${length(var.peer_vpc_ids)}"
+  route_table_id            = "${aws_route_table.internet.4.id}"
+  destination_cidr_block    = "${element(var.peer_cidrs, count.index)}"
+  vpc_peering_connection_id = "${element(aws_vpc_peering_connection.vpc_peer.*.id, count.index)}"
+}
```
alext: Why are these 2 blocks added? It should only be necessary to add to the json file to add a new peering connection.

LeePorte: I thought we needed a block per VPC added in addition to the VPC that PaaS runs in.

alext: This gets a little involved, but we need one route per VPC peering per route-table. Because this 2-dimensional arry-like thing cant be done directly with terraform, we've instead created a single array resource (using the `count` param) per AZ, effectively hard-coding the fact that we have 3 AZs (and therefore 3 route-tables).

The upshot is that this will only need to be changed if we change the number of route-tables we have.

There's more details in the docs here: https://github.com/alphagov/paas-cf/blob/master/doc/vpc_peering.md

terraform/prod.vpc_peering.json
```diff
@@ -10,5 +10,17 @@
         "account_id": "650575079077",
         "vpc_id": "vpc-82c913e5",
         "subnet_cidr": "172.17.0.0/22"
+    },
+    {
+        "peer_name": "dit_trade_prod",
```
alext: The t in dit stands for trade, so this name is redundant. `dit_prod` (and `dit_dev`) is probably fine here.

alext:
Could you add a link to the DeskPro ticket to the commit message please.

PR #1153, opened 2018-01-02
[Support] Override default certificate renewal cron

LeePorte:
## What

We are currently hitting rate limits from Let's Encrypt for certificate
renewals. This change will override the default and set it to run once
per week on a Sunday at midnight.

## How to review

Code review

## Who can review

Not @LeePorte


dcarley:
I don't think we should do this because we know that this is caused by incorrect entries in the database and the story to investigate/fix it is second down in the backlog: https://www.pivotaltracker.com/story/show/153672095

dcarley:
btw, you and @alext talk about this after standup? He might have some input.

alext:
I agree that this isn't the correct solution to the problem.

I'm not convinced that this would fix it anyway because I reckon the rate-limiting will happen over a much smaller window, and is down to the issue that's referenced in the story @dcarley linked to.

PR #1165, opened 2018-01-11
Refactor uaa_sync_admin_users

camelpunch:
## What

- Move network access out of constructor into memoized method. Objects are [easier to compose](http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/) if they don't do their work during object graph construction.
- Remove redundant 'self.' (which usually indicates an attribute write)
- Make description of options a bit more accurate, make it a required parameter (we never call it otherwise)
- Make things private where the test permits. Code is easier to change if you can tell at a glance what depends on it.
- Remove redundant comments
- Break some long lines
- Remove redundant parens in a comparison (operators have a lower precedence than other expressions)
- Rename `ua` to `uaa`
- Use `%()` to avoid escaping quotes

## How to review

Run tests (should have been done by Travis already), code inspection.

If this feels risky, you could ensure that removal still works. See https://github.com/alphagov/paas-cf/pull/1163 for instructions.

## Who can review

Not @camelpunch

### DIFF
--- a/scripts/lib/uaa_sync_admin_users.rb
+++ b/scripts/lib/uaa_sync_admin_users.rb
@@ -5,15 +5,13 @@ require 'time'
 require 'set'
 require 'uaa'
 
-require 'pp'
-
 # Utility class to create or delete admin users.
 #
 # Manages  users members from the groups listed in
 # UaaSyncAdminUsers::DEFAULT_ADMIN_GROUPS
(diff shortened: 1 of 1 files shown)


henrytk:
Travis is happy and I have run the script locally to create and delete users. Merging.

PR #1174, opened 2018-01-18
[#143584125] Enable password resets

henrytk:
## What

We want to allow tenants to be able to reset their passwords, but not enable account creation.

See commit messages for details.

## How to review

* deploy from this branch and make sure all tests pass
* check you can reset a user's password. You may need to verify your email with SES as our dev account is still in the sandbox:

  ```
  aws ses verify-email-identity --email-address your-email@example.com
  ```
* check all account creation endpoints are blocked. I am not going to tell you what they are, you have to try and hack it. Go on, I dare you.
* check the password rotation works. Related pull request in the team manual: https://github.com/alphagov/paas-team-manual/pull/176

## Who can review

Anyone but me, @chrisfarms, or @46bit

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1655,6 +1655,51 @@ jobs:
                   cf enable-feature-flag diego_docker
 
       - aggregate:
+        - task: block-create-account-endpoints
+          config:
+            platform: linux
+            image_resource:
+              type: docker-image
+              source:
+                repository: governmentpaas/cf-cli
+                tag: 465642da06051a55630d39c899697b678f66a7f7
+            inputs:
+              - name: paas-cf
+              - name: config
+              - name: bosh-CA
+            run:
+              path: sh
+              args:
+                - -e
+                - -u
+                - -c
+                - |
+                  ./paas-cf/concourse/scripts/import_bosh_ca.sh
+
+                  . ./config/config.sh
+                  cf login -a "${API_ENDPOINT}" -u "${CF_ADMIN}" -p "${CF_PASS}" \
+                    -o admin -s assets
+
+                  cd paas-cf/platform-tests/example-apps/create-account-holding-page/
+                  cf push
+
+                  if ! cf domains | grep -q "login.((system_dns_zone_name))"; then
+                    cf create-domain admin "login.((system_dns_zone_name))"
+                  fi
+                  if ! cf domains | grep -q "uaa.((system_dns_zone_name))"; then
+                    cf create-domain admin "uaa.((system_dns_zone_name))"
+                  fi
+
+                  cf create-route assets "login.((system_dns_zone_name))" --path create_account
+                  cf create-route assets "login.((system_dns_zone_name))" --path create_account.do
+                  cf create-route assets "uaa.((system_dns_zone_name))" --path create_account
+                  cf create-route assets "uaa.((system_dns_zone_name))" --path create_account.do
+
+                  cf map-route create-account-holding-page "login.((system_dns_zone_name))" --path create_account
+                  cf map-route create-account-holding-page "login.((system_dns_zone_name))" --path create_account.do
+                  cf map-route create-account-holding-page "uaa.((system_dns_zone_name))" --path create_account
+                  cf map-route create-account-holding-page "uaa.((system_dns_zone_name))" --path create_account.do
+
         - task: deploy-compose-service-broker
           config:
             platform: linux
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2956,6 +2956,7 @@ jobs:
       - get: cf-secrets
         passed: ['rotate-cloudfoundry-credentials']
         trigger: true
+      - get: cf-tfstate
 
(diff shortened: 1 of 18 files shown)


keymon:
I did a push force and changing a git commit to force travis to refresh the PR branch :thinking:

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -247,6 +250,11 @@ resources:
     source:
       interval: 5m
 
+  - name: smtp-credential-rotation-timer
```
keymon: I would challenge that we rotate the credentials in timely manner. That would force the generation of a new manifest every 720h, and then force a restart of UAA. 

If the requierement is not keep permanent AWS keys in the state file, we can use the property https://www.terraform.io/docs/providers/aws/r/iam_access_key.html#pgp_key by passing a random pregenerated GPG public key for which we do not keep the private key anyware.

By encrypting the AWS access key with a gpg key without private key, we are effectively not keeping those keys at all! The SES credentials would be accessible anyway.

manifests/cf-manifest/spec/manifest/base_properties_spec.rb
```diff
@@ -94,12 +94,19 @@
...
+      it { is_expected.to include("user" => "SmtpAccessTokenID") }
+      it { is_expected.to include("password" => "SmtpPassword") }
+    end
+
     describe "links" do
       subject(:links) { login.fetch("links") }
 
       it { is_expected.to include("passwd" => "https://login.#{terraform_fixture(:cf_root_domain)}/forgot_password") }
-      it { is_expected.to include("signup" => "https://login.#{terraform_fixture(:cf_root_domain)}/create_account") }
-      it { is_expected.to include("homeRedirect" => "https://www.#{terraform_fixture(:cf_root_domain)}/next-steps") }
```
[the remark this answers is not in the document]
henrytk: Sounds good. Done.

concourse/pipelines/create-cloudfoundry.yml
henrytk: We don't care that the keys are in the state file - that's okay. Deleting the key from the state file is a hack that causes Terraform to generate a new key on the next run. All unused keys will be cleared up after every deploy.

platform-tests/example-apps/create-account-holding-page/nginx.conf
```diff
@@ -0,0 +1,54 @@
...
+  server {
+    listen <%= ENV.fetch("PORT") %>;
+    server_name localhost;
+
+    location / {
+      root <%= ENV["APP_ROOT"] %>/public;
+      try_files $uri /index.html;
+    }
+
+    error_page 405 =200 @405;
```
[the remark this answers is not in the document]
henrytk: I chose to redirect to `GET` and provide a response body to make life easier for me as a tester. I can see in a browser the correct response. I gave little care to whether it was appropriate because this is an endpoint not advertised to tenants, shouldn't be used, and will only ever be seen by us or an attacker.

I can change it if you have strong feelings, but I don't think it is a good use of our time.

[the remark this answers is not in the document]
keymon: The main reason is that this way if something happens to call it, it would be more likely to fail properly. With a 200 we risk of having a parsing error somewhere and having issues troubleshooting.  

It is not a big change, but no strong feelings from me. If you have strong ones to keep it like this, I am OK with it :-D

concourse/pipelines/create-cloudfoundry.yml
keymon: But what is the motivation of generating new keys every X hours?

platform-tests/example-apps/create-account-holding-page/nginx.conf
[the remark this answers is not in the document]
henrytk: Understandable. I think we should keep it as is for now. This whole approach to route blocking is a temporary solution anyway.

henrytk:
I've updated the pipeline to make sure the key is removed from the state file when the manual credential rotation is triggered. This is in addition to the automatic rotation every 720 hours.

Ready for re-review

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2994,35 +3004,14 @@ jobs:
...
-            - -e
-            - -c
-            - |
-              cp cf-tfstate/cf.tfstate updated-tfstate/cf.tfstate
-              terraform state rm \
-                -state=updated-tfstate/cf.tfstate \
-                aws_iam_access_key.ses_smtp
-              echo "ses smtp access key no longer managed by terraform."
-              echo "next successful deployment run will revoke the expired keys"
+      file: paas-cf/concourse/tasks/forget-ses-smtp-access-key.yml
```
keymon: You need to pull the resource `paas-cf`, add a `get: pass-cf` before the task.

keymon: image

henrytk:
> The rotate keys does not work due missing resource get

@keymon soz, that was due to the latest refactor. Strangely it worked for me :/

henrytk:
> The rotate keys does not work due missing resource get

@keymon soz, that was due to the latest refactor. Fixed and pushed.

PR #1179, opened 2018-01-22
[#149928015] Use non-deprecated JWT signing key properties

camelpunch:
## What

Update UAA JWT signing key properties to latest format.

This modern format allows setting of multiple signing keys, which makes it possible to have overlapping certificates. Useful when we want to rotate internal certificates.

## How to review

Push a pipeline with the branch and ensure that deploying CF is a no-op.

## Who can review

Anyone but @keymon or @camelpunch.


manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -452,9 +452,11 @@ properties:
       client_secret: (( grab secrets.uaa_cc_client_secret ))
 
     jwt:
-      verification_key: (( grab secrets.uaa_jwt_verification_key ))
```
alext: If this secret is no longer used, can we remove it from the secret generator and anywhere else it's referenced.

camelpunch: Good call. I'll update this PR to do that, if indeed it's unused.

keymon: I think it is generated here, by extracting the public key from the cert: https://github.com/alphagov/paas-cf/blob/f24c36623d0b72be26c606f74e8248c6a8f1c966/concourse/pipelines/create-cloudfoundry.yml#L549-L550

But maybe is being used in datadog.

alext: I couldn't find any other references to it in this repo...

camelpunch:
The 'Remove generation' commit has passed through my dev env successfully. Ready for review again.

PR #1184, opened 2018-01-25
[#149928858] Implement logic to rotate CF internal certs

keymon:
[#149928858 Deal with internal TLS certs expiring before March] https://www.pivotaltracker.com/n/projects/1275640/stories/149928858

What?
----

We want to be able to rotate the CF internal certificates in a automated
fashion, in a similar way we rotate other CF credentials.

Some public facing certs, like app or sys domains, won't be rotated with
this operation.

How to review?
-------

 - Code review.  See commit for implementation details
 - Rotate the certs:
    1. Run the job rotate-cf-internal-certs
    2. rerun the deploy script. There should be no downtime in the apis but a pair of expected `CF-StatsUnavailable` (fixed by https://github.com/alphagov/paas-cf/pull/1183)
    3. Check you can use the platform, or run acceptance tests.
 - Delete old certs:
    1. run the job delete-old-cf-internal-certs
    2. redeploy and test the platform again
 -  You can check certstrap args with: `docker run -ti governmentpaas/certstrap:465642da06051a55630d39c899697b678f66a7f7 certstrap init -h` or `docker run -ti governmentpaas/certstrap:465642da06051a55630d39c899697b678f66a7f7 certstrap sign -h`

Who?
---

Anyone but @keymon or @LeePorte

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -62,6 +62,8 @@ groups:
       - rotate-cloudfoundry-credentials
       - clear-cloudfoundry-credentials
       - expire-aws-keys
+      - rotate-cf-internal-certs
+      - delete-old-cf-internal-certs
 resource_types:
 - name: s3-iam
   type: docker-image
@@ -544,7 +546,8 @@ jobs:
                   echo "Extracting extant certs"
                   tar -xvzf cf-certs/cf-certs.tar.gz -C certs
 
-                  ./paas-cf/manifests/cf-manifest/scripts/generate-cf-certs.sh certs bosh-CA/bosh-CA.tar.gz
+                  ./paas-cf/manifests/cf-manifest/scripts/generate-cf-certs.sh \
+                    certs bosh-CA/bosh-CA.tar.gz create
 
                   echo "Creating updated cert tarball"
                   cd certs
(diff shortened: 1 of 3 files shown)


keymon:
@46bit I rebased with master to fix a minor conflict in the pipeline groups

46bit:
Thanks for the instructions on how to check `certstrap` args 👍

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -62,6 +62,9 @@ groups:
       - rotate-cloudfoundry-credentials
       - clear-cloudfoundry-credentials
       - expire-aws-keys
+      - expire-ses-smtp-credentials
```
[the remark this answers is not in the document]
keymon: You are right, they renamed that job in ea67c65bf3993943e659998a3786d53b77c616c7, which caused the conflict, and then I added the line back by accident when fixing the conflict. 

I removed it

keymon:
I rebased with master to include the latest merge of #1183. Now that this is added, the certificate shall pass without downtime in the api availability tests. 

Note: you must deploy at least once without rotating to get the new loggregator release deployed.

46bit:
@keymon Even with your new rebase, when running the `rotate-cf-internal-certs` task and then `create-cloudfoundry` I seem to reliably get a `CF-StatsUnavailable` error. But unless you have any ideas, that doesn't seem sufficient reason to revisit this? https://deployer.miki.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/api-availability-tests/builds/151

keymon:
@46bit but did you first deploy the changes once first? :-m 

It is strange, I do not think it is related to the key rotation. Anyway, I think it is acceptable to have a pair of errors.

46bit:
I had. It seems repeatable that there's a very few errors but only after a rotation. I'm happy with that, so I'll merge this on Monday morning 👍

manifests/cf-manifest/manifest/020-cf-properties.yml
```diff
@@ -457,6 +457,8 @@ properties:
         keys:
           default:
             signingKey: (( grab secrets.uaa_jwt_signing_key ))
+          previous:
+            signingKey: (( grab secrets.uaa_jwt_signing_key_old || secrets.uaa_jwt_signing_key ))
```
alext: This line has a typo: it should be `secrets.uaa_jwt_signing_old_key`.

I only noticed this because both secrets were changed when rotating, which I wasn't expecting...

We should add a test for this.

keymon: Well spotted! Feel free to push it back and I would fix this and write a test.

PR #1188, opened 2018-01-29
[#154717287] Remove Tom Byers from merge keys

dcarley:
## What

This reverts commit 0d7ef00246beebc5dc683bae58021a3af189a015.

Tom has left the PaaS team.

## How to review

Confirm that the key is Tom's.

## Who can review

Anyone.


alext:
I think we should also re-encrypt everything without his key, otherwise the store is in a inconsistent state...

PR #1189, opened 2018-01-29
[WIP] [#185157352] Adopt use of fewer Elasticache Parameter groups

46bit:
## What

Use the [updated Elasticache broker](https://github.com/alphagov/paas-elasticache-broker/pull/8) that uses fewer Elasticache parameter groups. We can no longer specify the Redis `maxmemory-policy` when `cf create-service-instance`, so also add a plan for each of those that suit our expected usecases.

## How to review

* Check the plans are suitable
* Deploy to dev and check that you can provision yourself a nice Redis

## 🐝 Before merge!

Make sure to update the WIP commit to point to the merged release of the Elasticache broker.

## Who can review

Not @46bit @chrisfarms

### DIFF
--- a/manifests/cf-manifest/manifest/070-elasticache-broker.yml
+++ b/manifests/cf-manifest/manifest/070-elasticache-broker.yml
@@ -43,22 +43,55 @@ jobs:
               bindable: true
               plan_updateable: false
               plans:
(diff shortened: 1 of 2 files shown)


chrisfarms:
Closing as we have decided to keep the existing one param-group per instance.

PR #1190, opened 2018-01-29
[support] Remove failure-testing pipeline

dcarley:
## What

We don't deploy this anywhere by default and as such we haven't
been very good at maintaining it. We have applied minor changes to the
pipeline config, mostly copy/pasted from other pipelines, but I don't
think we've actually run it anywhere for a long time. I suspect that it
doesn't work in it's current form.

I think it's best that we remove it for now. If we need it again in the
future then we can restore it from history, or implement it in a
slightly different way (e.g. automatically based on job info from BOSH).

## How to review

Code review. See if my rationale makes sense and check whether I've missed anything.

## Who can review

Anyone.


alext:
I agree with the rationale here. I haven't had a chance to do code review though.

PR #1191, opened 2018-01-29
[#154477526] Run the smoke tests from paas-cf-smoke-tests

bandesz:
## What

We had to fix the smoke tests because of a tightly timed test which used to fail randomly.

In the "can be pushed, scaled and deleted" tests the application is upscaled to two instances. The tests wait until both application instances are in a running state and run up to 30 requests to check whether both instances get traffic.

The timing of these tests are really tight, as the route registration can be still pending even if the API returns a running state for the second instance and the requests are too quick (we are seeing only ~400ms for all 30 requests). In rare cases the tests fail although the route is eventually registered a little bit later (some hundreds of milliseconds).

We add some sleep between the requests, which adds an extra 5 seconds window for the new route registration to finish.

We used the previously applied pattern to override the cf-smoke-tests git submodule before running the tests and include a check whether we need to update or remove our fork if the submodule was updated in a new release.

## Related PRs

* paas-cf-smoke-tests: https://github.com/alphagov/paas-cf-smoke-tests/pull/1
* cf-smoke-tests: https://github.com/cloudfoundry/cf-smoke-tests/pull/39

## How to review

1. Review and merge https://github.com/alphagov/paas-cf-smoke-tests/pull/1 first
1. Delete the branch and create a tag on paas-cf-smoke-tests on gds_master called fix_app_scale_tests_154477526 (same name as the branch) - this is to avoid the branch accidentally being deleted
1. Deploy your CF from this branch successfully
    ```BRANCH=fix_app_scale_tests_154477526 make dev pipelines```

## Who can review

Not @henrytk or @bandesz

### DIFF
--- a/platform-tests/upstream/run_smoke_tests.sh
+++ b/platform-tests/upstream/run_smoke_tests.sh
@@ -2,6 +2,22 @@
 
 set -eu
 
+# FIXME: Remove this once https://github.com/cloudfoundry/cf-smoke-tests/pull/39 was merged
+# and our CF release contains the fix.
+(
+  cd  "$(pwd)/cf-release/src/smoke-tests/"
+  expected_commit_hash="1f219aa5cb8f3b6bfad312654ca66ff54d0cae5e"
+  current_commit_hash="$(git log --pretty=format:'%H' -n 1)"
+  if [ "${expected_commit_hash}" != "${current_commit_hash}" ]; then
+    echo "ERROR: Current commit for smoke-tests is different than expected one: ${expected_commit_hash} != ${current_commit_hash}"
+    echo "Double check if we still need to pull our branch"
+    exit 1
+  fi
+  git remote add alphagov https://github.com/alphagov/paas-cf-smoke-tests.git
+  git fetch alphagov
+  git checkout fix_app_scale_tests_154477526
+)
+
 export CONFIG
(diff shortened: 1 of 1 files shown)


keymon:
@andrasferenczszabo Thx. Actually in the story https://www.pivotaltracker.com/story/show/150673047 to remove the cf-release dependency, we will depend directly on the cf-smoke-tests, so we do not need that logic anymore, we can pull our own branch from the pipeline.

See branch https://github.com/alphagov/paas-cf/tree/remove-cf-release-dependency-150673047

keymon:
@samcrang @bandesz see https://github.com/alphagov/paas-cf/commit/24a576d7d84b96f6135e2db35f565add8294c782 . I think we can add the change in that commit, or create a separated PR to be merged after the remove cf-release story. thoughts?

bandesz:
@keymon @samcrang I think it's always better to keep things separate, so I would merge this PR. We can get rid of the smoke tests sooner as the other PR is still WIP.

bandesz:
@keymon fine with me then, but I leave it to @samcrang as he does the review

samcrang:
I don't love that we've managed to do work that conflicts but given that it's happened I think it probably makes most sense to close this PR and bundle the work up with the other PR.

PR #1192, opened 2018-01-30
[#149928858] Fix wrong name for old jwt signing key

keymon:
https://www.pivotaltracker.com/story/show/149928858

What?
-----

Follow up of https://github.com/alphagov/paas-cf/pull/1184

We made a typo in the name of the key when pulling the old jwt
signing key for UAA during the certificate rotation.

The name should be <cert-name> + _old + _key, as per [1].

The issue would be shadowed by spruce, as we pull a default value
if the key is not defined. We added a test for this specific
case in the manifest.

[1] https://github.com/alphagov/paas-cf/blob/0567d631f09a8b352fc8629375a81731f523069a/concourse/pipelines/create-cloudfoundry.yml#L1304-L1313

How to review?
----------

Code review, the tests in travis shall run.

Who?
----

Anyone but @keymon, likely @alext

### DIFF
--- a/manifests/cf-manifest/manifest/020-cf-properties.yml
+++ b/manifests/cf-manifest/manifest/020-cf-properties.yml
@@ -458,7 +458,7 @@ properties:
           default:
             signingKey: (( grab secrets.uaa_jwt_signing_key ))
           previous:
-            signingKey: (( grab secrets.uaa_jwt_signing_key_old || secrets.uaa_jwt_signing_key ))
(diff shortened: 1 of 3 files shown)


alext:
Tests failed because rubocop...

keymon:
@alext updated and pushed.

alext:
Confirmed that after rotation, the old key is retained as expected now. :+1:

PR #1198, opened 2018-01-31
Remove an older FIXME

camelpunch:
## What

Whilst approving the paas-usage-events-collector rename story, I noticed this FIXME lying around, with a comment asking for it to be reverted. It's a Concourse task that cleans up unused secrets.

## How to review

Ensure that this is still A Good Idea.

## Who can review

Anyone except @camelpunch.


alext:
Whoops, I think this is my bad...

PR #1202, opened 2018-01-31
[#154249348] Use ACM for public-facing TLS certs

keymon:
[#154249348 Use ACM for public-facing TLS certs](https://www.pivotaltracker.com/story/show/154249348)

What?
----

Following on from the spike (#142501669) we now want to proceed with the roll-out of Amazon Certificate Manager (ACM) for the public-facing certs.

In this PR we add the logic to use the ACM certificates using the terraform  ACM certificate datasource to automatically discover them. We also add
a temporary commit with logic to delete the old certificates, which can be reverted in the future.

Dependencies
------------

**IMPORTANT** We added the logic of creating the ACM certificates in alphagov/paas-bootstrap#129. That PR must be merged and applied first.

How to review
---------------

 1. Apply alphagov/paas-bootstrap#129
 2. Run this pipeline.
 3. Check that the certificates have been deleted:

```
DEPLOY_ENV=hector
aws iam list-server-certificates | grep ${DEPLOY_ENV}
aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/cf-certs.tfstate - # it shall fail
aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/cf-certs.tar.gz  - | tar -tvzf -  # It shall not include apps or system
```

You can try the code after reverting the temporary commit that deletes the old certs:
 1. Create a tmp branch, revert the last commit and push
 2. Rerun the pipeline.

Next steps.
---------

After merge, wait some time and revert the temporary commit.

Who?
----

Anyone but @keymon or @bandesz


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -406,7 +406,6 @@ jobs:
                 paas-cf/concourse/scripts/s3init.sh "((state_bucket))" google-oauth-secrets.yml paas-cf/concourse/init_files/zero_bytes
                 paas-cf/concourse/scripts/s3init.sh "((state_bucket))" cf.tfstate paas-cf/concourse/init_files/terraform.tfstate.tpl
                 paas-cf/concourse/scripts/s3init.sh "((state_bucket))" cf-certs.tar.gz paas-cf/concourse/init_files/empty.tar.gz
-                paas-cf/concourse/scripts/s3init.sh "((state_bucket))" cf-certs.tfstate paas-cf/concourse/init_files/terraform.tfstate.tpl
```
alext: Removing this line is going to cause the pipeline to stall in the `delete-old-cf-certs` task for new dev deployments. I think we should leave this in place until the cleanup PR.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1037,6 +978,104 @@ jobs:
...
+                    params:
+                      AWS_DEFAULT_REGION: ((aws_region))
+                      STATE_BUCKET: ((state_bucket))
+                    run:
+                      path: sh
+                      args:
+                        - -e
+                        - -c
+                        - |
+                          aws s3 rm "s3://${STATE_BUCKET}/cf-certs.tfstate"
```
alext: Likewise, removing this file is likely to cause future runs of the pipeline to stall.

concourse/pipelines/create-cloudfoundry.yml
keymon: :thinking: It is true for new deployments. I would do a different course of action, and completely remove the usage of a resource. I would implement this as a task that downloads the file using S3.

keymon:
@alext 
> Can we now remove the `CF_SKIP_SSL_VALIDATION` variable set for dev in the `Makefile`?

That is a good question and I have mix feelings. The variable is all over the place and that feature can be really useful when not using ELBs. For instance, we want to run the deployment on bosh-lite. Because that, I was thinking to simple set `CF_SKIP_SSL_VALIDATION=false` in dev

On the other hand, if it is not being used and tested, the feature will eventually be broken, and it would become obsolete code.

Other thing to take into account: we must keep it for the availability tests to work until the certs are replaced.

dcarley:
> On the other hand, if it is not being used and tested, the feature will eventually be broken, and it would become obsolete code.

👐

keymon:
>>     On the other hand, if it is not being used and tested, the feature will eventually be broken, and it would become obsolete code.

> :open_hands: 

OK, I would remove the CF_SKIP_SSL_VALIDATION then ;)

We can always review the git history if we want to add it back.

alext:
> Other thing to take into account: we must keep it for the availability tests to work until the certs are replaced.

> OK, I would remove the CF_SKIP_SSL_VALIDATION then ;)

Based on the issue with the availability tests, I think this should be done in a follow-up PR (it can be the same one that removes the cf-certs.tfstate cleanup.

keymon:
> Based on the issue with the availability tests, I think this should be done in a follow-up PR (it can be the same one that removes the cf-certs.tfstate cleanup).

I am preparing everything to be as easy as revert one commit.

keymon:
@alext I fixed all those comments, but I am fixing now the cloudfront configuration that relies on the system cert.

terraform/cf-certs/upload.tf
```diff
@@ -1,27 +0,0 @@
...
-  name_prefix       = "${var.env}-system-domain-"
-  certificate_body  = "${var.system_domain_crt}"
-  private_key       = "${var.system_domain_key}"
-  certificate_chain = "${var.system_domain_intermediate_crt}"
-  path              = "/cloudfront/${var.env}-system-domain/"
-
-  lifecycle {
-    create_before_destroy = true
-  }
-}
```
alext: Can we leave this terraform in place in this PR. I think it should remain in place until we remove the certs that this manages.

alext:
One comment above about retaining the terraform config for the moment. I think we should retain that, and the scripts to manually run it until we remove these certs.

Also, this needs a rebase onto master.

terraform/cf-certs/upload.tf
keymon: Did you check the PR #1211 where we add back the code in order to delete the certificate?

keymon: TBH I do not think it should be in place, if we remove it we remove it. We can revert the code or use the code in the branch #1211 to remove this dependency. 

Otherwise, following the same reasoning, we should also keep the generation and upload of the self signed certificates, and if we do so, we had basically removed lots of work from the story and delayed to the follow up story.

keymon: If we want to keep it, we should remove the commits that remove the generation & upload of apps&system domains: 43f221e404435a4bb36d8fb17ebc18467d6c5df4 and 53fcceb15f97c5357fe11b2bab82ff9bd5edacdd.

We can move them to #1211 or revert them and revert them back.

keymon:
Rebased.

terraform/cf-certs/upload.tf
alext: Thinking about this further, I think it's fine as is...

PR #1206, opened 2018-02-01
[#154760148] Unify post-deploy tasks

paroxp:
## What

At the moment, we have two places where some post-deploy tasks are run.
One is sort of inline with the cf-deploy and the other has a dedicated
separate task. We think this will get confusing as to where any future
tasks would come in.

Here, we're moving them back into cf-deploy as it requires less
refactoring to achieve what we need, which is running the tests in the
same time as cf-deploy and post-deploy tasks.

## How to review

- Run the pipelines and expect the `cf-deploy` task to succeed
- Make sure the tasks moved have been executed
- Make sure the tasks moved are in the right place

## Who can review

I'd like it to be someone from the initial pair that implemented a new task. /cc @chrisfarms @henrytk 

But I imagine anyone could pick that up.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -574,62 +574,6 @@ jobs:
             params:
               file: generated-cf-secrets/cf-secrets.yml
 
-  - name: post-deploy
-    serial: true
-    plan:
-    - aggregate:
-      - get: cf-tfstate
-      - get: pipeline-trigger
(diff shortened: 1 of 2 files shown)


paroxp:
Reworked everything. Tested in my dev - all works fine. Ready for re-review.

camelpunch:
:white_check_mark: cf-deploy is smaller
:white_check_mark: post-deploy is bigger
:white_check_mark: they both succeed

PR #1216, opened 2018-02-07
[Support] Change ES instance type to compute optimised

LeePorte:
## What

This PR changes the instance type used by Elasticsearch to a compute
optimised type as we are currently hitting the CPU limits of the general
purpose instance type.

## How to review

Code review and deploy this branch to dev

## Who can review

Not @LeePorte


alext:
It would be good to include some of the reasoning behind the choice of instance type in the commit message.

Also, any reason to pick a c4 over a c5?

LeePorte:
Updated commit message.

c5 forces the use of enhanced networking which we don't use elsewhere at present. I would like to test this further before using it. This would take it out of scope for a support task.

PR #1217, opened 2018-02-08
[#154959067] Add a system domain ELB

henrytk:
## What

We want to be able to serve UAA, the tech docs, and product page websites
from a single ELB.

The configuration for the ELB is based on the existing UAA ELB, because
it has been tuned to avoid UAA downtime during deploys.

We use a wildcard DNS for the following reasons:

* it is a simpler configuration with less Terraform to manage
* it allows for a two-phase deploy where we remove UAA subdomains later
* we want this to be the default route for system domain traffic
* if we change our mind on the previous point any new domains will
  take precedence over the wildcard, as they would be more specific.

We do not remove DNS entries here or take down unnecessary
infrastructure (such as the UAA ELB or the CloudFront distributions) as
they can be done in follow-up deployments to avoid downtime.

## How to review

* Deploy
* Check all tests pass
* Check wildcard DNS works:

  ```
  # Get the system ELB domain name:
  dig *.<system_domain>

  # Use it to send traffic to UAA via the new ELB
  curl -k -H "Host: uaa.<system_domain>" https://<system_elb_domain_name>/login
  ```

## Who can review

Anyone but me.


terraform/cloudfoundry/elbs.tf
```diff
@@ -93,6 +93,57 @@ resource "aws_proxy_protocol_policy" "cf_uaa_haproxy" {
   instance_ports = ["443"]
 }
 
+resource "aws_elb" "cf_system" {
```
alext: Can we name this something like `cf_router_system_domain` to make it clear that it's loadbalancing the routers.

terraform/cloudfoundry/elbs.tf
```diff
@@ -93,6 +93,57 @@ resource "aws_proxy_protocol_policy" "cf_uaa_haproxy" {
...
+  name                        = "${var.env}-cf-system"
+  subnets                     = ["${split(",", var.infra_subnet_ids)}"]
+  idle_timeout                = 19
+  cross_zone_load_balancing   = "true"
+  connection_draining         = true
+  connection_draining_timeout = 20
+
+  security_groups = [
+    "${aws_security_group.cf_api_elb.id}",
```
alext: I'm not sure this is the correct security group as this will be serving non-public things (eg tech docs and product pages)...

This might need a bit of thought...

terraform/cloudfoundry/elbs.tf
henrytk: Sensible. Done.

terraform/cloudfoundry/elbs.tf
alext: Upon thinking about it, I think this is fine as it is. in prod the two security groups are both open to the world, and in dev I don't think it's a problem if all system domain things are restricted to the office.

terraform/cloudfoundry/elbs.tf
henrytk: I had forgot to update the ELB name and bucket prefix, but that is now done, squashed, and force pushed.

alext:
This looks good to me, but I had the same issue as you with the proxy protocol after the rename. I'm 99% sure this is caused by the rename, but I'm going to test a fresh deploy on top of master to verify this.

PR #1218, opened 2018-02-08
[#154959067] Remove UAA, Product Page, and Tech Docs DNS

henrytk:
## What

Now that we have a system domain ELB, we can switch the DNS over (the wildcard DNS is already in place).

## How to review

~~⚠️ make sure https://github.com/alphagov/paas-cf/pull/1217 has at least started deploying to prod before merging.~~ Done.

You must have an existing deployment that includes https://github.com/alphagov/paas-cf/commit/e664eaffd27d27bdfabfac9b649d78ea0fdbbc83

It would also be useful to have CloudFront Distributions in place for the product page and tech docs.

⚠️ the Terraform needs to be run manually _before_ the deployment happens, to ensure the system domain apex DNS record remains in place. This is true for all environments, so you will have to sit and nurse this through staging and prod.

### Product Pages/Tech Docs

* Use the master branch to deploy the CloudFront Distributions. This will set up DNS. 
* Check traffic is routing correctly
* Switch to this branch and run the same Terraform in plan mode. It should want to destroy two DNS entries. Apply the changes. 
* Check with `dig` that the domain names are resolving to the new ELB.
* Hit the apps with requests to make sure the domains and redirects still work.

### UAA

Run the pipeline and traffic should still route to UAA (all tests should pass).

## Who can review

Preferably Alex, but really anybody but me.

### DIFF
--- a/terraform/cloudfoundry/dns-records.tf
+++ b/terraform/cloudfoundry/dns-records.tf
@@ -38,6 +38,18 @@ resource "aws_route53_record" "apps_wildcard" {
   records = ["${aws_elb.cf_router.dns_name}"]
 }
 
+resource "aws_route53_record" "system_apex" {
+  zone_id = "${var.system_dns_zone_id}"
+  name    = "${var.system_dns_zone_name}."
+  type    = "A"
+
+  alias {
+    name                   = "${aws_elb.cf_router_system_domain.dns_name}"
+    zone_id                = "${aws_elb.cf_router_system_domain.zone_id}"
+    evaluate_target_health = true
+  }
+}
+
 resource "aws_route53_record" "apps_apex" {
   zone_id = "${var.apps_dns_zone_id}"
   name    = "${var.apps_dns_zone_name}."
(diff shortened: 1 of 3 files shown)


alext:
As discussed offline, I think we also need to address the cloudfront redirect for the base system domain as well. Otherwise that will still be using the old cert.

I think it should be possible to handle this in the same way that the existing product page redirects are handled. It might be as simple as adding another route to that redirect app.

terraform/cloudfoundry/dns-records.tf
```diff
@@ -38,6 +22,14 @@ resource "aws_route53_record" "cf_ssh_proxy" {
   records = ["${aws_elb.ssh_proxy.dns_name}"]
 }
 
+resource "aws_route53_record" "system_domain" {
+  zone_id = "${var.system_dns_zone_id}"
+  name    = "${var.system_dns_zone_name}"
+  type    = "CNAME"
```
alext: You can't put a `CNAME` at the root of a zone. You'll have to use an `ALIAS` here. See `cf_apps_apex` below for an example.

We should probably also call this `system_apex` for consistency.

henrytk: Good spot. Done.

alext:
This has now been deployed to staging and prod.

PR #1220, opened 2018-02-08
Fix pipeline pinning for acceptance tests.

alext:
## What

These 2 should be pinned on the output of post-deploy and the 2
availability tests. Otherwise they will start running even if the
availability tests fail.

These look to have been erroneously removed in the refactoring in
3344249

## How to review

* Verify this makes sense.
* Push the pipeline to a concourse and verify the pinning looks correct.

## Who can review

Not me.


keymon:
LGTM!

PR #1221, opened 2018-02-12
[SUPPORT] Increase number of parser instances

LeePorte:
## What

We are increasing the number of parsers as we are seeing large queues of
log messages in production. This should give us a higher throughput and
increase fault tolerance.

## How to review

Code review

## Who can review

Not @LeePorte or @bandesz


alext:
From looking at the metrics, these VMs aren't maxed out on CPU. Why do we think that these are the bottleneck?

PR #1222, opened 2018-02-12
[SUPPORT] Add an additional IP address for each new parser

LeePorte:
## What

The parsers are set up with static IP addresses. This allocates another
one for the newly added instances.

## How to review

Code review

## Who can review

Not @LeePorte

### DIFF
--- a/manifests/cf-manifest/manifest/800-logsearch.yml
+++ b/manifests/cf-manifest/manifest/800-logsearch.yml
@@ -40,6 +40,7 @@ jobs:
   - name: cf
     static_ips:
       - 10.0.16.14
+      - 10.0.16.15
   properties:
     redis:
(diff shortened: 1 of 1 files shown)


manifests/cf-manifest/manifest/800-logsearch.yml
```diff
@@ -40,6 +40,7 @@ jobs:
   - name: cf
     static_ips:
       - 10.0.16.14
+      - 10.0.16.15
```
henrytk: How does this work in dev where we have just 1 instance, will it just ignore the extra static IP? \cc @alext @LeePorte

alext: Apparently not... We're going to revert these changes.

PR #1228, opened 2018-02-13
[#154959067] Redirect HTTP system domain traffic to HTTPS

henrytk:
## What

A regression was introduced (you're welcome) when we removed the
CloudFront distributions in front of the Product Page. This follows
the same pattern used by the apps domain ELB to defer the redirection
to HAProxy.

## How to review

* Deploy
* Check tests pass
* This should work:

```
curl http://www.${DEPLOY_ENV}.dev.cloudpipeline.digital -L
```

## Who can review

Anyone but me

### DIFF
--- a/platform-tests/src/platform/acceptance/plain_http_test.go
+++ b/platform-tests/src/platform/acceptance/plain_http_test.go
@@ -1,13 +1,17 @@
 package acceptance_test
 
 import (
+	"encoding/json"
 	"fmt"
-	"net"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
 )
 
 var _ = Describe("plain HTTP requests", func() {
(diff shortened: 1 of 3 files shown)


henrytk:
The custom acceptance tests will fail here because we were asserting HTTP traffic to the API fails:

```
------------------------------
• Failure [0.003 seconds]
	
plain HTTP requests
	
/tmp/build/3151afb2/paas-cf/platform-tests/src/platform/acceptance/plain_http_test.go:52
	
  to the API
	
  /tmp/build/3151afb2/paas-cf/platform-tests/src/platform/acceptance/plain_http_test.go:25
	
    has the connection refused [It]
	
    /tmp/build/3151afb2/paas-cf/platform-tests/src/platform/acceptance/plain_http_test.go:24
	
    should timeout
	
    Expected
        <bool>: false
    to be true
	
    /tmp/build/3151afb2/paas-cf/platform-tests/src/platform/acceptance/plain_http_test.go:23
------------------------------
```

the behaviour before the regression was that only the CloudFront requests were redirected to HTTPS.

platform-tests/src/platform/acceptance/plain_http_test.go
```diff
@@ -21,10 +21,14 @@ var _ = Describe("plain HTTP requests", func() {
...
-			Expect(err).To(HaveOccurred(), "should not connect")
-			Expect(err.(net.Error).Timeout()).To(BeTrue(), "should timeout")
+			req, err := http.NewRequest("GET", fmt.Sprintf("http://%s/v2/info", testConfig.ApiEndpoint), nil)
+			Expect(err).NotTo(HaveOccurred())
+			_, err = http.DefaultTransport.RoundTrip(req)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError(Or(
+				ContainSubstring("connection refused"),
+				ContainSubstring("connection reset by peer"),
+			)))
```
46bit: This relates to weirdness that @henrytk and I found. ELBs configured to not listen on port 80 still establish TCP connections on port 80, under some conditions. We've been talking to AWS support about it, and they seemed surprised themselves. Search your email inbox for `TCP connections to ports not covered by security groups` to read that AWS support thread.

Concourse gets `connection refused`. Running from our own computers gets `connection reset by peer`. I decided to allow both, as the connection being established but then immediately closed (`connection reset by peer`) is the behaviour that outside tenants will see.

46bit:
I'm done making changes. This is ready for re-review.

platform-tests/src/platform/acceptance/plain_http_test.go
dcarley: It'll be the transparent proxy in our office that's accepting the connections 🐼

46bit: 🤦‍♀️

samcrang: @46bit do you mind if I update the commit message with the nugget of @dcarley wisdom?

samcrang: I've taken an EXECUTIVE DECISION and done it.

46bit:
Following the decision we've made to not serve port 80 for the system domain, I'm going to close this PR and open another one.

PR #1232, opened 2018-02-16
[#153435666] Enable continuous smoke tests in dev

henrytk:
## What

We want to run continuous smoke tests in dev, because they can catch
errors caused by things outside of the normal pipeline run, such as
during credential rotation[1]. We use the existing pattern of checking
whether the healthcheck app is deployed to ensure they should be
running.

We use the standard Alpine container to check whether the smoke tests
need to run because this container has no special requirements and the
Docker Hub standard library[2] containers are officially maintained.

We use the govpaas-alerting-dev email address[3] because most people do
not subscribe to this and therefore reduces noise.

[1] https://www.pivotaltracker.com/n/projects/1275640/stories/153435666
[2] https://hub.docker.com/u/library/
[3] https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!forum/govpaas-alerting-dev;context-place=myforums

## How to review

* Have a kick-off for the story
* Make sure:
  * Smoke tests run in dev
  * You get an email if they fail (you can use `bosh stop api` to force quick failure).
* Code review to confirm they should run in staging and prod (we will get NO_DATA alerts if they don't).

## Who can review

Anyone but me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,11 +83,13 @@ globals:
 .PHONY: dev
 dev: globals ## Set Environment to DEV
 	$(eval export AWS_ACCOUNT=dev)
+	$(eval export PERSISTENT_ENVIRONMENT=false)
 	$(eval export ENABLE_BILLING_APP ?= false)
 	$(eval export ENABLE_DESTROY=true)
 	$(eval export ENABLE_AUTODELETE=true)
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
+	$(eval export ALERT_EMAIL_ADDRESS=govpaas-alerting-dev@digital.cabinet-office.gov.uk)
 	$(eval export SKIP_COMMIT_VERIFICATION=true)
 	$(eval export ENV_SPECIFIC_CF_MANIFEST=cf-default.yml)
 	$(eval export DISABLE_HEALTHCHECK_DB=true)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -83,11 +83,13 @@ globals:
 .PHONY: dev
 dev: globals ## Set Environment to DEV
 	$(eval export AWS_ACCOUNT=dev)
+	$(eval export PERSISTENT_ENVIRONMENT=false)
 	$(eval export ENABLE_BILLING_APP ?= false)
 	$(eval export ENABLE_DESTROY=true)
 	$(eval export ENABLE_AUTODELETE=true)
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
+	$(eval export ALERT_EMAIL_ADDRESS=govpaas-alerting-dev@digital.cabinet-office.gov.uk)
```
samcrang: Is this likely to get super-spammy? Can we pull out email addresses from an environment variable or sub-address this with our environment name so we can write email filters?

henrytk: It will have more spam than Monty Python at the Hollywood Bowl, but this is an email address made for spam. We don't expect developers to subscribe to this email group unless they are doing so for debugging purposes.

samcrang: /me unsubscribes from email group

alext: @samcrang Protip: you can remain subscribed to the group, but set it to not email you.

samcrang: I have it configure with a daily digest in case I ever need something to read.

PR #1234, opened 2018-02-16
[#154964608] Monitor Elasticache Redis instance count and parameter group count

keymon:
[#154964608 Monitor Elasticache Redis instance count and parameter group count](https://www.pivotaltracker.com/story/show/154964608)

## What

We need to monitor the number of Redis instances and  number of parameter groups in use. We might run out of these preventing tenants from being able to create new instances in a timely manner.

We asked Amazon about monitoring this in Trusted Advisor ([support ticket](https://console.aws.amazon.com/support/v1?region=eu-west-1#/case/?displayId=4843437621&language=en) (in prod account)). They advised that it's not currently possible to track this in Trusted Advisor, and recommend that we monitor this manually as a workaround. We'll need to manually keep track of the current limit in each AWS account to set the thresholds for the monitors.

We decided to implement an additional metric in our `metrics-exporter` service that would connect to AWS and query both nodes and cache parameter groups.

The limit in AWS is for nodes, both memcache and redis, which can be retrieved with the call elasticache:DescribeCacheClusters.

### Housekeeping

* We moved the different gauges into separate files. It's worth to review that commit separately.

## Dependencies

 - PR https://github.com/alphagov/paas-aws-account-wide-terraform/pull/115 shall be reviewed and merged first

## How to review

 - Code review
 - Test it:
   - Deploy the code with datadog enabled: `make dev pipelines BRANCH=$(git rev-parse --abbrev-ref HEAD)  SELF_UPDATE_PIPELINE=false ENABLE_DATADOG=true`
   - Go to datadog and check that the metrics aws.elasticache.* are there
   - check that the monitors for elasticache are configured

## Who can review it

Anyone but @bandesz or @keymon


46bit:
LGTM.

PR #1241, opened 2018-02-21
[#120268519] Add gorouter latency monitor to datadog

LeePorte:
## What

This sets up a gorouter datadog monitor with a warning threshold of
400ms and a critical threshold set at 800ms. These values have been
derived from reviewing the data held in datadog for the past month. If
we had them implemented during the load testing session then they would
have fired.

## How to review

Deploy from this branch with `ENABLE_DATADOG=true` and check the monitors for your deployment in datadog.

## Who can review

Not @LeePorte

### DIFF
--- a/terraform/datadog/router.tf
+++ b/terraform/datadog/router.tf
@@ -139,3 +139,21 @@ resource "datadog_monitor" "gorouter_healthy" {
 
   tags = ["deployment:${var.env}", "service:${var.env}_monitors", "job:router"]
 }
+
+resource "datadog_monitor" "gorouter_latency" {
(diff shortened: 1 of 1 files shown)


camelpunch:
This appears to work well under the somewhat contrived test conditions that I set up:

- Deploy
- `bosh ssh` to each router and `monit stop gorouter`
- DataDog monitor is triggered

LeePorte:
There is another monitor for gorouter not running. you might be better off hitting the healthcheck app with vagetta.

camelpunch:
@LeePorte hit it with my what now? Did I miss something?

LeePorte:
https://github.com/tsenart/vegeta

PR #1247, opened 2018-02-23
[#120268519] Monitoring gorouter latency

LeePorte:
## What

This commit increases the thresholds on the gorouter monitor in order to
avoid false positives.

The inital monor values were set too low and as a result were triggering
the alert in a flapping manner. This is not useful for us as a team.

## How to review

Code review should suffice.
If you wish you can deploy from this branch with `ENABLE_DATADOG=true`

## Who can review

Not @LeePorte

### DIFF
--- a/terraform/datadog/router.tf
+++ b/terraform/datadog/router.tf
@@ -143,16 +143,16 @@ resource "datadog_monitor" "gorouter_healthy" {
 resource "datadog_monitor" "gorouter_latency" {
   name                = "${format("%s gorouter latency", var.env)}"
   type                = "metric alert"
-  message             = "Gorouter latency too high."
+  message             = "${format("Gorouter latency too high. See: %s#Gorouter-high-latency-alerts", var.datadog_documentation_url)}"
   escalation_message  = "Gorouter latency still too high. Check the deployment."
   no_data_timeframe   = "7"
   require_full_window = true
 
-  query = "${format("avg(last_1m):avg:cf.gorouter.latency{deployment:%s,job:router} by {ip} > 800", var.env)}"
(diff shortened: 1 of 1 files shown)


alext:
Are we sure that these warnings are spurious? It could be that we already need to add router capacity.

I'd also note that as we're using a mean this can be distorted by small number of very slow requests. I think we may be better off using a median, or a percentile here.

dcarley:
> Are we sure that these warnings are spurious? It could be that we already need to add router capacity.

There's some context in the comments on https://www.pivotaltracker.com/n/projects/1275640/stories/120268519 from the first review/approval. I'm not super confident that the current routers are contended for resource and that adding another one will lower the response latency. One way to be more certain would be to do some load testing in dev but I'm not sure if it's worth the effort.

> I'd also note that as we're using a mean this can be distorted by small number of very slow requests. I think we may be better off using a median, or a percentile here.

I'm not sure we have that option. The monitor only supports avg/min/max/sum. It looks like the metric itself is already reported as an average - although I'm confused where it comes from, does gorouter publish to loggregator and datadog-nozzle picks it up?

dcarley:
I cloned the existing prod monitor, removed the tags so that it doesn't show on the dashboard, and changed the settings to match this PR: https://app.datadoghq.com/monitors#4253618?group=all&live=4h

The 10min smoothing brings it below our previous thresholds:
screen shot 2018-02-26 at 13 18 25

And still triggers for (what I assume was) load testing:
screen shot 2018-02-26 at 13 22 05

I think the new thresholds are probably alright, in that they're not too high, but I'm going to have a think about it for a bit.

PR #1253, opened 2018-03-01
[Support] Embiggen staging elasticsearch disks.

alext:
## What

Staging is [hovering around the 75%](https://app.datadoghq.com/metric/explorer?live=false&page=0&is_auto=false&from_ts=1514764800000&to_ts=1519948799999&tile_size=m&exp_metric=system.disk.in_use&exp_scope=deploy_env%3Astaging%2Cbosh-job%3Aelasticsearch_master%2Cdevice%3A%2Fdev%2Fxvdf1&exp_group=bosh-index&exp_agg=avg&exp_row_type=metric) mark at the moment, and has been for
nearly a month. Adding another 50% will give us some headroom until we
get to migrate away.

## How to review

Code review. Check my rationale makes sense.

## Who can review

Not me.

### DIFF
--- a/manifests/cf-manifest/env-specific/cf-staging.yml
+++ b/manifests/cf-manifest/env-specific/cf-staging.yml
@@ -3,5 +3,5 @@ meta:
   cell:
     instances: 6
   elasticsearch_master:
-    disk_size: 307400
+    disk_size: 460800
     instance_type: c4.2xlarge


manifests/cf-manifest/env-specific/cf-staging.yml
```diff
@@ -3,5 +3,5 @@ meta:
   cell:
     instances: 6
   elasticsearch_master:
-    disk_size: 307400
+    disk_size: 460800
```
46bit: I glanced at the metrics in Datadog:

* A 50% increase in disk should take the current 76% usage down to ~51%.
* We alert at 75% usage. That gives 24% headroom before the next alert.
* Disk usage has grown by 19% in 3 months, which is roughly 6.3% per month.
* A 50% increase in disk should take the growth rate down to 6.3/1.5 ≈ 4.2% per month.
* That 24% headroom will take 24/4.2 ≈ 5.7 months to use up.

This is fine provided we expect staging to be using Logit in August?

alext: I think that's fine. I would expect that we will be using Logit by then.

46bit:
Merged. This will hopefully deploy quickly, as since #1242 we no longer reallocate shards when taking our Elasticsearch nodes down for a rolling upgrade.

alext:
I'm afraid we still have to wait for all the data to be copied from the old disk to the new one.

PR #1261, opened 2018-03-05
[#155462110] Use alpine instead of bosh-cli where needed

keymon:
https://www.pivotaltracker.com/story/show/155462110

What?
----

We were using the bosh-cli docker image in some places where it was not
needed as we did not use the bosh cli.

In this commit we remove such dependency. Alpine is enough.

How to review?
------------

Code review. You can try to run the job.

Who?
---

Anyone but @keymon

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -130,7 +130,7 @@ jobs:
             type: docker-image
             source:
               repository: alpine
-              tag: "3.4"
+              tag: latest
           inputs:
             - name: paas-cf
           params:
@@ -145,7 +145,7 @@ jobs:
             type: docker-image
             source:
               repository: alpine
-              tag: "3.4"
+              tag: latest
           inputs:
             - name: bosh-CA
           outputs:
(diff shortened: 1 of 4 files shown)


concourse/pipelines/autodelete-cloudfoundry.yml
```diff
@@ -129,8 +129,8 @@ jobs:
           image_resource:
             type: docker-image
             source:
-              repository: governmentpaas/bosh-cli
-              tag: 465642da06051a55630d39c899697b678f66a7f7
+              repository: alpine
+              tag: "3.4"
```
henrytk: I don't think we should pin versions here. I know we do it in other places, but I think that is a mistake. Two reasons:

* We trust these standard libraries, so should be okay with whatever new versions they release
* There is maintenance involved in updating these tags
* We have no need for pinning, because we don't intend to run different versions of the same container like we do with our in-house images.

keymon:
I put 3.4 because it is what we use everywhere. I would then move to latest
in all the occurrences of alpine in a separated commit.

keymon:
@henrytk I moved to latest tags.

PR #1271, opened 2018-03-08
[#155463715] Move global manifest properties to specific jobs

keymon:
[#155463715 Move global manifest properties to specific jobs](https://www.pivotaltracker.com/story/show/155463715)

What?
----

In order to move closer to the manifests from [cf-deployment](https://github.com/cloudfoundry/cf-deployment) we want to move all the global properties
to each specific job in the manifest.

How to review
-------------

This change is meant to result in no real changes in the deployment,
a No-Op.

**To test the deployment**:

 1. Deploy from master (be sure this branch is rebased) 
 2. Deploy this branch
    - bosh would display a lot of changes in the properties diff, but then it would be a no-op, no VM would be updated at all.
    - all test should pass

**Verify the new manifest**

We have created a set of tool scripts in the repo https://github.com/keymon/paas-cf-release-reference/tree/master/scripts

You can check that the new manifest is equivalent by using the script  https://github.com/keymon/paas-cf-release-reference/blob/master/scripts/manifest_to_specific_properties.rb. This script, given a rendered cf-manifest.yml and a directory with all the releases spec (e.g. `/tmp/releases/capi-release/jobs/*/spec`) would generate a manifest with the properties in each job.

To use it:

```
git clone https://github.com/keymon/paas-cf-release-reference
cd paas-cf-release-reference
```

Generate the manifest based on master, and expand the variables into each specific job using the spec_job:

```
PAAS_CF_PATH=~/gds/paas-cf

( cd $PAAS_CF_PATH && git checkout master )

spruce merge \
    $PAAS_CF_PATH/manifests/cf-manifest/manifest/*.yml \
    <($PAAS_CF_PATH/manifests/cf-manifest/scripts/grafana-dashboards-manifest.rb $PAAS_CF_PATH/manifests/cf-manifest/grafana/) \
    <(ruby $PAAS_CF_PATH/terraform/scripts/generate_vpc_peering_manifest.rb "$PAAS_CF_PATH/terraform/prod.vpc_peering.json"
) \
    < $PAAS_CF_PATH/manifests/cf-manifest/stubs/datadog-nozzle.yml \
    < $PAAS_CF_PATH/manifests/cf-manifest/stubs/oauth.yml \
    > spruced-manifest.yml

RELEASES_DIR=./releases-spec/ ./scripts/manifest_to_specific_properties.rb ./spruced-manifest.yml > expanded-spruced-manifest.yml

```

The file `expanded-spruced-manifest.yml` will be our reference.

[...]

PR #1272, opened 2018-03-09
[#155630096] Fix rotation jobs

bandesz:
## What

I re-enabled the certificate and credentials rotation jobs.

Found the following issues:

* The smoke tests were passed the admin_user/admin_password variables but the config expects the user/password params instead. This meant the smoke tests were using the "admin" user and if the continuous smoke tests started during the rotation the failed login attempts eventually locked the admin user. (This was fixed in https://github.com/alphagov/paas-cf/pull/1271)
* The temporary test user is being deleted with the "admin" user but during a rotation this clean up can fail. Fortunately CF allows to delete a user with its own creds, so we'll self-destruct these test users.
* The UAA JWT signing keys were set with `default` and `previous` key ids in the configuration, but every issued JWT token contains the key id and can only be verified with that key. When we rotated the keys we updated the `default` key which caused all issued keys to be invalid and the `previous` key was never used (as no JWT token was ever issued with the `previous` key name). Instead of using `default` and `previous` we'll generate the SHA1 hash of these keys and use it as key id.

[1] https://github.com/cloudfoundry/uaa/blob/master/docs/UAA-APIs.rst#get-the-token-signing-key-get-token-key

## How to review

1. Run create-cloudfoundry from this branch:
    ```BRANCH=fix_rotation_jobs_155630096 make dev pipelines```

1. Run the credentials/rotate-cloudfoundry-credentials job.
   * You should be able to log in with the CF cli (scripts/cf_admin_login_dev.sh) after it

1. The continuous-smoke-tests job will fail until the cf-deploy finishes - this is expected.

1. Run the rotate-cf-internal-certs job and run the following script while running cf-deploy:
    ```while true; do date; cf curl '/v2/organizations/<admin org guid>/spaces?order-by=name&inline-relations-depth=1' | jq ".total_results"; sleep 1; done```

1. Run the delete-old-cf-internal-certs and run the same test

## Who can review

Not me.


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -279,7 +279,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-billing
-      tag_filter: v0.15.0
+      tag_filter: v0.17.0
```
alext: Did you mean to include this in this PR? I don't see how it's related to this (or #1271).

bandesz: It got in somehow when I rebased this branch from Hector's PR. A master rebase solved the issue - but this is still WIP.

PR #1273, opened 2018-03-12
[support] Replace concourse.ci domain in docs

dcarley:
## What

The website has moved to a new domain after the old domain as hijacked:

- https://www.cloudfoundry.org/blog/cve-2018-1227/

## How to review

Check that the domain matches the advisory and that I haven't missed any occurrences.

## Who can review

Anyone.


alext:
I've confirmed that this updates all referenced under our control. The only remaining references are in docs for the vendored concourse/atc library in the platform tests. I don't think these are important to update now as people are unlikely to be looking at our vendored version of the docs.

PR #1275, opened 2018-03-12
[#155936773] Use bosh-job instead of job in datadog queries

keymon:
What?
-----

Fix some issues caused in datadog after renaming the instance_groups:

 1. For the EC2 metrics monitors, after renaming the `instance_group` the VM did not get recreated and the AWS tag `job:` is still the old one. To solve it we use the tag `bosh-job:*` emitted by the datadog agent, as it will be always updated .

 2. The CC worker includes the `instance_group` in the process. We need to use a new process search include the job name from https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/24

How to review
-------------
 - Review first https://github.com/alphagov/paas-datadog-for-cloudfoundry-boshrelease/pull/24, wait for a new build in https://concourse.build.ci.cloudpipeline.digital/teams/main/pipelines/datadog-for-cloudfoundry-release and update the latest commit
 - Code review
 - deploy the new monitors, check that work.

Note: the monitor for CPU would not work because we do not have metrics from ec2 in development

Who?
---

Anyone but @keymon

### DIFF
--- a/terraform/datadog/aws.tf
+++ b/terraform/datadog/aws.tf
@@ -62,7 +62,7 @@ resource "datadog_monitor" "rds-failure" {
 resource "datadog_monitor" "ec2-cpu-utilisation" {
   name                = "${format("%s EC2 high CPU utilisation", var.env)}"
   type                = "metric alert"
-  query               = "${format("avg(last_1h):avg:aws.ec2.cpuutilization{deploy_env:%s,!job:diego-cell,!job:elasticsearch_master} by {bosh-job,bosh-index} > 90", var.env)}"
+  query               = "${format("avg(last_1h):avg:aws.ec2.cpuutilization{deploy_env:%s,!bosh-job:diego-cell,!bosh-job:elasticsearch_master} by {bosh-job,bosh-index} > 90", var.env)}"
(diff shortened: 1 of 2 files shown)


terraform/datadog/aws.tf
```diff
@@ -78,7 +78,7 @@ resource "datadog_monitor" "ec2-cpu-utilisation" {
 resource "datadog_monitor" "ec2-cpu-utilisation-es" {
   name                = "${format("%s EC2 high CPU utilisation - ElasticSearch", var.env)}"
   type                = "metric alert"
-  query               = "${format("avg(last_1h):avg:aws.ec2.cpuutilization{deploy_env:%s,job:elasticsearch_master} by {bosh-job,bosh-index} > 95", var.env)}"
+  query               = "${format("avg(last_1h):avg:aws.ec2.cpuutilization{deploy_env:%s,bosh-job:elasticsearch_master} by {bosh-job,bosh-index} > 95", var.env)}"
```
bandesz: Is there a reason to change this in this story? There are plenty of other `job:` filters that would need to be changed.

PR #1276, opened 2018-03-13
[#155463715] Move UAA oauth properties to the UAA job

bandesz:
## What

Previously we moved all global properties to the specific jobs. We tested everything on dev but we missed the oauth properties as those are only added in staging and prod.

❗️ we have to set up a new staging Google Ouath client before we merge this.

## How to review

Code review only.

I ran the following command and validated from the generated manifest, that the oauth config is properly added to the uaa job.

```
BRANCH=move_properties_to_specific_jobs_155463715_fix_oauth SELF_UPDATE_PIPELINE=false NEW_ACCOUNT_EMAIL_ADDRESS=andras.ferencz-szabo@digital.cabinet-office.gov.uk make dev pipelines
```

## Who can review

Not me or @keymon

### DIFF
--- a/manifests/cf-manifest/stubs/oauth.yml
+++ b/manifests/cf-manifest/stubs/oauth.yml
@@ -1,22 +1,26 @@
-properties:
-  login:
-    oauth:
-      providers:
-        google:
-          type: oidc1.0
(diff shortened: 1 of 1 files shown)


camelpunch:
LGTM. I guess this will become another task to remember in the Ops File story.

PR #1278, opened 2018-03-16
[#155592565] Bosh links phase 2

camelpunch:
## What

Part 2, in which we walk the other nats IP to the top of the static range.

## How to review

- Ensure https://github.com/alphagov/paas-cf/pull/1277 is merged and deployed first
- Rebase this on master. You should be left with one commit.
- Deploy on top of a dev env that had master applied
- Ensure no app availability tests failed

## Who can review

Not @camelpunch, @LeePorte or @bandesz.

PR #1279, opened 2018-03-16
[#155592565] Bosh links phase 3

camelpunch:
## What

Part 3, in which we remove the nats static IPs, adjust the static/dynamic ranges and start relying on implicit Bosh links to get addresses.

## How to review

- Ensure https://github.com/alphagov/paas-cf/pull/1278 is merged and deployed first
- Rebase this on master. There should be two commits left: a range adjustment and commit revert.
- Deploy ensuring that Bosh does nothing - it should 'choose' to use the existing nats IPs that are now in the dynamic range.

## Who can review

Not @camelpunch, @LeePorte or @bandesz.


samcrang:
I've merged #1278. I've rebased this off `master`.

PR #1281, opened 2018-03-19
[#156091458] Merge 000 and 010 CF manifests

camelpunch:
## What

Merge the 000 and 010 CF manifests. This brings us a little bit closer to [cf-deployment.yml](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml).

Adjust some whitespace to look a bit more like cf-deployment. Re-order some keys etc.

Remove a test that started failing. It asserted that the order matched upstream order. We'll soon be using upstream directly, so this won't be necessary.

## How to review

- Deploy master to dev env
- Deploy this to dev env
- Ensure bosh shows no changes

## Who can review

Anyone but @camelpunch


manifests/cf-manifest/spec/manifest/instance_group_spec.rb
```diff
@@ -61,20 +61,6 @@ def get_instance_group(instance_group_name)
...
-    end
-
-    it "has scheduler before the diego cells" do
-      expect("scheduler").to be_ordered_before("diego-cell")
-    end
-
-    it "has api before scheduler" do
-      expect("api").to be_ordered_before("scheduler")
-    end
-  end
```
alext: > We'll soon be using upstream manifests as a base, so the small aesthetic
> difference that ordering makes will be moot.

The reason for asserting on the ordering is not aesthetic. We had an incident a while back because we had the order different to upstream, and upgraded the cells before bbs. This test was added in de70ecd to help catch this in future.

I'm inclined to leave this test in place until we're actually consuming the upstream manifests directly.

camelpunch: :+1: OK I can re-order and re-instate the tests

keymon:
@camelpunch can you rebase this with master? all the previous PRs have been merged.

keymon:
LGTM, the diff is the same:

```
wget https://raw.githubusercontent.com/cloudfoundry/cf-deployment/master/cf-deployment.yml 

spruce diff <(spruce merge  <(git show master:manifests/cf-manifest/manifest/000-base-cf-deployment.yml) <(git show master:manifests/cf-manifest/manifest/010-cf-jobs.yml )) manifests/cf-manifest/manifest/000-base-cf-deployment.yml
```

But I still think that plain `diff` is not suitable to diff these files, it is much more convenient use `spruce diff`

PR #1283, opened 2018-03-20
[#154842336] Revert "Remove old route from the paas-billing app"

dcarley:
## What

This reverts commit 3d0bce5ef1a10e186e70c65424053f2d53fcb1e0.

The change has been deployed to production now.

## How to review

Code review. Make sure that I've reverted the right commit and code.

## Who can review

Anyone.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2195,11 +2195,6 @@ jobs:
 
                 cf push paas-billing
 
-                # TODO: remove after the new route was deployed to staging/prod
-                if cf app paas-billing | grep -q "paas-billing.${APPS_DNS_ZONE_NAME}"; then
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2195,11 +2195,6 @@ jobs:
 
                 cf push paas-billing
 
-                # TODO: remove after the new route was deployed to staging/prod
-                if cf app paas-billing | grep -q "paas-billing.${APPS_DNS_ZONE_NAME}"; then
-                  cf delete-route -f "${APPS_DNS_ZONE_NAME}" --hostname paas-billing
-                fi
-
```
alext: I don't think `APPS_DNS_ZONE_NAME` is used by this task any more, so it should probably be removed from `params`

dcarley: Good catch. I've pushed an additional commit to remove it.

keymon:
LGTM

PR #1284, opened 2018-03-21
[SUPPORT] Add new VPC peerings for DiT

LeePorte:
## What

This commit adds the new VPC peerings as requested via support.

## How to review

Code review and check against the details in ticket 9129

## Who can review

Not @46bit or @LeePorte


samcrang:
Looks legit

PR #1287, opened 2018-03-21
[#155856825] Switch DNS from CC ELB to system domain ELB.

henrytk:
## What

Now that we have the route for the API configured we can send traffic to it via the GoRouters.

## How to review

* deploy on top of https://github.com/alphagov/paas-cf/pull/1286
* make sure it doesn't redeploy any instance groups affected by the change of properties. Only default values were removed, so it should be a no-op.
* make sure all tests pass

### Before merging ⚠️ 

* merge https://github.com/alphagov/paas-cf/pull/1286
* rebase on master

## Who can review

Anyone but me.

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -503,17 +503,30 @@ instance_groups:
     provides:
       cloud_controller: {as: cloud_controller, shared: true}
     properties:
-      description: null
+      support_address: "https://docs.cloud.service.gov.uk"
+      router:
+        route_services_secret: ((secrets_route_services_secret))
       system_domain: ((terraform_outputs_cf_root_domain))
       system_domain_organization: admin
       app_domains:
       - ((terraform_outputs_cf_apps_domain))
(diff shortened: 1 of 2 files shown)


manifests/cf-manifest/manifest/000-base-cf-deployment.yml
```diff
@@ -1033,23 +988,20 @@ instance_groups:
...
           ca_cert: ((certs_bosh_ca_cert))
-          ssl:
-            port: 8443
-          port: 8080
           clients:
             cc-service-dashboards:
               secret: ((secrets_uaa_clients_cc_service_dashboards_password))
             cc_routing:
               secret: ((secrets_uaa_cc_routing_secret))
+          ssl:
```
bandesz: uaa.ssl.port can be removed

henrytk: This was left in on purpose due to consistency with upstream: https://github.com/cloudfoundry/cf-deployment/blob/v1.14.0/cf-deployment.yml#L971-L972 (upstream's inconsistency :) )

PR #1290, opened 2018-03-23
[#155856825] Revert: Switch DNS from CC ELB to system domain ELB.

bandesz:
## What

This reverts https://github.com/alphagov/paas-cf/pull/1287. 

We switched the CF API endpoint over to the cf_router_system_domain ELB but we overlooked the idle timeout which is much lower (19s) than it was on the API ELB (300s). The lower idle timeout caused `cf push` commands taking longer than 19s to be aborted.

We'll revert this change for now so we'll have time to figure out the right solution.

## How to review

Code review is enough. All the manifest property changes are a no-op as those are the default values in the job spec files.

The only real change is adding back the route 53 record for the API.

## Who can review

Not me.

### DIFF
--- a/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
+++ b/manifests/cf-manifest/manifest/000-base-cf-deployment.yml
@@ -503,30 +503,17 @@ instance_groups:
     provides:
       cloud_controller: {as: cloud_controller, shared: true}
     properties:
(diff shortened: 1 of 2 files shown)


keymon:
I think having the CC behind the gorouter would also cause API downtime when we restart the gorouters. We should check that.

PR #1291, opened 2018-03-23
[#155463385] Use ops files

bandesz:
## What

Generate the CF cloud-config and manifest fully with `bosh interpolate` using ops-files and get rid of spruce.

Changes:
 - we converted every CF manifest except the base one to an ops file
 - we changed the dynamic manifest generator scripts (for Grafana dashboards and VPC peering security groups) to generate ops files
 - we merged the base cloud config files into one main file
 - we converted all extra cloud config manifests to ops files
 - now the generation is only one step (bosh interpolate) in the Concourse pipeline
 - we moved the generation commands for cloud-config and the manifest into script files so we can test the same exact commands. This will make testing safer but sacrifices some clarity in the pipeline
 - we added additional tests to cover all aspects of the manifest generation (e.g. whether we add VPC peering for prod)

## How to review

1. Deploy your CF from master
    ```
    git checkout master
    ENABLE_DATADOG=true SELF_UPDATE_PIPELINE=false NEW_ACCOUNT_EMAIL_ADDRESS=x@digital.cabinet-office.gov.uk make dev pipelines
    ```
1. Deploy your CF from this branch:
    ```
    git checkout use_ops_files_155463385
    BRANCH=use_ops_files_155463385 ENABLE_DATADOG=true SELF_UPDATE_PIPELINE=false NEW_ACCOUNT_EMAIL_ADDRESS=x@digital.cabinet-office.gov.uk make dev pipelines
    ```

    In the cf-deploy/update-cloud-config and cf-deploy/cf-deploy jobs you should see no manifest diff and only the NATS instances (unreleated issue) should be redeployed

## Who can review

Not @henrytk or me.


samcrang:
Works as described in dev environment.

PR #1293, opened 2018-03-26
Don't let app or API availability tests overlap

camelpunch:
## What

Set `serial: true` on app and API availability test jobs.

I frequently have to manually stop these tests to avoid false negatives.
We don't ever want them to run on top of each other.

## How to review

Run in dev env, think about the consequences. It means that jobs might queue up instead of overlapping.

## Who can review

Anyone but @camelpunch

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -710,6 +710,7 @@ jobs:
                 exit 1
 
   - name: app-availability-tests
+    serial: true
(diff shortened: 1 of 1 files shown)


keymon:
LGTM

keymon:
Note that with this change, if you trigger the pipeline twice, the job `pre-deploy` would fail because the api or app availability tests are not running with the right trigger ID. 

Because that, we still need to manually kill the corresponding tasks.

PR #1300, opened 2018-04-03
Use medium VM type for queue VMs

henrytk:
## What

This increases the queue VMs from t2.small to m3.medium instances.
We have been seeing the queue VMs maxing out on CPU and running out
of credits. The m3.medium instances offer twice the vCPU and a fixed
performance.

~~There is the option to change this just in prod, as we only see the
high stress on the queue VM CPU in that environment, but we do not
have many instances of this VM, so the environment parity seems more
valuable than the cost saving.~~

**Update:**
    This has been made into an environment-specific configuration because
    we anticipate our logging stack to be scaled in numerous ways as we
    grow, so there will be large cumulative cost savings.

## How to review

Code review should be enough. I have deployed it and it didn't cause chaos and destruction.

## Who can review

Anyone but me

### DIFF
--- a/manifests/cf-manifest/env-specific/cf-default.yml
+++ b/manifests/cf-manifest/env-specific/cf-default.yml
@@ -3,3 +3,4 @@ cell_instances: 3
 elasticsearch_master_disk_size: 307400
 elasticsearch_master_instance_type: m4.large
 parser_instance_type: c4.large
+queue_vm_type: small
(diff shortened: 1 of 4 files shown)


henrytk:
Actually, I am having second thoughts about whether we make it env-specific. We are potentially going to scale several aspects of our logging stack, so the cumulative cost differences could add up.

keymon:
Env specific sounds good :+1:

henrytk:
Changed to environment-specific configuration.

keymon:
LGTM

PR #1292, opened 2018-03-23
[#155854402] Adapt internal certificates to cf deployment bosh variables

keymon:
[#155854402 Restructure internal certs to match cf-deployment](https://www.pivotaltracker.com/story/show/155854402)

## What

We want to start using certificates as they do in the [cf-deployment](https://github.com/cloudfoundry/cf-deployment):
 - Use the same naming convention
 - With [certs generated using bosh variables](https://bosh.io/docs/variable-types.html)

Of course, this change must:
 - Be without downtime.
 - Still allow rotate the certificates without downtime.

In this PR we will:
 - rename all the certificates as in cf-deployment
 - start using bosh-interpolate variable generation (state kept in a separated `cf-vars-certs.yml` file)
 - Implement logic to migrate from the existing `cf-certs.tgz` to the new `cf-vars-certs.yml`
 - Adapt the certificate rotation scripts to rotate the certs in `cf-vars-certs.yml`
 - Allow rotate also the dedicated CA certs without downtime.
 - Use the rotation of CAs and certs to stop using the bosh CA everywhere, and start with dedicated CA certs as in cf-deployment

### About certificate rotation

We implement certificate rotation as follows: 

 - All CA certs have two variables `<cert_name>` and `<cert_name>_old`. In the manifest we use both CA certs concatenated.
 - To rotate, we follow these 3 steps:
    1. Rename the old CA certs from `<cert_name>` to `<cert_name>_old`.
    2. Generate new CA certs.
    2. Deploy. Both new and old CA are used everywhere.
    3. Generate new certificates for all the other certs.
    4. Deploy. Old certs and new certs would work because both old and new CAs are deployed.
    5. Delete all the `_old` certs.
    6. Deploy.

### About the bosh-CA to dedicated-CA migration

 - When migrating from `cf-certs.tgz` to `cf-vars-store.yml`, we will use the `bosh-CA` as the value of every single CA.
 - Then we will do a normal certificate rotation: Rotate CAs, deploy both, generate new certs, deploy, and delete the old ones.

## How to review

The first time you run this branch it shall be almost a noop except:
 - The `.instance_groups[diego-cell].jobs[rep].properties.diego.rep.server_cert` and `.instance_groups[diego-cell].jobs[rep].properties.diego.rep.server_key` change. 
   This is because now we use the same certificate for the `rep` client and server. 
 - `.instance_groups[router].jobs[gorouter].properties.router.ca_certs` is added. This is to allow gorouter to validate the certificate of UAA.

First deployment:

[...]

PR #1302, opened 2018-04-03
[#156313150] Remove migrate-cf-certs job, remove unnecesary bosh-CA certificate, unit test cert rotation

keymon:
[#156313150 Clean up work after migrating cert generation to bosh](https://www.pivotaltracker.com/story/show/156313150)

from [#155854402 Restructure internal certs to match cf-deployment](https://www.pivotaltracker.com/story/show/155854402)

What?
----

After merging alphagov/paas-cf#1292 and deploy it to production,
we have migrated the old cf-certs.tgz file to the new vars-store
file. The task migrate-cf-certs added to support a zero-downtime migration[1]
is not needed anymore

We also stop using stop using bosh-CA if not needed, after we stopped using it as CA of the [API and apps endpoints](https://github.com/alphagov/paas-cf/pull/1202) and for the [cf internal certs](https://github.com/alphagov/paas-cf/pull/1292)

Finally, we add a unit test to verify that every CA certificate used is modified to append the old certificate during rotation.

[1] b1abe8806bde01baa8ed3b3e20ae4af3722839bf

How to review?
-----------

 - Code review
 - Push the pipeline, it should work as usual
 - ~~Be sure that we migrated the certs in  prod and all the envs~~ (this is done)

Who?
----

Anyone but @keymon


manifests/cf-manifest/spec/support/manifest_helpers.rb
```diff
@@ -52,6 +52,38 @@ def fetch(key, default_value = nil)
...
+          recursive_inject(acum2, x2, new_path) { |acum3, x3, path3|
+            yield(acum3, x3, path3)
+          }
+        }
+      else
+        yield(acum, x, path)
+      end
+    end
+
+    def inject(acum)
```
bandesz: Can you please explain what is this used for and how is it used?

manifests/cf-manifest/spec/support/manifest_helpers.rb
```diff
@@ -68,20 +100,30 @@ def workdir
     Cache.instance.workdir ||= $workdir
   end
 
+  def manifest_without_vars_store
```
bandesz: This doesn't seem to be used anywhere

keymon: I added a test in `manifests/cf-manifest/spec/manifest/rotate_certs_spec.rb`, I forgot to push, sorry.

manifests/cf-manifest/spec/support/manifest_helpers.rb
```diff
@@ -68,20 +101,30 @@ def workdir
...
+    Cache.instance.manifest_without_vars_store ||= \
```
bandesz: Originally this used ```Cache.instance.manifest_with_defaults``` which caused the test to use the wrong manifest.
I added a separate variable.

PR #1308, opened 2018-04-09
[Firebreak] Periodically dump infrastructure state to bucket

chrisfarms:
## What

We're building a (somewhat) [live infrastructure map dashboard](https://github.com/alphagov/paas-infrastructure-map/pull/1) to get an overview of deployed components.

We create a job that can be run in the pipeline to create a snapshot of the state of what the BOSH deployment looks like (via the BOSH director API) and a dump some EC2/ELB info which gets written to the state bucket as JSON.

The state should not be too sensitive, but making it public should be considered leaking too much information about the platform internals.... So to allow the generated JSON state dump to be consumed by an external service, a pre-signed URL is created (valid for 1 week) and stored in a secrets file in the state bucket. A deployment pipeline (like build/ci) can fetch the presigned secret url and use it to fetch the state until it expires.

## How to review

Push the pipeline for this branch... you should see a new `health-snapshot` job in the `health` group for the `create-cloudfoundry` pipeline.

The job should be on a timer and should periodically create a dump of json state that you can access (assuming you have aws creds) via something like:

```bash
$ export SNAPSHOT_URL=$(aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/health-snapshot-secrets.yml - | awk '/snapshot_url/ { print $2 }')
$ curl $SNAPSHOT_URL
```

You could also test using: https://github.com/alphagov/paas-infrastructure-map/pull/1

## Who can review

Not @chrisfarms or @henrytk

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -57,6 +57,7 @@ groups:
     jobs:
       - continuous-smoke-tests
       - check-var-store-certs
+      - health-snapshot
   - name: credentials
     jobs:
       - rotate-cloudfoundry-credentials
@@ -253,6 +254,24 @@ resources:
       region_name: ((aws_region))
       versioned_file: cloud-config.yml
 
+  - name: health-snapshot-data
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: health-snapshot-data.json
+      initial_version: "-"
+      initial_content_text: "{}"
+
+  - name: health-snapshot-secrets
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: health-snapshot-secrets.yml
+      initial_version: "-"
+      initial_content_text: ""
+
   - name: deployed-healthcheck
     type: s3-iam
     source:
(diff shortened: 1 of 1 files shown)


paroxp:
Would you be able to resolve conflicts?

chrisfarms:
Rebased on master and added the new `initial_content` and `initial_version` attrs to the s3 resources (due to the removal of s3init.sh).

Have not tested a dev deployment.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3325,3 +3349,139 @@ jobs:
...
+            - name: vpc-tfstate
+          outputs:
+            - name: health-snapshot-data
+          run:
+            path: sh
+            args:
+              - -e
+              - -u
+              - -c
+              - |
```
paroxp: Is it worth moving to a separate file? My pipeline diff looks terrible - I cannot sanity check easily without viewing the source.

chrisfarms: Hmmm ... if the script was moved into a file within paas-cf then that would mean this job would have to have a dependency on the paas-cf resource which means it would have to check it out each time just to run the script which seems like overkill.... is that what you are suggesting or am I missing something in concourse that allows you to break things out into separate files without storing it in a resource?

paroxp: This is indeed what I was suggesting, however I'm not in the position to block the delivery on that.

paroxp:
Tested in my environment. The data is reachable. Good firebreak idea 👍

chrisfarms:
Woo 🎉

PR #1311, opened 2018-04-11
[Support] Increase prod cells to 27

alext:
## What

We're currently hovering around 32% advertised capacity remaining on the
cells, which is below the threshold we're aiming for in ARD021[1].
Adding 3 cells beings this back up to around 40%.

[1]https://government-paas-team-manual.readthedocs.io/en/latest/architecture_decision_records/ADR021-cell-capacity-assignment-2/

## How to review

## Who can review

Not me.

### DIFF
--- a/manifests/cf-manifest/env-specific/cf-prod.yml
+++ b/manifests/cf-manifest/env-specific/cf-prod.yml
@@ -1,5 +1,5 @@
 ---
-cell_instances: 24
+cell_instances: 27
 elasticsearch_master_disk_size: 3072000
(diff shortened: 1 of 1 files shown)


46bit:
I agree this would bring it up to ~40%. It might only buy us about 2 weeks—I'm okay with that if you are.

screenshot 2018-04-11 14 27 32

alext:
I think that's enough for now. It's not a big deal to add more when they're needed. No point in paying for them before they're needed.

PR #1312, opened 2018-04-11
[#155783706] Add opsfile to create paas-admin UAA client

keymon:
What?
-----

```
 ☻/ This is Bill.
/▌  Bill wants to deploy a application integrated with UAA.
/\  Bill creates a dedicated UAA client for this app.
    That's because Bill knows that federated login does not work
    without a client in UAA.

    Bill is smart

```

Be like Bill... review and merge this opsfile that would create a new UAA client so that our paas-admin app can be integrated with UAA and we can authenticate our tenants using the same creds than the platform.

How to review?
-------------

 - code review
 - tested in hector environment
 - as in https://github.com/alphagov/paas-admin#prerequisites

Who?
----

Anyone but @keymon or Bill

### DIFF
--- a/manifests/cf-manifest/manifest/operations/200-paas-admin-uaa-client.yml
+++ b/manifests/cf-manifest/manifest/operations/200-paas-admin-uaa-client.yml
@@ -1,0 +1,16 @@
+- type: replace
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaa/clients/paas-admin?
+  value:
+    override: true
+    authorized-grant-types: authorization_code,client_credentials,refresh_token
+    autoapprove: true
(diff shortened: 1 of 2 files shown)


keymon:
For the record: Fair point, but we are moving to cf-deployment approach, were our base manifest is the one in cf-deployment https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml and then we modify everything with opsfiles.

Actually, soon we will play a story to move everything that differs iin our base manifest from the [cf-deloyment.yml](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml to opsfiles)

keymon:
@paroxp be like Bill

PR #1322, opened 2018-04-13
[WIP] Do not import AWS creds from an expired STS token

46bit:
## What

See commit message for details.

## How to review

Try using this and check it works well.

## Who can review

Not @46bit

### DIFF
--- a/scripts/create_sts_token.sh
+++ b/scripts/create_sts_token.sh
@@ -52,13 +52,15 @@ generate_new_token() {
   chmod 600 "${TOKEN_FILE}"
   trap 'rm ${TOKEN_FILE}' ERR
 
+  echo "if [ \$(date +%s) -lt ${expires} ]; then" >> "${TOKEN_FILE}"
   aws sts get-session-token \
     --serial-number "${token_arn}" \
     --duration-seconds "${STS_TOKEN_DURATION}" \
     --output text \
     --query "[Credentials.AccessKeyId,Credentials.SecretAccessKey,Credentials.SessionToken]" \
     --token-code "${mfa_token}" | \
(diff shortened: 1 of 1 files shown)


alext:
I'm not sure I understand in what case the token would be stale. This script always calls the `delete_stale_tokens` function before generating a new token, so the token should never be stale after calling this.

I'm wary of having the generated token files contain more than export statements.

Have you tried instead using this script via a function similar to the one documented here: https://government-paas-team-manual.readthedocs.io/en/latest/team/managing_aws_access/#temporary-credentials

Personally, I use a variation on that pattern with the following function in my .bash_login:
```sh
aws_key() {
  if [ -z "${1}" ]; then
    echo "Usage: aws_key <target>" 1>&2
    return 1
  fi
  local target=$1
  export AWS_DEFAULT_REGION=eu-west-1
  export AWS_ACCOUNT=$target
  eval $(pass "gds/aws/${target}")
  "${HOME}/paas/paas-cf/scripts/create_sts_token.sh" && source "${HOME}/.aws_sts_tokens/${AWS_ACCOUNT}.sh"
}
```

46bit:
> I'm not sure I understand in what case the token would be stale. This script always calls the delete_stale_tokens function before generating a new token, so the token should never be stale after calling this.

When I started on PaaS, various constraints meant I was importing `source ~/.aws_sts_tokens/dev.sh` directly in my `.zshrc`. The script may delete that file, but if its expired credentials are already loaded into your environment then the `aws` calls will fail.

46bit:
I'm going to close this, and try to bring this problem up at a Knowledgeshare instead.

PR #1325, opened 2018-04-17
Redirect to Paas Admin on Login

carolinegreen:
## What

Redirect to Paas Admin as previously on login the user would have been directed to next steps on the product page.  However this content has now been moved to Paas Admin, so redirect to the new location of next steps.

Depends on https://github.com/alphagov/paas-product-page/pull/63

## How to review

Make sure the pages are redirected.

## Who can review

Anyone who isn't me or @paroxp


LeePorte:
https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/spec/manifest/base_properties_spec.rb#L122 will need updating for the tests to pass.

PR #1327, opened 2018-04-19
WIP: [#154770854] Enable the Redis service for everyone.

alext:
## What

We want to make this generally available. This enables access to the
broker for all tenants on the platform.

With this being generally available, there's no longer a need to
explicitly enable access for the tests.

## How to review

* Deploy to a dev env
* Verify that the redis service is available in all orgs

## Who can review

Not me.


henrytk:
While looking at this I noticed the elasticsearch/tiny plan and the redis/tiny-clustered plan are both enabled to all, but no `cf enable-service-access` is mentioned in our codebase for them. Were these enabled manually? If so, we should probably move to a consistent approach on this.

PR #1328, opened 2018-04-19
[#148234139] Deploy cf-networking-release (enable private applications)

keymon:
[#148234139 Deploy cf-networking-release (enable private applications)](https://www.pivotaltracker.com/story/show/148234139)

What
----

We want to use cf-networking-release to enable private networking between
containers and apps in the platform.

In this PR:
 - We copy the config from cf-deployment, adapting it to use RDS, our own
   secrets, CA rotation, etc
 - Adapt the tests for the new feature.
 - Enable the related tests from acceptance test.
 - Add test to avoid vpc peering overlap.

How to review
-------

Code review. Run the deployment and acceptance tests.

You can test the private networking https://docs.cloudfoundry.org/devguide/deploy-apps/cf-networking.html#add-policy

Who can review
------
Anyone but @henrytk or @keymon


keymon:
While deploying from master I had one error in the API availability tests, but I do not think it was related.

alext:
The admin user in dev environments doesn't get updated to have the additional permissions. This is because the `sync_admin_users` job is skipped in dev environments.

alext:
Scratch that. It does get the permissions, I had to do a fresh login to pick them up though.

alext:
I've been testing rotating the certificates following [the team manual](https://government-paas-team-manual.readthedocs.io/en/latest/team/rotating_credentials/#rotating-certificates), and I've run into an issue after the first redeploy (after rotating the CAs). Bosh errored with:

```
Task 185 | 10:59:43 | Error: 'diego-cell/752b03a9-d9b3-4c68-8a4b-82118b308abc (0)' is not running after update. Review logs for failed jobs: vxlan-policy-agent
```

On the VM, the vxlan-policy-agent job isn't running. I see the following error in the logs:
```
{"timestamp":"1524483809.405450344","source":"cfnetworking.vxlan-policy-agent","message":"cfnetworking.vxlan-policy-agent.policy-client-get-policies","log_level":2,"data":{"error":"http client do: Get https://policy-server.service.cf.internal:4003/networking/v1/internal/policies: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"networkPolicyCA\")"}}
```

Looking on the VM, I see that /var/vcap/jobs/vxlan-policy-agent/config/certs/ca.crt only contains a single cert, whereas I'd expect it to include 2 while rotation was in progress. I think the `instance_groups.diego-cell.jobs.vxlan-policy-agent.properties.cf_networking.vxlan_policy_agent.ca_cert` needs to be included in `010-cert-rotation.yml`.

keymon:
Well spotted, I though I fixed all those cases.

There is a test to check every CA which has a corresponding `_old` CA certificate, is used in the manifest as `((<name>_ca))((<name>_ca_old))`.

But in this case I did not even add `network_policy_ca_old` and the test pass.

Shall we add a test that checks that every `<name>_ca` certificate must have a corresponding `<name>_ca_old` one?

alext:
> Shall we add a test that checks that every <name>_ca certificate must have a corresponding <name>_ca_old one?

I think that's a good idea given we missed this one.

PR #1330, opened 2018-04-20
[#156925097] Fix CF API url used by destroy-cloudfoundry

46bit:
## What

See commit message. This should fix the `remove-healthcheck-db` and `remove-billing-db` tasks of the `destroy-cloudfoundry` pipeline's `delete-deployment` job.

I am not certain if this fully fixes the `destroy-cloudfoundry`. This works fine in my temporary dev environment `mikie`, but it [fails in my permanent `miki` dev environment](https://deployer.miki.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/18) even when I thought I'd cleared out RDS instances beforehand.

## How to review

* Code review
* Try it in your dev environment (use `BRANCH=fix-destroy-cloudfoundry-api-url-156925097 make dev pipelines` to push these custom pipelines.) Check the `destroy-cloudfoundry` entirely succeeds (you'll need to destroy any service instances you've created manually, but not the billing or healthcheck DBs)

## Who can review

Not @46bit

### DIFF
--- a/concourse/tasks/get-cf-cli-config.yml
+++ b/concourse/tasks/get-cf-cli-config.yml
@@ -15,6 +15,7 @@ run:
   path: sh
   args:
     - -e
+    - -u
     - -c
     - |
       VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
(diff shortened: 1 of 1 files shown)


concourse/scripts/pipecleaner_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 var _ = Describe("PipeCleaner", func() {
-	const runTimeout = 2 * time.Second
+	const runTimeout = 5 * time.Second
```
dcarley: Could you split this change out to a separate commit that explains why it was necessary?

46bit: Hrm, the build now goes green without that change. I've removed it.

46bit:
I removed the timeout change and the build is still green. Ready for re-review.

PR #1331, opened 2018-04-25
[#156729524] Enable the syslog drain test in cf-acceptance-tests.

bandesz:
## What

We originally disabled the test in [1] because it was incompatible with running CATS from a Concourse container (instead of a BOSH errand).

The test would have prevented us from breaking syslog drain functionality in the incident [2].

Also I increased the PipeCleaner timeout, because both @46bit and I had a timeout on Travis the last week.

[1] https://github.com/alphagov/paas-cf/commit/3a7d54a5c1f1e50283c16e99eb0a70e374984dbb
[2] https://www.pivotaltracker.com/story/show/156683943

## How to review

1. Deploy CF from this branch, the acceptance tests should pass. Also validate that there is a test saying `forwards app messages to registered syslog drains` really runs.

    ```BRANCH=enable_syslog_acc_tests_156729524 make dev pipelines```

## Who can review

Not me.

### DIFF
--- a/concourse/scripts/pipecleaner_test.go
+++ b/concourse/scripts/pipecleaner_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 var _ = Describe("PipeCleaner", func() {
-	const runTimeout = 2 * time.Second
+	const runTimeout = 5 * time.Second
 
 	var (
 		command *exec.Cmd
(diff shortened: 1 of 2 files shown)


henrytk:
Rebased on master because the cf-networking pull request also change the `platform-tests/upstream/run_acceptance_tests.sh` file.

PR #1333, opened 2018-04-27
WIP: [#154438043] Migrate to cf-deployment manifests

keymon:
[#154438043 Migrate to cf-deployment manifests](https://www.pivotaltracker.com/story/show/154438043)

What
----

We want to consume the base manifests from cf-deployment, and use opsfiles to add our customisations on top.

In this PR we will add cf-deployment as a submodule as a easy way to track the pinned version of the project upstream. This works fine for travis and local development.

In this PR we start using the upstream cf-deployment.yml and we add multiple opsfiles for the different customisations.

 - `./manifests/cf-manifest/operations.d`: numbered files with the naming convention `[0-9][0-9][0-9]-<type>-<description>.yml`. **All files in this directory are automatically included.

 - `./manifest/cf.manifests/operations/`: files added adhoc, e.g. optional features.

**We aim to achieve a NoOp deployment**, get exactly the manifest. We will fix any outstanding issue in follow up stories. Because that:

 - We include every single property, even defaults or not necessary or irrelevant:
   - Defaults are separated files (numbered by a 9) like `099-bosh-set-migrated.yml`. They can be removed later.
   - The ones in doubt are marked with a `TODO` or `FIXME`

 - We include changes like remove link consumes, add migrated_from, etc.

 - We even include some incorrect config in the manifest, e.g. log-api provides is set in the wrong part of the manifest.

**The `XX9-*.yml`* opsfiles**

There are multiple opsfiles numbered as `XX9`, which require follow up work and will eventually be deleted.

How to review
-------------

 - Code review
 - Unit test should pass
 - Deployment should work with a noop
 - You can compare the generated manifest using the script `./diff-manifest.sh`. There should be only differences related to `meta`, manifest version, `addons` and a pair of UAA clients for brokers and graphite.

Who?
----

Anyone but @keymon or @bandesz


manifests/cf-manifest/operations.d/220-cf-disable-locket.yml
```diff
@@ -0,0 +1,27 @@
...
+- type: replace
+  path: /instance_groups/name=diego-cell/jobs/name=route_emitter/properties/locks?/locket?/enabled?
+  value: false
+
+- type: replace
+  path: /instance_groups/name=scheduler/jobs/name=auctioneer/properties/locks?/locket?/enabled?
+  value: false
+
+- type: remove
+  path: /instance_groups/name=scheduler/jobs/name=tps/properties/capi/tps/watcher
```
alext: How is this related to locket? Does it really belong in this file?

manifests/cf-manifest/operations.d/320-cc-set-loggregator-port.yml
```diff
@@ -0,0 +1,5 @@
+---
+# TODO: this value is changed in an opsfile from cf-deployment: aws.yml
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/doppler?/port?
+  value: 443
```
alext: I wonder whether we would be better off not using the `aws.yml` opsfile as all it does is change this port for us to have to change it back. What do you think?

manifests/cf-manifest/operations.d/520-datadog-add-instance-group-checks.yml
```diff
@@ -0,0 +1,151 @@
...
+    name: datadog-consul-agent-client
+    release: datadog-for-cloudfoundry
+
+- type: replace
+  path: /releases/-
+  value:
+    name: datadog-for-cloudfoundry
+    sha1: 09f70cbd87ec6045d44aa346e23fd6dd9bf14bfa
+    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/datadog-for-cloudfoundry-0.1.23.tgz
+    version: 0.1.23
```
alext: This might be better at the top of this file.

manifests/cf-manifest/operations.d/320-cc-set-loggregator-port.yml
keymon: We added it back because maybe one day they might add some stuff to `aws.yml` and we would miss it.

manifests/cf-manifest/operations.d/520-datadog-add-instance-group-checks.yml
keymon: :+1:

manifests/cf-manifest/operations.d/320-cc-set-loggregator-port.yml
alext: Fair enough, that makes sense.

keymon:
>    I think it would be better to keep all the ipsec related stuff together in one opsfile.

:+1: 

>   It would be good to put a README in the operations.d directory to explain the naming/numbering scheme

I had mixed feelings about this. It is really easy to miss a README in a subdir, and I was not sure if it will add too much value. I will add it anyway.

keymon:
I resolved the comments above. Pending do a `git rebase -i --autosquash master` to autosquash the `fixup!` comments.

PR #1336, opened 2018-05-01
[#155399638] Add a third router to z3

keymon:
https://www.pivotaltracker.com/story/show/155399638

What
----

We want o add an additional router to production. The reasons are:

 - The router is a very critical component of the platform, and
   if it is unavailable it would immediatelly affect all the tenant
   traffic.
 - We recently had an incident where one of the gorouter processes
   stopped serving, jeopardising all the deployment availability.
 - During deployment, we have only one router running, which will have
   more load and be a single point of failure.

Adding an additional router will increase our HA capabilities, ensure
that there is always at least 2x routers during normal operations
(including deployments). The cost is minimal.

In this PR:
 - we add the new AZ available for the router network in cloud config
 - we make the number of router instances customizable,
   per environment, in the same way we do for the cells.
   So we will still use 2 instances in development.

How to review
-------------

Do code review, specially of the parameter to change the number of
instances for staging/prod, as it is different for
different environments.

Testing it:

 1. Deploy this branch. It contains a temporary commit to enable 3x
    routers in development.
 2. Check it deploys OK, test pass, etc.
 3. Check the 3rd router is there, it is in the ELB, etc.
 4. Remove the temporary commit.

Dependencies
------------

After merge, remove the tmp commit

Who can review
--------------

Anyone but @keymon


manifests/shared/spec/fixtures/terraform/cf.yml
```diff
@@ -7,6 +7,7 @@ terraform_outputs_cell2_subnet_id: subnet-98765432
 terraform_outputs_cell3_subnet_id: subnet-98765433
 terraform_outputs_router1_subnet_id: subnet-09876543
 terraform_outputs_router2_subnet_id: subnet-98765432
+terraform_outputs_router2_subnet_id: subnet-87654321
```
alext: typo: should be `router3`.

The fact that nothing failed would suggest that we have a gap in our tests.

PR #1337, opened 2018-05-01
[SUPPORT] Increase prod cells to 30

keymon:
What
----
We're currently hovering around 32% advertised capacity remaining on the
cells, which is below the threshold we're aiming for in ARD021[1].
Adding 3 cells beings this back up to around 40%.

Same than [2]

[1] https://alphagov.github.io/paas-team-manual/architecture_decision_records/ADR021-cell-capacity-assignment-2/#decision
[2] 5cf3323d5773a245dd9406f65824dde8342b0a99

How to review
-------------

Code review.

Who can review
--------------

not @keymon

### DIFF
--- a/manifests/cf-manifest/env-specific/cf-prod.yml
+++ b/manifests/cf-manifest/env-specific/cf-prod.yml
@@ -1,5 +1,5 @@
 ---
-cell_instances: 27
+cell_instances: 30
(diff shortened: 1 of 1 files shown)


keymon:
This will buy again 2 weeks only. Maybe we want to go for 33?

PR #1341, opened 2018-05-02
[#157219548] Configure logsearch parse to consume both queues

keymon:
https://www.pivotaltracker.com/story/show/157219548

What
----

We were configuring two different parsers consuming messages from
the specific redis queue of each AZ.

But there is no really a reason for this, and having this setup
can cause issues, as only one queue might start receiving messages
when the ingestors are redeployed and marked as unhealthy in the
ELBs. When this happens, one of the parsers will be really busy
and unable to cope with the backlog meanwhile the other is idle.

If the two parsers consume both queues, the load would be evenly
distributed across both of them and they will not be saturated.

Analysing the parser template[1], we found out that we might able
to define multiple redis "inputs" and override specific options in
each, specifically `host`.

With this change we
do not need to configure each parser differently and we can
merge the two instance groups `parser_z1` and `parser_z2` in
one `parser`.

[1] https://github.com/cloudfoundry-community/logsearch-boshrelease/blob/v205.0.0/jobs/parser/templates/config/input_and_output.conf.erb#L3-L21

**Note:** deploying this might cause somedowntime of the logs, but they will
be stored in the queue and processed later.

How to review
-------------

Code review. To test:

 1. Deploy this
 2. SSH into any parser and check the config `/var/vcap/jobs/parser/config/input_and_output.conf`. The parser logs report about the 2 inputs.
 3. Also check that the logs are still flowing Elasticsearch and available in kibana.

Who can review
--------------

Anyone but @keymon

### DIFF
--- a/manifests/cf-manifest/operations.d/710-logsearch.yml
+++ b/manifests/cf-manifest/operations.d/710-logsearch.yml
@@ -37,16 +37,23 @@
 - type: replace
   path: /instance_groups/-
   value:
-    name: parser_z1
+    name: parser
     release: logsearch
-    azs: [z1]
+    azs: [z1, z2]
     jobs:
     - name: parser
       release: logsearch
       properties:
         redis:
-          host: ((queue_static_ips_first))
+          host: dummy # This value would get overriden in each plugin below
         logstash_parser:
+          inputs:
+          - plugin: redis
+            options:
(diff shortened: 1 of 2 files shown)


alext:
Is it worth adding a `migrated_from` to make the deploy more efficient?

PR #1343, opened 2018-05-03
Add billing URL to PaaS Admin

paroxp:
What
----

We've broken our pipelines, as we're deploying an application that is
going to fail due to missing environment variable.

This should be addressed in already backlogged story that will enable
tagging for that application.

How to review
-------------

- See if the changes make sense
- I'd say run through pipelines, but it's a blocker for staging. Perhaps unblock the pipeline and test on staging?


dcarley:
Haven't tested, but makes sense and I'll watch staging.

PR #1344, opened 2018-05-03
[support] Add logsearch query for CC failed jobs alert

dcarley:
What
----

So that you can easily find out what jobs have failed. Similar to the CC
errors alert that's next to it.

I've removed the text about checking the API because I don't think it's
useful. Errors can happen even if CC API is fine. The error logs will
tell you the actual problem.

How to review
-------------

I haven't tested that this applies and I suggest you don't either because enabling DataDog takes too long. Check the changes for syntax and that the search query returns the right things in Logsearch.

Who can review
--------------

Anyone.


alext:

PR #1345, opened 2018-05-03
Remove redundant environment variables

paroxp:
What
----

We've made use of `/v2/info` to reduce the amount of environment
variables we need to provide to run the application. This is less error
prone, as we can simply read the same configuration from different
places.

We're also sneaking in a quick fix for the session secret which
previously has been set to default.

How to review
-------------

- See if it makes sense
- ⚠️ Merge after alphagov/paas-admin#44 ⚠️


alext:
Can I suggest that the fix for the session secret should be in a separate commit.

paroxp:
Separated. Good call.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2106,9 +2106,9 @@ jobs:
                       OAUTH_CLIENT_SECRET: "(( grab uaa_clients_paas_admin_secret ))"
                       API_URL: "https://api.${SYSTEM_DNS_ZONE_NAME}"
                       BILLING_URL: "https://billing.${SYSTEM_DNS_ZONE_NAME}"
-                      UAA_URL: "https://uaa.${SYSTEM_DNS_ZONE_NAME}"
                       NOTIFY_API_KEY: "${NOTIFY_API_KEY}"
                       NOTIFY_WELCOME_TEMPLATE_ID: "${NOTIFY_WELCOME_TEMPLATE_ID}"
+                      SESSION_SECRET: "session-(( grab uaa_clients_paas_admin_secret ))"
```
dcarley: Can you explain (in the commit message) what this change will do?

Will it invalidate people's existing sessions?

If it's important that this is set, should we prevent `paas-admin` from starting when it's not?

PR #1350, opened 2018-05-10
[#157190168] Cleanup manifests following cf-deployment migration (part 1)

alext:
What
----

When migrating to consume the cf-deployment manifests, we aimed to generate an output manifest that matched our existing one as closely as possible. To achieve this we added in several properties that weren't strictly necessary, or were set to the default values.

This is part 1 of the cleanup, and removes all the unnecessary properties that match the defaults from boshreleases etc. This should be a no-op deploy.

Part 2 will follow with more changes that will require changes to VMs. We've separated this into 2 PRs to make the review process easier.

How to review
-------------

Deploy this branch on top of a deployment from current master. Verify the deploy is a no-op.

Who can review
--------------

Not @bandesz or myself.


keymon:
I testing this on the latest stemcells update in [my env](https://deployer.hector.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/pipeline-lock/builds/181)

keymon:
Things we might include in this PR or next. But I am happy to merge this now.

 - https://github.com/alphagov/paas-cf/blob/e42d52c0593928b87cd32109c551a0d7ced5814a/manifests/cf-manifest/operations.d/300-consul.yml#L3-L7

- https://github.com/alphagov/paas-cf/blob/e42d52c0593928b87cd32109c551a0d7ced5814a/manifests/cf-manifest/operations.d/320-cc-add-cutoff-age.yml#L8-L11

- https://github.com/alphagov/paas-cf/blob/e42d52c0593928b87cd32109c551a0d7ced5814a/manifests/cf-manifest/operations.d/320-cc-add-quotas.yml#L41-L43

- Maybe all the file: https://github.com/alphagov/paas-cf/blob/manifest_cleanup_157190168_part1/manifests/cf-manifest/operations.d/329-cc-defaults.yml

 - https://github.com/alphagov/paas-cf/blob/e42d52c0593928b87cd32109c551a0d7ced5814a/manifests/cf-manifest/operations.d/360-doppler.yml#L17-L19

alext:
> Things we might include in this PR or next. But I am happy to merge this now.

I think all of those will trigger a redeploy. I've already got some of them in part 2, and I haven't finished looking through all the `FIXME`s yet.

PR #1361, opened 2018-05-18
Remove Cloudfoundry group from pipeline

alext:
What
----

This was identical to the 'All' group except for the pipeline-lock job
which isn't included. It therefore doesn't offer much value, so let's
remove it.

This also renames the 'all' group to 'depoly' because it doesn't contain
all jobs, so the name is misleading.

This is leftover from when this pipeline used to deploy both bosh and
cloudfoundry.

How to review
-------------

* Deploy the pipelines (`make dev pipelines`)
* Verify the new layout looks good and you can see/access everything

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1,25 +1,8 @@
 ---
 groups:
-  - name: all
+  - name: deploy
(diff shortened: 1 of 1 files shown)


keymon:
:+1:

PR #1366, opened 2018-05-23
[#153119909] Upgrade Ruby to 2.5

bandesz:
## What

Update Ruby version to 2.5

We decided to use the same (latest) Ruby version everywhere. We still used 2.2 in multiple places which reached EOL on 2018-03-31 (#1), so we decided to migrate to the latest stable version.

We had to upgrade govuk-lint as older versions didn't support Ruby 2.5. Thanks to the upgrade we've got new style check violations but we decided to ignore them (at least for now), as this would unnecessarily blow up the story and increase the risk.

By upgrading the cf-cli Docker images there is a side-effect as the CF cli was upgraded to 6.34.1 (https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/112) but we never updated the image tags in the pipelines. Probably it's better if we clean this up, so I applied some fixes (see commits) because of the CLI upgrade too.

#1 https://www.ruby-lang.org/en/downloads/branches/

## How to review

1. Create a new CF env from this branch

    ```
    BRANCH=upgrade_ruby_153119909 SELF_UPDATE_PIPELINE=false ENABLE_DATADOG=true make dev pipelines
    ```

1. Make sure all tests pass, including the ones in the health group.

1. Run the create-cloudfoundry pipeline again on this env or on your existing env.

1. Destroy the CF environment. (do the bosh cli tests from the paas-bootstrap PR first)

## How to merge

❗️ The PR contains a WIP commit which needs to be updated with the final Docker image tags.

## Who can review

Not me.


.rubocop.yml
```diff
@@ -4,13 +4,14 @@ AllCops:
...
-  TargetRubyVersion: 2.2
-Metrics/BlockLength:
-  # FIXME: Remove when upstream PR is released:
-  # https://github.com/alphagov/govuk-lint/pull/87
-  Enabled: true
-  Exclude:
-    - "**/spec/**/*"
+  TargetRubyVersion: 2.5
 Layout/IndentHeredoc:
-  # FIXME: Remove when we've upgraded to Ruby 2.3
```
henrytk: Now this has been removed we have linting errors which are silenced indefinitely. What are we going to do about these?

bandesz: I don't believe in strongly opinionated style checking, so you might need to ask someone else.

henrytk:
The 'check-var-store-certs' in the health group failed with:

```
Successfully pulled ruby@sha256:086c979c15e425e60061e36a242db7fd8c02be4cd6216b891b35de74ce4628ec.
	
./paas-cf/concourse/scripts/check-vars-store-certificates.rb:12:in `<main>': undefined method `each' for nil:NilClass (NoMethodError)
```

Just investigating now.

samcrang:
Updated the commit to use the proper Docker images.

PR #1371, opened 2018-05-29
[#157083954] Deploy Log Cache service

keymon:
https://www.pivotaltracker.com/story/show/157083954

What
----

We want to deploy log-cache https://github.com/cloudfoundry/log-cache to
transfer and cache tenants metrics. One of the harder problems with
collecting metrics (and logs) and exposing them to tenants is if/how to
cache them, transport them, drop them, and scale delivery of them to
authenticated tenants. Loggregator already solves this problem so it
would be nice to take advantage of that.

log-cache allows us to retrieve this metrics and cache them reducing the
overhead caused on the dopplers when there are several consumers. It is
proposed as a new addition to the loggregator for the normal log
shipping.

We tried this service in a previous spike [#156731599](https://www.pivotaltracker.com/n/projects/1275640/stories/156731599)
and it probably fits our needs.

In this PR we deploy the service, following the pattern in the latest
experimental opsfile in cf-deployment, which colocates the service
with the doppler instances.

We manually copy that opsfile (still not available in the cf-deployment version that we are tracking) and tune it for our deployment. We also had to add more config to override the links (`cloud_controller_ng` links). 

For testing, we will run the "apps" [cf-acceptance tests with log-cache enabled](https://github.com/cloudfoundry/cf-acceptance-tests/commit/923a0fe9335bf0a1e75286801e952f248491cd38), although a pair of tests had to been disabled as they were not working (reported upstream). 

Dependencies:
---------------

 1. Merge PR https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/129 first
 2. Wait for the new container https://hub.docker.com/r/governmentpaas/cf-acceptance-tests/ to be built
 2. Update the WIP commit with the right container tag

How to review
-------------

 - Code review
 - Tests should pass

Test it with the cf log-cache-cli plugin  https://github.com/cloudfoundry/log-cache-cli:

```
go get code.cloudfoundry.org/log-cache-cli/...
go build -o  $GOPATH/bin/log-cache-cli code.cloudfoundry.org/log-cache-cli/cmd/cf-lc-plugin/
cf install-plugin $GOPATH/bin/log-cache-cli
```

And then try log-cache:
 - tail the logs of any app `cf tail <appname>`
 - Get meta information of the metrics stored in log-cache: `cf log-meta`
Who can review
--------------

Anyone but @keymon or @paroxp

### DIFF
--- a/manifests/cf-manifest/operations.d/540-log-cache.yml
+++ b/manifests/cf-manifest/operations.d/540-log-cache.yml
@@ -189,6 +189,19 @@
     name: log-cache-scheduler
     release: log-cache
 
+# Add certification rotation support
+- type: replace
+  path: /instance_groups/name=doppler/jobs/name=log-cache-cf-auth-proxy/properties/uaa/ca_cert
+  value: ((uaa_ca.certificate))((uaa_ca_old.certificate))
+- type: replace
+  path: /variables/-
+  value:
+    name: log_cache_ca_old
+    options:
+      common_name: log-cache
+      is_ca: true
+    type: certificate
+
 # FIXME: Rename the doppler secrets to our legacy secrets variables
 - type: replace
   path: /instance_groups/name=doppler/jobs/name=log-cache-cf-auth-proxy/properties/uaa/client_secret
--- a/manifests/cf-manifest/operations.d/540-log-cache.yml
+++ b/manifests/cf-manifest/operations.d/540-log-cache.yml
@@ -1,0 +1,191 @@
+# FIXME: Copied from cf-deployment experimental opsfile in the latest
+# version of cf-deployment
+#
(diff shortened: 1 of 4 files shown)


manifests/cf-manifest/operations.d/540-log-cache.yml
```diff
@@ -0,0 +1,228 @@
...
+
+- type: replace
+  path: /instance_groups/name=doppler/jobs/-
+  value:
+    name: log-cache
+    properties:
+      egress_port: 8080
+      health_addr: localhost:6060
+      tls:
+        ca_cert: ((log_cache.ca))
```
alext: This CA cert (and 2 other below) are missing the `_old` variant, meaning that CA cert rotation currently breaks log-cache. These were missed because the existing test for CA certs doesn't account for access via the `cert.ca` variables.

I'll add a fix for this (and update the test) in #1379.

PR #1372, opened 2018-05-29
[#156224014] Don't disable the uaa.require_https flag.

bandesz:
What
----

The default is true upstream but we've set it to false in [1].

If require_https is disabled the Current-User cookie won't have the secure flag which came up in our latest pentest.

There is no reason to disable this anymore. Whatever caused issues previously it doesn't seem to be present anymore.

[1] https://github.com/alphagov/paas-cf/commit/9a4dda67ead70bd39d811cc373d8fd10dbb9df44

How to review
-------------

1. Deploy your CF from this branch

2. Make sure the tests pass. Log in to UAA on the web.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/cf-manifest/operations.d/330-uaa.yml
+++ b/manifests/cf-manifest/operations.d/330-uaa.yml
@@ -86,10 +86,6 @@
   value: 8
 
 - type: replace
-  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaa/require_https?
-  value: false
-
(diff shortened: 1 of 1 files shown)


alext:
This doesn't actually fix the secure flag for the Current-User cookie. It does fix it for the X-Uaa-Csrf cookie though.

alext:
I've updated the commit message to reference the correct cookie.

PR #1375, opened 2018-05-30
[#157470481] Rename service plans according ADR025

keymon:
What
----

We want to consolidate the naming convention for all the backing services
we offer, as it is described in ADR025: https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR025-service-plan-naming-conventions/

We also want to only offer encrypted plans and deprecate the non
encrypted ones. We will not remove them, but make them unavailable

To disable the plans, we add a script in the task that enable the broker to
specifically disable those.

How to review
-------------

 - Code review
 - Deploy this, let the tests pass
 - Check the marketplace as a normal (non admin) user:

```
cf create-user testuser password123
cf set-org-role testuser govuk-paas OrgManager
cf auth testuser password123
cf marketplace
cf marketplace mysql
cf marketplace postgres
cf marketplace redis
```

Who can review
--------------

Anyone but @keymon or @paroxp

### DIFF
--- a/platform-tests/src/platform/acceptance/mysql_service_test.go
+++ b/platform-tests/src/platform/acceptance/mysql_service_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
 	"github.com/cloudfoundry-incubator/cf-test-helpers/helpers"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/workflowhelpers"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gbytes"
@@ -16,7 +17,7 @@ import (
 var _ = Describe("MySQL backing service", func() {
 	const (
 		serviceName  = "mysql"
-		testPlanName = "tiny-unencrypted-5.7"
+		testPlanName = "tiny-5.7"
 	)
 
 	It("should have registered the mysql service", func() {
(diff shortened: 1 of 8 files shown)


platform-tests/src/platform/acceptance/mysql_service_test.go
```diff
@@ -16,7 +17,7 @@ import (
 var _ = Describe("MySQL backing service", func() {
 	const (
 		serviceName  = "mysql"
-		testPlanName = "Free"
+		testPlanName = "tiny-unencrypted-5.7"
```
chrisfarms: should this be the `tiny-5.7` plan instead otherwise won't it fail due to not being enabled?

chrisfarms: ...same comment for the postgres one I guess

keymon:
>>     You cannot disable access to a service plan for an org if the plan is currently available to all orgs. You must first disable access for all orgs; then you can enable access for a particular org.
> I was wondering if we need to revoke access from all orgs first ... but it seems to work as-is, so I guess not.

I do not know, but it seems to work indeed. I can add a temporary line to do it anyway.

platform-tests/src/platform/acceptance/mysql_service_test.go
keymon: fixed

alext:
> I was wondering if we need to revoke access from all orgs first ... but it seems to work as-is, so I guess not.

I think this relates to when you restrict access to a service by org (ie using the `-o` flag), not when you're restricting which plans are available. I think those 2 things are orthogonal.

PR #1379, opened 2018-05-31
[#157032509] Fix CA certification rotation

bandesz:
## What

We use the BOSH cli to generate certificates into a variable stare - based on the manifest variable definitons.
There is a known issue if you try to rotate the CA certificates outlined in [1].

The root of the problem is that when OpenSSL is configured to trust multiple CAs, and two of them have the same subject name, OpenSSL will only verify certificates against the first one (see [2] in OpenSSL code).

One solution for the CA certificate rotation problem is to set the Subject Key Identifiers so OpenSSL will be able to handle multiple certificates with the same subject name.

We patched the bosh-cli to generate certificates with an SKI. This will solve the problem long-term but the first rotation _would_ still cause downtime as OpenSSL is still not able to find the right CA unless all CAs with the same common name have SKIs.

To achieve the first rotation without downtime (when old CAs don't have SKIs but new CAs have it) we've created a Concourse job `credentials/regenerate-cf-certs-cas-with-skis` which needs to be run once (+ a CF deploy).

[1] https://docs.google.com/document/d/1vKxziTEvIgKHubukoyAGaJzGqrMBjun7JffbunlLBPg/edit#heading=h.wftyivqbaag4
[2] https://github.com/openssl/openssl/blob/49f6cb968ff63793f6671d9026fb2a7034dad79a/crypto/x509/x509_lu.c#L613-L617

## How to review

1. Review the following PRs first
    * https://github.com/alphagov/paas-config-server/pull/2
    * https://github.com/alphagov/paas-bosh-cli/pull/1
    * https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/131

1. Make sure you've updated for CF env from the latest master and this PR is rebased from master

1. Follow the `How to deploy instructions` on your dev env

1. Make sure all tests pass at every stage.

### 1. Regenerate the existing certificates with SKIs

1. Run the `credentials/regenerate-cf-certs-cas-with-skis`
1. Run create-cloudfoundry to deploy the updated CAs with SKIs
1. Run the `rotate-cf-certs-leafs` delete existing non-CA certs
1. Run create-cloudfoundry to generate new non-CA certs with SKIs against the new CAs and deploy them

### 2. Run the regular CA rotation process

1. Run rotate-cf-certs-cas to copy existing CAs into _old suffixes
1. Run create-cloudfoundry to generate new CAs and deploy them alongside the old CAs
1. Run rotate-cf-certs-leafs to delete existing non-CA certs
1. Run create-cloudfoundry to generate new non-CA certs against the new CAs and deploy them
1. Run delete-old-cf-certs to delete old CAs
1. Run create-cloudfoundry to remove the old CAs

## How to merge

❗️ The TMP commit needs to be udpated with the final Docker image tags for bosh-cli-v2.

1. Merge https://github.com/alphagov/paas-config-server/pull/2
1. Update and merge https://github.com/alphagov/paas-bosh-cli/pull/1
1. Build a final BOSH cli binary (instructions in the PR)
1. Update and merge https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/131
1. Update the Docker image tags in the TMP commit and merge this PR.

## Who can review

Not me

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -156,7 +156,7 @@ jobs:
             type: docker-image
             source:
               repository: governmentpaas/bosh-cli-v2
-              tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+              tag: c88f3e0b03558c987693fad3f180d9052b77342c
           run:
             path: sh
             args:
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -48,6 +48,7 @@ groups:
       - expire-aws-keys
       - rotate-cf-certs-cas
       - rotate-cf-certs-leafs
+      - regenerate-cf-certs-cas-with-skis
       - delete-old-cf-certs
 resource_types:
 - name: s3-iam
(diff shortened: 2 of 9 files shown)


alext:
I've tested the rotation process, and it all works well.

While doing do, I discovered a problem with the log-cache certificates that didn't include the old CA certs, so I've added a commit to address that.

I've also added a commit to check that all the certificates in the vars-store have a SKI present. This should help us catch any regressions around this.

I'm in the process of merging all the other PRs - I've merged config-server, and bosh-cli, and am not waiting for DockerHub to do it's thing...

keymon:
I am happy with the code related to test the `cert.ca` vs `cert_ca`  :+1:

PR #1383, opened 2018-06-05
[#157965597] Upgrade cf-deployment and loggregator

henrytk:
What
----

- We've upgraded to cf-deployment v1.38.0
- We've upgraded to the latest loggregator release to resolve multiple CVEs (see commit).

How to review
-------------

We've deployed this a couple of times to dev environments. You can do so too if you want, however it probably isn't necessary and you can just eyeball the changes and lean on staging.

Who can review
--------------

Not me. Not Henry.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -3403,7 +3403,7 @@ jobs:
         - get: deployed-healthcheck
         - get: cf-smoke-tests-release
         - get: paas-cf
-          passed: ['post-deploy']
+          passed: ['pipeline-lock']
         - get: cf-manifest
           passed: ['post-deploy']
         - get: cf-secrets
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -267,12 +267,8 @@ resources:
   - name: cf-acceptance-tests
     type: git
     source:
-      #FIXME: We have forked the acceptance tests due to:
-      # https://github.com/cloudfoundry/cf-acceptance-tests/pull/290
-      # Remove this once we are using a version of the acceptance
-      # tests that include the above fix.
-      uri: https://github.com/alphagov/paas-cf-acceptance-tests/
-      branch: cf1.28-allow-http-redirect
+      uri: https://github.com/cloudfoundry/cf-acceptance-tests
+      branch: cf1.38
 
   - name: cf-smoke-tests
     type: git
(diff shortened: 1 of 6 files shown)


alext:
This [introduces a new secret](https://github.com/cloudfoundry/cf-deployment/compare/cae6b134d5ac1ec57ee329d6f873f28565695e5c...683f5044781f600104f6cc9e5a4e13d200d35c77#diff-f39617caec1965e56248ce1bf556db4dR347) - the UAA encryption key passphrase. Can we add this to our secret generation scripts, otherwise we'll end up with a mixture of secrets generated by bosh and ones generated by our scripts. We should probably also blacklist this one in the secret rotation for now until we work out that process.

alext:
Note: we have [a story](https://www.pivotaltracker.com/story/show/155463404) to address moving all these secrets over to bosh generated ones.

henrytk:
@alext 

> otherwise we'll end up with a mixture of secrets generated by bosh and ones generated by our scripts.

In my opinion this is not a problem if we are not rotating it ourselves. It also means more work for this story and more stuff to undo when we move away from generating our own secrets.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3403,7 +3401,7 @@ jobs:
         - get: smoke-tests-timer
           trigger: true
         - get: deployed-healthcheck
-        - get: cf-smoke-tests
+        - get: cf-smoke-tests-release
         - get: paas-cf
           passed: ['post-deploy']
```
alext: I'm wondering whether we should unpin `paas-cf` here. As things stand the continuous smoke tests fail once `self-update-pipelines` has run. They only start passing again once `post-deploy` has completed. The error is:
```
missing inputs: cf-smoke-tests
```

The reason for this is that the task that's used comes from the `paas-cf` repo, and therefore (due to the pinning) doesn't match the inputs defined in the pipeline (the version of `paas-cf` being used is still referencing `cf-smoke-tests` instead of `cf-smoke-tests-release`).

I think we should fix this, otherwise we'll get spurious alerts in Pagerduty.

alext: I've added a commit to update the pinning. I'll get someone else to sanity-check it.

bandesz: I think it makes sense that we want to use the current version of paas-cf right after the pipeline update as we use three different tasks there. Otherwise the pipeline version would be different from the tasks and it could fail every time when we change them.

PR #1403, opened 2018-06-19
[#156164640] Create staging and prod London deployments.

alext:
What
----

This adds the targets etc for deploying the staging and prod environments in the London region.

How to review
-------------

As per the paas-bootstrap PR, other than code review, the best way to test this is probably to try deploying the London staging deployment. Given the fiddly nature of this process it's probably best to pair with @henrytk or myself while deploying.

* Perform review steps from https://github.com/alphagov/paas-bootstrap/pull/168 to get a deployer concourse
* checkout this branch
* Load staging AWS keys (if you haven't already)
* Checkout the credstore branches: https://github.com/alphagov/paas-credentials/pull/121 and https://github.com/alphagov/paas-credentials-high/pull/37
* Upload the secrets: `COMPOSE_PASSWORD_STORE_HIGH_DIR=/path/to/paas-credentials-high make stg-lon upload-all-secrets`
* Push the pipelines pinned to this branch: `BRANCH=london make stg-lon pipelines`, the `create-cloudfoundry` pipeline should trigger automatically
* While it's running, trigger the `generate-git-keys` job from the operator tab, and then add the generated ssh public key (included in the job output) to the deploy keys for paas-cf.
* Verify the pipeline runs to completion, and that the deployed environment works.
* Create the Pingdom checks: `make stg-lon pingdom ACTION=apply`

## After merging

* Checkout master
* Point the pipeline back at the master branch again: `make stg-lon pipelines`
* Deploy the `prod-lon` environment as above, except don't point the branch at london when pushing the pipelines.

Who can review
--------------

Not @alext or @henrytk

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ showenv: check-env ## Display environment information
 .PHONY: upload-datadog-secrets
 upload-datadog-secrets: check-env ## Decrypt and upload Datadog credentials to S3
 	$(eval export DATADOG_PASSWORD_STORE_DIR?=${HOME}/.paas-pass)
-	$(if ${AWS_ACCOUNT},,$(error Must set environment to staging/prod))
+	$(if ${MAKEFILE_ENV_TARGET},,$(error Must set MAKEFILE_ENV_TARGET))
 	$(if ${DATADOG_PASSWORD_STORE_DIR},,$(error Must pass DATADOG_PASSWORD_STORE_DIR=<path_to_password_store>))
 	$(if $(wildcard ${DATADOG_PASSWORD_STORE_DIR}),,$(error Password store ${DATADOG_PASSWORD_STORE_DIR} does not exist))
 	@scripts/upload-datadog-secrets.sh
@@ -212,7 +212,7 @@ upload-datadog-secrets: check-env ## Decrypt and upload Datadog credentials to S
 .PHONY: upload-compose-secrets
 upload-compose-secrets: check-env ## Decrypt and upload Compose credentials to S3
 	$(eval export COMPOSE_PASSWORD_STORE_DIR?=${HOME}/.paas-pass)
-	$(if ${AWS_ACCOUNT},,$(error Must set environment to dev/staging/prod))
+	$(if ${MAKEFILE_ENV_TARGET},,$(error Must set MAKEFILE_ENV_TARGET))
 	$(if ${COMPOSE_PASSWORD_STORE_DIR},,$(error Must pass COMPOSE_PASSWORD_STORE_DIR=<path_to_password_store>))
 	$(if $(wildcard ${COMPOSE_PASSWORD_STORE_DIR}),,$(error Password store ${COMPOSE_PASSWORD_STORE_DIR} does not exist))
 	@scripts/upload-compose-secrets.sh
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -77,13 +76,14 @@ list_merge_keys: ## List all GPG keys allowed to sign merge commits.
 .PHONY: globals
 PASSWORD_STORE_DIR?=${HOME}/.paas-pass
 globals:
-	$(eval export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION})
 	$(eval export PASSWORD_STORE_DIR=${PASSWORD_STORE_DIR})
 	@true
 
 .PHONY: dev
 dev: globals ## Set Environment to DEV
```
bandesz: is there is a reason why we don't call check-env here?

concourse/scripts/pipelines-cloudfoundry.sh
```diff
@@ -93,6 +93,7 @@ pipeline_name: ${pipeline_name}
 enable_destroy: ${ENABLE_DESTROY:-}
 self_update_pipeline: ${SELF_UPDATE_PIPELINE:-true}
 aws_account: ${AWS_ACCOUNT:-dev}
+makefile_env_target: ${MAKEFILE_ENV_TARGET}
```
bandesz: On line 76 and 82 we should replace AWS_ACCOUNT with MAKEFILE_ENV_TARGET

bandesz:
In tag-release.sh we use `GIT_USER="gov-paas-${AWS_ACCOUNT}"`, won't we have a git user per deploy env?

concourse/scripts/pipelines-cloudfoundry.sh
alext: Fixed.

Makefile
alext: Not particularly, It's checked lower down by all the other tasks that do actions (eg pipelines), so it's not strictly necessary here. It's also not something that's been changed in the PR.

I guess we could add it here for consistency with paas-bootstrap, but I'm not sure it's necessary.

alext:
> We have to discuss what to do with VPC peering, because currently we would apply the same peering config to eu-west-2, but the connecting VPCs can't have peering connections with VPCs having the same IP ranges.

I've added a commit to make the VPC peerings be per deploy-env. I'm not sure we really have a need to do cross-region VPC peerings as these are currently used for backing services where you want to keep the latency low. Also given the reason for creating the London deployments is to have the data stored in the UK, it makes sense that the tenants will also want to move their peered VPCs to the London region as well.

> I suggest we choose a different CIDR for the London VPC (e.g. 10.1.0.0/16), so we can still create the existing peering connections and that will make the migration for any existing tenant easier.

I don't think it's worth going this far. Changing the IP range of the VPC isn't a small change as that range is coded in lots of places. Parameterising that would be a large change, and not worth the effort until we have a clear need to do so.

alext:
> In tag-release.sh we use GIT_USER="gov-paas-${AWS_ACCOUNT}", won't we have a git user per deploy env?

I added a commit to change this to use the `DEPLOY_ENV`. I'd note that this is merely a cosmetic thing as it's merely the user part of the committer email for the tag that's created.

PR #1405, opened 2018-06-20
[#158386152] Add missing datadog notifications

bandesz:
## What

It seems we forgot to add the email notification for many alerts and we didn't get emails for alerts which should've been investigated, like the daily increase in the gorouter latency caused by a Verify perf test.

I didn't enable the email notifications for the DNS response time monitor as we know this monitor alerts a lot, so it would cause a lot of noise.

Also I removed the escalation_message field from the Datadog monitors. This field is only needed if we enable renotify_interval, but we don't use it (except for 2 Compose monitors).

## How to review

1. Code review

1. Deploy the Datadog monitors from this branch and check if they deploy cleanly (You can speed up review if you manually enable Datadog only for the datadog-terraform-apply in the pipeline)

## Who can review

Not me.


henrytk:
I've done a quick code review but not deployed this, because Andras had already done so and it would waste time in my opinion.

PR #1423, opened 2018-06-27
[#155463404] Bosh generate secrets cleanup

henrytk:
What
----

Remove unnecessary code after https://github.com/alphagov/paas-cf/pull/1422

How to review
-------------

* Deploy on top of https://github.com/alphagov/paas-cf/pull/1422.
* I don't think we needed the `do` block, so replaced it with a different indentation. Just to be sure, check the `generate-cf-secrets` task does not run in parallel with the rest of the job.

## Before merging ⚠️ 

Make sure https://github.com/alphagov/paas-cf/pull/1422 has made it through to prod before merging this.

Who can review
--------------

Not me or @keymon


samcrang:
Rebased.

PR #1433, opened 2018-07-03
[#158525694] Add a script for reporting org usage

henrytk:
What
----

The intention of this script is to provide a CSV output for analysing
whether some orgs are unused and can be deleted.

Example usage:

cf login ...
./scripts/org-usage-report.sh

How to review
-------------

This is the minimum scripting required to deliver the required information. In my opinion, the review should only be concerned with making sure the information is mostly accurate. There is little need for precision, because the output of this script is just advising us which orgs we should think about decommissioning.

Note: continuous-smoke-tests result in orgs, spaces, and apps disappeared during script execution, which occassionally creates erroneous data.

Who can review
--------------

Anyone but me or @LeePorte

### DIFF
--- a/scripts/org-usage-report.sh
+++ b/scripts/org-usage-report.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+set -eu
+
+UAA_TOKEN=$(cf oauth-token)
+UAA_API=$(jq -r '.UaaEndpoint' < ~/.cf/config.json)
+
+# Assert logged in
+cf target > /dev/null
+
+TARGET=$(jq -r '.Target' < ~/.cf/config.json)
+echo "Targetting CF API: $TARGET" 1>&2
+
+paginate() {
+  next_url=
+  page=1
+  rpp="$1"
+  path="$2"
+  while [ "$next_url" != 'null' ]; do
+    url="${path}?results-per-page=${rpp}&page=${page}"
+    organizations=$(cf curl "$url")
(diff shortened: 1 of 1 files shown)


samcrang:
I get a bunch of `jq: error (at <stdin>:1): Cannot iterate over null (null)` errors when running this but it appears(?) as though you still get the correct results if you `2>/dev/null`. Is this expected?

```
$ jq --version
jq-1.5
$ cf --version
cf version 6.37.0+a40009753.2018-05-25
```

henrytk:
@samcrang 

> I get a bunch of jq: error (at <stdin>:1): Cannot iterate over null (null) errors when running this but it appears(?) as though you still get the correct results if you 2>/dev/null. Is this expected?

I didn't investigate fully, but I am pretty sure it was the result of orgs, spaces, apps _et cetera_ disappearing due to continuously running smoke tests. If the data is correct I don't really care.

alext:
> I didn't investigate fully, but I think it was the result of orgs, spaces, apps et cetera disappearing due to continuously running smoke tests.

Is it worth adding some exceptions to skip the smoke/acceptance test orgs?

henrytk:
@alext 

> Is it worth adding some exceptions to skip the smoke/acceptance test orgs?

Possibly. I was conscious of how much time I was spending on this. The error output doesn't seem to affect the data produced, so the script has already delivered its value.

samcrang:
Imma just merge it—we can deal with making it betterer another time.

PR #1444, opened 2018-07-12
[#114110115] Fix error running

keymon:
What
----

In https://github.com/alphagov/paas-cf/pull/1422 we introduced a bug
that prevents terraform destroy from running.

That is because we have a variable with a `-` that fails when
converting it to `TF_VAR_` env vars.

We fix the script causing the issue to replace `-` with `_` but also
remove parsing the cf-vars-store file for terraform destroy as
it is not needed.

How to review
-------------

Push pipelines.

Run the destroy pipeline??? maybe that is too much

Who can review
--------------

not me


keymon:
Fixed ruby lint

concourse/scripts/extract_tf_vars_from_yaml.rb
```diff
@@ -23,7 +23,7 @@ def process_yaml(yaml_tree, prefix_chain)
       process_yaml(v, prefix_chain + [name])
     }
   else
-    puts "#{prefix_chain.join('_')}='#{yaml_tree}'"
+    puts "#{prefix_chain.join('_').tr('-', '_')}='#{yaml_tree}'"
```
alext: I wonder whether we should substitute all characters that aren't valid in shell variable names here...

Also, I think it's getting to the point where it's worth adding tests for this script.

keymon: > Also, I think it's getting to the point where it's worth adding tests for this script.

You know that I would also say that is the point to be more smart handling secrets and variables, rather than parsing and converting `.yml` files :) 

I did the very minimum. TBH the ruby script change is not needed anymore because we do not consume this file in terraform destroy from the other commit. but it was a simple change.

I think if we enable credhub, which seems something we will do soon, we will refactor how we do all this stuff. Being more ambitious, if we start deploying other tech stacks for containers, we might have to solve how to share secrets&variables, and likely we will refactor all of this too... 

So I wouldn't really add a test for this

tlwr: This unbroke stuff for me and looks legit. I'm going to merge this because it will stop terraform destroy from breaking.

Hector's suggestions seem sensible to the novice eye.

PR #1451, opened 2018-07-18
[#157737884] Deploy Aiven broker

henrytk:
What
----

This deploys the Aiven broker and modifies the Compose Elasticsearch acceptance tests to use Aiven instead.

How to review
-------------

* Merge https://github.com/alphagov/paas-aiven-broker/pull/5 and update the temporary commit on this branch.
* Upload the Aiven secrets:

  ```
  make dev upload-aiven-secrets
  ```

* Deploy from this branch. Make sure the custom acceptance tests pass.

### Before merging ⚠️ 

* Merge https://github.com/alphagov/paas-aiven-broker/pull/5 and update the temporary commit on this branch.
* Aiven secrets need to be uploaded for each environment.

Who can review
--------------

Anyone but me or @jpluscplusm

### DIFF
--- a/config/service-brokers/aiven/config.json
+++ b/config/service-brokers/aiven/config.json
@@ -1,0 +1,42 @@
+{
+  "port": "8080",
+  "log_level": "info",
+  "catalog": {
+    "services": [
+      {
+        "id": "1b45c99b-c90d-45b8-918d-9fb7dcb4beec",
+        "name": "elasticsearch",
+        "description": "Elasticsearch instances provisioned via Aiven",
+        "bindable": true,
+        "plan_updateable": true,
+        "metadata": {
+          "displayName": "Elasticsearch",
+          "providerDisplayName": "GOV.UK PaaS",
+          "supportUrl": "https://www.cloud.service.gov.uk/support"
+        },
+        "plans": [
+          {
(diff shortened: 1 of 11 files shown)


platform-tests/example-apps/healthcheck/elasticsearch.go
```diff
@@ -26,8 +26,7 @@ func elasticsearchHandler(w http.ResponseWriter, r *http.Request) {
 
 func testElasticsearchConnection(tls bool) error {
 	var credentials struct {
-		URI          string `json:"uri"`
-		CACertBase64 string `json:"ca_certificate_base64"`
+		URI string `json:"service_uri"`
```
alext: Is there a reason this is different from all the other brokers? I think it would be good for this to be consistent and use the `uri` key in the credentials JSON. This is also what the documentation describes here: https://docs.cloudfoundry.org/services/binding-credentials.html

config/service-brokers/aiven/config.json
```diff
@@ -0,0 +1,43 @@
+{
+  "port": "8080",
+  "log_level": "info",
+  "cloud": "aws-eu-west-1",
```
alext: I don't think we should include this here any more now that this config value is taken from the environment, it will only cause confusion.

PR #1454, opened 2018-07-19
[#157737884] Add check for Aiven API token in pipelines script

alext:
## What

This means that the pipeline will fail early if the API token hasn't
been uploaded, instead of waiting until the post-deploy task before
failing.

How to review
-------------

* Run `make dev pipelines` without aiven secrets uploaded, and verify you get a helpful error message.
* Upload the secrets (`make dev upload-aiven-secrets`), and repeat. Verify it now works.

Who can review
--------------

Not me.


46bit:
The code looks good to me having done something similar in #1329, but don't have a dev env to try it in practice.

PR #1456, opened 2018-07-20
[support] Increase ElastiCache parameter group number

LeePorte:
What
----

This commit adds in the variables with the correct values that are used
for calculating the ElastiCache alerts in datadog.

How to review
-------------

Code review against patterns use in other env vars

Who can review
--------------

Not @LeePorte


LeePorte:
Yes raised in both via support

PR #1465, opened 2018-07-31
[#156911499] paas-admin acceptance tests accepts all aggrements

keymon:
What
----

When running the paas-admin acceptance tests, we create test users.
These users must have any agreement accepted before we can do
testing of the paas-admin app.

Dependency
---------

https://github.com/alphagov/paas-admin/pull/82=

How to review
-------------

Code review, and test should pass.

Who can review
--------------

Anyone


keymon:
I already tested it, if you trust me, a mere code review is enough. The change is fairly safe to merge anyway, any issue would be detected at staging.

PR #1469, opened 2018-08-02
Alert on 25+% swap utilisation

46bit:
What
----

See commit message.

How to review
-------------

I suggest code review and I'll fix any problems in staging.

Who can review
--------------

Who can review.


alext:
This check looks good to me. I don't think it belongs in this file though as it's not specific to aws, more a general monitor for any VM. I think it would be better placed in `general.tf` and named something like `vm-swap-utilisation`.

Having said that, I think there are other checks in there that also don't really belong either...

46bit:
@alext I did the thing. Ready for re-review/merge.

PR #1471, opened 2018-08-06
[#159136450] Use Bosh DNS

LeePorte:
What
----

We have enabled Bosh DNS using the 1.8 release. This is an essential precursor to enabling service discovery. Service discovery is not enabled yet, this only changes the DNS mechanism from Consul to Bosh DNS. Consul has not been removed as part of this work - this will be played in a subsequent story.

Since we'll be using bosh-dns in the future and we only have monitoring for consul we've also added a simple process check to check that the bosh-dns process is running on all of the machines that we used to check the consul agent on.

How to review
-------------

1. Code review
1. Follow the steps in https://github.com/alphagov/paas-bootstrap/pull/178
1. Deploy to a dev environment with DATADOG_ENABLED=true
1. Check that the `$yourEnv Bosh-dns process running` monitor is in datadog 🐶 

Who can review
--------------

Not @LeePorte or @richardTowers


alext:
> Due to our network configuration we are unable to use upstreams opsfile,
this is due to the expectation that all VMs are within the same network
name.
>
> As a result the bosh interpolate is failing to select the correct
network name for each instances location.

Could we not use the upstream opsfile, and then add an additional opsfile to do the additional renaming necessary. I think this would make it much clearer how we differ from upstream, and would also insulate better us against future upstream changes.

There's also noting in the commit messages documenting why we're using 1.8.0 as opposed to 1.6.0 that's included in upstream. This is the sort of thing that's very important to track the why for. Ideally, this would be bumped in a separate commit along with the explanation.

keymon:
I agree on this. I suggest for do a PR to address these two comments, explaining it accordingly.

The change shall be no-op

LeePorte:
The reason for bumping to 1.8 was due to the following;

Fixed in 1.7
No longer return unchecked/unknown instances in results for healthy instances

Fixed in 1.8
Fix an issue where bosh-dns entries would stop resolving after 1000 bosh-dns entries had been checked for healthiness

These got lost in the squashing.

PR #1476, opened 2018-08-09
Increase the collection interval

paroxp:
What
----

We're now scraping a lot of metrics every 1 minute. We presume that's
OK, however this includes querying the AWS CloudWatch, which is really
expensive. For the time being in order to save some moneys, we'd like to
run the queries every 10 minutes. A downside is that at the current
state of an application it has to apply these for all metric collectors.

This should be a temporary change, and ideally be reverted back and
sorted in a different way.

How to review
-------------

- Make sure the change makes sense

I haven't tested this. Base on https://github.com/alphagov/paas-rds-broker-boshrelease/blob/master/jobs/rds-metric-collector/spec#L40-L42


paroxp:
Updated the story.

Force pushed to make travis pass.

keymon:
LGTM. To add a different interval per collector shouldn't be difficult.

PR #1479, opened 2018-08-10
Appease newer versions of Shellcheck.

blairboy362:
What
----

Once shellcheck has been installed to a version newer than 0.4.6 a bunch
of new alerts started getting thrown. Make those checks pass.

How to review
-------------

1. Code review
1. Run tests locally (with shellcheck > 0.4.6)
1. Run it through your dev pipeline

Who can review
--------------

Anyone but me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bin/terraform*
 platform-tests/pkg
 platform-tests/src/results
 terraform/.terraform/
+.venv/
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ bin/terraform*
 platform-tests/pkg
 platform-tests/src/results
 terraform/.terraform/
-.venv/
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint_terraform: dev ## Lint the terraform files.
 	@terraform/scripts/lint.sh
 
 lint_shellcheck:
-	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' | xargs shellcheck
+	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' -not -path './.venv/*' | xargs shellcheck
 
 lint_concourse:
 	cd .. && SHELLCHECK_OPTS="-e SC1091" python paas-cf/concourse/scripts/pipecleaner.py --fatal-warnings paas-cf/concourse/pipelines/*.yml
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint_terraform: dev ## Lint the terraform files.
 	@terraform/scripts/lint.sh
 
 lint_shellcheck:
-	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' -not -path './.venv/*' | xargs shellcheck
+	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' | xargs shellcheck
 
 lint_concourse:
 	cd .. && SHELLCHECK_OPTS="-e SC1091" python paas-cf/concourse/scripts/pipecleaner.py --fatal-warnings paas-cf/concourse/pipelines/*.yml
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ bin/terraform*
 platform-tests/pkg
 platform-tests/src/results
 terraform/.terraform/
+.venv/
```
jpluscplusm: This *feels* like a non-repo-specific change that should live in your per-user .gitignore.

Makefile
```diff
@@ -46,7 +46,7 @@ lint_terraform: dev ## Lint the terraform files.
 	@terraform/scripts/lint.sh
 
 lint_shellcheck:
-	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' | xargs shellcheck
+	find . -name '*.sh' -not -path '*/vendor/*' -not -path './platform-tests/pkg/*'  -not -path './manifests/cf-deployment/*' -not -path './.venv/*' | xargs shellcheck
```
jpluscplusm: What pulls in the venv - why's it there?

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1554,7 +1554,7 @@ jobs:
                 . ./config/config.sh
                 echo | cf login -a "${API_ENDPOINT}" -u "${CF_ADMIN}" -p "${CF_PASS}"
 
-                if cf service-brokers | grep "rds-broker\s"; then
+                if cf service-brokers | grep "rds-broker\\s"; then
```
jpluscplusm: Assuming shellcheck is happy with it, it feels cleaner (and more obvious to the next dev) to single-quote these, so we can drop the double-backslash. Ditto every other `grep "string with no variables"`, below.

manifests/cf-manifest/scripts/generate-manifest.sh
```diff
@@ -17,7 +17,7 @@ if [ "${DISABLE_USER_CREATION}" = "false" ] ; then
 fi
 
 opsfile_args=""
-for i in ${PAAS_CF_DIR}/manifests/cf-manifest/operations.d/*.yml; do
+for i in "${PAAS_CF_DIR}"/manifests/cf-manifest/operations.d/*.yml; do
```
jpluscplusm: I'd prefer `"${PAAS_CF_DIR}/manifests/cf-manifest/operations.d/*.yml"`, but I don't know how shellcheck would feel about that.

terraform/scripts/lint.sh
```diff
@@ -16,7 +16,7 @@ chmod +x "${WORKING_DIR}"/terraform-provider-pingdom
 
 cd "${WORKING_DIR}"
 
-for dir in ${PAAS_CF_DIR}/terraform/*/ ; do
+for dir in "${PAAS_CF_DIR}"/terraform/*/ ; do
```
jpluscplusm: Same as above - I'd prefer the whole thing was in quotes.

.gitignore
blairboy362: TIL per-user gitignore files are a thing...I will update!

Makefile
blairboy362: I couldn't get the python stuff to work using the global (or per-user) python environment. Using a virtualenv was the only way I could get `make test` to work locally. The problem then was: shellcheck threw literally hundreds of errors checking the contents of .venv (which contain lots of shell scripts). So I wanted to exclude those, and couldn't find a better way.

manifests/cf-manifest/scripts/generate-manifest.sh
blairboy362: Shellcheck was fine. But changing that as suggested changes the behaviour of the `for` loop...or more specifically the argument to the `for` loop. Shell interpolation gets in the way and the subsequent `opsfile_args` gets very much the wrong value.

terraform/scripts/lint.sh
blairboy362: As above; the shell interpolation gets in the way if you put the whole line in quotes.

Makefile
jpluscplusm: You could instantiate the venv outside the repo, I *believe* ...

manifests/cf-manifest/scripts/generate-manifest.sh
jpluscplusm: Good catch. I'd forgotten that chunks them all into one space-containing word.

Makefile
blairboy362: TL;DR: I've taken the change out of this PR.

I thought about this over the weekend. Putting a venv outside of the repo is not idiomatic and certainly not something I want to encourage. So I think the best thing to do here is to add some scripting and documentation so that everyone then uses a venv in the same place, so a change like this is not so out of place. I'll do that in a separate PR when I get a moment.

PR #1481, opened 2018-08-15
[#159136450] Use upstream opsfile to configure bosh-DNS

alext:
What
----

Following on from #1471 this updates the manifest generation to use the upstream opsfile to configure bosh-DNS.

I've also pulled out a separate commit to set the version to 1.8.0 so that we can track the reasons for using this version instead of the version used in upstream cf-deployment.

How to review
-------------

* Code review
* run the pipeline as far as the `generate-cf-config` job and verify the generated manifest is identical to the one from master (this change should be a no-op)

Who can review
--------------

Not me.


manifests/cf-manifest/operations.d/030-bosh-dns-overrides.yml
```diff
@@ -0,0 +1,17 @@
+---
+
+- type: replace
+  path: /addons/name=bosh-dns/jobs/name=bosh-dns/properties/aliases/_.cell.service.cf.internal
+  value:
+    - _.diego-cell.cell.((deployment_name)).bosh
+    - _.isolated-diego-cell.cell.((deployment_name)).bosh
```
keymon: Can we say why this needs to be overrode?

keymon:
LGTM, this renders the same manifest than master in my environment.

PR #1484, opened 2018-08-15
[support] raise aws elasticache param group limits in prod

chrisfarms:
What
----

We have raised our limits for number of elasticache param groups in
ireland and london. This bumps the var to keep our monitor in parity
with reality.

How to review
-------------

Code review should suffice

Who can review
--------------

Not @chrisfarms


alext:
Looks good to me. I can confirm that the new number matches what AWS support said we now have.

PR #1489, opened 2018-08-21
[#159789897] Use Config from cf-acceptance-tests

richardTowers:
(... instead of cf-test-helpers)

What
----

The Config in cf-test-helpers uses mixed units - some things have timeouts
in seconds, some in minutes. This is different to the config in
cf-acceptance-tests which causes some difficult situations.

For example, if we try and reuse the same config for the upstream
acceptance tests (which don't use the cf-test-helpers config type) and our
custom acceptance tests we end up with timeouts that are a factor of 60
different.

By using the type from cf-acceptance-tests everywhere we can reuse the same
config and all fields are in seconds.

This change will introduce one difference - the cf-acceptance-tests config
type sets a default TimeoutScale of 2.0, which means that all the
configured timeouts are doubled before being used. As a result, all the
timeouts for the custom-acceptance-tests have effectively been doubled with
this change. We've decided that this is acceptable, as all it does is bring
these timeouts inline with the upstream acceptance tests.

How to review
-------------

* Code review
* Run the changes down a pipeline (or look at
https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/pipeline-lock/builds/69)

Who can review
--------------

Anyone other than Rich and Alex.


alext:
Given @richardTowers and I paired on this part of the story, I'm going to go ahead and merge this.

PR #1491, opened 2018-08-22
Fix unbound variable

paroxp:
What
----

Now that we have implemented the slim dev deployment functionality, we
should make sure it doesn't break any of our persisting environments.

Falling back to nothing with bash should ensure the desired behaviour,
where it has failed on us in previous staging deployment.

How to review
-------------

- See if the thing makes sense
- Probably no need to deployment to dev. The only way to find out if it works, is to deploy it :D


alext:
Doh! Thanks for picking this up.

PR #1497, opened 2018-08-23
Re-enable the self-signed ssl test.

blairboy362:
What
----

The issue has been fixed by the 3rd party maintainers of the site:
https://github.com/chromium/badssl.com/issues/359
This test can now be re-enabled.

How to review
-------------

1. Code review
1. Run the tests locally (`make test`)

Who can review
--------------

Anyone but me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,8 @@ lint_ruby:
 
 .PHONY: lint_posix_newlines
 lint_posix_newlines:
-	git ls-files | grep -v vendor/ | xargs ./scripts/test_posix_newline.sh
+	@# for some reason `git ls-files` is including 'manifests/cf-deployment' in its output...which is a directory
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -57,7 +57,8 @@ lint_ruby:
 
 .PHONY: lint_posix_newlines
 lint_posix_newlines:
-	git ls-files | grep -v vendor/ | xargs ./scripts/test_posix_newline.sh
+	# for some reason `git ls-files` is including 'manifests/cf-deployment' in its output...which is a directory
```
alext: Having this comment here causes make to print it on each run. Prefixing the comment with an `@` prevents this.

blairboy362: Good spot. I've added the `@` and squashed it into the same commit.

PR #1498, opened 2018-08-24
[#SUPPORT] Use Eventually when waiting for ES instance

keymon:
What
----

In the custom acceptance tests for the elasticsearch cluster, maybe
it takes a little bit longer to create the test index by the
healthcheck than the time the test takes to check for it,
causing some flakiness.

To fix this we wrap the check of the index around a
Eventually() that would timeout after 5 seconds.

How to review
-------------

Code review

Run the custom acceptance tests.

Who can review
--------------

Not me


keymon:
Talking with @alext this is a code smell, and actually we should not rely on the existance of a `test_index` . We should change the test app to create an index for testing.

PR #1501, opened 2018-08-29
[#133323615] Use per-environment Pingdom contacts

henrytk:
What
-----

We have dedicated email addresses (Google Groups) for alerting. They
are: govpaas-alerting-{dev,staging,prod}@digital.cabinet-office.gov.uk.
These addresses have been added as contacts in Pingdom, so we can use
them instead of the team email address. This will allow us to control
our subscription to Pingdom alerts with greater granularity.

How to review
-------------

- check contact IDs match contacts for the correct environment (Pingdom login details are in paas-credentials).
- check contacts have the correct email address
- apply the Terraform changes. This is still a manual process. Use the documentation in this repository to figure out how to do it. If you can't do it using the docs, we need to update the docs in this pull request
- after applying we may need to delete the contact 'Government PaaS Team Maillist' (contact ID 14270737). If it isn't used anywhere else its existence will cause confusion in future.

Who can review
--------------

Anyone with prod access. The reviewer may need to delete a redundant Pingdom contact after we have applied this, so they may need high credentials store access.

### DIFF
--- a/terraform/dev.tfvars
+++ b/terraform/dev.tfvars
@@ -12,8 +12,8 @@ cdn_db_skip_final_snapshot = "true"
 cdn_db_maintenance_window = "Mon:07:00-Mon:08:00"
 support_email="govpaas-alerting-dev@digital.cabinet-office.gov.uk"
 
-# 14270737 - Government PaaS Team Maillist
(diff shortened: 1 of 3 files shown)


alext:
I've applied this to all 4 environments.

PR #1502, opened 2018-08-30
[#158665945] Update paas-billing release

LeePorte:
What
----

This PR updates pass-billing to a new version that uses the broker
plan ID to select pricing plans rather than the Cloud Foundry ID. This
enables pass-billing to work in multiple environments including London
and in dev.

How to review
-------------

Deploy from this branch and check the admin org bill in your dev env. There will be billing entries for services on the bill.

Before merge
----

Update the commit to use the tag generated after merging https://github.com/alphagov/paas-billing/pull/49

Who can review
--------------

Not @LeePorte or @keymon

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -352,7 +352,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-billing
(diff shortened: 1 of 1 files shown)


henrytk:
I've updated this to point to the right tag and rebased it on master.

PR #1507, opened 2018-08-31
[#158913461] Improve showing secrets in showenv task

alext:
## What

This currently fails noisily if the pipeline hasn't been run, and
therefore the vars-store files don't exist in the S3 bucket. This wasn't
a problem for the existing show-cf-secrets.sh script only because we
hadn't `set -e` etc in it.

This takes the existing script and generalises it to show secrets from
any vars-store file and to fail gracefully if the secrets file doesn't
exsit.

How to review
-------------

* Code review
* Run `make dev showenv` and verify it still works
* Test `make dev showenv` when a vars store file doesn't exist in the bucket.

Who can review
--------------

Not me.


tlwr:
LGTM

PR #1510, opened 2018-09-03
[Support task] Upload Logit secrets

henrytk:
What
----

These are required for sending platform logs to Logit.

These were uploaded as part of paas-bootstrap, but nothing in that
repository uses these secrets now that Concourse logs aren't sent to
Logit.

How to review
-------------

Use it to upload secrets to your dev environment.

Related pull request: https://github.com/alphagov/paas-bootstrap/pull/180

Who can review
--------------

Anyone but me.


henrytk:
@alext 

> While we're here, do you think it's worth adding a guard to the pipelines-cloudfoundry.sh script to fail early if these secrets aren't present (similar to what it does for notify, aiven etc.)?

Yes. Done.

PR #1512, opened 2018-09-03
[#157053149] Tweak generate_logit-filters.sh for clarity

LeePorte:
What
----

This PR makes tweaks to the script for generating logit filters to make it clearer to know where to copy and paste from by adding start and end flag lines. There is an additional change to the README to add clarification.

How to review
-------------

* Code review and follow the process in https://github.com/alphagov/paas-team-manual/pull/245.
* Apply the generated filters to all the stacks in logit.
* The latest generated filters have been added in their own commit to make it easier to see the diff. Best viewed with whitespace changes ignored.

Who can review
--------------

Not @LeePorte, @blairboy362 and @samcrang


scripts/generate_logit_filters.sh
```diff
@@ -2,8 +2,12 @@
 
 set -euo pipefail
 
+echo "********************************* Start copy below this line ********************************************"
+
 echo "filter {"
 sed 's/^/  /' < config/logit/10_base.conf
 sed 's/^/  /' < config/logit/20_logsearch_for_cloudfoundry.conf
 sed 's/^/  /' < config/logit/30_custom_filters.conf
```
henrytk: We need to document/automate/script the generation of the contents of these files. When the source (logsearch-for-cloudfoundry) changes we will not get the updates and need to know how to regenerate these.

LeePorte: Ah I misunderstood.

scripts/generate_logit_filters.sh
```diff
@@ -1,9 +1,45 @@
...
+    exit 1
+fi
+
+set -u
+
+apk update && apk add git
+
+cd /tmp
+git clone https://github.com/cloudfoundry-community/logsearch-for-cloudfoundry.git
+cd logsearch-for-cloudfoundry/src/logsearch-config
```
alext: I think we should pin the version of logsearch-for-cloudfoundry we're using as well. This ensures we get a consistent output.

blairboy362: The files we reference are via a tag passed in as an argument...I'm not sure I follow here. Are you suggesting that instead of allowing someone to specify the version when we run it they have to update the script every time they want it to change? I was trying to avoid having to remember to update a version in a script to run it. Could you elaborate?

alext: I was pointing out that we're consuming 2 different git repos in this script, but only pinning a version for one of them - the script will always use current master of `logsearch-for-cloudfoundry`.

As for pinning versions in the repo, I think it's fine to have this script parameterised, but I think we need to pin the versions somewhere in code (maybe in the `Makefile`) so that it's clear exactly what was used to generate the current filters.

PR #1515, opened 2018-09-05
[#160233844] Remove Compose collector

henrytk:
What
----

Stop collecting Compose audit events

We are no longer using Compose, therefore we do not need to collect
Compose audit events. This upgrades to a version of the billing app in
which the Compose collector has been removed[1].

[1] https://github.com/alphagov/paas-billing/pull/51

How to review
-------------

Review in conjunction with https://github.com/alphagov/paas-billing/pull/51

## Before merging ⚠️ 

Once you are ready to merge bump the version of paas-billing, as per
the code comment, and rebase this commit. See commit message.

Make sure it deploys okay.

Who can review
--------------

Anyone but me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -352,7 +352,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-billing
-      tag_filter: v0.39.0
+      tag_filter: v0.41.0
 
   - name: paas-accounts
     type: git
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2043,7 +2042,6 @@ jobs:
                     'CF_CLIENT_SECRET' => '${CF_CLIENT_SECRET}',
                     'CF_CLIENT_REDIRECT_URL' => 'https://billing.${SYSTEM_DNS_ZONE_NAME}/oauth/callback',
                     'CF_API_ADDRESS' => '${API_ENDPOINT}',
-                    'COMPOSE_API_KEY' => '${COMPOSE_API_KEY}',
```
alext: It seems that this is not sufficient to remove the env var from the app. Because this is using a simple `cf push` below and not a zero-downtime strategy, env vars won't get removed when they're removed from the manifest.

henrytk: Good spot. I've put a fix in which can be reverted once it reaches prod. However, in doing this I realised we are getting rid of Compose soon anyway, so I am not sure if we should even care.

PR #1519, opened 2018-09-06
[support] Add the quota name into the org usage script

LeePorte:
What
----

The addition of the quota name enables filtering by default quotas to
find orgs that should now be being billed.

How to review
-------------

code review

login to an environment and run `./scripts/org-usage-report.sh >org_usage.csv` check that quota names are in the csv file

Who can review
--------------

Not @LeePorte


scripts/org-usage-report.sh
```diff
@@ -84,10 +84,17 @@ find_latest_login() {
   echo "$last_logon_times" | tr ' ' '\n' | sort | tail -n 1
 }
 
+quota_name() {
+  guid="$1"
+  quota_url=$(cf curl "/v2/organizations/${guid}" | jq -r '.entity.quota_definition_url')
```
keymon: In `${org}` you still have the json object of the org, that would save one `curl` per org

```diff
diff --git a/scripts/org-usage-report.sh b/scripts/org-usage-report.sh
index 271d4ee2..11b5c14a 100755
--- a/scripts/org-usage-report.sh
+++ b/scripts/org-usage-report.sh
@@ -85,10 +85,8 @@ find_latest_login() {
 }
 
 quota_name() {
-  guid="$1"
-  quota_url=$(cf curl "/v2/organizations/${guid}" | jq -r '.entity.quota_definition_url')
-  name=$(cf curl "$quota_url" | jq -r '.entity.name')
-  echo "$name"
+  quota_url="$(jq -r '.entity.quota_definition_url')"
+  cf curl "$quota_url" | jq -r '.entity.name'
 }
 
 orgs=$(get_resources 100 "/v2/organizations")
@@ -123,7 +121,7 @@ for guid in $org_guids; do
   last_logon_time=$(find_latest_login "$users")
 
   # quota
-  quota=$(quota_name "$guid")
+  quota=$(echo "${org}" | quota_name "$guid")
 
   # Space-separated list of emails of org managers
   org_manager_emails=$(echo "$managers" | jq -r ".resources[].entity.username")
```

henrytk:
The two scripts being introduced here take file inputs which don't necessarily exist. See comments on https://github.com/alphagov/paas-team-manual/pull/247

scripts/abandoned_orgs.js
```diff
@@ -0,0 +1,75 @@
+const NotifyClient = require('notifications-node-client').NotifyClient;
+
+const orgs = require('./abandoned_orgs.json');
```
henrytk: This should be `'../abandoned_orgs.json'`

scripts/billing_orgs.js
```diff
@@ -0,0 +1,75 @@
+const NotifyClient = require('notifications-node-client').NotifyClient;
+
+const orgs = require('./old_orgs.json');
```
henrytk: This should be `'../old_orgs.json'`

scripts/billing_orgs.js
```diff
@@ -0,0 +1,75 @@
...
+const NOTIFY_API_KEY = process.env['NOTIFY_API_KEY'];
+const NOTIFY_TEMPLATE_ID = process.env['NOTIFY_TEMPLATE_ID'];
+
+function setupNotify() {
+  const notify = new NotifyClient(NOTIFY_API_KEY);
+
+  return notify;
+}
+
+async function organisationsWithManagers(organisations, remaining) {
```
henrytk: This fails for me with:

```
(node:42178) UnhandledPromiseRejectionWarning: TypeError: organisations.reduce is not a function
    at organisationsWithManagers (/Users/henryknott/dev/go/src/github.com/alphagov/paas-cf/scripts/abandoned_orgs.js:15:24)
    at sendEmails (/Users/henryknott/dev/go/src/github.com/alphagov/paas-cf/scripts/abandoned_orgs.js:56:31)
    at /Users/henryknott/dev/go/src/github.com/alphagov/paas-cf/scripts/abandoned_orgs.js:74:9
    at Object.<anonymous> (/Users/henryknott/dev/go/src/github.com/alphagov/paas-cf/scripts/abandoned_orgs.js:75:3)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:191:16)
    at bootstrap_node.js:612:3
(node:42178) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 4)
(node:42178) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Same goes for the other script.

henrytk:
Both of the scripts introduced here are operating on the same data structure, so we should probably use a single script and parameterise the JSON file it is operating on.

.gitignore
```diff
@@ -5,3 +5,8 @@ bin/terraform*
 platform-tests/pkg
 platform-tests/src/results
 terraform/.terraform/
+node_modules/
+package-lock.json
```
tlwr: IMO lockfiles should be checked in to git

keymon: But there shouldn't be a `package-lock.json` or `package.json` at all in the project. Should it?

henrytk: it does seem weird this is at the root of the repository. Perhaps a sub folder for org management or something?

alext:
Comments here have raised the question for me of why we're using Node for these scripts. All the other scripts in this repo are either bash or ruby. I'm reluctant to add another language here without a good reason.

Is there a good reason to be using Node for these?

.gitignore
LeePorte: As it is only there for running the contact_orgs.js script, I decided to add it as an exemption. @paroxp has stated that this should be a feature in Pazmin, and as a result it is only short lived.

paroxp: I'd remove the `package-lock.json` from `.gitignore` to avoid potential problems in the future.

Hector's right. We're not using `package.json` file - which may be a problem as I imagine people don't just have `notifications-node-client` npm package installed globally, nor is it documented that it's required (I think) - therefore it won't work out of the box sadly.

As Lee mentioned, this should some day become a feature of paas-admin - but we should make sure it works for people who might be using it.

paroxp:
@alext I agree. It doesn't look like a good place for these.
But, if it's meant to help and continue with some stories, I have few good reasons for it to be in Node:

1. This code has already been written a while ago - why rewrite it in something else.
2. Bash doesn't have the Notify Client :D Though we could use curl and/or ruby client instead - this is more work though.
3. These JS scripts are meant to go into `paas-admin` some day - which is written in Node.

We could restart the story and work that functionality into paas-admin already? Otherwise we may need to do it all again (probably next quarter).

henrytk:
This story has changed in scope. We are only automating notification for abandoned orgs. We will do this in the Notify UI for now, and so won't require the controversial Node script. See https://github.com/alphagov/paas-team-manual/pull/247 for details of the process.

PR #1522, opened 2018-09-10
[#159730005] Update cf-deployment to v3.6.0

jpluscplusm:
What
----

We updated our upstream, `cf-deployment`, to v3.6.0 from v2.6.0.

This PR *must not* be merged&deployed before alphagov/paas-bootstrap#182, and *must* be merged&deployed immediately after it. This is because the combination of these 2 PRs moves bosh-dns from `paas-cf` to `paas-bootstrap` and, unless their their deployment is tightly coupled, significant breakage will occur. Bosh-dns being moved from per-deployment to per-BOSH-director, via the runtime-config mechanism, is a requirement of `cf-deployment` v3.0.0+.

Bosh-dns being moved to the director's runtime-config means that a deployment's bosh-dns aliases can't be configured at the bosh-dns addon level. Instead, we include the `bosh-dns-aliases` job in the CF deployment (as an addon) and use its more verbose syntax [1].

All CF components are at the same version that `cf-deployment` v3.6.0 defines - we don't use any later versions. We removed some ops files which are either no-ops or irrelevant, and started using some of those which upstream consolidated (`dc945ca`). We removed our previous vendoring of ops files which install log-cache, since after cf-deployment v3.0.0 it's included by default.

We added our usual CA rotation tactic, and updated some of our tests to allow for some BOSH packages having `-<stack>` suffixes not surfaced in their containing jobs or releases. e.g. `staticfile-buildpack-cflinuxfs2` instead of `staticfile-buildpack`.

[1] https://bosh.io/jobs/bosh-dns-aliases?source=github.com/cloudfoundry/bosh-dns-aliases-release&version=0.0.2#p%3daliases

Review
-------

This was paired on throughout, and is being released by @richardTowers, @tlwr and myself, with FIXME donating their access to prod.

Deployment plan: https://hackmd.cloudapps.digital/nZAnzvldRLqtxcEXbXoa6w?view#Deloyment-to-Production


tlwr:
As of https://github.com/alphagov/paas-cf/pull/1523 this is ZDD, as above: deploy and run https://github.com/alphagov/paas-bootstrap/pull/182 first

PR #1523, opened 2018-09-10
[#159730005] Cert rotation and uaa_ca on gorouter for v3.6.0

tlwr:
```
For cf-deployment v3.6.0

We have TLS between the gorouter and UAA, so we need tell the gorouter
to trust whatever CAs upstream has, AS WELL AS the CA that we use for
UAA.

We also do certificate rotations on all of these CA certs so we need to
add in the appropriate `${ca}_old` values.
```

What
----

We need to get gorouter to trust cloud controller (amongst others)

We're separating out our #159730005 v3.6.0 into separate releases so we do not get HTTP 526s when rolling out v3.6.0

How to review
-------------

Code review

Who can review
--------------

Not me or Richard

### DIFF
--- a/manifests/cf-manifest/operations.d/310-router.yml
+++ b/manifests/cf-manifest/operations.d/310-router.yml
@@ -34,9 +34,14 @@
     enable_tls: true
     private_key: ((router_ssl.private_key))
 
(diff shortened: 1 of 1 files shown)


alext:

PR #1532, opened 2018-09-12
[#159730005] Ensure current buildpacks are installed.

tlwr:
In 88ad0226095eafc4c969828d3d98446065551247 we "fixed" a test which asserted
the wrong thing, in cf-deployment v3.6.0 they changed the package names,
however we override these because we pin our buildpack versions (because we
like our tenants).

The result of this was that creating a new environment left us with a grand
total of 1 buildpack (where the buildpack version we had pinned was the same as
v3.6.0 - binary_buildpack: they had the same package name)

This PR does three things:

- Revert the test which we fixed, this can be reverted in one weeks time when
we have upgraded the buildpacks (after we have notified our tenants).

- Add an ops file which changes the package names that we look for to be the
same as they were pre v3.6.0 upgrade. This means that they will be installed
when creating a new environment.

- Adds a test which blindly asserts that the buildpacks that we expect to exist
do, in fact, exist.

Either code review, or:

- Spin up a new CF
- It goes green

Not @tlwr @jpluscplusm

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2819,6 +2819,74 @@ jobs:
                     ACCOUNTS_PASSWORD="$(awk '/^secrets_paas_accounts_admin_password:/ { print $2 }' < ../cf-vars-store/cf-vars-store.yml)" \
                       npm run test:acceptance
 
+          - task: ensure-buildpacks-exist
+            config:
+              platform: linux
+              image_resource:
+                type: docker-image
+                source:
+                  repository: governmentpaas/cf-cli
+                  tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+              inputs:
+                - name: paas-cf
+                - name: cf-vars-store
+              params:
+                CF_ADMIN: admin
+                SYSTEM_DNS_ZONE_NAME: ((system_dns_zone_name))
+              run:
+                path: bash
+                args:
+                  - -c
+                  - |
+                    set -ueo pipefail
+
+                    CF_PASS=$(awk '/cf_admin_password/ {print $2}' cf-vars-store/cf-vars-store.yml | tr -d '"')
+                    API_ENDPOINT="https://api.${SYSTEM_DNS_ZONE_NAME}"
+                    echo | cf login -a "${API_ENDPOINT}" -u "${CF_ADMIN}" -p "${CF_PASS}" > /dev/null
+
+                    SORT="sort"
+
+                    cat <<-BUILDPACKS | "$SORT" > desired_buildpacks.txt
+                    binary_buildpack
+                    staticfile_buildpack
+                    java_buildpack
+                    ruby_buildpack
+                    nodejs_buildpack
+                    go_buildpack
+                    python_buildpack
+                    php_buildpack
+                    BUILDPACKS
+
+                    cf buildpacks \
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2819,6 +2819,74 @@ jobs:
...
+                      echo 'Could not find one of the following buildpacks:'
+                      cat buildpacks_not_found.txt
+                      echo '-----'
+                      echo 'List of buildpacks found:'
+                      cat actual_buildpacks.txt
+                      exit 1
+                    else
+                      echo 'All buildpacks present'
+                      exit 0
+                    fi
```
richardTowers: Bike shedding, but I think this would be a bit easier to read with a loop (at the cost of being `O(n^2)`)

```
cf_buildpacks="$(cf buildpacks)"
has_all_buildpacks=true
for buildpack in binary staticfile java ruby nodejs go python php; do
  if grep -vq "${buildpack}_buildpack" <<< "$cf_buildpacks"; then
    echo "Could not find buildpack $buildpack"
    has_all_buildpacks=false
  fi
done
if ! $has_all_buildpacks; then
...
else
...
fi
```

richardTowers: Actually thinking about it this isn't much better 🤷‍♀️

tlwr:
would
```
if grep -q . buildpacks_not_found.txt; then
  echo 'Could find the following buildpacks:'
  comm -23 desired_buildpacks.txt actual_buildpacks.txt
  echo '-----'
  echo 'List of buildpacks found:'
  cat actual_buildpacks.txt
  exit 1
else
  echo 'All buildpacks present'
  exit 0
fi
```
be better in your eyes

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2819,6 +2819,74 @@ jobs:
...
+                    cat <<-BUILDPACKS | "$SORT" > desired_buildpacks.txt
+                    binary_buildpack
+                    staticfile_buildpack
+                    java_buildpack
+                    ruby_buildpack
+                    nodejs_buildpack
+                    go_buildpack
+                    python_buildpack
+                    php_buildpack
+                    BUILDPACKS
```
alext: Is there a way we could extract the list of expected buildpacks from the manifest instead of having to hardcode them here? Maybe by extracting the names from the `install_buildpacks` property.

tlwr: IMO the point of this test is that we explicitly have a list of buildpacks that we want.

I would want this test to fail if upstream removed say the `php_buildpack`.

alext: There are 2 assertions here.

1. Assert that the buildpacks in the manifest are actually present on the platform
2. Assert that the a specific list of buildpacks are included.

IMO, 2 can be better covered by a manifest unit test, which leaves this test covering the first case to catch issues like the ones we encountered with the package renaming etc.

jpluscplusm:
@samcrang and I are adding a unit test to catch this problem earlier in the testing cycle.

jpluscplusm:
https://deployer.welch.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/tag-release/builds/3 passed on @tlwr's dev env, which is the last check I believe we need before merging this.

PR #1547, opened 2018-09-18
[#160582942] Enable dotnet-core buildpack

tlwr:
What
----

We didn't think we had the capability to support dotnet-core as a
buildpack, but for various reasons we don't think this decision is
valid anymore. The reasons include but are not limited to:

- the team now has more dotnet experienced people
- we can't stop people using custom buildpacks anyway
- our users want it
- it seems to be stable

How to review
-------------

1. Code review
2. See [green pipeline](https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry?groups=deploy)
3. Log into api.towers.dev.cloudpipeline.digital and push an app

Who can review
--------------

Not me


keymon:
LGTM

PR #1549, opened 2018-09-19
[#159136931] Stop relying on Consul

henrytk:
## What

BBS had been using Consul for cell presence.  Therefore, despite Consul
not being the primary system for locks, its removal can still cause
problems.

How to review
-------------

Ensure this deploys with passing tests and no anticipated downtime.

Who can review
--------------

@tlwr


tlwr:
happy with this but waiting for travis

PR #1551, opened 2018-09-19
[#158108701] Upgrade buildpacks to CF v3.6 versions

alext:
What
----

This upgrades the buildpacks to the versions included in CF v3.6.0.

How to review
-------------

* Verify the versions being upgraded to match the versions in [the announcement](https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!topic/gov-uk-paas-announce/vXHub0FpsDw). (Note: I've also included upgrading the .NET Core buildpack in this. This wasn't in the announcement because it hasn't previously been enabled on the platform.)
* Test that a deployment from this branch works.
* Check my reasoning for removing the `install_buildpacks` test (see commit message).

I've already tested a new deployment of CF to verify the buildpacks are installed, so there's probably no need to repeat this test.

Who can review
--------------

Not me.


tlwr:
this is green on my pipeline

PR #1557, opened 2018-09-24
[#158180318] Update the chrome plugin

paroxp:
What
----

Now that we have upgraded concourse, some of the frontend has changed.
This means our plugin is out of date and should be updated if we'd like
to keep on using it on our dashboard TV.

In the new concourse version, upstream has decided to separate the
grouping bar into it's own place. This is a simple change really and
requires some minimal effort to hide new additional element.

Some other changes have been made, purely as a maintenance task to
keep the linter happy as well as the code more modern.

See commit for more details.

How to review
-------------

- Code review
- Copy contents of the file
- Open the deployer concourse on staging
- Open the `Inspect` console on Chrome (Cmd + Alt + J)
- Paste the contents of the file
- See no errors and expect the dashboard to be clean

After merge
-----------

Make sure to update the plugin on the TV's Chrome. See README for more details.

### DIFF
--- a/misc/chrome_plugins/clean_concourse_pipeline/src/inject.user.js
+++ b/misc/chrome_plugins/clean_concourse_pipeline/src/inject.user.js
@@ -8,27 +8,25 @@
 // @version     1
 // @grant       none
 // ==/UserScript==
-var readyStateCheckInterval = setInterval(function() {
-    if (document.readyState === "complete") {
-        clearInterval(readyStateCheckInterval);
+const readyStateCheckInterval = setInterval(() => {
+  if (document.readyState === 'complete') {
+    clearInterval(readyStateCheckInterval);
 
-        console.log("Monitor mode is go");
-        var element = document.getElementsByClassName("legend")[0];
-        element.parentNode.removeChild(element);
+    console.log('Monitor mode is go');
+    const $legend = document.querySelector('.legend');
+    $legend.style.display = 'none';
 
-        var element = document.getElementsByClassName("lower-right-info")[0];
-        element.parentNode.removeChild(element);
+    const $infoBox = document.querySelector('.lower-right-info');
+    $infoBox.style.display = 'none';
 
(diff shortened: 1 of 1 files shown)


alext:
This change seems to conflate 2 things - there are some JS style changes, and the functional changes to reflect the newer concourse. Could they be separated into 2 commits so that we can more easily see what the functional changes are.

paroxp:
I was hoping this would go unseen 👀 

Sure!

misc/chrome_plugins/clean_concourse_pipeline/src/inject.user.js
```diff
@@ -8,27 +8,27 @@
...
+const readyStateCheckInterval = setInterval(() => {
+  if (document.readyState === 'complete') {
+    clearInterval(readyStateCheckInterval);
 
-        console.log("Monitor mode is go");
-        var element = document.getElementsByClassName("legend")[0];
-        element.parentNode.removeChild(element);
+    console.log('Monitor mode is go');
+    const $legent = document.querySelector('.legend');
+    $legent.parentNode.removeChild($legent);
```
richardTowers: `$legent` typo?

paroxp: thanks

richardTowers:
The script looks like it works (it removes the chrome as you'd expect), but after it's run I see a bunch of errors from the elm:

```
elm.min.js?id=18701b041d48f77c150ba19dedd849da:1 Uncaught TypeError: Cannot read property 'childNodes' of undefined
```

Presumably we're removing some elements that concourse still thinks it owns. Would it be better to set `display: none;` on things instead of removing them?

Do we care about these errors? I wouldn't have raised them, but you said "See no errors" in your instructions :)

paroxp:
> I wouldn't have raised them, but you said "See no errors" in your instructions

You're right, my bad :D

In general, I didn't really care for it. But it does mean the console is cleaner and would make it easier to debug things.

PR #1575, opened 2018-10-11
[#160556015] Rotation for cc db encryption key - Part 2

henrytk:
What
----

This removes the temporary rotation pipeline added in #1597 

How to review
-------------

Code review should be sufficient.

Who can review
--------------

Anyone but me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1290,28 +1290,6 @@ jobs:
                 ./paas-cf/concourse/scripts/bosh_login.sh bosh-secrets/bosh-secrets.yml
                 bosh -n deploy cf-manifest/cf-manifest.yml
 
-      - task: rotate-cc-db-encryption-keys
-        config:
-          platform: linux
-          image_resource: *gov-paas-bosh-cli-v2-image-resource
-          inputs:
-            - name: paas-cf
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3250,6 +3251,37 @@ jobs:
         params:
           file: updated-tfstate/cf.tfstate
 
+  - name: rotate-cc-db-encryption-key
+    serial: true
+    plan:
+    - aggregate:
+      - get: paas-cf
```
alext: We should probably pin this to the version thats passed cf-deploy to ensure we're using a version that matches what's been deployed...

henrytk: Sensible, but paas-cf is only imported for the use of the `./paas-cf/concourse/scripts/bosh_login.sh` script, which is stable; additionally, this pipeline is temporary and will be removed immediately after it is run everywhere. I gave some thought to whether it is worth keeping, but it means speculating about our future implementation of the rotation mechanism.

henrytk: Also, the diff here is weird at the moment because it is a branch off of a branch.

alext:
I'm going to close this for now as we're not going to run this task in the pipeline until after the next CF ugprade.

PR #1578, opened 2018-10-15
Rotate Miki's GPG key for a new Yubikeyed one

46bit:
What
----

See commit messages.

How to review
-------------

1. 🐝🐝🐝 **The first commit changes the existing public key file. This seems odd to me. I'd like ideas on why?** 🐝🐝🐝
2. Verify this is changing my GPG key from CAAD002F768C0A74 to BB97A18439B31647;
3. Let me know if you know of anywhere else in paas-bootstrap that needs changing.

Who can review
--------------

Not @46bit. Ideally someone with context on alphagov/paas-cf#1564


46bit:
This won't work because Pay split up the private keys into separate subkeys. I'll figure out something else.

46bit:
It would still be good to see if `make update_merge_keys` run by someone else (@alext? @henrytk?) causes changes to the YAML.

46bit:
Actually, I think we're golden. But someone will need to update my key ID and resign `paas-credentials`--I can't easily because I've already revoked my old key.

henrytk:
@46bit 

> The first commit changes the existing public key file. 

That's okay, and unfortunately expected.

I see your PR and raise you this counter PR: https://github.com/alphagov/paas-team-manual/pull/255

henrytk:
@46bit 

> But someone will need to update my key ID and resign paas-credentials--I can't easily because I've already revoked my old key.

Can you raise a 'small' task for this in Pivotal? Ping a PM/DM and have them prioritise it for before your next suport rota.

PR #1583, opened 2018-10-17
[#161146691] Add logic to pause the kick-off after kick-off

keymon:
What
----

We want to not trigger the deploy pipeline automatically for every
environment unless the developer opts in the day before.

To do so, we simply add an additional job that would pause the
timer that triggers the kick-off after the timer is triggered.

The developer must unpause the resource manually to get the deployment
the next day.

Known flaws: 
 - The first time you deploy from scratch, the feature is not disabled by default. I think we can accept this instead of add more complexity.

How to review
-------------

Code review.

To test: Change the branch in concourse/pipelines/deployment-kick-off.yml and
   the time window of the trigger to test it.

Who can review
--------------

Not me

### DIFF
--- a/concourse/pipelines/deployment-kick-off.yml
+++ b/concourse/pipelines/deployment-kick-off.yml
@@ -69,8 +69,6 @@ jobs:
           inputs:
             - name: paas-cf
           params:
-            AWS_ACCOUNT: ((aws_account))
-            AWS_DEFAULT_REGION: ((aws_region))
             DEPLOY_ENV: ((deploy_env))
             SKIP_AWS_CREDENTIAL_VALIDATION: true
           run:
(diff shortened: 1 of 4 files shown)


concourse/pipelines/deployment-kick-off.yml
```diff
@@ -82,3 +82,32 @@ jobs:
...
+            AWS_DEFAULT_REGION: ((aws_region))
+            DEPLOY_ENV: ((deploy_env))
+            SKIP_AWS_CREDENTIAL_VALIDATION: true
+          run:
+            path: sh
+            args:
+              - -e
+              - -c
+              - |
+                make -C ./paas-cf "${AWS_ACCOUNT}" pause-kick-off
```
henrytk: We have a variable `((makefile_env_target))` which we use in situations like this. It is used precisely for this purpose: differentiating `make` targets that use the same AWS account (for example, prod and prod-lon). I know we don't use this deployment kick-off in prod, but I think we should be consistent.

concourse/pipelines/deployment-kick-off.yml
```diff
@@ -82,3 +82,32 @@ jobs:
...
+            source:
+              repository: governmentpaas/self-update-pipelines
+              tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+          inputs:
+            - name: paas-cf
+          params:
+            AWS_ACCOUNT: ((aws_account))
+            AWS_DEFAULT_REGION: ((aws_region))
+            DEPLOY_ENV: ((deploy_env))
+            SKIP_AWS_CREDENTIAL_VALIDATION: true
```
henrytk: I see why using these three AWS variables is necessary, but it is starting to feel weird how dependent this is on AWS, given all it needs to do is get the Concourse ATC password from an S3 bucket (for which it can use its own instance profile).

No action required here, just thought I would comment because it is weird.

concourse/scripts/pause-kick-off.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
henrytk: These two scripts could probably be one script with an argument: `(pause|unpause)`, but I will survive if you leave it as it is.

keymon:
@henrytk changed that

PR #1584, opened 2018-10-17
[#SUPPORT] Script the creation of orgs, not users

richardTowers:
What
----

Since we've got paazmin now we should probably use that for inviting
users
and we can avoid a bunch of complexity in this script. We don't have the
ability to add new orgs in paazmin yet though, so we still need a script
for that.

The script now just creates an empty org / space and prompts the user to
go to paazmin and invite the new users instead of scripting it all.

How to review
-------------

* Code review (see 32c2a0a for the actual diff)
* Use the script to create an org against a dev env

Who can review
--------------

Anyone other than @richardTowers


scripts/create-org.sh
```diff
@@ -0,0 +1,121 @@
...
+    >&2 echo 'Expected guid not to be empty. Aborting'
+    exit 1
+  fi
+  echo '******************************************'
+  echo 'Organisation created.'
+  echo
+  echo 'Please create invite the required users to this organisation using paas admin by visiting:'
+  echo
+  echo -e -n "${COL_BLUE:-}"
+  echo "https://admin.cloud.service.gov.uk/organisations/${guid}/users/invite"
```
keymon: The url of the admin portal should be based on the current targeted deployment:

```suggestion
  echo "$(cf target | sed -n 's|api endpoint: *https://api.|https://admin.|p'
)/organisations/${guid}/users/invite"
```

scripts/create-org.sh
```diff
@@ -0,0 +1,122 @@
...
+    >&2 echo 'Expected guid not to be empty. Aborting'
+    exit 1
+  fi
+  echo '******************************************'
+  echo 'Organisation created.'
+  echo
+  echo 'Please create invite the required users to this organisation using paas admin by visiting:'
+  echo
+  echo -e -n "${COL_BLUE:-}"
+  echo "$(cf target | sed -n 's|api endpoint: *https://api.|https://admin.|p'
```
keymon:
```suggestion
  echo "$(cf target | sed -n 's|api endpoint: *https://api.|https://admin.|p')/organisations/${guid}/users/invite"
```

scripts/create-org.sh
```diff
@@ -0,0 +1,122 @@
...
+)/organisations/${guid}/users/invite"
```
keymon:
```suggestion
```

scripts/create-user.sh
```diff
@@ -1,337 +0,0 @@
...
-SUBJECT_RESET='Reset your password on GOV.UK PaaS'
-NOTIFICATION='
-As a new account has been created please remember to update the
-gov-uk-paas-announce mailing list. You can do that by inviting the user to the
-group by using this URL:
-
-https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!managemembers/gov-uk-paas-announce/invite
-
-As a welcome message you can use the text from here:
-https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!forum/gov-uk-paas-announce
```
alext: With the update to the process, I don't see where we invite the user to the announce list.  Having said that, This feels like a gap in the process for users that are invited by an org manager anyway. This needs some thought.

keymon: Indeed because actually I think I have forgotten this part of the process for a lot of tenants. 

We should automate that process.

richardTowers: Yeah - I just missed this bit and would have forgotten it. I'll raise another PR to add these instructions back into the script (after the other prompt)

PR #1586, opened 2018-10-18
[#161146691 ] Small fixes to auto-kickoff process and docs

jpluscplusm:
What
----

Fixes:

- The `unpause-kick-off` Makefile target should be phony. This is almost certainly a typo from a copy/paste.
- The docs refer to the triggering resource by a slightly wrong name.

Both of these should be uncontroversial.

How to review
-------------

Code review.

Who can review
--------------

Anyone. @henrytk reviewed the previous PR, so the context of that process would probably be handy.


keymon:
sorry :angel:

PR #1589, opened 2018-10-19
[#SUPPORT] Add script to list all users

richardTowers:
What
----

Uses the current `cf-cli` user's oauth token to query the uaa API
directly.

Could have used the uaa client instead of calling the API, but it's a
bit of a pain to use.

How to review
-------------

* Code review
* Run the script

Who can review
--------------

PaaS admins

### DIFF
--- a/scripts/list-all-users.sh
+++ b/scripts/list-all-users.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e -u -o pipefail
+
+uaac_endpoint=$(cf curl /v2/info | jq -r '.authorization_endpoint')
+oauth_token=$(cf oauth-token)
+
+get_number_of_users() {
+  curl --silent --fail "$uaac_endpoint/Users?count=0" -H "Authorization: $oauth_token" | jq -r '.totalResults'
+}
+
+get_users() {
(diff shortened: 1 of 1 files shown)


keymon:
The alternative approach with uaac:

```bash
#!/bin/bash

set -e -u -o pipefail

env="${1:-}"

case ${env} in
  "prod")
    uaac_target=https://login.cloud.service.gov.uk
  ;;
  "prod-lon")
    uaac_target=https://login.london.cloud.service.gov.uk
  ;;
  *)
    echo "Usage $0 <env>"
    exit 1
  ;;
esac

login_and_target() {
  uaac target "${uaac_target}"
  uaac me > /dev/null || bundle exec uaac token sso get cf -s ""
}

get_all_users () {
  total_results="$(uaac users | sed -n 's/.*totalresults: //p')"
  for s in $(seq 1 500 "${total_results}"); do
    uaac users --count 500 --start "${s}" -a username | sed -n 's/.*username: //p'
  done
}

login_and_target
get_all_users
```

keymon:
I like your approach :+1:

richardTowers:
It was heavily based on yours 😸

PR #1595, opened 2018-10-22
[#160175644] Enable bosh prometheus exporter

henrytk:
What
----

We want to use the Bosh exporter to gather VM metrics. This is an alternative to running the node exporter on every VM.

Related
--------
https://github.com/alphagov/paas-bootstrap/pull/202

How to review
-------------

* Deploy from this branch.
* Check in the Grafana UI the Bosh dashboards are there and populated by data.
* Check in AlertManager to see the correct alerts are set up

Who can review
--------------

@henrytk and @keymon


keymon:
Closing for the time being until we solve the problem of how to create read-only users in bosh.

PR #1606, opened 2018-10-29
[#161411167] Remove health snapshot

henrytk:
## What

This reverts 86653a3c811f9169fc4b0bf12fd19d99f12dd695. It was not
reverted using `git revert` due to conflicts.

This health snapshot was broken when we enabled UAA in Bosh and would
have taken significant development time to fix. It is a maintenance
burden that is not currently delivering value, so it has been removed.

How to review
-------------

Code review should be sufficient.

Who can review
--------------

@keymon


keymon:
LGTM

PR #1611, opened 2018-11-02
[#161647174] Add new plans for postgres v10.5 (experimental)

keymon:
What
----

We want to enable the new version of postgres 10 for some tenants that
are requesting it. They also require it with the "citext" extension.

We add the new plans, tagged as "alpha".

This will give us some prompt feedback for the new version.

We will enable access to these plans adhoc and manually to the needed
organisations once created.

How to review
-------------

Code review.
It has been deployed in my environment.

Who can review
--------------

Anyone.


manifests/cf-manifest/spec/manifest/rds_broker_spec.rb
```diff
@@ -178,16 +177,23 @@
           "xlarge-9.5",
           "xlarge-ha-unencrypted-9.5",
           "xlarge-ha-9.5",
+
```
LeePorte: This empty line appears to be making Travis fail.

keymon: Sorry, removed.

LeePorte: 👍

PR #1612, opened 2018-11-05
[#161716019] Pre-configure a wildcard hostname '*'

keymon:
What
----

We want to pre-assign a host '*' route for the main shared domains
(non-internal) into our admin org, so that no other tenant would be able to
register the wildcard domain.

This would prevent a potential issue if someone registers the wildcard for
any other application, potentially hijacking the traffic of apps being
unavailable, or any route not mapped yet.

For the time being the route remains unmapped, in the future we might want
to assign this to a custom application with some message and custom HTTP
status code.

We also add a custom acceptance test for this case.

How to review
-------------

 - Code review
 - Run the pipeline
 - Check you cannot create a '*' route as a normal user:

```
DEPLOY_ENV=...
cf create-user testuser pass123
cf set-space-role testuser govuk-paas sandbox SpaceDeveloper

cf auth testuser pass123
cf target -o govuk-paas -s sandbox

cf create-route public ${DEPLOY_ENV}.dev.cloudpipelineapps.digital --hostname '*'
```

Who can review
--------------

Not me

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1484,10 +1484,12 @@ jobs:
                 cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname www
                 cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname api
                 cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname status
+                cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname '*'
 
                 # Reserve the london app name so that we don't cause conflicts
                 # when we roll out the London region.
                 cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname london
+
 
       - task: register-rds-broker
         config:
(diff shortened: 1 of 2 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1489,6 +1489,27 @@ jobs:
                 # when we roll out the London region.
                 cf create-route docs "${APPS_DNS_ZONE_NAME}" --hostname london
 
+      - task: configure-wildcard-routes
```
henrytk: We already have a place for this: https://github.com/alphagov/paas-cf/blob/894afdd5ac60220448ba12635f082f0e66f51845/concourse/pipelines/create-cloudfoundry.yml#L1481-L1486

keymon: Fixed. I put it in `govuk-paas/docs` now

richardTowers:
Code review: looks good - just going to run it down my pipeline to test.

PR #1614, opened 2018-11-06
Check that platform tests compile in test

richardTowers:
What
----

The staging pipeline failed because of a missing colon in the platform
tests - this sort of thing should be caught at unit test time. Because
we can't run the platform tests, we weren't even trying to compile them.

Using `go test -run ^$` makes go compile the test code, but doesn't run
any tests (because they don't match the empty string pattern).

Also fixing the missing colon that made the tests fail and an unrelated
error about `%s` that occurs with my version of the comiler.

How to review
-------------

* Code review
* Check the Travis build output to confirm it ran the new check

Who can review
--------------

Not @richardTowers


richardTowers:
Output of the new check: https://travis-ci.org/alphagov/paas-cf/builds/451383466#L1038-L1051

PR #1620, opened 2018-11-14
[Support] Disable Prometheus cf_exporter

alext:
What
----

We've discovered that this was DoSing the API. We're not replying on any
metrics from it at the moment, so we're going to disable it for the time
being until we have a need for it, at which point we can work out how to
configure this better.

How to review
-------------

* Code review is probably enough.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/prometheus/spec/operations/monitor_cf_spec.rb
+++ b/manifests/prometheus/spec/operations/monitor_cf_spec.rb
@@ -1,0 +1,16 @@
+RSpec.describe "Monitor CF" do
+  it "adds the firehose instance_group" do
+    firehose = manifest_with_defaults.get("instance_groups.prometheus2")
+    expect(firehose).not_to be_nil
+    expect(firehose["vm_type"]).to eq("medium")
+    expect(firehose["networks"]).to eq([{ "name" => "cf" }])
+  end
+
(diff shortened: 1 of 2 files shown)


alext:
I've manually stopped this job in prod for the moment, so this can be merged in the morning if necessary.

manifests/prometheus/spec/operations/monitor_cf_spec.rb
```diff
@@ -0,0 +1,16 @@
+RSpec.describe "Monitor CF" do
+  it "adds the firehose instance_group" do
+    firehose = manifest_with_defaults.get("instance_groups.prometheus2")
+    expect(firehose).not_to be_nil
+    expect(firehose["vm_type"]).to eq("medium")
+    expect(firehose["networks"]).to eq([{"name" => "cf"}])
```
richardTowers:
```suggestion
    expect(firehose["networks"]).to eq([{ "name" => "cf" }])
```
☝️ to make the linter happy

alext: Yeah, rubocop... sigh... It goes too far sometime IMO.

I've rebased this change in...

PR #1628, opened 2018-11-21
WIP: [#162118725] Upgrade stemcells to 170.9, and cflinuxfs3

alext:
What
----

Fixes a bunch of CVEs

Also includes an update to cflinuxfs3 which isn't currently in use, but still worth updating for the future.

How to review
-------------

Code review is probably enough

Who can review
--------------

Not me.


alext:
I've run it down my dev environment with the updates, and it looked good there.

PR #1629, opened 2018-11-21
[#161316533] Deploy 2 separate apps from paas-billing

jpluscplusm:
What
----

As of https://github.com/alphagov/paas-billing/pull/56 there are 2 apps for the API and the data collector. This commit deploys both during `post-deploy`, using the appropriate strategy to deliver HA and non-HA uptime as required by the apps' consumers.

How to review
-------------

- Code review
- Run down a dev pipeline, probably as part of validating https://github.com/alphagov/paas-billing/pull/56 
- After this has run in prod, the `FIXME` line in the concourse pipeline can be PR'd away. It's harmless to leave there, however.

Who can review
--------------

Anyone except @jpluscplusm or @bandesz

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -342,7 +342,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-billing
-      tag_filter: v0.44.0
+      tag_filter: v0.46.0
 
   - name: paas-accounts
     type: git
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1863,19 +1863,38 @@ jobs:
...
+                    app['env'] = app['env'].merge(env)
+                    app['services'] = ['billing-db']
+                  }
+                  File.write('manifest-collector.yml', collector.to_yaml)
                 "
 
-                cf push paas-billing
+                # FIXME: remove this once all production envs have run this at
+                # least once. It's harmless to leave here, however.
+                cf rename paas-billing paas-billing-collector || true
```
keymon: Can we do something more like:

```
apps="$(cf apps)"
if echo "${apps}" | grep -qe '^paas-billing ' ; then 
   cf rename paas-billing paas-billing-collector
fi
``` 

That would fail if there is any problem in the call `cf apps` (e.g. network timeout), only do the rename if the app is called `paas-billing` and again fail and stop  if `cf rename` fails.

bandesz: fixed

PR #1630, opened 2018-11-22
[Support] Make installed cf cli plugins available in subshell script

alext:
## What

This script points CF_HOME to a temporary directory to create an
indipendent login for the subshell. This has the side-effect of hiding
any cli plugins that are installed. To avoid this, this PR adds a
symlink back to plugins directory so that they can be found.

How to review
-------------

* Code review
* use the cf_subshell_scoped_login scipt, and verify that `cf plugins` still returns any plugins you have installed.

Who can review
--------------

Not me.


bandesz:
Merging this without triggering a Concourse deploy.

PR #1636, opened 2018-11-23
WIP: [#162157392] Add alert for Syslog Drains in prometheus

keymon:
What
----

Here we add a basic example of one alert using syslog drains on
prometheus and alertmanager.

We include a new directory `manifests/prometheus-manifest/alerts.d`
where one file per alert can be added.

We also setup receivers for email and pagerduty, that would allow to
configure notification for any alert. To do so, the alert must
add a new label `notify: ` including the words `email` or `pagerduty`
or both.

How to review
-------------

Deploy this. Try to trigger the alert by, for instance, doing:

```
cf target -o admin -s billing
for i in $(seq 1..300); do
    cf cups test-syslog-$i -l syslog://127.0.0.1.xip.io
    cf bind-service test-syslog-$i paas-accounts
fone
```

To trigger alerts and test only the receivers, you can use
the following script:

https://gist.github.com/keymon/2dcfc653bd0e2099fda589cb19e983a3

Who can review
--------------

Anyone

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,6 +262,13 @@ upload-pagerduty-secrets: check-env ## Decrypt and upload pagerduty credentials 
 	$(if $(wildcard ${PAGERDUTY_PASSWORD_STORE_DIR}),,$(error Password store ${PAGERDUTY_PASSWORD_STORE_DIR} does not exist))
 	@scripts/upload-pagerduty-secrets.sh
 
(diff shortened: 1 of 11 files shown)


LeePorte:
This PR is being closed in favour of https://github.com/alphagov/paas-cf/pull/1639

PR #1638, opened 2018-11-26
[#162013441] Fine-tune "Cloud Controller latency as reported by gorouter" DataDog monitor

bandesz:
# What

The first version of the monitor was too sensitive and was making a lot of
noise.

We decided to increase the thresholds by 50 ms and only consider the average of
all gorouter latencies.

# How to review

Code review should be enough.

# Who can review

Not me.


jpluscplusm:
LGTM. Merging after Travis.

PR #1639, opened 2018-11-26
[#162157392] Add alert for Syslog Drains in prometheus

LeePorte:
What
----

Here we add a basic example of one alert using syslog drains on
prometheus and alertmanager.

We include a new directory `manifests/prometheus-manifest/alerts.d`
where one file per alert can be added.

We also setup receivers for email, that would allow to
configure notification for any alert. To do so, the alert must
add a new label `notify: ` including the words `email`.

How to review
-------------

Deploy this. Try to trigger the alert by, for instance, doing:

```
cf target -o admin -s billing
for i in $(seq 1 300); do
    cf cups test-syslog-$i -l syslog://127.0.0.1.xip.io
    cf bind-service paas-accounts test-syslog-$i
fone
```

To trigger alerts and test only the receivers, you can use
the following script:

https://gist.github.com/keymon/2dcfc653bd0e2099fda589cb19e983a3

Who can review
--------------

Anyone

### DIFF
--- a/manifests/prometheus/alerts.d/600-syslog_drains.yml
+++ b/manifests/prometheus/alerts.d/600-syslog_drains.yml
@@ -13,3 +13,13 @@
       annotations:
         summary: Syslog drains
         description: "Consider scaling the adapters to cope with the number of syslog drains."
+    - alert: CFSyslogDrainsCritical
+      expr: firehose_value_metric_cf_syslog_drain_scheduler_drains > 300
+      labels:
+        deployment: ((metrics_environment))
+        severity: critical
+        notify: email
+        runbook: "https://github.com/cloudfoundry/cf-syslog-drain-release/tree/v6.4#syslog-adapter"
+      annotations:
+        summary: Syslog drains
+        description: "Consider scaling the adapters to cope with the number of syslog drains."
--- a/manifests/prometheus/alerts.d/600-syslog_drains.yml
+++ b/manifests/prometheus/alerts.d/600-syslog_drains.yml
@@ -8,6 +8,7 @@
       labels:
         deployment: ((metrics_environment))
         severity: warning
+        notify: email
         runbook: "https://github.com/cloudfoundry/cf-syslog-drain-release/tree/v6.4#syslog-adapter"
       annotations:
         summary: Syslog drains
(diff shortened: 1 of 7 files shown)


manifests/prometheus/scripts/generate-manifest.sh
```diff
@@ -11,10 +11,22 @@ for i in "${PAAS_CF_DIR}"/manifests/prometheus/operations.d/*.yml; do
   opsfile_args+="-o $i "
 done
 
+alerts_opsfile_args=""
+for i in "${PAAS_CF_DIR}"/manifests/prometheus/alerts.d/*.yml; do
+  alerts_opsfile_args+="-o $i "
+done
+
+vars_files=""
```
richardTowers: I think it would be more consistent to call this `varsfile_args`, which might be different enough to `VARS_FILES` that you don't need to disable the shellcheck warning anymore

manifests/prometheus/operations.d/310-alertmanager-routes.yml
```diff
@@ -0,0 +1,32 @@
+- type: replace
+  path: /instance_groups/name=alertmanager/jobs/name=alertmanager/properties/alertmanager/route?
+  value:
+    receiver: default-receiver
+    group_by:
+    # FIXME: not sure about these values
```
richardTowers: `# FIXME`

richardTowers:
🤦‍♂️  was just about to say "this didn't work in my pipeline", but I deployed the wrong branch 🤦‍♂️ 

I'll let it deploy overnight and see how it looks in the morning.

manifests/prometheus/alerts.d/600-syslog_drains.yml
```diff
@@ -0,0 +1,25 @@
...
+  path: /instance_groups/name=prometheus2/jobs/name=prometheus2/properties/prometheus/custom_rules?/-
+  value:
+    name: syslog-drains
+    rules:
+    - alert: CFSyslogDrainsWarning
+      expr: firehose_value_metric_cf_syslog_drain_scheduler_drains > 250
+      labels:
+        deployment: ((metrics_environment))
+        severity: warning
+        notify: email,pagerduty
```
richardTowers: Remove pagerduty

PR #1640, opened 2018-11-26
[#161387488] Send metrics to Prometheus from paas-metrics

bandesz:
# What

We extended the paas-metrics app to send metrics to Prometheus while still sending the metrics to Datadog.

Changes:
 * Added a /metrics endpoint which exposes the Prometheus metrics for scraping
 * We added a Unit field to the collected metrics, so we can use it as a suffix for the Prom. metrics (e.g. if the Unit is "seconds" and the metric is "foo.bar", then the Prom. metric name will be "paas_foo_bar_seconds")
 * The CF_SKIP_SSL_VALIDATION env var was removed from the app as it's not needed anymore
 * Use Go 1.11

# How to review

1. Deploy your CF pipeline from this branch with DataDog enabled.
  ```
  ENABLE_DATADOG=true BRANCH=paas_metrics_prometheus_161387488 make dev pipelines
  ```

2. Check if the metrics app is serving the metrics:

```
https://paas-metrics.${DEPLOY_ENV}.dev.cloudpipelineapps.digital/metrics
```

3. Check if Prometheus has the new metrics (prefixed with `paas_`)

```
https://prometheus.${DEPLOY_ENV}.dev.cloudpipeline.digital/graph?g0.range_input=1h&g0.expr=paas_aiven_estimated_cost_pounds&g0.tab=0
```

4. Check if these metrics are also available in DataDog for your env.

# Who can review

Not me.


bandesz:
> Despite the PR comment, I _shall_ Review this :-)

what PR comment? ;)

jpluscplusm:
> what PR comment? ;)

`Who can review` ... :-P

tools/metrics/gauges_cf.go
```diff
@@ -52,30 +52,35 @@ func QuotaGauge(c *Client, interval time.Duration) MetricReadCloser {
...
 				Name:  "op.quota.memory.allocated", // memory allocated to apps
 				Value: float64(allocatedMemory),
+				Unit:  "megabytes",
 			},
 			{
 				Kind:  Gauge,
 				Time:  time.Now(),
 				Name:  "op.quota.routes.reserved", // number of routes reserved
 				Value: float64(reservedRoutes),
+				Unit:  "megabytes",
```
jpluscplusm:
```suggestion
				Unit:  "routes",
```

or 

```suggestion
				Unit:  "count",
```

... though do we want to get into a metric taxonomy discussion? Probably not :-)

tools/metrics/main.go
```diff
@@ -17,54 +21,81 @@ import (
...
+	}
+
+	return 8080
+}
+
+func runHTTPServer(port int, metricsHandler http.Handler) {
+	addr := fmt.Sprintf(":%d", port)
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "max-age=0,no-store,no-cache")
```
jpluscplusm: Perhaps include `s-maxage=0` here, too, as we're behind CloudFront.

tools/metrics/main.go
```diff
@@ -86,24 +117,38 @@ func Main() error {
 		}, 30*time.Second),
 		CDNTLSValidityGauge(logger, tlsChecker, cfs, 1*time.Hour),
 		ElasticCacheInstancesGauge(logger, ecs, 5*time.Minute),
+		//TestCounterGauge(10 * time.Second),
```
jpluscplusm:
```suggestion
```

tools/metrics/main.go
```diff
@@ -86,24 +117,38 @@ func Main() error {
...
+			AppKey:        os.Getenv("DATADOG_APP_KEY"),
+			Logger:        logger,
+			BatchSize:     100,
+			BatchInterval: 60 * time.Second,
+			Tags:          []string{"deploy_env:" + os.Getenv("DEPLOY_ENV")},
+		})
+		multiWriter.AddWriter(dataDogreporter)
+	}
+
+	if os.Getenv("DEBUG") == "1" {
```
jpluscplusm: Nice.

tools/metrics/prometheus_reporter.go
```diff
@@ -0,0 +1,87 @@
...
+// NewPrometheusReporter creates a new PrometheusReporter instance
+func NewPrometheusReporter(registry prometheus.Registerer) *PrometheusReporter {
+	return &PrometheusReporter{
+		registry:   registry,
+		metricVecs: map[string]interface{}{},
+	}
+}
+
+// WriteMetrics implements MetricWriter. Metrics are buffered
+// in batches not written immediately to rate limit api calls
```
jpluscplusm: JOOI whose API are we talking about protecting here?

tools/metrics/Gopkg.lock
```diff
@@ -3,11 +3,14 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:79b84ded24b1338dedccd54b7f67ec966407ce4634bd5ccb2560da762619387e"
```
jpluscplusm: I'm not sure quite what to *suggest*, but these SHA->SHA changes (throughout Gopkg.lock) seem quite arbitrary. Can we embed any more info around them, and what they actually reflect? e.g. tags, branches, dates? As potentially the next-person-to-work-on-this, I don't have great visibility of when I should be modifying these (or allowing tooling to modify them) next ... !

tools/metrics/aiven.go
```diff
@@ -91,6 +90,7 @@ func AivenCostGauge(client *AivenClient, interval time.Duration) MetricReadClose
 					Time:  time.Now(),
 					Name:  "aiven.estimated.cost",
 					Value: currentCost,
+					Unit:  "pounds",
```
jpluscplusm: I've checked the Aiven API docs and can't see a unit, here. Are we sure it's £££ not $$$?

tools/metrics/main.go
```diff
@@ -86,24 +117,38 @@ func Main() error {
...
+		multiWriter.AddWriter(StdOutWriter{})
```
jpluscplusm: (Nice.)

tools/metrics/gauges_cf.go
bandesz: This was a mistake, I'll change it to count.

tools/metrics/main.go
bandesz: Originally we weren't, did this change recently? Also this is a copy-paste from the original code and I don't think it matters as this is checked by gorouter internally as part of the http check and there is no cache in play. (Or I'm badly mistaken)

tools/metrics/prometheus_reporter.go
bandesz: I don't even know where I was copy pasting this from, I'll remove the unrelated bits.

tools/metrics/Gopkg.lock
bandesz: I think these are mostly caused by the newer version of dep and this project wasn't touched in a long time. As the lock file contains internal details we shouldn't care about most of the changes.

bandesz: (Just to make it clear these changes were fully generated by a `dep ensure` command)

tools/metrics/aiven.go
bandesz: I've checked it in the Aiven console and the invoices are in pounds.

tools/metrics/main.go
alext: Minor point - this won't be behind CloudFront. CloudFront is only used by cdn-routes it's not in place for *.cloudapps.digital.

tools/metrics/gauges_cf.go
bandesz: fixed

tools/metrics/main.go
bandesz: I'll leave this as is

tools/metrics/main.go
bandesz: removed

tools/metrics/prometheus_reporter.go
bandesz: updated

bandesz:
@jpluscplusm All comments were resolved.

tools/metrics/main.go
jpluscplusm: @alext Thanks - I'd made an assumption we had traffic parity between BYOD and non-BYOD apps.

bandesz: It's always dangerous to make assumptions, so don't make them :)

tools/metrics/prometheus_reporter.go
```diff
@@ -0,0 +1,88 @@
...
+func NewPrometheusReporter(registry prometheus.Registerer) *PrometheusReporter {
+	return &PrometheusReporter{
+		registry:   registry,
+		metricVecs: map[string]interface{}{},
+	}
+}
+
+// WriteMetrics converts the received metrics into Prometheus metrics.
+// Any metrics are registered in the Prometheus registry the first time they created and we update these metrics later.
+// This was we can manage counters correctly.
```
jpluscplusm:
```suggestion
// This way we can manage counters correctly.
```

jpluscplusm:
LGTM:

- Tests pass locally
- pipeline deployment results in data in the `jonmtest` env's Prometheus
- pipeline deployment results in data in the `jonmtest` env's Datadog metrics
- autodelete pipeline works, overnight

Merging.

PR #1641, opened 2018-11-26
[#161981619] Upload paas-accounts documents from the pipeline

richardTowers:
What
----

paas-accounts has a database containing documents that users should
agree to when they start using PaaS.

Because the documents are uploaded to the API manually in each
environment the environments don't all match.

Currently the terms of service is present in Ireland, but not London.

This adds the essential documents (that you would expect to exist in
every environment) to paas-accounts through the pipeline.

The API call is a PUT, so it should be safe to call it every time the
pipeline runs (although it will overwrite any edits made through the API
between deployments).

How to review
-------------

* Code review
* Run this down a dev pipeline
* Invite a new user in paas-admin for you dev env (use your email with a + on the end)
  and observe that when you log in you see the terms and conditions.

Who can review
--------------

Not @richardtowers

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1968,6 +1968,7 @@ jobs:
                   done
                 fi
 
+                (
                 cd paas-accounts
 
                 ruby -ryaml -e "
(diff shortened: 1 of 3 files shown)


LeePorte:
This currently creates new documents on each pipeline run.

PR #1647, opened 2018-11-28
[#161981619] Upload paas-accounts documents from the pipeline

richardTowers:
What
----

paas-accounts has a database containing documents that users should agree
to when they start using PaaS.

Because the documents are uploaded to the API manually in each environment
the environments don't all match.

Currently the terms of service is present in Ireland, but not London.

This adds the essential documents (that you would expect to exist in every
environment) to paas-accounts through the pipeline.

The API call is a PUT, so it should be safe to call it every time the
pipeline runs (although it will overwrite any edits made through the API
between deployments).

How to review
-------------

* Code review
* Run this down a pipeline and observe that new users have to agree to
  the doc
* Run the post-deploy job a second time to ensure users don't have to
  agree to the doc a second time

Who can review
--------------

Not @richardtowers (perhaps Lee would be best?)

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -348,7 +348,7 @@ resources:
     type: git
     source:
       uri: https://github.com/alphagov/paas-accounts
-      tag_filter: v0.1.0
+      tag_filter: v0.3.0
 
   - name: paas-admin
     type: git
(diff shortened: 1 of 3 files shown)


keymon:
Rebased with master

keymon:
LGTM, tested in my env. 

I rebased, waiting for travis and I will merge.

PR #1650, opened 2018-11-28
[#160175665] Add job to rotate prometheus credentials

keymon:
What
----

We want to be able to rotate the prometheus secrets generated
by bosh during the prometheus deployment.

As we store the secrets in a vars-store.yml file,
we reuse the script used for the cf-manifest:

   manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb

This script requires a version of the manifest without the variables
expanded, so we update the render-manifest.sh script to be able
to render the manifest without variables, and we pass the
right arguments as needed.

The rendered manifest without variables is stored as a resource
in the S3 bucket.

Note that currently the prometheus-vars-store only contains a set
of plain passwords, we do not use the `*_old` vars, and we do not
keep any variable. That is effectivelly like deleting all the
vars-store.yml file. But we decided to use the
rotate-vars-store-secrets.rb script as it is more flexible.

How to review
-------------

For the UAA secrets:
 - Not really required, but you can rotate the cf secrets and then
 redeploy prometheus.

For the prometheus secrets:
 1. Rerun prometheus-deploy
 2. Run `rotate-prometheus-credentials`
 3. Rerun prometheus-deploy to change the secrets
 4. Check that you can login in all services, and that metrics carry on
 working.

Who can review
--------------

Not me

### DIFF
--- a/manifests/prometheus/scripts/generate-manifest.sh
+++ b/manifests/prometheus/scripts/generate-manifest.sh
@@ -21,11 +21,15 @@ for i in ${VARS_FILES}; do
   varsfile_args+="--vars-file $i "
 done
 
+vars_store_args=""
+if [ -n "${VARS_STORE:-}" ]; then
+  vars_store_args=" --var-errs --vars-store ${VARS_STORE}"
+fi
+
 # shellcheck disable=SC2086
 bosh interpolate \
-  --var-errs \
-  --vars-store "${VARS_STORE}" \
   ${varsfile_args} \
   ${opsfile_args} \
   ${alerts_opsfile_args} \
+  ${vars_store_args} \
   "${PROM_BOSHRELEASE_DIR}/manifests/prometheus.yml"
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1120,10 +1120,11 @@ jobs:
 
                   tar -xzf ./ipsec-CA/ipsec-CA.tar.gz -C ./ipsec-CA
                   cp cf-vars-store/cf-vars-store.yml cf-vars-store-updated/
-                  ./paas-cf/manifests/cf-manifest/scripts/generate-manifest.sh \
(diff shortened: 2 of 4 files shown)


manifests/prometheus/scripts/generate-manifest.sh
```diff
@@ -23,9 +23,7 @@ done
 
 # shellcheck disable=SC2086
 bosh interpolate \
-  --var-errs \
-  --vars-store "${VARS_STORE}" \
   ${varsfile_args} \
   ${opsfile_args} \
   ${alerts_opsfile_args} \
-  "${PROM_BOSHRELEASE_DIR}/manifests/prometheus.yml"
+  "${PROM_BOSHRELEASE_DIR}/manifests/prometheus.yml" $@
```
alext: Shellcheck is complaining about this.

I'm not sure I like this pattern where more and more of the options to `bosh interpolate` get moved outside the script. When we were using spruce, we ended up with the manifest generation script being a very small wrapper around spruce with all the important details being out in the pipeline. This meant you had to verify manually that the tests were calling it in the same way as the pipeline...

Can we instead add a feature-flag to control the vars-store behaviour (eg if the `SKIP_VARS_STORE` env var is set, it skips the corresponding flags).

keymon: Sure, I can add the flag :+1: 

I shall do the same for the `cf-manifest.yml`

keymon: @alext have a look, we do the vars-store if the variable `VARS_STORE` is set

keymon: If you are OK, I can do the same in the cf-manifest stuff

alext: Yeah, that makes sense. I like this approach, although we don't need the `$@` on the last line any more...

keymon:
@alext have a look now, pls

PR #1652, opened 2018-11-29
[#161184357] Use Terraform to create ACM certs.

alext:
What
----

This moves the ACM cert management into Terraform because it's much more robust than our set of scripts.

Previously, the apps cert was generated by paas-bootstrap so that we didn't need to duplicate the scripts into this repo. That's no longer a concern when using Terraform, so it makes sense for the apps cert to be generated here as this is where it's used...  The system domain cert is used by both this repo and paas-bootstrap, so that remains in paas-bootstrap.

This additionally includes a temporary commit to import the existing apps cert into Terraform to prevent it attempting to create a second one.

How to review
-------------

* https://github.com/alphagov/paas-bootstrap/pull/222 should be reviewed and merged/deployed first.
* Run the pipeline from this branch
* Verify terraform updates correctly and doesn't create a second apps domain cert.
* Run the pipeline a second time to verify the import process is idempotent.

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -839,6 +839,39 @@ jobs:
 
                 cat vpc-peering-tfvars/vpc-peers.tfvars
 
+      # FIXME: remove this when it's run everywhere.
+      - task: extract-acm-cert-arn
+        config:
+          platform: linux
+          image_resource: *awscli-image-resource
+          params:
+            AWS_DEFAULT_REGION: ((aws_region))
+          inputs:
+            - name: cf-tfstate
+          outputs:
+            - name: apps-cert
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -839,6 +839,34 @@ jobs:
...
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
+              tf_cert_details=$(jq -r '.modules[].resources."aws_acm_certificate.apps"' cf-tfstate/cf.tfstate)
+              if [ "${tf_cert_details}" = "null" ]; then
+                echo "Finding cert to import"
+                aws acm list-certificates \
```
bandesz: Creating a new environment would fail, because the `aws acm list-certificates` returns an empty string if the certificate doesn't exist. We have to handle this case too.

alext: Added a check for non-existing cert.

PR #1661, opened 2018-12-06
[#162303736] Deploy the yace CloudWatch exporter and scrape it with Prometheus

bandesz:
# What

Deploy https://github.com/alphagov/paas-yet-another-cloudwatch-exporter to collect CloudWatch metrics and make them available for Prometheus.

Additional changes:
 - add the `deploy_env` tag to all EC2 instances in the CF and Prometheus Bosh deployments (this is not used yet as the instances needs to be recreated to get the new tag)

For now we use the `deployment` EC2 tag for filtering, but in dev one env will get the metrics from all Prometheus, Concourse and Bosh instances. Later if a stemcell update recreates all instances we can depend on the new `deploy_env` tag.

# How to review

❕ https://github.com/alphagov/paas-aws-account-wide-terraform/pull/144 - already reviewed and merged.

1. Review: https://github.com/alphagov/paas-bootstrap/pull/225
2. Run the create-cloudfoundry pipeline from this branch
3. You should see new metrics, like `aws_ec2_cpucreditbalance_minimum` in your Prometheus
4. Check if all relevant RDS/EC2 instance metrics are reported
5. Test the key rotation by running the "expire-aws-keys" job and run the pipeline. In the cf-terraform job a new key pair should be generated and the app should be redeployed successfully.

# Who can review

Anyone

### DIFF
--- a/concourse/tasks/forget-access-keys.yml
+++ b/concourse/tasks/forget-access-keys.yml
@@ -26,4 +26,8 @@ run:
         aws_iam_access_key.metrics_exporter
       echo "metrics exporter access key no longer managed by terraform."
 
+      terraform state rm -state=updated-tfstate/cf.tfstate \
+        aws_iam_access_key.yace
+      echo "yace access key no longer managed by terraform."
+
       echo "next successful deployment run will delete the unused keys"
--- a/config/cloudwatch-exporter/config.yml
+++ b/config/cloudwatch-exporter/config.yml
@@ -1,0 +1,35 @@
+---
+discovery:
+  exportedTagsOnMetrics:
+    ec2:
+      - Name
+    rds:
+      - Name
+  jobs:
+    - type: "ec2"
+      region: (( grab $AWS_REGION ))
+      searchTags:
+        - Key: deploy_env
+          Value: (( concat "^" $DEPLOY_ENV "$" ))
+      metrics:
+        - name: CPUCreditBalance
+          statistics:
+            - Minimum
+          period: 60
+          length: 600
+    - type: "rds"
+      region: (( grab $AWS_REGION ))
+      searchTags:
+        - Key: deploy_env
+          Value: (( concat "^" $DEPLOY_ENV "$" ))
+      metrics:
+        - name: FreeStorageSpace
+          statistics:
+            - Minimum
+          period: 60
+          length: 600
+        - name: CPUCreditBalance
+          statistics:
+            - Minimum
+          period: 60
+          length: 600
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -377,6 +377,12 @@ resources:
     source:
       service_key: ((pagerduty_integration_key))
 
+  - name: paas-yet-another-cloudwatch-exporter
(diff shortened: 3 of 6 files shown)


concourse/tasks/forget-access-keys.yml
```diff
@@ -26,4 +26,8 @@ run:
         aws_iam_access_key.metrics_exporter
       echo "metrics exporter access key no longer managed by terraform."
 
+      terraform state rm -state=updated-tfstate/cf.tfstate \
+        aws_iam_access_key.yace
+      echo "yace access key no longer managed by terraform."
+
```
alext: We also need to add a corresponding section to the `remove-unused-ses-smtp-access-keys` task in the pipeline. This is also currently badly named because it currently handles both the ses-smtp user and the metrics user.

config/cloudwatch-exporter/config.yml
```diff
@@ -0,0 +1,35 @@
...
+    ec2:
+      - Name
+    rds:
+      - Name
+  jobs:
+    - type: "ec2"
+      region: $AWS_REGION
+      searchTags:
+        - Key: deployment
+          Value: ^($DEPLOY_ENV|bosh|concourse|prometheus)$
```
alext: With the changes to add the `deploy_env` tag, can we now filter on that instead of this set of deployments?

config/cloudwatch-exporter/config.yml
```diff
@@ -0,0 +1,35 @@
+---
+discovery:
+  exportedTagsOnMetrics:
+    ec2:
+      - Name
+    rds:
+      - Name
+  jobs:
+    - type: "ec2"
+      region: $AWS_REGION
```
alext: I think it would be clearer that this is replaced with `sed` if we used a different delimiter (ie not a `$`).  Maybe something like `__AWS_REGION__`.

bandesz: We'll use spruce as it's available in the Docker image.

config/cloudwatch-exporter/config.yml
bandesz: Updated.

concourse/tasks/forget-access-keys.yml
bandesz: I renamed the task to `remove-unused-iam-access-keys` and added the extra command for yace.

bandesz:
@alext I addressed all comments and I'll test these on my dev env.

config/cloudwatch-exporter/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      searchTags:
+        - Key: deploy_env
+          Value: (( concat "^" $DEPLOY_ENV "$" ))
+      metrics:
+        - name: FreeStorageSpace
+          statistics:
+            - Minimum
+          period: 60
+          length: 600
+        - name: CPUCreditUsage
```
alext: We want the balance, not the usage here...

bandesz: Fixed

alext:
This is missing the alerts. We can add this in a follow-up PR though. This is good to go.

PR #1662, opened 2018-12-06
[#162199780] Alert on average total diego-cell memory used

jpluscplusm:
What
----

In `528417be` we record a metric and alert on it per-diego-cell, but we actually want to alert on the average *aggregate* memory that the cells have free.

We can do this by recording a new derived metric (per-VM, recorded for all VMs regardless of purpose), averaging it across all our diego cells and comparing the resulting percentage against the existing 50% and 55% thresholds.

NB This is only valid as long as our diego cells are (memory-)homogenous!

How to review
-------------

- code review
- push to a dev env (`jonmtest` has it running currently, if you want to speed this up)
  - maybe push a `[DELETE]` commit with 100%/99% thresholds to force it to fire. `jonmtest` env has already done this, however. Your call.

Who can review
--------------

Anyone except @jpluscplusm.

### DIFF
--- a/manifests/prometheus/alerts.d/diego_cell_available_memory.yml
+++ b/manifests/prometheus/alerts.d/diego_cell_available_memory.yml
@@ -5,27 +5,21 @@
   value:
     name: BoshDiegoCellAvailableMemory
     rules:
-      - record: bosh_job:mem_percent:diego_cell:avg5m
-        expr: 100 - avg_over_time(bosh_job_mem_percent{environment="((metrics_environment))",bosh_job_name="diego-cell"}[5m])
       - &alert
         alert: BoshDiegoCellAvailableMemory_Warning
-        expr: bosh_job:mem_percent:diego_cell:avg5m < 55
+        expr: avg(bosh_job_mem_percent{bosh_job_name="diego-cell"}) >= 45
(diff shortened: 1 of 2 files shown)


manifests/prometheus/alerts.d/diego_cell_available_memory.yml
```diff
@@ -5,27 +5,34 @@
...
+            {{ $labels.bosh_job_name }}
             memory free was low {{ $value | printf "%.0f" }}%
             in the last two hours on average.
             Review if we need to scale...
           url: https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR021-cell-capacity-assignment-2/#decision
 
       - <<: *alert
         alert: BoshDiegoCellAvailableMemory_Critical
-        expr: bosh_job:mem_percent:diego_cell:avg5m < 50
+        expr: >-
```
bandesz: this should be similarly replaced: `avg(bosh_job_mem_percent{bosh_job_name="diego-cell"}) >= 50`

bandesz:
I added the suggested changes and also fixed the cell idle cpu alert, as it had the same problem

bandesz:
Ready to be reviewed

PR #1664, opened 2018-12-07
[#161511720] Apps Apex domain DNS and Cert SAN checks

tnwhitwell:
What
----

Added some tests to check that the CDN broker's assumptions about the apps apex domain are valid:
  - DNS A records are the same as the `apps.` domain
  - Certificate for apex domain is valid, and contains the correct SAN

How to review
-------------

- Code Review
- Ensure tests pass in pipeline (Use [this](https://deployer.tnw.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/8) as an example if you want!)

Who can review
--------------

Not @tnwhitwell / @richardTowers

### DIFF
--- a/platform-tests/src/platform/acceptance/init_test.go
+++ b/platform-tests/src/platform/acceptance/init_test.go
@@ -41,7 +41,9 @@ func TestSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	testConfig, err = config.NewCatsConfig(os.Getenv("CONFIG"))
-	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	httpClient = &http.Client{
 		Transport: &http.Transport{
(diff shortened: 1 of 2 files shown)


platform-tests/src/platform/acceptance/init_test.go
```diff
@@ -41,7 +41,9 @@ func TestSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	testConfig, err = config.NewCatsConfig(os.Getenv("CONFIG"))
-	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
```
bandesz: what is the reason for this change?

bandesz: as it was explained the Ginkgo/Gomega suite is not initialised at this point yet and the Expect will fail miserably.

PR #1665, opened 2018-12-07
[#SUPPORT] Diego cell available memory alert uses average

keymon:
What
----

In the original monitor in datadog we were averaging across all
cells, not checking each cell.

How to review
-------------

Code review

Who can review
--------------

not me


keymon:
DUPE, closing

PR #1670, opened 2018-12-10
[#162448351] Finalise Prometheus emails

bandesz:
# What

Changes:
 * Group the email alerts by alert name
 * Set the email subject to `[<env>] [<severity>] <alert name>`
 * Send alert emails by default
 * Separate receivers for warnings/critical alerts
 * Disable emails for resolved alerts
 * Remove the now unneeded `notify` labels from our alerts
 * Set the external urls for Prometheus/Alertmanager/Grafana which means the links in Alertmanager and the alert emails are working now 

What I didn't change:
 * I didn't touch the email body as it didn't seem to worth the effort for only some minor optimisations
 * I didn't add a way to disable email alerts - mostly because there is no need for it currently and maybe a more appropriate action is to remove the alert. (We can still silence one temporarily in alertmanager)

Up for debate:
 * All alerts should have the severity tag, but ones without one will trigger a critical alert to be on the safe side.
 * The alerts will be repeated only every 24h as an insurance.

# How to review

1. Deploy Prometheus from this branch
2. Trigger an alert (e.g. stop a process with monit on one of your instances and wait 5 minutes) and check if it looks ok (govpaas-alerting-dev list).
3. Check if all alert related links work correctly in Alertmanager and in the emails
4. Resolve the alert and check whether you get a resolve email (you shouldn't).

# Who can review

Not me.

### DIFF
--- a/manifests/prometheus/alerts.d/aiven-cost.yml
+++ b/manifests/prometheus/alerts.d/aiven-cost.yml
@@ -12,7 +12,6 @@
         for: 1h
         labels:
           severity: warning
-          notify: email
         annotations:
           summary: "Aiven estimated cost is high, possibly need to alert finance."
           description: "The estimated monthly cost of Aiven is currently £{{ $value | printf \"%.2f\" }}, possibly need to alert finance."
@@ -22,4 +21,3 @@
         expr: paas_aiven_estimated_cost_pounds > 3100
         labels:
           severity: critical
-          notify: email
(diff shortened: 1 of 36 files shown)


tnwhitwell:
Deploying in my environment, and will merge when it has run

tnwhitwell:
Confirmed working - stopped cdn-broker, and received an alert 6 minutes later on the mailing list.

PR #1680, opened 2018-12-12
[#162604166] Add a test to check we only offer TLS 1.2

keymon:
What
----

We want to ensure that our SSL endpoints for apps and system
only offer TLS 1.2.

We add a custom acceptance test that would check for all the old
disabled protocols, and expect failure.

How to review
-------------

Code review, test by changing the policy of the ELB in your environment.

Who can review
--------------

Not me

### DIFF
--- a/platform-tests/src/platform/acceptance/tls_version_test.go
+++ b/platform-tests/src/platform/acceptance/tls_version_test.go
@@ -1,0 +1,83 @@
+package acceptance_test
+
+import (
+	"crypto/tls"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Checking the TLS version", func() {
+
+	testHostPortGetters := map[string]func() string{
+		"apps_apex": func() string {
+			return testConfig.GetAppsDomain() + ":443"
+		},
+		"apps": func() string {
+			return "healthcheck." + testConfig.GetAppsDomain() + ":443"
+		},
+		"system": func() string {
+			return testConfig.GetApiEndpoint() + ":443"
+		},
+	}
+
+	for domainClass, hostPortGetter := range testHostPortGetters {
+		domainClass, hostPortGetter := domainClass, hostPortGetter
+
+		Context(domainClass+" domain", func() {
+			It("should not allow SSL 3.0", func() {
+				tlsConfig := &tls.Config{
+					MinVersion: tls.VersionSSL30,
+					MaxVersion: tls.VersionSSL30,
+				}
+				hostPort := hostPortGetter()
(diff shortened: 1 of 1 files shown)


platform-tests/src/platform/acceptance/tls_version_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Checking the TLS version", func() {
+
+	testHostPortGetters := map[string]func() string{
+		"apps_apex": func() string {
+			return testConfig.GetAppsDomain() + ":443"
```
alext: This block is running outside an `It` or a `Before*`, and therefore is executed before the `BeforeSuite` has run. I would therefore expect `testConfig` to be nil at this point.  Additionally `Expect` calls can't be used in this context - if they fail, they cause a runtime panic because there's nothing in place to handle them.

alext: Actually, scratch that. Now that I read this more closely I understand what's going on, and this is returning a function that is then executed in the relevant context.

keymon: ?

jpluscplusm:
Given the subtle and relatively invisible bug this is addressing, might it be worth including some known-bad endpoints to assert the correctness of the tests?

badssl.com, for instance, exposes these:

- https://tls-v1-0.badssl.com:1010/
- https://tls-v1-1.badssl.com:1011/
- https://tls-v1-2.badssl.com:1012/

They could be added to the `testHostPortGetters` map, possibly optionally included by an envvar in case the badssl service disappears.

platform-tests/src/platform/acceptance/tls_version_test.go
keymon: You crossed your text. Answering anyway, it is a map of getters, functions that return a string. 

The reason is that the values of `testConfig` are set in the `BeforeSuite` as you mention, so I need to delay the evaluation. Each test call the getter to get the right value.

keymon:
> Given the subtle and relatively invisible bug this is addressing, might it be worth including some known-bad endpoints to assert the correctness of the tests?
> https://badssl.com, for instance, exposes these:

That is a good Idea, I will refactor the test to do so

keymon:
> > Given the subtle and relatively invisible bug this is addressing, might it be worth including some known-bad endpoints to assert the correctness of the tests?
> > https://badssl.com, for instance, exposes these:
> 
> That is a good Idea, I will refactor the test to do so

At the end we decided that test the test is too much and we were doing it too complicated, we will rely in the code review of this PR.

PR #1685, opened 2018-12-14
[#162604166] Add Doppler and concourse to TLS testing

LeePorte:
What
----

This PR adds in the doppler and concourse endpoints to the TLS tests. This is to verify that our ELBs are configured with the correct policies.

How to review
-------------

Run it through dev

Who can review
--------------

@alext

### DIFF
--- a/platform-tests/src/platform/acceptance/tls_version_test.go
+++ b/platform-tests/src/platform/acceptance/tls_version_test.go
@@ -8,7 +8,6 @@ import (
 )
 
 var _ = Describe("Checking the TLS version", func() {
-
 	testHostPortGetters := map[string]func() string{
 		"apps_apex": func() string {
 			return testConfig.GetAppsDomain() + ":443"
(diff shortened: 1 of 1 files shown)


platform-tests/src/platform/acceptance/tls_version_test.go
```diff
@@ -19,6 +18,12 @@ var _ = Describe("Checking the TLS version", func() {
 		"system": func() string {
 			return testConfig.GetApiEndpoint() + ":443"
 		},
+		"doppler": func() string {
+			return "doppler." + GetConfigFromEnvironment("SYSTEM_DNS_ZONE_NAME") + ":443"
+		},
+		"concourse": func() string {
+			return "concourse." + GetConfigFromEnvironment("SYSTEM_DNS_ZONE_NAME") + ":443"
```
alext: This should be `deployer` instead of `coucourse`, otherwise it's testing the system router domain (due to the wildcard DNS)

PR #1686, opened 2018-12-14
Fix quoting in some alert descriptions

alext:
## What

They were producing the following error when rendering:
```
<error expanding template: error parsing template __alert_GorouterLatency_Warning: template: __alert_GorouterLatency_Warning:1: unexpected "\\" in operand>
```
The string isn't quoted, and yet the embedded quotes are escaped
resulting in them being rendered in the config file as:
```
Gorouter latency is too high ({{ $value | printf \\\"%.0f\\\" }} ms\") on {{ $labels.bosh_job_ip }}
```
There's also a spurious " in the GoRouter alert.

How to review
-------------

Who can review
--------------

Not me.


jpluscplusm:
LGTM.

PR #1687, opened 2018-12-18
[#162381976] Remove DataDog part 1: Remove all monitors

alext:
## What

In order to remove DataDog without causing loads of noise this runs a
DataDog terraform destroy to remove all the monitors ahead of upcoming
changes to remove the agents etc that generate the metrics.

This also removes the config step to genetate the list of job_instances
that's only used in the datadog terraform apply task.

How to review
-------------

I've tested this in my dev environment, so code review is probably enough.

Who can review
--------------

Not me.


keymon:
LGTM

PR #1692, opened 2018-12-20
[#158179141] Implement rotation of DB encryption keys for CC and UAA

keymon:
What
----

Cloud Foundry is gradually starting to support multiple encryption keys for easier rotation. There are two types of encryption key for which this is supported we know of:

1. [UAA 4.13.0 introduced a new, required, passphrase for encryption keys used to encrypt UAA's MFA tokens](https://github.com/cloudfoundry/uaa-release/releases/tag/v58).
2. [CF release 2.9.0 changed](https://github.com/cloudfoundry/cf-deployment/releases/tag/v2.9.0) how the cloudcontroller database encryption key is specified.

Both of these support specifying multiple keys.

In this PR we implement what is needed to support the rotation of these
credentials:

 1. We add a new dedicated job `rotate-db-encryption-key`
 2. Configure the required errands to rotate the secrets in DB.
 3. Configure the manifest to support multiple encryption keys (old and new)
 4. Add logic to migrated from the default keys provided by cf-deplyoment.

To support the rotation, we use a strategy similar to the one used for
jwt verification keys for UAA[1], where the label/name of each key
is a password itself. That way we rotate both the label/name and
the value of the key in the rotation scripts.

Comments:

 - We add a dedicated job for rotation of DB keys, as it would be run
   less frequently and might be more complicated to troubleshoot. We
   had to extend the rotate-vars-store script.

 - CC rotation keys are configured in several jobs/vms. We change all
   and we add a test to double check that the key is not used in any
   new place.

 - For UAA we still do not have any encrypted value, as it is used for
   MFA. Because that I do not provide any way to check that the values
   are updated in DB, it should be checked in the future when MFA are
   implemented.

 - The rotation would be done with dedicated errands, that would
   create a VM in each run. But the bosh flag `--when-changed` is used
   so the errand only runs on config changes.

[1] https://github.com/alphagov/paas-cf/pull/1184

How to review
-------------

Code review.

Also testing it your environment. You can check how the keys change
in DB for the API DB.

The easiest is hijack the `init-db` job when `cf-terraform` is running

```
cd paas-cf
./bin/fly -t hector hijack -j create-cloudfoundry/cf-terraform -s init-db sh

# Then run
. ./cf-terraform-outputs/cf.tfstate.sh

. ./terraform-variables/cf-secrets.tfvars.sh

export PGPASSWORD=${TF_VAR_secrets_cf_db_master_password}


psql -h "${TF_VAR_cf_db_address}" -U dbadmin -d postgres <<EOF
\connect api

SELECT encryption_key_label, count(*) FROM "apps" group by encryption_key_label;

\connect uaa

select encryption_key_label, count(*) FROM user_google_mfa_credentials group by encryption_key_label;

EOF

```

As said, UAA would have no entries.

To test it:

 1. check the state in DB. It should be using the default keys.
 2. Deploy this pipeline once. Check how the errands run in
 `post-deploy`
 3. Check that the state in DB changed and it is using new keys.
 4. rerun `post-deploy`, the errands should not run.
 5. Run the  `rotate-db-encryption-keys` job in `credentials` group.
 6. Rerun all the pipeline, the errands should run.
 7. Check that the state in DB has changed again.

Dependencies
------------

A follow up PR is required to stop using the default keys from
cf-deployment. It would basically revert the latest commit.

https://github.com/alphagov/paas-cf/pull/1693

Who can review
--------------

Not me

### DIFF
--- a/manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
+++ b/manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
@@ -24,6 +24,9 @@ def parse_args
   parser.on("--preserve VAR", "variables to not rotate") { |v|
     options[:vars_to_preserve] << v
   }
+  parser.on("--rotate VAR", "variables to do rotate. Will rotate all if not set.") { |v|
+    options[:vars_to_rotate] = (options[:vars_to_rotate] || []) << v
+  }
   parser.parse!
 
   if options[:vars_store].nil? || options[:manifest].nil?
@@ -75,10 +78,13 @@ def rotate(manifest, vars_store,
            passwords: false,
            rsa: false,
            ssh: false,
+           vars_to_rotate: nil,
            vars_to_preserve: [],
            delete: false)
 
-  vars = manifest.fetch("variables").reject { |v| vars_to_preserve.include?(v["name"]) }
+  vars = manifest.fetch("variables")
+  vars = vars.select { |v| vars_to_rotate.include?(v["name"].gsub(/_old$/, '')) } unless vars_to_rotate.nil?
+  vars = vars.reject { |v| vars_to_preserve.include?(v["name"]) }
 
   if delete
     return delete_old(vars, vars_store)
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2310,32 +2310,58 @@ jobs:
                 export BOSH_CLIENT_SECRET
 
                 bosh -n clean-up --all
+      - do:
+        - task: rotate-cc-db-encryption-keys
+          config:
+            platform: linux
+            image_resource: *gov-paas-bosh-cli-v2-image-resource
+            inputs:
+              - name: paas-cf
+              - name: bosh-secrets
+              - name: bosh-CA-crt
+            params:
+              BOSH_ENVIRONMENT: ((bosh_fqdn))
+              BOSH_CA_CERT: bosh-CA-crt/bosh-CA.crt
+              BOSH_DEPLOYMENT: ((deploy_env))
+            run:
+              path: sh
+              args:
+                - -e
+                - -c
+                - |
+                  VAL_FROM_YAML=$(pwd)/paas-cf/concourse/scripts/val_from_yaml.rb
+                  BOSH_CLIENT=admin
+                  BOSH_CLIENT_SECRET=$($VAL_FROM_YAML secrets.bosh_admin_password bosh-secrets/bosh-secrets.yml)
+                  export BOSH_CLIENT
+                  export BOSH_CLIENT_SECRET
 
-      - task: rotate-cc-db-encryption-keys
-        config:
-          platform: linux
-          image_resource: *gov-paas-bosh-cli-v2-image-resource
-          inputs:
-            - name: paas-cf
-            - name: bosh-secrets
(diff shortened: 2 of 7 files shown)


manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
```diff
@@ -24,6 +24,9 @@ def parse_args
   parser.on("--preserve VAR", "variables to not rotate") { |v|
     options[:vars_to_preserve] << v
   }
+  parser.on("--rotate VAR", "variables to do rotate. Will rotate all if not set.") { |v|
+    options[:vars_to_rotate] = (options[:vars_to_rotate] || []) << v
```
bandesz: would be nicer to initialise vars_to_rotate on line 13, as we do it with vars_to_preserve

manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
```diff
@@ -75,10 +78,13 @@ def rotate(manifest, vars_store,
            passwords: false,
            rsa: false,
            ssh: false,
+           vars_to_rotate: nil,
```
bandesz: should this be have a default value of "[]"? I expect we would get an error if we call `<nil>.include?'

manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
```diff
@@ -75,10 +78,13 @@ def rotate(manifest, vars_store,
...
            vars_to_preserve: [],
            delete: false)
 
-  vars = manifest.fetch("variables").reject { |v| vars_to_preserve.include?(v["name"]) }
+  vars = manifest.fetch("variables")
+  vars = vars.select { |v| vars_to_rotate.include?(v["name"].gsub(/_old$/, '')) } unless vars_to_rotate.nil?
```
bandesz: shouldn't we throw an error if a variable ends with `_old` instead of silently replacing it?

keymon: That code picks any variable that is defined in the `variables` that is in the list of `vars_to_rotate`, and also the corresponding `_old` if exists.

For instance:

```
vars= [{'name'=>'one'},{ 'name'=>'two'}, {'name'=>'one_old'}]
vars_to_rotate=['one']

vars = vars.select { |v| vars_to_rotate.include?(v["name"].gsub(/_old$/, '')) } unless vars_to_rotate.nil?
```

That picks: `[{"name"=>"one"}, {"name"=>"one_old"}]`

manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
keymon: There is a `unless vars_to_rotate.nil?` just before.

being `vars_to_rotate=nil` means rotate all the vars.

manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
bandesz: Ah ok, I understand now, thanks.

manifests/cf-manifest/scripts/rotate-vars-store-secrets.rb
bandesz: we can leave this as is

PR #1694, opened 2018-12-20
[#162265546] Clean up left-over CF users

bandesz:
# What

We have a lot of users in CF which don't exist in UAA. These users don't have a username field in the CF API response.

Most of these users are created by the continuous smoke tests. When the tests finish we delete the user with itself but we failed to recognise that the user is not fully deleted.

We'll change the helper tasks to delete the temporary admin user with the global admin.

If the admin user's credentials are rotated the smoke tests could fail, therefore we make sure the creds rotation and the smoke tests don't run at the same time and also the continuous smoke tests always use the latest credentials/manifest variables.

I created a script to clean up the left-over users under `scripts/clean_up_cf_users.sh`.

A user:
 * will be skipped if a guid is not a UUID (some UAA clients)
 * will be skipped if it has a username
 * will be deleted otherwise

# How to review

1. Update your pipeline and run the continuous smoke tests
2. Run the script without any options on your dev and check if the numbers match.
3. Run the script with the `--delete` option on your dev env

After the clean up no left-over users should be present and no new ones should be created.

# Who can review

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -71,6 +71,7 @@ groups:
       - bump-patch-version
       - pipeline-check-lock
       - pipeline-release-lock
+      - cleanup-deleted-cf-users
   - name: tests
     jobs:
       - app-availability-tests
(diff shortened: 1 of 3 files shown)


scripts/clean_up_cf_users.sh
```diff
@@ -0,0 +1,93 @@
...
+
+if [ "${delete}" == "yes" ]; then
+  cnt=$(echo "${all_guids_to_delete}" | wc -l | tr -d ' ')
+  echo
+  echo "Deleting ${cnt} users"
+  echo
+
+  while read -r guid; do
+    echo "Deleting ${guid}"
+    cf curl -X DELETE "/v2/users/${guid}"
```
keymon: I feel a little bit nervous about this, I suggest we add a small safety check like this one:

```diff
$ git diff
diff --git a/scripts/clean_up_cf_users.sh b/scripts/clean_up_cf_users.sh
index 7ab633f6..fb72337c 100755
--- a/scripts/clean_up_cf_users.sh
+++ b/scripts/clean_up_cf_users.sh
@@ -80,7 +80,11 @@ if [ "${delete}" == "yes" ]; then
 
   while read -r guid; do
     echo "Deleting ${guid}"
-    cf curl -X DELETE "/v2/users/${guid}"
+    if [ "$(cf curl "/v2/users/${guid}/organizations" | jq -r .total_results)" == "0" ]; then
+      cf curl -X DELETE "/v2/users/${guid}"
+    else
+      echo "Not deleting ${guid}, it is member of some organisations"
+    fi
   done <<< "${all_guids_to_delete}"
 
   echo
```

keymon:
> I am not sure about removing the dependency of continuous smoke tests to the deployment jobs, as we might want to run based on the latest manifest deployed successfully. We should discuss this.

After talking offline about this, I agree that makes sense. @bandesz pointed that we barely rely on the `manifest.yml` but to get the UAA endpoint and the admin password, all to create a temporary user. Also, the admin password is reseted immediately if the rotate job runs, so it makes sense to get the latest. @bandesz suggested we could generate the UAA url based on the system domain.

keymon:
We discussed to add a task to run this job from concourse.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3338,9 +3338,7 @@ jobs:
         - get: paas-cf
           passed: ['pipeline-lock']
         - get: cf-manifest
-          passed: ['post-deploy']
```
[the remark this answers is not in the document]
bandesz: It doesn't really matter as we only get the system domain and the uaa url from it. It left it untouched eventually.

scripts/clean_up_cf_users.sh
bandesz: It would be really slow and there are some (probably manually) deleted users who still have an organisation.

bandesz:
I created a Concourse task for this, ready to be reviewed again.

keymon:
Happy to merge this. Pending to run that manual step.

PR #1700, opened 2019-01-03
[#162753042] Use cf-deployment.yml manifest to check buildpacks

keymon:
What
----

We had a small script to check that all the expected buildpacks
defined in the manifest are deployed, but we do not install
the buildpacks using manifests anymore.

We will replace the manifest used for reference, and pull from
the upstream cf-deployment.yml. We expect the same buildpacks
are installed.

How to review
-------------

Code review

Who can review
--------------

Not me


keymon:
It passed here: https://deployer.hector.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/35

PR #1713, opened 2019-01-16
[#163250921] Add larger Redis instance to plans

LeePorte:
What
----

DiT have requested larger Redis instances to be available in the London
region. This PR enables that to happen.

How to review
-------------

Code review
Deploy from this branch, enable redis access and then create a `redis medium-clustered-3.2` service. Bind it to an app and check the env vars after restaging it should be using a `clustercfg.<instance ID>.x.x.cache.amazonaws.com:6379` address

Who can review
--------------

Not @LeePorte


config/billing/pricing_data.yml
```diff
@@ -33,6 +33,7 @@ eu-west-1:
   rds_mysql_m4_4xlarge: "1.545"
   rds_mysql_m4_4xlarge_ha: "3.09"
   elasticache_t2_micro: "0.018"
+  elasticache_m5_large: "0.172"
```
alext: You'll also need to update the `config.json.erb` template to add a new plan that references these values.

manifests/cf-manifest/operations.d/730-elasticache-broker.yml
```diff
@@ -71,6 +71,12 @@
                       free: true
                       metadata:
                         displayName: Redis Non-clustered Tiny
+                    - id: b6949ea7-5c98-4c69-b981-4b5318ea8b7c
+                      name: medium-clustered-3.2
+                      description: 6.37GB RAM, clustered (1 shard), 3 nodes, failover, daily backups
```
alext: Do we need 3 nodes, or would 2 be sufficient? This will presumably have a cost impact.

config/billing/pricing_data.yml
LeePorte: I've updated the config template now.

tnwhitwell:
Deployed from the branch, and confirm the instance has been created as expected, and env vars are correct.

manifests/cf-manifest/operations.d/730-elasticache-broker.yml
LeePorte: It will have a cost impact, though I'm happy with 3 as the request has been about both size and failover capability.

alext: I was thinking that 2 should be enough for failover (with RDS the HA instances only have 2 nodes, I was expecting this would be similar)

LeePorte: 3 prevents a split brain scenario in the event of network partitioning though.

manifests/cf-manifest/operations.d/730-elasticache-broker.yml
```diff
@@ -93,6 +99,16 @@
...
                   maxmemory-policy: volatile-lru
                   reserved-memory: '0'
+              b6949ea7-5c98-4c69-b981-4b5318ea8b7c:
+                instance_type: cache.m5.large
+                replicas_per_node_group: 2
+                shard_count: 1
+                snapshot_retention_limit: 7
+                automatic_failover_enabled: true
+                parameters:
+                  cluster-enabled: 'yes'
```
alext: I don't think this is what we want. I think what we want for this plan is non-clustered Redis with a single replia to allow automatic failover.

Cluster mode looks to be about having multiple node groups, each with a master and replicas for use when you're sharding things (https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis-RedisCluster.html). IIRC the only reason we enabled cluster mode for the initial plan was because AWS don't support backups for non-clustered t2.micro instances. Clustered mode doesn't play nicely with things like Sidekiq, hence we had to create the non-backed up unclustered plan.

alext: I note that since we initially created these plans(2b9da87), the limitations seems to have changed, and it's only `cache.t1.micro` instances that the restriction applies to now (https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/backups.html#backups-constraints). I'm going to raise a story to look at enabling backups for both plans (and perhaps deprecating the clustered plan).

manifests/cf-manifest/operations.d/730-elasticache-broker.yml
```diff
@@ -71,6 +71,12 @@
                       free: true
                       metadata:
                         displayName: Redis Non-clustered Tiny
+                    - id: b6949ea7-5c98-4c69-b981-4b5318ea8b7c
+                      name: medium-clustered-3.2
+                      description: 6.37GB RAM, clustered (1 shard), 3 nodes, failover, daily backups
+                      free: true
```
alext: I don't think we want this to be marked as free - this will make it available to trial orgs.

alext:
I've reworked this PR to use a non-clustered config for the new plan with a single replica node to enable automatic failover in the event of a failure.

This now also enables backups for the existing tiny unclustered plan now that AWS support them, and renames it to `tiny-3.2` to be more in line with [our naming scheme](https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR025-service-plan-naming-conventions/). Now that this has backups, there's no need for the `tiny-clustered-3.2` plan, so I've marked it as deprecated.

richardTowers:
Rebased onto master, and amended the date to point at Monday.

I've tested this in my pipeline and it looks good. I'll merge this on Monday morning.

PR #1716, opened 2019-01-18
Configure buildpack versions with YAML

tnwhitwell:
What
----

Updating the pipeline every time we need to update the buildpacks is not very sustainable, and is quite confusing to do without missing something.

`config/buildpacks.yml` is a configuration file that describes the state of our buildpacks, with each described as follows:

```yaml
- name: binary_buildpack
  repo_name: binary-buildpack
  stack: cflinuxfs2
  version: v1.0.27
  sha: 90e04df12df78f6d06ef127aa9992f505b7e51e21cb5ac97b68bc7e742193635
  filename: binary-buildpack-cflinuxfs2-v1.0.27.zip
  url: https://github.com/cloudfoundry/binary-buildpack/releases/download/v1.0.27/binary-buildpack-cflinuxfs2-v1.0.27.zip
  dependencies: []
```

Concourse will now run `parse_buildpacks_and_run_uploader.rb` which parses buildpacks.yml, and runs the existing `upload-buildpack.sh` script to download from GitHub and upload to cloudfoundry.

How to review
-------------

* Sanity check that all versions and SHAs are the same between the original configuration in the pipeline and `config/buildpacks.yml`
* Make a note of current buildpacks in cloudfoundry: `cf buildpacks`
* Delete the buildpacks:
    ```sh
    cf delete-buildpack -s cflinuxfs2 -f staticfile_buildpack
    cf delete-buildpack -s cflinuxfs2 -f java_buildpack
    cf delete-buildpack -s cflinuxfs2 -f ruby_buildpack
    cf delete-buildpack -s cflinuxfs2 -f dotnet_core_buildpack
    cf delete-buildpack -s cflinuxfs2 -f nodejs_buildpack
    cf delete-buildpack -s cflinuxfs2 -f go_buildpack
    cf delete-buildpack -s cflinuxfs2 -f python_buildpack
    cf delete-buildpack -s cflinuxfs2 -f php_buildpack
    cf delete-buildpack -s cflinuxfs2 -f binary_buildpack
    cf delete-buildpack -s cflinuxfs3 -f staticfile_buildpack
    cf delete-buildpack -s cflinuxfs3 -f java_buildpack
    cf delete-buildpack -s cflinuxfs3 -f ruby_buildpack
    cf delete-buildpack -s cflinuxfs3 -f dotnet_core_buildpack
    cf delete-buildpack -s cflinuxfs3 -f nodejs_buildpack
    cf delete-buildpack -s cflinuxfs3 -f go_buildpack
    cf delete-buildpack -s cflinuxfs3 -f python_buildpack
    cf delete-buildpack -s cflinuxfs3 -f php_buildpack
    cf delete-buildpack -s cflinuxfs3 -f binary_buildpack
    ```

* Deploy this branch to your environment and confirm the same versions are reinstalled

Who can review
--------------

Not @tnwhitwell or and @richardTowers


richardTowers:
Note that the `dependencies: []` in the buildpack.yml aren't needed for this PR, but we'll like to track them so we can auto-generate upgrade notifications. There's a script to generate this file in https://github.com/alphagov/paas-cf/pull/1717

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1508,117 +1508,7 @@ jobs:
...
-                    https://github.com/cloudfoundry/ruby-buildpack/releases/download/v1.7.27/ruby-buildpack-cflinuxfs3-v1.7.27.zip \
-                    f934d2008543c6fa8a6ca471c71ef3d176d4ec598a4f7ded3913c4b52f96108a
-
-                ./paas-cf/concourse/scripts/upload-buildpack.sh \
-                    staticfile_buildpack cflinuxfs3 \
-                    staticfile-buildpack-cflinuxfs3-v1.4.32.zip \
-                    https://github.com/cloudfoundry/staticfile-buildpack/releases/download/v1.4.32/staticfile-buildpack-cflinuxfs3-v1.4.32.zip \
-                    7473873333aa4c18ffe4f92cce9ad270770881563a91dd7d0c4e13b8ea20efc7
-
+                ruby -ryaml ./paas-cf/concourse/scripts/parse_buildpacks_and_run_uploader.rb
```
alext: We shouldn't need to do the `-ryaml` here, instead the script should require what it needs.  Also, if we give the script a `#!` line and make it executable we won't need the `ruby` call here either, which will make it consistent with other ruby scripts.

concourse/scripts/parse_buildpacks_and_run_uploader.rb
```diff
@@ -0,0 +1,8 @@
+yaml = YAML.safe_load(File.read("paas-cf/config/buildpacks.yml"))
```
alext: It would probably be better to pass in the path to the config file instead of hard-coding it here.

concourse/pipelines/create-cloudfoundry.yml
tnwhitwell: the key is named the same way as in the [buildpack manifest](https://github.com/cloudfoundry/python-buildpack/blob/master/manifest.yml).

Issue with removing it is that the first email we generate won't have any dependancy change information, unless we go off-process this time?

agreed on the ruby script - will modify that.

concourse/scripts/parse_buildpacks_and_run_uploader.rb
```diff
@@ -0,0 +1,16 @@
...
+require 'yaml'
+
+unless ARGV[0]
+  puts "USAGE: #{$PROGRAM_NAME} /path/to/buildpacks.yml"
+  exit 1
+end
+
+yaml = YAML.safe_load(File.read(ARGV[0].to_s))
+yaml['buildpacks'].each do |bp|
+  system("./paas-cf/concourse/scripts/upload-buildpack.sh", \
```
richardTowers: Do we need the backslashes on the end of these lines?

alext: Good spot, we don't - this is ruby, not bash...

tnwhitwell: This was originally all inlined. On it.

alext:
Testing this in my dev env, I got an error fetching the java buildpack: 
```
Downloading https://github.com/cloudfoundry/java-buildpack/releases/download/v4.16.1/java-buildpack-v4.16.1.zip...
Error: java-buildpack-v4.16.1.zip at https://github.com/cloudfoundry/java-buildpack/releases/download/v4.16.1/java-buildpack-v4.16.1.zip checksum differ. Expected  got b02b6fe9f71fbdcc060e81860723a752a5e455dd63b37ca3d41caa63c130b563
```

Additionally, the pipeline didn't error, it just continued without the java buildpack. We should fix both of these issues.

tnwhitwell:
> Testing this in my dev env, I got an error fetching the java buildpack:
> 
> ```
> Downloading https://github.com/cloudfoundry/java-buildpack/releases/download/v4.16.1/java-buildpack-v4.16.1.zip...
> Error: java-buildpack-v4.16.1.zip at https://github.com/cloudfoundry/java-buildpack/releases/download/v4.16.1/java-buildpack-v4.16.1.zip checksum differ. Expected  got b02b6fe9f71fbdcc060e81860723a752a5e455dd63b37ca3d41caa63c130b563
> ```
> Additionally, the pipeline didn't error, it just continued without the java buildpack. We should fix both of these issues.

This should be resolved now

PR #1718, opened 2019-01-18
Automate buildpack update notification creation

tnwhitwell:
What
----

Rather than picking through release pages on GitHub, this will allow generation of the notification email, and creation of a markdown file to describe the latest buildpack update

This details:

* The buildpacks that have been updated
* The version they are now at
* Any changes to the included dependencies
* Release notes for all the interim release notes in the buildpack's repository

This is run with the script `scripts/create_buildpacks_email.sh`. By default, it targets the last time `buildpacks.yml` changed, and compares that to now.
If comparison to another point in time is required, it can be run with the commit hash as the first argument.

Running this will generate the email text on stdout, and will write a file to `doc/buildpack_release.md` which can then be committed.

How to review
-------------

* Code Review
* Run `scripts/create_buildpacks_email.sh`.
    * Ensure `doc/buildpack_release.md` is created, and email content is printed to the console.
        * Ensure the versions in these match those in `config/buildpacks.yml` and that they are the same between the email and the markdown output.
        * Ensure the dependency changes are correct between the different versions of the buildpacks
        * Ensure links to all the relevant release notes are present in both.

Who can review
--------------

Anyone but @tnwhitwell or @richardTowers


alext:
This PR seems to duplicate and conflict with #1716. Ideally, this would be based of that one, and add the additional scripts etc.

Additionally, it would be good to put the vendoring of dependencies into a separate commit so that it's easier to see the actual code changes without all the extra noise.

tnwhitwell:
Decided to merge with #1717 as having two PRs for this very similar task was making things confusing and miserable.

PR #1721, opened 2019-01-23
[#163285897] - Add timeout to automatic smoke tests

tnwhitwell:
What
----

Multiple timeouts have been added to the steps of this job, selected based on the average total job time over the last 14k automatic smoke test runs:

| minimum time    | maximum time    | mean time       |
|-----------------|-----------------|-----------------|
| 2.8166666666667 | 9.7666666666667 | 3.1423720930233 |

Retries are being ignored at this point, as we will be removing those as part of #162014355 - with the retries still there, this will still be a significant improvement over the tests never completing.

How to review
-------------

* BAU: Deploy to pipeline, ensure smoke tests continue as normal

* Test timeout functionality: create branch from this, change timeout of `smoke-tests-run` task to 1 minute, ensure that smoke tests are aborted at 1 minute.

Who can review
--------------

Anyone but @tnwhitwell

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -3457,11 +3457,13 @@ jobs:
               fi
       - do:
         - task: create-temp-user
+          timeout: 5m
           file: paas-cf/concourse/tasks/create_admin.yml
           params:
             PREFIX: cont-smoketest-user
 
         - task: smoke-tests-config
+          timeout: 1m
           file: paas-cf/concourse/tasks/generate-test-config.yml
           params:
             TEST_PROPERTIES: smoke_tests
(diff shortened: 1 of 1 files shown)


jpluscplusm:
What % of the test steps in your `smoke-test-runs.txt` dataset would hit the timeout?

On how many occasions would a have timeout struck 2 or 3 times in succession? (I know this isn't strictly representative of the end result of this PR, due to delays between tests runs that *aren't* present between attempts ...)

tnwhitwell:
> What % of the test steps in your `smoke-test-runs.txt` dataset would hit the timeout?

0.09% (13/13975)

So, I've just realised a flaw with my logic here - I've gone for full runs, which includes setup and all timeouts (if applicable).

Do you know of a way for me to check the length of runs of individual `step`s?

(slack may be helpful here)

PR #1723, opened 2019-01-23
[Support] Fix post deploy bosh lock conflict

alext:
What
----

This fixes an issue with a bosh lock conflict between the bosh cleanup task and the rotate-*-encryption-keys tasks by moving bosh-cleanup later in the post-deploy job.

I've also taken the opportunity to fix the serialisation of the aiven and elasticache broker tasks.

How to review
-------------

Code review is enough. (Adding `?w=1` to the diff URL should make it easier).

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2327,85 +2327,84 @@ jobs:
                   bosh run-errand rotate-uaa-database-key --when-changed
 
     - aggregate:
-      - do:
-        - task: deploy-aiven-service-broker
-          config:
-            platform: linux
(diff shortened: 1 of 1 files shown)


bandesz:
I think we should take the opportunity and move the bosh-cleanup out of post-deploy as I regularly ran into the issue that it deleted the prometheus-boshrelease while prometheus-deploy was running

alext:
Actually, I think this fix isn't sufficient as it still causes conflicts with the prometheus-deploy. I'm going to raise a separate PR to remove the bosh cleanup for the time being.

alext:
Doh!, @bandesz I just realised that you commented essentially the same thing, and I failed to read it...

PR #1731, opened 2019-01-28
[Support] Update Aiven threshold, and parameterise elasticache node alert.

alext:
What
----

This makes 2 tweaks to the alerts:

* Update Aiven cost alert with new threshold (and merge into single alert)
* Parameterise the elasticache node count alert.

How to review
-------------

Code review is probably enough - I've tested this in my dev env.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/prometheus/alerts.d/aiven-cost.yml
+++ b/manifests/prometheus/alerts.d/aiven-cost.yml
@@ -6,18 +6,11 @@
   value:
     name: AivenEstimatedCostHigh
     rules:
-      - &alert
-        alert: AivenEstimatedCostHigh_Warning
-        expr: paas_aiven_estimated_cost_pounds > 2600
+      - alert: AivenEstimatedCostHigh
+        expr: paas_aiven_estimated_cost_pounds > 6500
         for: 1h
         labels:
-          severity: warning
+          severity: critical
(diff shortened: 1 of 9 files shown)


manifests/prometheus/alerts.d/aiven-cost.yml
```diff
@@ -6,18 +6,11 @@
   value:
     name: AivenEstimatedCostHigh
     rules:
-      - &alert
-        alert: AivenEstimatedCostHigh_Warning
-        expr: paas_aiven_estimated_cost_pounds > 2600
+      - alert: AivenEstimatedCostHigh
+        expr: paas_aiven_estimated_cost_pounds > 6500
```
richardTowers: I've been thinking about changing the alert expr here to `delta(paas_aiven_estimated_cost_pounds[1d])*30 > 0` or something, so we’re not alerting on a sawtooth.

That data looks like this:

image

... so I wonder if 6,500 is still a little too low?

alext: I think that makes sense, but I think it should also be a separate story. This was intended to be a quick fix to get the board green while we work out what the longer-term solution to this alerting is.

PR #1735, opened 2019-01-30
[#163520707] A script to get all app stack versions

tnwhitwell:
What
----

A script to get all app stack versions for all organisations and spaces in a cloudfoundry instance

How to review
-------------

Code Review (sorry!)

Log in to cloudfoundry as an admin user, and run the script - check that the data that is returned is correct and exhaustive (all orgs, spaces and apps should be included).

Who can review
--------------

Not @tnwhitwell

### DIFF
--- a/scripts/get-all-app-stacks.sh
+++ b/scripts/get-all-app-stacks.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cflinuxfs2_uuid=$(cf stack cflinuxfs2 --guid)
+cflinuxfs3_uuid=$(cf stack cflinuxfs3 --guid)
+all_orgs_next_url="/v2/organizations?order-by=name"
+echo "org,space,appname,stack,"
+while [ "${all_orgs_next_url}" != "null" ]
+do
+    orgs_response=$(cf curl "${all_orgs_next_url}" | jq)
+    orgs=$(echo "${orgs_response}" | jq -r '.resources[] | [.metadata.guid, .metadata.url] | @csv')
+    while read -r org
+    do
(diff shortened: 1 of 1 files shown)


scripts/get-all-app-stacks.sh
```diff
@@ -0,0 +1,59 @@
...
+                            stack="cflinuxfs2"
+                        elif [ "${app_stack_guid}" == "${cflinuxfs3_uuid}" ]
+                        then
+                            stack="cflinuxfs3"
+                        else
+                            stack="docker"
+                        fi
+                        echo "'${org_name}','${space_name}','${app_name}','${stack}',"
+                    done <<< "${space_apps}"
+                    next_space_app_url=$(echo "${space_apps_response}" | jq '.next_url')
```
richardTowers:
```suggestion
                    next_space_app_url=$(echo "${space_apps_response}" | jq -r '.next_url')
```

scripts/get-all-app-stacks.sh
```diff
@@ -0,0 +1,59 @@
...
+                done
+            done <<< "${org_spaces}"
+            next_space_url=$(echo "${org_spaces_response}" | jq '.next_url')
```
richardTowers:
```suggestion
            next_space_url=$(echo "${org_spaces_response}" | jq -r '.next_url')
```

scripts/get-all-app-stacks.sh
```diff
@@ -0,0 +1,59 @@
...
+        done
+    done <<< "${orgs}"
+    all_orgs_next_url=$(echo "${orgs_response}" | jq '.next_url')
```
richardTowers:
```suggestion
    all_orgs_next_url=$(echo "${orgs_response}" | jq -r '.next_url')
```

PR #1736, opened 2019-01-30
[#162611438]Update Logstash filters to deserialize JSON messages

AP-Hunt:
What
----
Logstash filters were looking for component names beginning with "vcap." before
deserializing JSON. No components had this prefix.

Removes "vcap" from generated Logstash filters file inside the make task using
sed.

How to review
-------------

Check paas-dev ELK stack for events whose original JSON message (see `@raw`) has been broken down in to a field whose name matches the value of `@source.component`.

To deploy, copy and past `config/logit/output/generated_logit_filters.conf` to the relevant ELK stack logstash filters in Logit.

Who can review
--------------
Not @mogds or @AP-Hunt

### DIFF
--- a/config/logit/output/generated_logit_filters.conf
+++ b/config/logit/output/generated_logit_filters.conf
@@ -747,7 +747,7 @@ filter {
   ##--------------------------
   # Uaa conf. Parses uaa logs.|
   ##--------------------------
-  if [@source][component] == "vcap.uaa" {
+  if [@source][component] == "uaa" {
 
       # ---- Parse UAA events (general)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -289,4 +289,11 @@ logit-filters:
 		-v $(CURDIR)/config/logit/output:/output:rw \
 		-w /mnt \
 		jruby:9.1-alpine ./scripts/generate_logit_filters.sh $(LOGSEARCH_BOSHRELEASE_TAG) $(LOGSEARCH_FOR_CLOUDFOUNDRY_TAG)
+	@echo "Making changes to $(CURDIR)/config/logit/output/generated_logit_filters.conf"
```
richardTowers: Can we do this in the `./scripts/generate_logit_filters.sh` file instead?

AP-Hunt: Yep, good call. Resolved in the amended commit.

richardTowers:
Just had a look in Kibana - so, so much better 😻

:shipit:

PR #1747, opened 2019-02-08
[#162755282] Fix the RouterTotalRoutesDrop alert

alext:
What
----

The existing expresion was causing the following error in Prometheus:

> Error executing query: many-to-many matching not allowed: matching
> labels must be unique on one side

In fact because we're using the same series on both sides of the /
operator, there's no need for the on() group_left() functions at all.

I initially fixed this by adding 'bosh_job_name,bosh_job_id' to the on()
function, but then realised that the whole thing was unnecessary.

How to review
-------------

* Code review.
* Try viewing the updated expression in the prometheus console in a production env.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/prometheus/alerts.d/router_total_routes.yml
+++ b/manifests/prometheus/alerts.d/router_total_routes.yml
@@ -9,8 +9,8 @@
         expr: avg_over_time(firehose_value_metric_gorouter_total_routes[5m])
       - record: gorouter:total_routes:delta5m
         expr: delta(firehose_value_metric_gorouter_total_routes[5m])
-      - &alert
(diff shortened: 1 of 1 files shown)


alext:
> I think the `min_over_time` is also unnecessary, it will just make the alert to alert longer.

I think that's the intention. The idea is that if there is a sudden large drop in the number of routes, the alert would fire for at least 30 minutes meaning we're more likely to notice it.  Having said that, I'm not 100% convinced that this is a good way to deal with the problem of the alert potentially being missed.

alext:
I've pushed another commit to remove the `avg_over_time`.

PR #1751, opened 2019-02-11
[#163364037] Deploy S3 broker as private beta

tnwhitwell:
What
----

Deploy s3-broker as private beta.

 * Add release + instance group to manifest
 * Add s3-broker to load balancer and add policy to instance group
 * Register service only for 'admin' org in each environment
 * Add custom acceptance tests
 * Set up billing (we simply set it to $5/month)

✅  (done) Before merging make sure you update the s3-broker release metadata with the final build details.

How to review
-------------

1. Review https://github.com/alphagov/paas-s3-broker/pull/7 (done)
1. Review https://github.com/alphagov/paas-s3-broker-boshrelease/pull/1 (done)
1. Code review of this PR
1. Deploy your CF from this branch
1. Make sure the custom acceptance tests pass

Who can review
--------------

Not @tnwhitwell or @bandesz


manifests/cf-manifest/operations.d/750-s3-broker.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: s3-broker
+        release: s3-broker
+        properties:
+          s3-broker:
+            port: 80
+            basic_auth_username: s3-broker
+            basic_auth_password: ((secrets_s3_broker_admin_password))
+            log_level: INFO
+            aws_region: "((terraform_outputs_region))"
+            resource_prefix: "paas-s3-broker-"
```
alext: We should include the `deploy_env` in the prefix. We want to be able to tell which env a bucket belongs to when looking in the AWS console. Otherwise it'll be hard to clean up orphaned buckets from dev environments etc.

bandesz: We add the deploy_env tag to the buckets and the IAM users. Also we don't add the prefix to any other services (RDS, Elasticache), so I'm inclined not to add it here either.

alext: We have started adding it to other instances (eg Aiven). We don't add it to Elasticache because we're severely restricted in the length of instance names there (we couldn't even fit a UUID in). I think the RDS broker was one we missed, and it's hard to change now.

Is there a reason not to include it? At the moment it's somewhat tedious to see which environment a bucket belongs to when looking at the list in the S3 console (the tags are many clicks in, and if you're dealing with a long list, that will get old quickly).

bandesz: Nothing against it. We need to to change the IAM policies in account wide terraform to add a "*-" prefix.

alext: Or we could use `paas-s3-broker-${env}` as the prefix... which shouldn't require any changes.

bandesz: We usually prefix everything, so I wouldn't put it as a suffix just to avoid an extra small change :)

tnwhitwell: Added the environment - this will apply both to bucket names and IAM usernames. Running down my pipeline to confirm.

PR #1753, opened 2019-02-12
[Support] Add a test for the generated concourse keys file

alext:
What
----

This tests that the generated gpg keys file matches the keys defined in
the .gpg-id. This will catch instances where keys are added to .gpg-id,
but `make update_merge_keys` isn't run.

As part of this I've updated the `.gpg-id` file to use full gpg fingerprints as this makes the testing easier (as well as being more secure)

I've also tweaked the test setup so that it's possible to use extra dependencies (like the openpgp library) without having to include them in `.travis.yml`.

How to review
-------------

* Code review. Verify Travis passes
* Verify that I've correctly updated the key fingerprints (although the `gpg-keys.yml` file hasn't changed, and the new test should catch any errors)

Who can review
--------------

Not me.


alext:
Note: once this is reviewed, I intend to make an equivalent change to paas-bootstrap.

PR #1756, opened 2019-02-14
[#163524147] Update the cdn-broker with TLS 1.2 policy

alext:
## What

This pulls in an updated version of the cdn-broker that configures
CloudFront to only support TLS 1.2

🚨 **Do not merge before 4th March** 🚨 

How to review
-------------

* Deploy from this branch
* Create a cdn-route instance.
* Once it's fully deployed, test is in an SSL tester (eg https://www.ssllabs.com/ssltest) to verify it only supports TLS 1.2

## 🚨 Before merging 🚨 

Update the TMP commit with the release created after https://github.com/alphagov/paas-cdn-broker-boshrelease/pull/13 is merged.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/cf-manifest/operations.d/720-cdn-broker.yml
+++ b/manifests/cf-manifest/operations.d/720-cdn-broker.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: cdn-broker
-    version: 0.1.9
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/cdn-broker-0.1.9.tgz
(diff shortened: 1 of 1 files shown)


alext:
> Interestingly https://www.ssllabs.com/ssltest/analyze.html?d=cdntest.andras.dev.cloudpipeline.digital&s=99.84.231.5 shows it's able to connect on TLS1.0, but I think this happened because the CF distribution was still being updated after the SSL cert was set up.

I saw the same when testing. I think this is because the default *.cloudfront.net cert is configured to allow everything, and it still being served while everything gets updated. Once it was all fully updated, ssltest was no longer reporting TLS1.0 as available.

PR #1762, opened 2019-02-18
[Support] bump ipsec release to fix xenial compilation

alext:
What
----

The version of ipsec-release we're using doesn't compile correctly on Xenial. We've only just noticed this because we've disabled the bosh compiled package cache due to an issue with it reusing packages across stemcells incorrectly (https://github.com/alphagov/paas-bootstrap/pull/249). This means that we've been using the ipsec-release compiled on trusty up until now.

This pulls in an upstream change to the packaging script that fixes the compilation on xenial.

How to review
-------------

I've already run this down my dev environment, and it fixed this issue there, so code review is probably enough.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/cf-manifest/operations.d/530-ipsec.yml
+++ b/manifests/cf-manifest/operations.d/530-ipsec.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: ipsec
-    sha1: 6a9e252162519e50f6c511c489ba304a286d496a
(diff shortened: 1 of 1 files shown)


jpluscplusm:
I've not directly tested 0.1.4 against my dev env, but this LGTM on the understanding it's not changed under the hood since the 0.0.x PR-time build.

PR #1745, opened 2019-02-07
[#161647315] Self-Enable RDS Broker Extensions

mogds:
What
----

The RDS broker has been updated so users have the option to enable supported extensions. This PR adds a Postgres pg_stat_statements enabled parameter group to be consumed by the RDS broker.

How to review
-------------

- Deploy this branch to a dev env and go through a database service lifecycle, with and without different extensions enabled, using the `-c '{"enabled_extensions": ["pg_stat_statements"]}'` argument to `cf create-service`
- Confirm the correct parameter group has been applied, and connect to the RDS instance to confirm the correct extensions have been enabled

**Merge Steps**

- Merge the [RDS broker PR](https://github.com/alphagov/paas-rds-broker/pull/95)
- Remove the [branch line](https://github.com/alphagov/paas-rds-broker-boshrelease/pull/77/files#diff-8903239df476d7401cf9e76af0252622R4) in `.gitsubmodules` for [the bosh release](https://github.com/alphagov/paas-rds-broker-boshrelease/pull/77) and update the submodule
- Update the WIP commit in this PR with the final rds broker bosh release and merge

Who can review
--------------

Not @mogds @AP-Hunt @LeePorte

PR #1765, opened 2019-02-20
[#164104586] Stop specifying Postgres10 minor version for broker

alext:
## What

The Postgres versioning scheme has changed starting with Postgres 10[1],
and it now only uses a 2 part version number instead of the 3 part
number that was used for 9.x and earlier.

For Postgres 9.x we're only specifying the first 2 parts so that we
don't have conflicts when RDS auto-upgrades the minor versions. We want
the same behaviour for Postgres 10.x, so we need this to only specify on
version part.

With this change, the plan names no longer reflect reality as RDS will
use postgres 10.6. We'l rename these in a separate PR becuase this
change will require notifying tenants.

[1]https://www.postgresql.org/support/versioning/

How to review
-------------

Code review may be enough. I've run this down my dev environment and verified that it works. I've also verified that doing an update-service on a 10.6 instance works correctly where previously it would attempt to downgrade it which would be rejected by RDS.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/cf-manifest/operations.d/713-rds-broker-postgres10-plans.yml
+++ b/manifests/cf-manifest/operations.d/713-rds-broker-postgres10-plans.yml
@@ -15,7 +15,7 @@
       vpc_security_group_ids:
         - ((terraform_outputs_rds_broker_dbs_security_group_id))
       engine: "postgres"
-      engine_version: "10.5"
+      engine_version: "10"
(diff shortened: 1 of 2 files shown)


AP-Hunt:
We will need to re-apply this change when the [RDS broker change PR](https://github.com/alphagov/paas-cf/pull/1745) gets merged

PR #1769, opened 2019-02-20
[Support] Remove bosh cleanup task

alext:
What
----

This replaces #1723 

We've seen this conflicting with the rotate-*-encryption-keys jobs when
trying to get a lock in the bosh director.

We've also seen this break the prometheus deploy job (which runs in
parallel to this)  because it deletes the prometheus releases between
them being uploaded, and the deployment happening.

Given it's not essential to run the cleanup, this removes it.

While doing this, I noticed that the aiven and elasticache post-deploy tasks were serialised, which they don't need to be, so this PR also fixes that.

How to review
-------------

* Code review (adding `?w=1` to the diff URL may help)
* Run the pipeline from this branch and verify all looks good (I've run this in my dev env already).

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2354,85 +2354,84 @@ jobs:
                   bosh run-errand rotate-uaa-database-key --when-changed
 
     - aggregate:
-      - do:
-        - task: deploy-aiven-service-broker
-          config:
-            platform: linux
-            image_resource: *cf-cli-image-resource
-            inputs:
-              - name: paas-cf
-              - name: paas-aiven-broker
-              - name: config
-            params:
-              AIVEN_API_TOKEN: ((aiven_api_token))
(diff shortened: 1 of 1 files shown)


bandesz:
Should we put the bosh clean-up task into a separate job and trigger it every night maybe?

bandesz:
From the documentation: "Cleans up releases, stemcells, orphaned disks, and other unused resources."

I guess some unnecessary costs might accrue from orphaned EBS volumes.

This means we might end up with orphaned EBS volumes which could accrue some unnecessary costs

alext:
I considered that, but I don't think that's actually necessary. Orphaned volumes are cleaned up regularly anyway, and the other things are just old releases etc that are files in a S3 bucket.

I reckon if things start to get big enough for us to notice, we can add this back in, but I'd be inclined to leave it until we know we need it.  This was originally added in #558, but looks to be based on a reckon as opposed to a concrete need.

PR #1770, opened 2019-02-20
[support] Place S3 service access in a conditional block

LeePorte:
What
----

This is currently causing conflicts on second runs of the pipeline as it has already been enabled for the admin org

How to review
-------------

deploy a current master in dev, try and run the pipeline a second time, this will fail, then deploy this branch on top and it will pass.

Who can review
--------------

anyone

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1694,7 +1694,11 @@ jobs:
                 fi
 
                 # TODO: remove '-o admin' when we are happy to release this to the world
-                cf enable-service-access aws-s3-bucket -o admin
+                if cf service-access | grep "aws-s3-bucket\s"; then
+                  echo "already enabled"
+                else
+                  cf enable-service-access aws-s3-bucket -o admin
+                fi
 
       - task: set-security-groups-from-manifest
         config:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1694,7 +1694,11 @@ jobs:
                 fi
 
                 # TODO: remove '-o admin' when we are happy to release this to the world
-                cf enable-service-access aws-s3-bucket -o admin
+                if cf service-access | grep "aws-s3-bucket\s"; then
+                  echo "already enabled"
+                else
+                  cf enable-service-access aws-s3-bucket -o admin
+                fi
```
alext: Do we need to do this at all?  I'm not clear why we need this enabled in the admin org...

LeePorte: true it could come out completely.

bandesz: I already fixed this in a different way, where did that go? o_O

bandesz: It was accidentally overridden with a force push :(

https://github.com/alphagov/paas-cf/compare/5e559ee5c099f760bff381ccd9e4b73f5d69dba5..c5863b7864bf68d94059c6826d51f34565848c13

LeePorte: @bandesz I've switched this out for your method as I prefer it.

PR #1771, opened 2019-02-21
[Support] Test that we have an equal number of cell per AZ

alext:
What
----

As a follow-up to #1749, this codifies the fact that the number of cells needs to be equal in each AZ into a unit test so that we don't rely on institutional memory for this fact.

I've also done a bit of refactoring of the test helpers while I was there. 

How to review
-------------

* Code review. Verify the tests pass in Travis.

Who can review
--------------

Not me.


bandesz:
Great idea, merged with [skip ci].

PR #1772, opened 2019-02-21
[support] Add larger Elasticsearch Plan

LeePorte:
What
----

DiT have requested a larger Elasticsearch plan as they are currently filling up the medium sized plan. This adds the plan and updates the billing information so that it is available to be billed.

How to review
-------------

Code review

Who can review
--------------

Anyone

### DIFF
--- a/config/billing/config.json.erb
+++ b/config/billing/config.json.erb
@@ -1231,6 +1231,19 @@
 			]
 		},
 		{
+			"name": "elasticsearch large-ha-6.x",
+			"valid_from": "2018-10-01",
+			"plan_guid": "90f12775-0b60-4a90-b2bc-c8c28a1ca85e",
+			"components": [
+				{
+					"name": "instance",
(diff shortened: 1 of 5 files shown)


alext:
You need to run the `generate_config.rb` script to update the actual billing configs.

Thinking about this, we should probably run this in the pipeline instead of storing the results in git. It's caught people out almost every time we update the billing.

config/billing/config.json.erb
```diff
@@ -1230,6 +1230,19 @@
...
+			"plan_guid": "90f12775-0b60-4a90-b2bc-c8c28a1ca85e",
+			"components": [
+				{
+				    "name": "instance",
+					"formula": "ceil($time_in_seconds/3600) * <%= price("aiven_elasticsearch_business_16") %>",
+					"currency_code": "USD",
+					"vat_code": "Standard"
+                }
+            ]
+        },
```
alext: The indentation's gone a bit odd here...

PR #1777, opened 2019-02-25
[support] Scale up API instances in London

LeePorte:
What
----

We are hitting the limits of the current number of API servers in the London region.

This will solve the limits and give us some headroom for growth.

How to review
-------------

Code review

Who can review
--------------

anyone

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -1,4 +1,4 @@
 ---
 cell_instances: 15
 router_instances: 3
-api_instances: 2
+api_instances: 3


alext:
We have 4 APIs in prod. Any reason not to match that here.  I think the APIs are only across 2 AZs, so it probably makes sense to keep them balanced.

LeePorte:
We scaled up the API instances in Ireland after the cf-exporter was causing it to lock up, as we'd not got that running I didn't see any particular reason to increase it. Cost is the only implication for going up to 4. I'm fine with either.

alext:
Let's roll this out as it is. We can always further refine later once things are happy again.

PR #1778, opened 2019-02-25
[#163592114] Deprecate Elasticsearch 5.x

AP-Hunt:
What
----
Update the descriptions and displayNames for all Elasticsearch 5 plans to indicate that they are deprecated.

The amended blocks are removed in another PR, which will be merged after the tenants notice period.

How to review
-------------
Code review should be enough. The changes made only affect the marketplace output.

Who can review
--------------
Not @tnwhitwell or @AP-Hunt


bandesz:
We should wait until the Aiven support gets back with the exact dates.

PR #1780, opened 2019-02-25
[#163592114] Remove Elasticsearch 5.x from billing

tnwhitwell:
What
----

As Elasticsearch 5.x is now EOL and no tenants are using it, it can be removed from the billing, so it is no longer visible in the calculator.

How to review
-------------

* Code review
* Ensure no tenants are running ES5 instances (instructions are available on https://github.com/alphagov/paas-cf/pull/1779)
Who can review
--------------

Not @tnwhitwell or @AP-Hunt

### DIFF
--- a/config/billing/config.json.erb
+++ b/config/billing/config.json.erb
@@ -1153,19 +1153,6 @@
 			]
 		},
 		{
-			"name": "elasticsearch tiny-5.x",
-			"valid_from": "2018-09-01",
-			"plan_guid": "07264114-c666-440b-934a-015508be4475",
-			"components": [
-				{
-					"name": "instance",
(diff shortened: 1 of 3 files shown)


AP-Hunt:
`paas-billing` requires that plans exist in order to consolidate bills each month, and to reconsolidate old bills should be the need arise, and `paas-admin` reads the pricing plans via the API to build the calculator. 

In the past, we've deprecated Compose plans but not removed them from billing. Instead, the plans were blacklisted in `paas-admin`. For ES 5.x we're also going to leave the plans in place for the time being, so that we don't break billing.

Closing this PR, and raising a story to address how to deprecate plans properly in the future.

PR #1782, opened 2019-02-26
[#164363210] Set API limits per user for Cloud Controller

LeePorte:
What
----

This PR sets API limits for Cloud Controller in order to prevent Denial of Service caused by one or two users.

Currently we have no API rate limiting in place. this sets global limits based upon https://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/capi-release&version=1.77.0#p%3dcc.rate_limiter

If we require further fine tuning we can use https://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/capi-release&version=1.77.0#p%3dcc.nginx_rate_limit_zones

I have arrived at these numbers by taking the defaults then reducing the reset time and then dividing the defaults by the time reduction factor and then applied a bit of rounding.

How to review
-------------

Who can review
--------------

Describe who can review the changes. Or more importantly, list the people
that can't review, because they worked on it.

### DIFF
--- a/manifests/cf-manifest/env-specific/default.yml
+++ b/manifests/cf-manifest/env-specific/default.yml
@@ -4,3 +4,4 @@ router_instances: 2
 api_instances: 2
 doppler_instances: 6
 log_api_instances: 2
+cc_hourly_rate_limit: 10000
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -4,3 +4,4 @@ router_instances: 3
 api_instances: 4
 doppler_instances: 12
 log_api_instances: 6
+cc_hourly_rate_limit: 15000
--- a/manifests/cf-manifest/env-specific/prod.yml
+++ b/manifests/cf-manifest/env-specific/prod.yml
@@ -4,3 +4,4 @@ router_instances: 3
 api_instances: 4
 doppler_instances: 12
 log_api_instances: 6
+cc_hourly_rate_limit: 55000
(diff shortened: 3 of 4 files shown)


jpluscplusm:
JOOI do these numbers come from any log analysis, or a Reckon, or wot?

bandesz:
The upstream default config is:
```
cc.rate_limiter.enabled:
    description: "Enable rate limiting for UAA-authenticated endpoints per user or client"
    default: false
  cc.rate_limiter.general_limit:
    description: "The number of requests a user or client is allowed to make for all endpoints that do not have a custom limit over the configured interval"
    default: 2000
  cc.rate_limiter.unauthenticated_limit:
    description: "The number of requests an unauthenticated client is allowed to make over the configured interval"
    default: 100
  cc.rate_limiter.reset_interval_in_minutes:
    description: "The interval in minutes, after which, a user's available api requests will be reset"
    default: 60
```

LeePorte:
Updated the PR message with how I for to the numbers.

richardTowers:
Interesting - we should discuss this in the incident retro I guess - I hadn't realised that this was an option though, and it feels like a good idea.

We'd probably want to do some further analysis of our metrics to work out whether anyone would be above these rate limits in current usage.

Also interested in whether the limits apply to users with Admin creds, e.g. for the cf_exporter that we disabled because of load issues.

tlwr:
Commit
---

    50000 is a limit that is generally too high, and only applicable in
    Ireland as we can see from Kibana.
    
    - makes Cloud Controller rate limit environment specific, because only
    ireland needs such a high rate limit
    - changes the rate limit from 50000 to 55000 in Ireland (just to be
    safe, although we do not expect Observe to increase in usage)
    - sets a default rate limit to 10000 over 1 hour, as a reasonable
    failsafe
    - sets the rate limit in London to 15000, which is comfortably above
    current usage

What
---

- See ⬆️ 

How to review
---

- Code review

@46bit and I are going to run this through a pipeline

---

Also I've rebased onto master

tlwr:
Behaviour of my dev environment

Screen Shot 2019-04-11 at 09 01 28

tlwr:
This seems to be working fine, the spikes above are Capi getting ddossed

PR #1785, opened 2019-02-28
[#164248877] RDS-broker reject unknown params

alext:
What
----

This updates the broker to a version that rejects unrecognised user
params - alphagov/paas-rds-broker#99

How to review
-------------

Deploy from this branch and verify everything works. Attempt to update a service instance with some unrecognized params (eg `cf update-service my-db -c '{"foo": "bar"}'`) and verify it correctly returns an error.

## 🚨 Before merging 🚨 

Update the TMP commit with the version built after https://github.com/alphagov/paas-rds-broker-boshrelease/pull/80 is merged.

Who can review
--------------

Not me.

### DIFF
--- a/manifests/cf-manifest/operations.d/710-rds-broker.yml
+++ b/manifests/cf-manifest/operations.d/710-rds-broker.yml
@@ -3,9 +3,9 @@
   path: /releases/-
   value:
     name: rds-broker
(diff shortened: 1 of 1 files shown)


alext:
This will need rebasing after #1783 is merged.

PR #1789, opened 2019-03-04
[#164338027] Remove Alex from CF and from merge keys

LeePorte:
What
----

Alex is no longer on the team so doesn't need access

How to review
-------------

Make sure it is Alex's key

Who can review
--------------

Not @LeePorte


henrytk:
:*(

richardTowers:
😭

PR #1798, opened 2019-03-08
[SUPPORT] Scale up doppler in production

tnwhitwell:
What
----

* Number of doppler instances in production have been scaled to 12 in prod, and 6 in all other instances.
* Test has been added to ensure this is the case

How to review
-------------

* Code Review
* Check unit tests pass
* Check the number of doppler instances do not change when run down Dev pipeline.

Who can review
--------------

Not @tnwhitwell

### DIFF
--- a/manifests/cf-manifest/env-specific/stg-lon.yml
+++ b/manifests/cf-manifest/env-specific/stg-lon.yml
@@ -2,3 +2,4 @@
 cell_instances: 6
 router_instances: 3
 api_instances: 2
+doppler_instances: 3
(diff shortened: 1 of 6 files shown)


richardTowers:
Extra context: this will increase our AWS spend by ~$502/month (if we leave it at 12 instances - in reality we might walk it back to 10 if it looks happy again).

manifests/cf-manifest/env-specific/stg-lon.yml
```diff
@@ -2,3 +2,4 @@
 cell_instances: 6
 router_instances: 3
 api_instances: 2
+doppler_instances: 6
```
LeePorte: I'd be inclined to drop this to 3, as we don't need the number in staging and it will still catch any issues caused during upgrade stories.

tnwhitwell: agreed. i've made the change

PR #1805, opened 2019-03-18
Add 3 cells to prod-lon

richardTowers:
What
----

The memory usage in prod-lon has been climbing, and we're now below our
threshold again.

Judging from this graph:

https://prometheus-1.london.cloud.service.gov.uk/graph?g0.range_input=2w&g0.expr=rep_memory_capacity_pct%3Aavg5m&g0.tab=0

Last time we increased our scale by 3 instances we got an extra 23%
rep_memory_capacity_pct. We're now running at 23%, which is below our
threshold of 33%. Increasing the cell instances by 3 should give us
another 20% or so, which should take us above the threshold again.

I'll raise a separate PR to scale back Ireland, which seems to be
swimming in free memory now.

How to review
-------------

* Code review should be fine

Who can review
--------------

Not @richardtowers

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -1,5 +1,5 @@
 ---
-cell_instances: 18
+cell_instances: 21
 router_instances: 3
 api_instances: 4
 doppler_instances: 12


bandesz:
Be careful with the calculations as the graph is misleading. The last time we've got an alert at 32.8%, but the capacity dropped to ~20% temporarily. Shortly before the extra instances were added it went back to ~33% and after the scaling capacity was at 45%. (So the diff is only 12)

Here is the correct calculation for the previous time:

```
15 * (100-32.8) / 100 = 10.08 "units" used previously

(18 - 10.08) / 18 * 100 = 44% will be free after scaling.
```

Currently the capacity is at 23%:
```
18 * (100-23) / 100 = 13.86 "units" used

(21 - 13.86) / 21 * 100 = 34% will be free after scaling.
```

We're barely going to be over the alert limit with this increase.

richardTowers:
Okay, I hadn't spotted that in the graph. I guess we'll have to scale again once this is out - we'll be able to observe how much capacity this gives us more directly.

PR #1810, opened 2019-03-19
Estimate Aiven cost at the end of the month

richardTowers:
What
----

The value of `paas_aiven_estimated_cost_pounds` is the current spend for
the month according to Aiven, so it goes up in a saw tooth from the
beginning of the month.

This uses a linear approximation that the cost for the month will be the
cost for the last 24 hours multiplied by 30 (for the days in the month).

The delta function just naively subtracts the value at the beginning of
the time period from the value at the end. You might think that the
`increase()` function would be a better fit, but sadly that only works
for counters that always strictly increase over time or reset to zero -
this one occasionally drops by a few pounds, which confuses prometheus.

For the first day of each month the value of
`delta(paas_aiven_estimated_cost_pounds[24h])` will be massively
negative as the counter resets to zero.

Previously we were just alerting when our spend for a month crossed a
particular threshold, which tended to happen towards the end of every
month.

This also changes the threshold significantly so the alert will stop firing - we'll have the conversation about whether we need to talk to finance again separately, we don't need a permanent red alert reminding us.

How to review
-------------

* Code review
* Look at these screenshots and check they look sensible:

Before:

image

After:

image

Who can review
--------------

* Not @richardTowers

### DIFF
--- a/manifests/prometheus/alerts.d/aiven-cost.yml
+++ b/manifests/prometheus/alerts.d/aiven-cost.yml
@@ -7,8 +7,7 @@
     name: AivenEstimatedCostHigh
     rules:
       - alert: AivenEstimatedCostHigh
-        expr: paas_aiven_estimated_cost_pounds > 6500
-        for: 1h
+        expr: delta(paas_aiven_estimated_cost_pounds[24h]) * 30 > 20000
         labels:
           severity: critical
         annotations:
--- a/manifests/prometheus/spec/alerts/aiven-cost.test.yml
+++ b/manifests/prometheus/spec/alerts/aiven-cost.test.yml
@@ -9,12 +9,16 @@ tests:
   - interval: 1h
     input_series:
       - series: "paas_aiven_estimated_cost_pounds"
-        values: 0+100x48
+        values: 0+100x24 2500+10x24
 
     alert_rule_test:
+
+      # Alert should not be firing initially
       - eval_time: 1h
         alertname: AivenEstimatedCostHigh
-      - eval_time: 25h
+
+      # Alert should be firing after we've been spending £100 / hour for the last 24 hours
+      - eval_time: 24h
         alertname: AivenEstimatedCostHigh
         exp_alerts:
           - exp_labels:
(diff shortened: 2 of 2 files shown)


manifests/prometheus/spec/alerts/aiven-cost.test.yml
```diff
@@ -0,0 +1,24 @@
+---
+rule_files:
+  # See alerts_validation_spec.rb for details of how this gets set:
+  - spec/alerts/fixtures/rules.yml
+
+evaluation_interval: 1h
+
+tests:
+  - interval: 1h
```
AP-Hunt: I think I would add a second test here that shows that it doesn't fire when ..
1) there's enough data to make a projection, and 
2) the projection doesn't go above the threshold

The tests at the moment only show it doesn't fire when there's not enough data, and when the condition is met.

manifests/prometheus/alerts.d/aiven-cost.yml
```diff
@@ -7,8 +7,7 @@
     name: AivenEstimatedCostHigh
     rules:
       - alert: AivenEstimatedCostHigh
-        expr: paas_aiven_estimated_cost_pounds > 6500
-        for: 1h
+        expr: delta(paas_aiven_estimated_cost_pounds[24h]) * 30 > 20000
```
bandesz: If we want to use the correct number of months:
```
delta(paas_aiven_estimated_cost_pounds[24h]) * days_in_month(timestamp(paas_aiven_estimated_cost_pounds)) > 0
```

Or simply we can change it to 31 so it always considers the highest possible number of days.

Also should we add some smoothing, so the graph is not that spikey?

manifests/prometheus/spec/alerts/aiven-cost.test.yml
richardTowers: Yep - will add one.

manifests/prometheus/alerts.d/aiven-cost.yml
richardTowers: Good suggestions - but I think it's actually nicer to assume all months are 30 days so the rate is constant. Smoothing would be good too, but I don't think it's worth the effort to implement.

manifests/prometheus/spec/alerts/aiven-cost.test.yml
richardTowers: ✅

AP-Hunt: Tests looks good :+1: I might be tempted to swap the last two around, because the convention established in other tests is
1. Not enough data
2. Enough data but not passing the condition
3. Enough data and passing the condition

It doesn't matter that much though, so I'll approve it and merge it in the morning.

PR #1838, opened 2019-04-09
Support: Use logstash prune high cardinality fields

tlwr:
I haven't tested this at all, it is a PoC to start a discussion

What
----

We have guids in some of our field names

As of the time of writing we have over 4k fields in our index in
elasticsearch, which makes kibana hard to use

How to review
-------------

Code review

Test in the dev stack in logit

Who can review
--------------

Not @tlwr

### DIFF
--- a/config/logit/output/generated_logit_filters.conf
+++ b/config/logit/output/generated_logit_filters.conf
@@ -1022,4 +1022,11 @@ filter {
       "[@source][director]"
     ]
   }
+
+  prune {
+    blacklist_names = [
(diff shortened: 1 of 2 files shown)


tlwr:
Having received a verbal thumbs up from 👑 in Slack, I will apply this to Dev

config/logit/output/generated_logit_filters.conf
```diff
@@ -1022,4 +1022,11 @@ filter {
       "[@source][director]"
     ]
   }
+
+  prune {
+    blacklist_names = [
```
tlwr:
```suggestion
    blacklist_names => [
```

tlwr:
This is not necessary if we clear out the fields in Kibana (when old things expire the GUIDs will go away), logging stacks with low retention do not have the GUID fields.

PR #1839, opened 2019-04-09
Support: Add 4xlarge and 8xlarge quotas

tlwr:
What
----

We have some tenants who are running many things, and thus they require
a larger quota.

We will one of our tenants to the 4xlarge and have the 8xlarge on
standby in case they need to scale at short notice

How to review
-------------

- YAML review
- Check the quotas make sense

Who can review
--------------

Not @tlwr


richardTowers:
👍 happy to merge once Travis completes.

PR #1855, opened 2019-04-16
Use GitHub to provide Concourse with GPG pubkeys

jpluscplusm:
What
----

This makes the "which users are trusted by Concourse?" GPG-id to pubkey translation script clearer in operation, as it references usernames, and continues the direction of authZ moving to Github.

It munges a list of GH usernames into a YaML array of the contents of each `https://github.com/<user>.gpg`. e.g. https://github.com/jpluscplusm.gpg

It also now creates stable output, though this is perhaps at the cost of not dealing with multiple keys. For example, my `jpluscplusm` account has 2 GPG keys associated with it, but my https://github.com/jpluscplusm.gpg page only contains one key at the time of writing. Perhaps it's a cache inside GH; perhaps it's a GH feature to only return 1; perhaps something funky and PKI is happening and what's returned is both my keys merged. I didn't check - it's 10PM and you don't pay me these days :-(

It also doesn't have a local "was this key found in my local keystore" step. I can see that being useful if it were definitely, 100% adhered to, and checked back against new team members; but to my recollection any missing keys were just imported from the key server service. At least with this PR's method, that service is somewhat more formally trusted, as GDS is paying its owners ...

How to review
-------------

- Be content with using GH is a trusted authz service
- Validate the GH username list
- Use the script to generate a new `concourse/vars-files/gpg-keys.yml` file
- Visually validate parity between the old and new sets of keys (due to the change in user order)
- Nudge team members to fix their GPG keys uploaded to GH, in line with the unacceptable contents of the `gpg-keys.yml` file
- Rinse & Repeat until happy

Who can review
--------------

Probably folks with Opinions about security and authorisation.


46bit:
> I didn't check - it's 10PM and you don't pay me these days :-(
We miss you too, Jonathan! 🌃

> this is perhaps at the cost of not dealing with multiple keys. For example, my jpluscplusm account has 2 GPG keys associated with it, but my https://github.com/jpluscplusm.gpg page only contains one key at the time of writing. Perhaps it's a cache inside GH; perhaps it's a GH feature to only return 1; perhaps something funky and PKI is happening and what's returned is both my keys merged.

It does have both your keys 🎉:

```
$ wget https://github.com/jpluscplusm.gpg
$ gpg --import jpluscplusm.gpg
gpg: key 0x19C114EE13FE6BA2: public key "Jonathan Matthews <jonathan.matthews@digital.cabinet-office.gov.uk>" imported
gpg: key 0x3031E77392117D1B: public key "Jonathan Matthews <contact@jpluscplusm.com>" imported
gpg: Total number processed: 2
gpg:               imported: 2
```

[My URL](https://github.com/46bit.gpg) includes a comment: `Note: The keys with the following IDs couldn't be exported and need to be reuploaded 3EBF8EB394FCBCBB, CAAD002F768C0A74`. At least one of those has been revoked, so it makes sense.

jpluscplusm:
> We miss you too

Likewise :-) But I took those as read ;-)

> [github.com/46bit.gpg](https://github.com/46bit.gpg) includes a comment:

If you remove either your revoked or otherwise unacceptable/unused keys from https://github.com/settings/keys, does the note disappear?

I *kinda* expect the `Note` not to invalidate the file contents.

Perhaps because their "user has no keys at all" page looks like this: https://github.com/fdsfds.gpg. I *reckon* they built it so that one could safely concat/consume multiple "keys" like that and not have to check for empty blocks. Ditto for the "revoked" keys notes. But that's all I'm basing it on - perhaps it actually does invalidate things :-)

46bit:
> If you remove either your revoked or otherwise unacceptable/unused keys from https://github.com/settings/keys, does the note disappear?
>
> I kinda expect the Note not to invalidate the file contents.

It does. And GPG happily loads it despite the note.

I like this PR a lot. Ideally I'd like this to be continuously deployed so that CI checks in real-time if the GPG key matches anyone's Github profile. But sadly that means a compromised Github user could definitely deploy bad things, so the saved file will still be needed.

jpluscplusm:
> perhaps it actually does invalidate things

I should have thunked first. It's actually using the "Armor Header" feature of https://tools.ietf.org/html/rfc4880#section-6.2:

"The format of an Armor Header is that of a key-value pair" ...

So any 

```
----- begin pubkey -----
Foo: bar
Note: user is wearing no pants

actual-key-contents
----- end pubkey -----
```

... blocks are RFC-compliant.

tlwr:
This is a good idea.

However, our current trust model relies on GPG and signing commits, and not on Github. This change uses Github as the trust source, and GPG to do some double-book accounting. As Miki mentioned above, a compromised Github account could introduce untrusted code.

There is also some additional complexity introduced (multiple GPG keys from Github; API calls being made, etc).

PR #1858, opened 2019-04-17
[#164763735] Bump paas-admin release

mogds:
What
----

Enables strict null checks and updates orgQuota (https://github.com/alphagov/paas-admin/pull/126)

Who can review
--------------

Not me


richardTowers:
We should probably wait until we've resolved https://status.cloud.service.gov.uk/incidents/dx69549mcgc5 before merging this.

PR #1861, opened 2019-04-18
[#165245540] Remove cflinuxfs2

mogds:
### Blocked: please do not merge

What
----

Cflinuxfs2 (trusty) is due to go out of support at the end of the month. This removes the cflinuxfs2 stack our deployment.

Before Merging
------------

- Ensure all apps are running cflinuxfs3 using the [stack auditor](https://docs.cloudfoundry.org/adminguide/stack-auditor.html)
- Delete the cflinuxfs2 buildpacks
- Delete the the cflinuxfs2 stack 

How to review
-------------

- Deploy to your dev env

Who can review
--------------

Not me


richardTowers:
I don't think the commit that touches `config/buildpacks.yml` is necessary - this file is just a cache of the current buildpack manifests (maybe shouldn't be in version control in the first place), and it's auto-generated by `./scripts/update_buildpacks.sh`. I'll drop the commit in a force-push.

richardTowers:
My previous comment was incorrect, as confirmed by trying to run this down my dev environment.

`config/buildpacks.yml` is actually used by the pipeline to work out which buildpacks it needs to upload. If it still has the cflinuxfs2 ones then they'll fail to upload once the stack is removed.

I've pushed another commit up to remove the cflinuxfs2 buildpacks from the config.

PR #1862, opened 2019-04-18
Support: Increase CC unauthenticated rate limit to 1800

tlwr:
What
----

Bump the CC unauthenticated rate limit

How to review
-------------

Code review (read the commit messages)

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/operations.d/320-cc-set-api-limits.yml
+++ b/manifests/cf-manifest/operations.d/320-cc-set-api-limits.yml
@@ -4,4 +4,4 @@
     enabled: true
     general_limit: ((cc_hourly_rate_limit))
     reset_interval_in_minutes: 60
-    unauthenticated_limit: 1800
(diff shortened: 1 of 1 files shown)


manifests/cf-manifest/operations.d/320-cc-set-api-limits.yml
```diff
@@ -4,4 +4,4 @@
     enabled: true
     general_limit: ((cc_hourly_rate_limit))
     reset_interval_in_minutes: 60
-    unauthenticated_limit: 100
+    unauthenticated_limit: 1800
```
46bit:
```suggestion
    unauthenticated_limit: 3700
```

Since we're picking these limits arbitrarily, I'd like `3700`. This means that someone can be doing 1 unauthenticated request per second and still have some unauthenticated requests left over. In the case we've just (1 authenticated request per second) this would have prevented the issue from showing up to block people.

PR #1900, opened 2019-05-15
Switch user origin script

tlwr:
What
----

- Shows UAA admin token in `showenv` `make` task
- Adds a script to change the origin of a user (this will be documented in the team manual)

How to review
-------------

Code review

Attempt to change the origin of a user (in a dev env)

Who can review
--------------

Not @tlwr

### DIFF
--- a/scripts/update-user-origin.rb
+++ b/scripts/update-user-origin.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+require 'json'
+
+def usage
+  <<~USAGE
+    Usage: #{$0} user-guid desired-origin
+
+    e.g. #{$0} 00000000-0000-0000-0000-000000000000 google
+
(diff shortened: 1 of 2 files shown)


scripts/update-user-origin.rb
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env ruby
+require 'json'
+
+def usage
+  <<~USAGE
+    Usage: #{$0} desired-origin user-guid
```
tnwhitwell: _Nitpick_ - wouldn't it be more semantically correct if it were `#{$0} user-guid desired-origin` - the aim of the script is to set the origin of the user, not modify the origin to include that user?

tlwr: True, I should use flags :trollface: 

Updated

PR #1901, opened 2019-05-15
Switch to using "candidate" stemcell

richardTowers:
What
----

bosh haven't released an official stemcell that patches the recent Intel
CPU vulnerabilities yet.

This commit switches to using one of their "candidate" stemcell releases
(from [this commit](https://github.com/bosh-io/stemcells-core-index/commit/ef0d9d9c281cb072d238ad9fa5b6814ae631fd96) ).
This won't have had the benefit of whatever testing upstream usually do
before releasing stemcells, but it does have the linux kernel update we
need to mitigate the CVEs.

Also note this is a "heavy" stemcell, rather than the usual "light"
stemcell we use (this means it includes the machine image, rather than
using an AMI that's already been pushed to AWS).

How to review
-------------

* Code review
* Confirm it ran down a dev env successfully (see https://deployer.leeporte.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/resources/pipeline-trigger 0.0.4)

Who can review
--------------

Richard and Toby have paired on it, so probably Toby.


tlwr:
This has run down both my pipeline and Lee's with different commits but the same code

4e8a3506ec74dc2bbd0d3bec8eb7cea58284f2dc and this one

PR #1910, opened 2019-05-20
Refactor metric tags to a slice of structs

AP-Hunt:
What
----
Applying the boy scout rule. This was a problem hit during an earlier story, with an easy fix.

Previously, metric tags were a slice of strings, which were required to be "key:value" pairs. This wasn't clear from the interface alone, and was only obvious on seeing the error message in the prometheus reporter (which made the assumption about string format). In this commit, I refactored the tags to be a slice of the `MetricTag` struct, and added some stringer methods. This should make the interface clearer, and this particular error avoidable in future.

How to review
-------------
Code review
`cf push` to a dev env and check the metrics are the same as before.

Who can review
--------------
Anyone

### DIFF
--- a/tools/metrics/gauges_cf.go
+++ b/tools/metrics/gauges_cf.go
@@ -1,13 +1,13 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
-	cfclient "github.com/cloudfoundry-community/go-cfclient"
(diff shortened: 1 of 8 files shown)


46bit:
Screen Shot 2019-06-06 at 15 53 54

I'm confident this doesn't remove anything we were relying on, and it does make errors much more visible. Rather than have a big discussion I think we're best to merge this and see how it affects behaviour. 👍

PR #1912, opened 2019-05-21
Introduce deploy_env specific user accounts, and always create users

tlwr:
What
----

We are using single sign on everywhere, so we should use it

This PR allows us to only create user accounts for individual users, depending on environment, and enables user account creation everywhere, as we want it enabled always.

How to review
-------------

Run this branch down your pipeline, and google oauth yourself into admin

Who can review
--------------

Not @tlwr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ dev: globals ## Set Environment to DEV
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
 	$(eval export ALERT_EMAIL_ADDRESS?=govpaas-alerting-dev@digital.cabinet-office.gov.uk)
-	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS=${ALERT_EMAIL_ADDRESS})
+	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS?=the-multi-cloud-paas-team+dev@digital.cabinet-office.gov.uk)
 	$(eval export ENABLE_ALERT_NOTIFICATIONS ?= false)
 	$(eval export SKIP_COMMIT_VERIFICATION=true)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -129,6 +129,7 @@ dev: globals ## Set Environment to DEV
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
 	$(eval export ALERT_EMAIL_ADDRESS?=govpaas-alerting-dev@digital.cabinet-office.gov.uk)
+	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS=${ALERT_EMAIL_ADDRESS})
```
richardTowers: Why did we need to set this here? Doesn't look like we've changed the usage elsewhere in the PR.

tlwr: https://github.com/alphagov/paas-cf/blob/3a9e9d0325ec75f5848929243cbd7b90ba44b44b/concourse/scripts/pipelines-cloudfoundry.sh#L140

If this isn't set then we do not do user account creation, which before this PR we did not do in `dev`

tlwr:
This has run down my dev env, here: https://deployer.tlwr.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/post-deploy/builds/81

Makefile
richardTowers: Is it ever going to not be set now?

richardTowers: Assuming this is now actually needed, because we're always enabling `user_creation`?

Is the default of `govpaas-alerting-dev@digital.cabinet-office.gov.uk` a bit weird though? Are we just using this as the email because it's already SES validated and we don't have a better one? If so, maybe worth a comment to that effect.

tlwr: This is needed, it turns out (I didn't check before) if we had the multi-cloud email in SES, but we do. So I have added a new commit which changes this

tnwhitwell:
> I'd like to create the team users in every `dev` env, but I recognise that would be more messy. Approved! ✅

I guess this could be done with yaml anchors, but agreed - this could get grubby

PR #1913, opened 2019-05-22
[#166056703] Remove certs that are no longer required

LeePorte:
What
----

We have several certs in the pipeline that are no longer required, these include the Consul certs, which were left unused after https://github.com/alphagov/paas-cf/commit/7fbe371cfe4fe04fc54213b320d95d7fb27e6809 was merged. And, metron related certs that are no longer required due to changes in the loggregator sub-system.

This commit will remove these certs from the vars store and enable our cert checking tests to pass again. The test has been failing as these carts are no longer part of the cert rotation job and have been left in the vars store.

How to review
-------------

code review

download one of the prod cf-vars-store-yaml and modify the paths in the script and run it locally. then verify that the certs have gone in the cf-vars-store-updated.yml

Who can review
--------------

Anyone with prod access.


tlwr:
Lee and I talked about this, but for any other reviewers: this needs a bit more done otherwise more certs than necessary are deleted

PR #1915, opened 2019-05-22
Show DNS Resolution times on User Impact dashboard

46bit:
What
----

In an incident on 2019-05-20, we took awhile to realise that network problems on one of the VMs running Cloud Controller was the cause of elevated API error rates. We have seen this problem before. As this is a repeated and open issue it is definitely worth trying to improve our response next time.

This commit adds DNS Resolution times to the User Impact dashboard. It is highly visible in the team area and commonly looked at out-of-hours to know if there is a problem. Despite the name, the User Impact dashboard is not directly about user-facing problems.

The JSON is copied from manually editing this dashboard in `prod-lon`.

How to review
-------------

Here's what I'd have seen during Monday night's incident with this change applied. I think the DNS Resolution graph strongly attracts attention here:

Screen Shot 2019-05-22 at 14 59 49

The new graph is inherently spiky. Here's `prod-lon` right now:

Screen Shot 2019-05-22 at 15 25 17

Real problems do stand out, though:

Screen Shot 2019-05-22 at 14 59 56

Ideas very welcome on improvements.

Who can review
--------------

Not @46bit.

### DIFF
--- a/manifests/prometheus/dashboards.d/user-impact.json
+++ b/manifests/prometheus/dashboards.d/user-impact.json
@@ -16,7 +16,8 @@
         "graphTooltip": 0,
         "id": null,
         "links": [],
-        "panels": [{
+        "panels": [
+            {
                 "aliasColors": {},
                 "bars": false,
                 "dashLength": 10,
--- a/manifests/prometheus/dashboards.d/user-impact.json
+++ b/manifests/prometheus/dashboards.d/user-impact.json
@@ -412,7 +412,7 @@
                 "fill": 0,
                 "gridPos": {
                     "h": 8,
-                    "w": 12,
+                    "w": 6,
                     "x": 0,
                     "y": 16
                 },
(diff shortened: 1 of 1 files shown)


tnwhitwell:
I don't know.. I wonder if this would be better placed on another dashboard, as it's already quite information-dense.

I wonder if it's time for us to create a '💩💨❋' dashboard?

richardTowers:
Yeah, the dashboard is not super-well named.... But this graph seems like it's more useful than a lot of the other ones on the dashboard (I'm looking at you "UAA requests/sec", and don't think I haven't seen you hiding in the middle "Application Instances").

I think I'd be happier having this on the dashboard, maybe with a view to shrinking or removing some of the other things later to address the information density blackhole we'll otherwise accrue.

46bit:
The graph showed a problem (150ms DNS resolution times with a `log-api` VM) which let me try this graph out. I've re-pushed with a change that when hovering over the graph it lists instances starting with the longest resolution times. Ready for review again.

46bit:
I chatted with Tom and we've decided not to merge this.

PR #1931, opened 2019-06-06
Add initial SLI dashboard

richardTowers:
What
----

We've decided to measure "uptime" based on metrics from the CloudFront CDN. CloudFront sits on the outside the platform, so any errors due to the platform will be seen in the cloud front metrics.

At the time of writing there are about 40 CloudFront CDNs on the platform (across London and Ireland). They cover both production and non-production services.

Our SLI should not be affected if a few services have high error rates - this probably indicates an issue with the individual applications.
If a large number of services have high error rates at the same time we should assume there's a problem with the platform.

Initially our application uptime SLI is:

```
The average (mean) percentage of 5XX responses from CloudFront, ignoring the 5 distributions with the highest percentage of errors
```

**Known issues**

* If more than 5 services have high error rates due to non-platform issues (e.g. bad application code), this will unfairly affect our SLI
* If a platform issue causes high error rates for fewer than 5 services our SLI will not be affected
* The `5xx_errorrate` metric is a percentage, so low-traffic services with a few requests per minute are weighted equally with high-traffic services which have many requests per second

How to review
-------------

* View existing dashboard in Ireland: https://grafana-1.cloud.service.gov.uk/d/zIQ-cPMZk/service-level-indicators?orgId=1
* Sanity check the JSON

Who can review
--------------

Not @richardtowers or @tnwhitwell

### DIFF
--- a/manifests/prometheus/dashboards.d/sli.json
+++ b/manifests/prometheus/dashboards.d/sli.json
@@ -1,0 +1,253 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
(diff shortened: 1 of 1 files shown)


tlwr:
:shipit:

PR #1932, opened 2019-06-06
Adjust user impact dashboard to be more legible

tnwhitwell:
Why
----

The graphs showing 2xx, 4xx and 5xx were scaled according to the highest value (usually the 2xx requests), and so 4xx and 5xx were not easily visible. Also, the colours blended together quite a bit

What
----
The request type with the most requests has now been moved to the right
axis, which means that 4xx and 5xx are autoscaled separately.

Also, the 2xx results were a little too BIG and BOLD, which drew
attention away from the 'point' of the graph (showing the problems).
These are now much less bold, so the eye is drawn to the issues.

FINALLY, Colours for the request graphs have been set manually, so
they merge together less.

How to review
-------------

Do a deploy from this branch and behold the majesty of my creation

(This has been deployed to my environment already, so I guess you could just look there!)

Who can review
--------------

Anyone!

### DIFF
--- a/manifests/prometheus/dashboards.d/user-impact.json
+++ b/manifests/prometheus/dashboards.d/user-impact.json
@@ -17,12 +17,16 @@
         "id": null,
         "links": [],
         "panels": [{
-                "aliasColors": {},
+                "aliasColors": {
+                    "4xx Responses": "orange",
+                    "5xx Responses": "red",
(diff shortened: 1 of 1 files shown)


46bit:
> Do a deploy from this branch and behold the majesty of my creation

👑

PR #1933, opened 2019-06-06
Scale up gorouters in `prod` (IE) from 3 to 6 [do not merge until June 10 8AM)

tlwr:
What
----

A tenant is doing load testing on June 10/11, this is commit scales up
the routers for this event

How to review
-------------

Cod review 🐟

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/env-specific/prod.yml
+++ b/manifests/cf-manifest/env-specific/prod.yml
@@ -1,6 +1,6 @@
 ---
 cell_instances: 24
-router_instances: 3
+router_instances: 6
 api_instances: 4
 doppler_instances: 12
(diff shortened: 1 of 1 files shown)


manifests/cf-manifest/env-specific/prod.yml
```diff
@@ -1,6 +1,7 @@
 ---
 cell_instances: 24
 router_instances: 3
+router_instances: 6
```
richardTowers: This line's a duplicate of the key above

tlwr:
:facedesk:

46bit:
🛫 **Merge from 8am on Monday June 10th** 🛫

PR #1935, opened 2019-06-07
[#166348084] Upgrade cf-deployment to 9.2 from 7.9

LeePorte:
What
----

Upgrades cf-deployment to 9.2 from 7.9

In cf-deployment 8.0 cflinuxfs2 was removed, so this incorporates some changes from upstream with existing changes we made.

We had a custom opsfile for working around the route service change in routing release v0.188.0 which we have now replaced with a reference to upstream.

Upgrades cf-acceptance-tests to 9.2 from 7.9, whilst doing this we also upgraded the version of go to 1.12.5. We had some trouble debugging some tests to do with the logging pipeline, so we also added two tests that fail obviously when `cf logs` does not receive application logs and router logs.

Split the custom acceptance tests job in the pipeline out into two:
- custom-acceptance-tests
- custom-broker-acceptance-tests
And reduce the parallelism for the tests, add ginkgo allowances for flakey tests. We split them out so that we rerun fewer tests when they flake, and so when developing/doing a cf upgrade, we can run platform custom tests separately from our broker tests. Hopefully this will encourage us to add more tests.

In more recent versions of CF the reverse log proxy (and reverse log proxy gateway) are generally available. We added an ALB (like Prometheus, unlike the classic ELBs for everything else) for `log-api.((system_domain))` and `log-search.((system_domain))` (one LB, two DNS records). This exposes the RLP part of loggregator to tenants, where auth is done using UAA auth token. The load balancer is provisioned via terraform but managed by Bosh (AWS CPI).

We added some more certificates to the certificate rotation ops file, because more certificates were added. And some more components now _require_ TLS, so we had to update a port in our doppler load balancer. This raised a well-known bug with classic ELBs and SSL policies, for which we used terraform `random_pet` to force the updating of TLS policies on ELBs. However people may experience failing custom acceptance tests due to a regression with SSL policies. This can be fixed by changing it through the Amazon console. The real fix is in another story which is to use ALBs.

We experienced some flakey tests, so we reduced the parallelism, or fixed them in individual commits.

We found quite a few places in our pipelines where governmentpaas docker images were not from the latest commit, so we added a test which checks for that, as well as that they are well formed YAML (if there is such a thing)

Who/How to review
-------------

This was done as a pair between @tlwr and @LeePorte


tlwr:
Just waiting on `s/go-1.12.5/$hash/` for latest docker images

PR #1941, opened 2019-06-12
[#166440328] Change cells to `r5.xlarge` EC2 instance type

46bit:
What
----

We want to utilise Reserved Instances to reduce our EC2 costs. We want to reserve the newest type of instances so that we are investing effectively.

Until now our cells have used the `r4.xlarge` EC2 instance type. This commit changes us to use the `r5.xlarge` EC2 instance type. These new instances have more memory (32GiB rather than 30.5GiB [1]) and so this commit updates where that number is used.

[1] https://www.ec2instances.info/?min_memory=20&min_vcpus=4&region=eu-west-2&selected=r4.xlarge,r5.xlarge

How to review
-------------

Code review. This is deployed to the `miki` dev environment and everything went green [2].

[2] `custom-acceptance-tests` failed, but the only failing test is because I still haven't deleted the `cflinuxfs2` buildpacks from the `miki` dev environment.

Who can review
--------------

Not @46bit.

### DIFF
--- a/manifests/cf-manifest/spec/manifest/base_properties_spec.rb
+++ b/manifests/cf-manifest/spec/manifest/base_properties_spec.rb
@@ -178,10 +178,10 @@ RSpec.describe "base properties" do
     describe "executor" do
       subject(:executor) { manifest["instance_groups.diego-cell.jobs.rep.properties.diego.executor"] }
 
-      it "should have a memory_capacity_mb of at least 30G" do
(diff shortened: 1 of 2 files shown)


tlwr:
This looks good to me, I don't want to approve/merge it yet until I go through and ensure that we have adequate capacity limits in EC2 in London and Ireland in all accounts.

tlwr:
I have raised the support tickets.

PR #1942, opened 2019-06-13
Bump the Instance Types for the Databases

paroxp:
## What

M5 instances are the latest generation of General Purpose Instances and
provide improved performance over M4.

Also comes with a nice discount of ~2.600473% :D

This should not change the current instances automatically. It will only
be applied to new instances, unless the `cf update-service` command has
been executed, which would mark the database to be automatically
upgraded during the next maintenance window.

Partially tested on Pay - no CF interaction.

How to review
-------------

- Code review - I may be rusty on CF :trollface: 
- Spin up dev environment from `master`
- [Set up a PostgreSQL service](https://docs.cloud.service.gov.uk/deploying_services/postgresql/#set-up-a-postgresql-service)
- Spin up dev environment from `bump-rds-instance-type`
- Change [PostgreSQL maintenance times](https://docs.cloud.service.gov.uk/deploying_services/postgresql/#postgresql-maintenance-times) to something soon
- Wait and see successful upgraded service 🤔

### DIFF
--- a/manifests/cf-manifest/operations.d/711-rds-broker-postgres95-plans.yml
+++ b/manifests/cf-manifest/operations.d/711-rds-broker-postgres95-plans.yml
@@ -21,21 +21,21 @@
       allowed_extensions: ["uuid-ossp", "postgis", "pg_stat_statements"]
 
     tiny_plan_rds_properties: &tiny_plan_rds_properties
-      db_instance_class: "db.t2.micro"
+      db_instance_class: "db.t3.micro"
       allocated_storage: 5
       backup_retention_period: 0
       skip_final_snapshot: true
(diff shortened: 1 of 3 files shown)


tlwr:
I think we have to create new plans, with new prices, and encourage tenants to upgrade, rather than change the plans.

paroxp:
I agree, that was my first take on this.

But then, we are reducing prices, and there is no impact on existing users. I'm happy to have a bigger kick-off for it :)

richardTowers:
Yeah - needs a discussion - if it's possible for our tenants not to have to worry about instance types that would be great, but it depends on how smooth we can make the upgrade process.

Thanks for the PR!

tlwr:
We have a story to do this, closing this

PR #1945, opened 2019-06-14
Update grafana org name and org users role

tlwr:
What
----

- Make `configure-grafana` task into a `task` file rather than inline ruby
- When configuring Grafana, add the new users into the org
- When configuring Grafana, ensure all users are org admins

This is so everyone gets the `Explore` button in the sidebar without having to fiddle with their user account.

How to review
-------------

- Run this down your pipeline
- Using the `admin` account, delete your user
- Run `prometheus-deploy` twice

Who can review
--------------

Not @tlwr


richardTowers:
> Looks fine to me. One comment (which you can choose not to address if you like):
> 
> There are now a few things that do set-like operations, but don't use sets. e.g.
> 
> ```
>         users_to_create = users_that_should_exist.reject do |email|
>           grafana_users.any? { |u| u.dig('email') == email }
>         end
> ```
> 
> It would save a bit of noise if you did `.to_set` at the end of each chain and then used `-` and `+` rather than `.select` and `.reject`

Ah, although now I see you're rejecting objects based on string values, so probably not possible.

tlwr:
Unfortunately the types are different so taking the difference of sets is not possible, as you say :(

PR #1947, opened 2019-06-14
Scale up the number of API instances

LeePorte:
What
----

Since DiT have reenabled prometheus in the staging org we are triggering alerts.

How to review
-------------

code review

Who can review
--------------

anyone


LeePorte:
I've doubled it as we're seeing around double the amount of traffic

PR #1954, opened 2019-06-18
Allow subshell scope logins for dev envs [skip ci]

AP-Hunt:
What
----

In the same spirit as `scripts/cf_subshell_scoped_login.sh`, this commit adds
an option to `scripts/cf_admin_login_dev` to allow users to scope the login to
a subshell. It uses the same method to do that as the former.

This is intended as a quality of life improvement, for those times when devs
are working across multiple spaces and the global nature of the `cf` tool is a
hindrance.

How to review
-------------

1. Code review
2. Try it out

Who can review
--------------
Not @AP-Hunt

### DIFF
--- a/scripts/cf_admin_login_dev.sh
+++ b/scripts/cf_admin_login_dev.sh
@@ -2,7 +2,22 @@
 
 set -euo pipefail
 
+usage() { echo "Usage: $0 [-s]" 1>&2; exit 1; }
+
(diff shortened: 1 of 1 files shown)


AP-Hunt:
Done a final check (see below), and the script isn't used in any of our pipelines. Only mentioned in the `.git` directory.

```
$ cf paas-cf && find . -type f -not -path "*vendor*" -exec grep 'cf_admin_login_dev' {} \;
```

PR #1959, opened 2019-06-24
Update the pausing and unpausing of kickoff

LeePorte:
What
----

With Concourse 5.x series pausing and unpausing of resources has been replaced by pinning and unpinning of resources. This PR reinstates the behaviour we had previously.

How to review
-------------

Code review
Deploy in your env and let it pin

unpin later in the day and see you have an env ready in the morning

Who can review
--------------

Anyone


tlwr:
Haven't deployed but Code review seems fine

PR #1965, opened 2019-07-04
Add "group" to pricing plans along with name

richardTowers:
What
----

Sometimes we want to ask a question like:

"How much has this organisation spent on postgres?"
"How much has this organisation spent on compute (including staging and tasks)?"

At the moment this is more difficult than it needs to be because we have
to add up all the various postgres plans the organisation might be
using. This currently involves doing string processing to work out which
plans are postgres based on which names start with "postgres", which
feels super hacky.

The plan would be to add a group field to the pricing plans in the DB,
and then expose this through the /pricing_plans API in paas-billing.
This would allow us to remove the code that does string processing on
"name".

This commit adds the following groups:

* aws-s3-bucket
* cloudfront
* compute (includes app, staging and task)
* elasticsearch
* mongodb
* mysql
* postgres
* prometheus
* redis
* unknown

I think it's best to add the groups to the existing plans so they're
there for all history. The [paas-billing code](https://github.com/alphagov/paas-billing/blob/94f16d71fda09ae19278e213e5e93743231deadb/eventstore/store.go#L210-L224)
replaces all the pricing plans when it starts up. Once that code is
updated to include groups (separate PR to be created later) we'll be
able to start exposing them in queries.

How to review
-------------

* Firstly think about whether this is a good idea in general (is there
  some complexity I'm missing?).
* If so, is "group" the best name for the
  new key, or would something else ("category"?) be better? Do the
  groupings I've come up with make sense?
* Confirm that merging this isn't going to break paas-billing somehow

Who can review
--------------

Not @richardtowers - might be relevant to @46bit 's interests?

### DIFF
--- a/config/billing/config.json.erb
+++ b/config/billing/config.json.erb
@@ -32,6 +32,7 @@
 	"pricing_plans": [
 		{
 			"name": "unknown-plan",
+			"group": "unknown",
 			"valid_from": "1970-01-01",
 			"plan_guid": "d5091c33-2f9d-4b15-82dc-4ad69717fc03",
 			"components": [
@@ -45,6 +46,7 @@
 		},
 		{
 			"name": "prometheus",
+			"group": "prometheus",
 			"valid_from": "2017-01-01",
 			"plan_guid": "cc65268f-98e9-40ef-aa8c-8e595204064d",
 			"components": [
@@ -58,6 +60,7 @@
 		},
 		{
 			"name": "prometheus",
+			"group": "prometheus",
 			"valid_from": "2017-01-01",
 			"plan_guid": "7ab58e11-3959-49b4-9218-807e3e78f9af",
 			"components": [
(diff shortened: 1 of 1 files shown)


46bit:
I note that https://github.com/alphagov/paas-billing/blob/94f16d71fda09ae19278e213e5e93743231deadb/eventstore/store.go#L210-L224 you linked does not do anything with the `group` field, and you'll probably need to a `Group` field to its pricing plan struct. So there is probably a second PR to do there.

richardTowers:
Yeah, there's a second PR needed (I was going to merge it into this one, but I couldn't work up the enthusiasm to determine whether the deployment order mattered). Agree we'll need some more metadata fields in this PR to actually be useful though - your comment in https://github.com/alphagov/paas-admin/pull/159 suggests `serviceName` `planName` `planVersion` and `planVariant` - maybe we should add them all?

46bit:
@richardTowers Now that you've rebuilt the calculator, is this still useful?

richardTowers:
I think so (although this PR still isn't complete, as mentioned above).

My calculator still uses the `/pricing_plans` endpoint, which still has this problem.

tlwr:
Closing this as we now have additionalMetadata

PR #1975, opened 2019-07-12
Add PaaS-admin to Pingdom

tlwr:
What
----

I accidentally deleted the route in CF for PaaS-admin, because I am a fool and because `cf d -r -f app` which shares a route with another app deletes routes for both apps.

I would like to avoid this happening again, by making Pingdom tell us when PaaSmin goes down.

How to review
-------------

Code review

Who can review
--------------

Not @tlwr

### DIFF
--- a/terraform/pingdom/pingdom.tf
+++ b/terraform/pingdom/pingdom.tf
@@ -66,3 +66,18 @@ resource "pingdom_check" "cf_api_healthcheck" {
   notifywhenbackup         = true
   contactids               = ["${var.pingdom_contact_ids}"]
 }
+
+resource "pingdom_check" "paas_admin_healthcheck" {
+  type                     = "http"
+  name                     = "PaaS CF API - ${var.env}"
+  host                     = "admin.${var.system_dns_zone_name}"
+  url                      = "/calculator"
+  shouldcontain            = "costs"
+  encryption               = true
+  resolution               = 1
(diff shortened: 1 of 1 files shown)


terraform/pingdom/pingdom.tf
```diff
@@ -66,3 +66,18 @@ resource "pingdom_check" "cf_api_healthcheck" {
   notifywhenbackup         = true
   contactids               = ["${var.pingdom_contact_ids}"]
 }
+
+resource "pingdom_check" "paas_admin_healthcheck" {
+  type                     = "http"
+  name                     = "PaaS CF API - ${var.env}"
+  host                     = "admin.${var.system_dns_zone_name}"
+  url                      = "/calculator"
```
richardTowers: Presumably `/calculator` so this covers a bit of the billing API too? (It does GET /pricing_plans I think) Or just because it's a page that doesn't need auth?

tlwr: Just because it doesn't need auth, although I was thinking of adding a health endpoint instead

richardTowers: I think /calculator is fine for now. In a way it's nice that it covers billing a little bit, although that will make it a bit less reliable.

tlwr:
Deployed this manually

PR #1976, opened 2019-07-12
Rename PaaS admin pingdom check to be correct

tlwr:
What
----

Fix typo, currently we have two checks with the same name

How to review
-------------

Code review

Who can review
--------------

@richardTowers


richardTowers:
Oops

PR #1977, opened 2019-07-15
Revert "Merge pull request #1961 (Microsoft SSO)"

richardTowers:
What
----

There's some issue with UAA's configuration that seems to only manifest
occasionally (so it managed to evade us in some dev environments).

This reverts commit 2e7495454342b613b7ff263bc6fe8eb66f8f5aab, reversing
changes made to 259f208907790b44b3da5faa8f603d9373cbc85a.

How to review
-------------

* Git review should be enough (e.g. `git diff prod-0.2.355` should be empty)

Who can review
--------------

Not @richardTowers


richardTowers:
We think we've identified the issue, so we're going to roll forward with a patched UAA.

PR #1982, opened 2019-07-15
Update uaa release to be zdd with v73.4.0

tlwr:
***do not merge***

What
----

We want to release a version of UAA which uses the
DelegatingPasswordEncoder which is zero downtime, once UAA understands
BOTH password hashes with the "{bcrypt}" prefix and without, then we can
release a version which runs migrations to ensure that all password
hashes start with '{bcrypt}'

How to review
-------------

Be @richardTowers


richardTowers:
Code review wise, this is https://github.com/alphagov/paas-uaa/compare/a8f574db5a48ecb6db80b344f0b81542238c4a7d...689ab59170e0b09c8873b578cfa99ca59ca2e2fe

tlwr:
This has been down my pipeline and went green

PR #1984, opened 2019-07-16
Scale down cells

46bit:
What
----

This reduces our number of cells to match what we actually need. See commit messages for full information.

I couldn't identify quite what caused the change in demand but my suspicion is that this is because of continued moves away from Ireland, the end of some tenant load testing and a pause in Portfolio's testing.

### 📉 Number of cells in Ireland: 24 ➡️ 18

Screen Shot 2019-07-16 at 11 37 12

### 📉 Number of cells in London: 30 ➡️ 27

Screen Shot 2019-07-16 at 11 37 32

### 📉 Money saved: yes 💸 

How to review
-------------

Code review.

Who can review
--------------

Not @46bit. Thanks to @venusbb for highlighting that we could do this.

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -1,5 +1,5 @@
 ---
-cell_instances: 30
+cell_instances: 27
 router_instances: 3
 api_instances: 8
 doppler_instances: 12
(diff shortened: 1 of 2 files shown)


46bit:
It turns out that we’ve already reserved the current number of EC2 instances, so we wouldn’t save any money by doing this. I'm sure we can find a better use for them in time but for now since we're paying for this many machines we may as well enjoy the reliability benefits.

PR #1990, opened 2019-07-17
User changes

tlwr:
What
----

- Remove Chris; different team, no more access
- Remove Rafal; different team, no more access
- Give Mo access to production; security clearance acquired and confirmed
- Add Emma (`cf_admin: false`; only SSO to Grafana)

How to review
-------------

Code review

Who can review
--------------

Not @tlwr

### DIFF
--- a/config/users.yml
+++ b/config/users.yml
@@ -50,3 +50,7 @@
 - email: "paul.dougan@digital.cabinet-office.gov.uk"
   origin: "google"
   deploy_envs: ['prod', 'prod-lon', 'stg-lon']
+
+- email: "emma.pearce@digital.cabinet-office.gov.uk"
+  origin: "google"
+  deploy_envs: ['prod', 'prod-lon', 'stg-lon']
--- a/config/users.yml
+++ b/config/users.yml
@@ -16,7 +16,7 @@
 
 - email: "mohamed.deerow@digital.cabinet-office.gov.uk"
   origin: "google"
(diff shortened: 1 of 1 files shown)


config/users.yml
```diff
@@ -60,3 +50,7 @@
 - email: "paul.dougan@digital.cabinet-office.gov.uk"
   origin: "google"
   deploy_envs: ['prod', 'prod-lon', 'stg-lon']
+
+- email: "emma.pearce@digital.cabinet-office.gov.uk"
+  origin: "google"
+  deploy_envs: ['prod', 'prod-lon', 'stg-lon']
```
paroxp: I know the default is `cf_admin: false`, but I'm debating weather it may be beneficial to make it clear.

config/users.yml
```diff
@@ -21,12 +16,7 @@
 
 - email: "mohamed.deerow@digital.cabinet-office.gov.uk"
   origin: "google"
-  deploy_envs: ['modeerow']
-  cf_admin: true
-
-- email: "rafal.proszowski@digital.cabinet-office.gov.uk"
```
paroxp: Sad times... I already miss you.

I shall presume it will not remove the account?

tlwr: It will delete the account https://github.com/alphagov/paas-cf/blob/master/scripts/sync-admin-users.rb#L38

PR #2000, opened 2019-07-24
[Spike] Add Grafana and Metric Store

tlwr:
This is a spike for people to deploy to their development environments, so that they can test [metric-store](https://github.com/cloudfoundry/metric-store-release).

What
----

Metric Store is an experimental Loggregator component which provides a Prometheus API compatible Cloud Foundry auth aware interface for your CF platform metrics.

Currently it only works for admins, but it is [theoretically possible](https://github.com/cloudfoundry/metric-store-release/issues/39) for it to be exposable to tenants.

This pull request does two things:

1. Includes `metric-store-release` using the upstream experimental ops file, with some custom config (multi-AZ in non-dev, configure machine type)

2. Adds a Grafana, which uses CF UAA for auth, which can connect to Metric Store (masquerading as Prometheus)

_Please note that metric-store is deployed to metric-store on the `((app_domain))` as well as the `((system domain))`, because in our developer environments we lock down ingress to GDS IP space, which prevents Grafana from talking to metric-store._

How to get started
-------------

1. Deploy this to your development environment

2. Navigate to `https://observe.env.dev.cloudpipeline.digital`

3. Sign in using `GOV.UK PaaS`

4. Explore, example query for a tenant app would be `http_count{source_id="56da9ba4-dd0a-44df-bbf5-c3c3ec862424"}` where `source_id` is an app GUID.

### DIFF
--- a/manifests/cf-manifest/operations.d/100-cf-update-vm-types.yml
+++ b/manifests/cf-manifest/operations.d/100-cf-update-vm-types.yml
@@ -29,6 +29,9 @@
 - type: replace
   path: /instance_groups/name=rotate-cc-database-key/vm_type
   value: errand
+- type: replace
+  path: /instance_groups/name=metric-store/vm_type
+  value: medium
 
 # TODO: Remove the router vm_type. Not needed with extensions
 - type: replace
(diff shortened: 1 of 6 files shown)


46bit:
https://content.pivotal.io/blog/metric-store-a-cloud-native-time-series-database-for-cloud-foundry provides more information on Metric Store (as of April.)

tlwr:
Closing this as the prototype has moved on significantly

PR #2016, opened 2019-08-02
Monitor exchange rates using Prometheus

tlwr:
What
----

Add USD currency metrics to paas-metrics

Events have caused the value of our charging currency (GBP) to change not insignificantly.

The exchange rate is not something we keep in our heads regularly, but has a significant impact on cross-charging and re-charging for infrastructure; a large portion of our infrastructure costs are in USD, so we risk losing money.

This PR does the following:

- Add 2 metrics to paas-metrics
  - One which is hardcoded to the same value as our exchange rate
  - Another uses the European Central Bank currency rates to get a GBP/USD exchange rate.
- Adds a test to ensure that we keep our paas-metrics and config/billing rates the same (eventually expose the exchange rates inside paas-billing with an endpoint maybe?)
- Adds a Prometheus alert which tells us when our exchange rate has drifted

How to review
-------------

Code review

Travis

See [tlwr paas-metrics](https://paas-metrics.tlwr.dev.cloudpipelineapps.digital/metrics) metrics:

- `paas_currency_configured_ratio`
- `paas_currency_real_ratio`

See [tlwr Prometheus Alerts](https://prometheus-1.tlwr.dev.cloudpipeline.digital/alerts) to see that the alert is firing

Who can review
--------------

Not @tlwr


46bit:
e> eventually expose the exchange rates inside paas-billing with an endpoint maybe?

There also is an endpoint, added in https://github.com/alphagov/paas-billing/pull/70. 🎉 You can use it as `/currency_rates?range_start=2019-08-05&range_stop=2019-08-06`, giving:

```
[{"code":"GBP","valid_from":"1970-01-01T00:00:00+00:00","rate":1},{"code":"USD","valid_from":"1970-01-01T00:00:00+00:00","rate":0.8}]
```

manifests/prometheus/alerts.d/exchange-rate.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+- type: replace
+  path: /instance_groups/name=prometheus2/jobs/name=prometheus2/properties/prometheus/custom_rules?/-
+  value:
+    name: ExchangeRatesDrifting
+    rules:
+      - alert: ExchangeRatesDrifting_Critical
+        expr: "paas_currency_real_ratio > paas_currency_configured_ratio"
```
46bit: It reads like this will alert on any decrease in purchasing power. I wonder if we're better off detecting anything outside a 5–10% difference, as otherwise we will see this alarm go off very often (and from temporary changes.) That's only my opinion though.

tlwr: My reasoning for this value was, if we're eating money with the rate that the ECB gives us, we're probably eating money from whatever exchange rate AWS provides, but yeah your comment makes sense the more I think about it

In your opinion would:

```
(paas_currency_real_ratio / paas_currency_configured_ratio) > 1.025
```

Screen Shot 2019-08-06 at 17 22 23

46bit: Looks good. I'd gently nudge that we should be updating the exchange rate when it changes in our favour too.

tlwr:
Rebased this on master with the new alerts

manifests/prometheus/alerts.d/exchange-rate.yml
tlwr: There is now an alert for this as well :)

tlwr:
So this turned into a bit of a monster :/

I ended up refactoring most of paas-metrics and a lot of the tests.

I broke out most of the things into separate packages so we didnt have one huge `main` package. This work isn't completely finished IMO (we could definitely move the metrics into their own packages more).

This PR now does the following:

- Adds a metric which tracks the live exchange rate between USD and GBP
- Adds metrics for the currently configured exchange rates in paas-billing (i.e. no more `plan_guids.json`)
- Adds an alert when the configured currency is 2.5% higher or lower than the configured exchange rate
- Adds some more tests
- Makes some old tests use Ginkgo/Gomega instead of `testing` directly

How to review:

- Run the tests
- Deploy to your development environment
- Look at this query: `(paas_currency_real_ratio / paas_billing_currency_configured_ratio)`
- Look at [tlwr](https://paas-metrics.tlwr.dev.cloudpipelineapps.digital/metrics)
  - Search for: `paas_billing_currency_configured_ratio`
  - Search for: `paas_currency_real_ratio`

PR #2019, opened 2019-08-07
[#167565279] Change the default cdn-broker cache TTL to 0

tnwhitwell:
## What

Change the default caching TTL in new CloudFront distributions from 86400 (1 day) to 0 (0 days).

Allow a custom default TTL to be set via the `default_ttl` config value

## Why

We advertise the CDN broker as being the way to use a custom domain with the PaaS. It comes with caching enabled with a 1 day TTL by default, which is a bit daft as people would not expect this.

## How to review

* Deploy to a dev env and test that a CloudFront distribution is created with a 0 TTL
* Maybe do some updates to see if it changes properly?

## What do do before merging

Ensure that https://github.com/alphagov/paas-cdn-broker-boshrelease/pull/19 has been merged. Update `manifests/cf-manifest/operations.d/720-cdn-broker.yml` so that the release version is the one built after that merge

## Who can review
Not me - I don't want to see this again.

### DIFF
--- a/manifests/cf-manifest/operations.d/720-cdn-broker.yml
+++ b/manifests/cf-manifest/operations.d/720-cdn-broker.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: cdn-broker
-    version: 0.1.16
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/cdn-broker-0.1.16.tgz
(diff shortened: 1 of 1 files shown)


tlwr:
Closing this in favour of https://github.com/alphagov/paas-cf/pull/2022

~continuous delivery~

PR #2021, opened 2019-08-09
[#167804765] Update our forked UAA to upstream's v74.0.0

46bit:
What
----

Mitigate https://www.cloudfoundry.org/blog/cve-2019-11274/ by using https://github.com/alphagov/paas-uaa-release/pull/10 which uses https://github.com/alphagov/paas-uaa/pull/7.

How to review
-------------

See the testing plan in https://github.com/alphagov/paas-uaa/pull/7.

🐝🐝🐝 **Before merging, update the `[WIP]` commit to use the build from merging https://github.com/alphagov/paas-uaa-release/pull/10.** 🐝🐝🐝

Who can review
--------------

Not @46bit

PR #2024, opened 2019-08-13
[#167862846] [CVE-2019-9893] Upgrade to cf deployment v11.0.0

46bit:
What
----

Upgrades our `cf-deployment` from `v10.0.0` to [`v11.0.0`](https://github.com/cloudfoundry/cf-deployment/releases/tag/v11.0.0).

ℹ️ This would include UAA `74.0.0` except for our fork. In https://github.com/alphagov/paas-cf/pull/2021 we have already upgraded our fork to be based on that version.

ℹ️ This would upgrade our buildpacks except I've been told we pin them.

How to review
-------------

* Code review;
* See if the tests pass;
* Try deploying to your dev env and check everything passes;
* Check that the buildpacks were not upgraded (as we'd need to do tenant comms.)

So long as support stays quiet, Miki is going to deploy it and test it to speed things up.

Who can review
--------------

Not @46bit

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -302,13 +302,14 @@ resources:
     type: git
     source:
       uri: https://github.com/cloudfoundry/cf-acceptance-tests
-      branch: cf10.0
+      branch: master
+      tag_filter: v3.5.0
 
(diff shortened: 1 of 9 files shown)


tlwr:
This has tests passing in my dev environment

PR #2030, opened 2019-08-20
[#167572519] Use credhub to deploy paas-apps

mogds:
What
----

- Updates image tags
- Adds a task to copy vars from S3 to credhub
- Updates the create-cloudfoundry pipeline to use credhub values when deploying paas-admin, paas-billing, paas-metrics and paas-accounts
- Removes retrive-config task from the above

How to review
-------------

- Deploy to your dev env

Who can review
--------------

Not me


tlwr:
Works in my pipeline

PR #2031, opened 2019-08-21
Switch from UnblendedCost to AmortizedCost

richardTowers:
What
----

Because we reserve instances, the UnblendedCost appears artificially low
(since most of our instances were paid for back in July).

This makes comparing the money we're spending on AWS with the money
we're getting from billing very difficult - it would be much easier if
we reported the AmortizedCost metric instead.

How to review
-------------

* Code review
* Run it down a pipeline and check the metrics still work

Who can review
--------------

Not Richard


46bit:
👏 for thinking of this

PR #2032, opened 2019-08-21
Fixes for billing alerts

richardTowers:
What
----

These are currently alerting because we're only billing for ~70% of our
RDS usage, but this may actually be reasonable because we run some RDS
instances ourselves.

The EC2 alerts were pointing at the wrong metrics, which themselves were not being calculated correctly given our reserved instances (fixed in #2031).

I've also added a "layer: billing" label to them so we can hide them from Grafana.

How to review
-------------

* Code review should be enough

Who can review
--------------

Not Richard

### DIFF
--- a/manifests/prometheus/alerts.d/paas-billing-costs.yml
+++ b/manifests/prometheus/alerts.d/paas-billing-costs.yml
@@ -7,12 +7,12 @@
     name: PaaSBillingDBCostsExceedRDSCosts
     rules:
       - alert: PaaSBillingDBCostsExceedRDSCosts
-        expr: sum(increase(paas_billing_total_costs_pounds{name=~"mysql.+|postgres.+"}[24h])) / sum(paas_aws_cost_explorer_by_service_pounds{service="Amazon Relational Database Service"}) > 1.1
+        expr: sum(increase(paas_billing_total_costs_pounds{name=~"mysql.+|postgres.+"}[24h])) / sum(paas_aws_cost_explorer_by_service_pounds{service="Amazon Relational Database Service"}) > 1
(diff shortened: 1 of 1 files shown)


manifests/prometheus/alerts.d/paas-billing-costs.yml
```diff
@@ -46,9 +49,10 @@
     name: PaaSBillingAppCostsFallsShortOfEC2Costs
     rules:
       - alert: PaaSBillingAppCostsFallsShortOfEC2Costs
-        expr: sum(increase(paas_billing_total_costs_pounds{name="app"}[24h])) / sum(paas_aws_cost_explorer_by_service_pounds{service=~"^EC2.+"}) < .9
+        expr: sum(increase(paas_billing_total_costs_pounds{name="app"}[24h])) / sum(paas_aws_cost_explorer_by_service_pounds{service=~"Amazon Elastic Compute Cloud - Compute"}) < .9
```
tlwr:
```suggestion
        expr: sum(increase(paas_billing_total_costs_pounds{name="app"}[24h])) / sum(paas_aws_cost_explorer_by_service_pounds{service=~"(Amazon Elastic Compute Cloud - Compute)|(^EC2.+)"}) < .9
```

I would err on the safe side with ⬆️ but theoretically we shouldn't care about "EC2 - Other"

:shipit:

richardTowers: I think "EC2 - Other" is small enough to ignore, and it keeps the thing simpler just having one.

If the metric changes its name then we'll notice fairly quickly.

PR #2033, opened 2019-08-22
Increase memory allocation to Envoy proxy

tlwr:
No story

What
----

We have seen some weird timeouts with some of our tenants, during all
parts of HTTP requests.

Hypothesis is that Envoy is running out of memory which causes the
timeouts

Increase memory 32mb to 48mb

How to review
-------------

Run it down your pipeline in an environment where `SLIM_DEV_DEPLOYMENT` is `false`

Who can review
--------------

Not @tlwr


tlwr:
Mo reported 99.7 in API availability and good for app availability when reviewing

PR #2038, opened 2019-08-27
Revert "Update exchange rate to match current average"

richardTowers:
What
----

This reverts commit f0ad7e6670ac066be39020b5543d7d6d07599d22.

It turns out that the extra complexity that this adds to billing's
queries causes it to fall over.

We could consider changing the `epoch` exchange rate and relying on the
consolidation to keep the past consistent, but for now let's just try
and fix the thing.

GIF of a cat entering a bathtub

How to review
-------------

* Code review should be enough

Who can review
--------------

Not @richardTowers

### DIFF
--- a/config/billing/config.json.erb
+++ b/config/billing/config.json.erb
@@ -9,11 +9,6 @@
 			"code": "USD",
 			"valid_from": "epoch",
 			"rate": 0.8
-		},
-		{
-			"code": "USD",
-			"valid_from": "2019-08-01",
-			"rate": 0.82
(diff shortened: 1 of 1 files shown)


richardTowers:
Some more images that illustrate the effect of this change:

Burst balance:

burst balance graph showing a huge spike on 24/08/2019

IOPS:

IOPS graph showing a huge spike on 24/08/2019

CPU:

CPU graph showing a huge spike on 24/08/2019

PR #2044, opened 2019-09-05
[#167986184] Put Doppler behind an ALB

AP-Hunt:
What
----
Points doppler.((system_domain)) at the Loggregator ALB

The Loggeregator ALB already has a pair of listener rules, for its 443 listener, which route traffic to either the reverse log proxy gateway or doppler based the host header. Doppler had previously been on its own classic ELB in SSL mode, in order to support web sockets. However, ALBs support web sockets, so routing the doppler traffic through the ALB should be sufficient to support web sockets with fewer load balancers.

🚨There will be a second PR which remove the old ELB 🚨 

How to review
-------------
1. Code review
2. Run it down your pipeline
3. Log in to your env and tail the logs for an app. Check it works.

Who can review
--------------
Anyone

### DIFF
--- a/manifests/cf-manifest/operations.d/360-log-api.yml
+++ b/manifests/cf-manifest/operations.d/360-log-api.yml
@@ -25,11 +25,9 @@
 
 - type: replace
   path: /releases/name=loggregator
-  # We cannot use a version of loggregator past 105.5 until we get rid of ELBs
-  # This is because of TLS cipher suite mismatch
   value:
     name: "loggregator"
-    version: "105.5"
-    url: "https://bosh.io/d/github.com/cloudfoundry/loggregator-release?v=105.5"
-    sha1: "62b9210def7c5179f6088b9449f551f57f1a1ee8"
+    version: "105.6"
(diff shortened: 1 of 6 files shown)


tlwr:
I think this might have to be split into two releases, as it is now, but slightly differently

When dealing with cloud config we have to be quite careful, as cloud config is applied across the entire deployment, and it is applied outside the context of a `bosh deploy` step

In this case, if the bosh deploy fails then the deployment is going to be in a bad state.

I think this PR (1️⃣) needs to only the addition, and use, of the new target group, but without DNS

Then there should be a DNS pr/release (2️⃣), as terraform happens before the bosh deployment, this has to happen after doppler instances are in the target group (1️⃣)

Then there should be the PR you raised (3️⃣ #2050) to, as it already does, remove the CLB, but also do the target group removal from this PR

AP-Hunt:
> I think this might have to be split into two releases, as it is now, but slightly differently

Yeah that seems like a reasonable approach. I'll adjust these PRs once I've resolved the problem with the custom acceptance tests.

AP-Hunt:
Closing this PR in favour of 2 other PRs which can be merged in sequence to perform a safer rollout.

PR #2039, opened 2019-08-28
[167571906] Migrate cf-vars-store and prometheus-vars-store to Credhub

mogds:
What
----

- Migrates cf-vars-store and prometheus-vars-store to credhub.
- Updates the cf and prometheus deployments to no longer use the vars-store
- Updates task that copies secrets to the concourse namespace to copy from the bosh namespace instead of the vars-store
- Updates relevant jobs to use credhub vars and removes cf-vars-store inputs

How to review
-------------

- Review and deploy https://github.com/alphagov/paas-bootstrap/pull/300
- Code review
- Deploy to your dev env

Who can review
--------------

Not me

PR #2054, opened 2019-09-09
[167571906] Migrate cf-vars-store and prometheus-vars-store to Credhub

mogds:
What
----

Following on from the reverted https://github.com/alphagov/paas-cf/pull/2039. This adds an additional task to remove any deleted certs with blank values from cf-vars-store. 

The PR above failed during `migrate-secrets-to-credhub` with the following
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x7ab049]
```

How to review
-------------

- Code review
- Deploy to your dev env

Who can review
--------------

Not me

### DIFF
--- a/concourse/pipelines/autodelete-cloudfoundry.yml
+++ b/concourse/pipelines/autodelete-cloudfoundry.yml
@@ -21,14 +21,6 @@ resources:
       tag_filter: ((paas_cf_tag_filter))
       commit_verification_keys: ((gpg_public_keys))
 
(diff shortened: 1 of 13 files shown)


AP-Hunt:
Code looks good. Testing it down pipelines.

AP-Hunt:
All ran down my pipeline just fine. Approving.

PR #2059, opened 2019-09-11
Add VPC peer for dit-staging / datasci-staging

michalc:
Adding a VPC peer for dit-staging / datasci-staging.

### DIFF
--- a/terraform/prod-lon.vpc_peering.json
+++ b/terraform/prod-lon.vpc_peering.json
@@ -28,5 +28,11 @@
         "account_id": "165562107270",
         "vpc_id": "vpc-0911316657ad779ee",
         "subnet_cidr": "172.18.4.0/22"
+    },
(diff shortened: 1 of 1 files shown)


terraform/prod-lon.vpc_peering.json
```diff
@@ -23,6 +23,12 @@
         "vpc_id": "vpc-074a7e7c7b4eccc8c",
         "subnet_cidr": "172.18.0.0/22"
     },
+    {
+        "peer_name": "dit-staging_datasci-staging",
```
AP-Hunt: The subnet CIDR looks right to me, but could you swap it with the next definition?  It'd be good to keep the list ordered numerically by subnet CIDR.

michalc: Done

PR #2061, opened 2019-09-12
[#167971884] Switch paas-pass secrets from s3 to credhub

richardTowers:
~**Conflicts with #2057 - we either need to review / merge that first and update this or the other way around**~ *Done*.

What
----

Change the upload scripts to write the secrets from our passwordstore to
credhub instead of S3.

Update the concourse pipeline and bosh manifests to read secrets from
credhub instead of S3 (this mostly just involves using the `((secret))`
syntax).

Add a make target to remove the secrets from S3 once this has deployed
(NOTE: this script should only be run once the pipeline is reading from
credhub).

How to review
-------------

* Review alongside https://github.com/alphagov/paas-credentials/pull/160 and https://github.com/alphagov/paas-credentials/pull/159
* [x] Run the `upload-all-secrets` make target from this branch against your dev env
* [x] Run the `crosscheck-s3-credhub-secrets` make target from this branch against your dev env and check there are no significant changes
* [x] Run the `create-cloudfoundry` pipeline from this branch against your dev env
* [x] Run the `clean-up-s3-secrets` make target from this branch
* [x] Run the `create-cloudfoundry` pipeline again to confirm that no secrets are being read from S3

## How to release

* [ ] Run the `upload-all-secrets` make target from this branch against:
  * [ ] `stg-lon`;
  * [ ] `prod`;
  * [ ] `prod-lon`.
* [ ] Run the `crosscheck-s3-credhub-secrets` make target to check the secrets in S3 and CredHub are not significantly different in:
  * [ ] `stg-lon`;
  * [ ] `prod`;
  * [ ] `prod-lon`.
* [ ] Merge this PR and wait for the pipelines to finish;
* [ ] Run the `clean-up-s3-secrets` make target against:
  * [ ] `stg-lon`;
  * [ ] `prod`;
  * [ ] `prod-lon`.
* [ ] Once people have done this in their dev environment, review and merge #2068 to remove the temporary Make targets added by this PR.

Who can review
--------------

Not Richard or Tom

PR #2072, opened 2019-09-19
Remove copying logit secrets to credhub concourse ns

mogds:
What
----

We no longer need to copy logit secrets via the pipeline as it can be done via the Makefile.

How to review
-------------

Code review

Who can review
--------------

Not me


46bit:
After some more thought, existing dev environments will be fine because this code will have long-since have gotten Logit creds into CredHub. Then #2061 will fix things permanently. Woo.

PR #2083, opened 2019-09-26
[#168427527] - use credhub for cf-secrets

richardTowers:
What
----

Currently we store a bunch of database passwords in S3 in an object called cf-secrets.yml. This PR takes these secrets, writes them to credhub instead and updates the pipeline to read the secrets from credhub. If the secrets don't exist in the S3 object they'll be generated with `credhub generate --no-overwrite`, which supports the case where we're bringing up a new cloud foundry.

When I looked into the rotation process for these secrets I found that they're all skipped. This is because rotating database passwords in a zero-downtime way is tricky, so over they years we've added all of them to the list of creds that we don't rotate. Since this doesn't work anyway I've just removed the rotation code.

All the old secrets generation code is deleted because it's not used anymore 🎉 

How to review
-------------

* Code review
* Make sure you've run create-bosh-concourse in your dev env from (near) paas-bootstrap@master
* Run this branch down your dev env and observe that `credhub find -p /` now contains the cf-secrets (I've already done this in my dev env)
* Run the destroy-cloud-foundry pipeline in your dev env and then re-run the create-cloudfoundry pipeline (I've already done this in my dev env)
* Merge this PR and wait for it to deploy in staging / prod
* Review #2084 - the follow up PR to fully remove the S3 version of cf-secrets and fix the destroy-cloudfoundry pipeline

Who can review
--------------

Not @richardTowers


tlwr:
This ran down my pipeline okay, and everything worked, although I experienced a problem which I think is localised to my developer environment wherein I could not stage apps.

Recreating the BBS and the cells solved this problem. This was caught in multiple places by the pipeline. So I think we are cool to push this out because I don't think it was related.

In terms of code review, it was nice to be able to delete so much. I think when you reach for bash arrays it is an indicator of needing a proper programming language, but it is fine in this case IMO.

Once the other PRs are out, then I will deploy this

PR #2091, opened 2019-10-08
Fix VPC peering for traderemedies-services

tlwr:
What
----

There are two rules called dit-services_datasci, the later of which is
actually the VPC peering config for dit-services/datasci

The former rule is actually intended for traderemedies-services/datasci

See https://govuk.zendesk.com/agent/tickets/3823235 for context

For vpc peering configurations, the last entry with the correct name
wins

How to review
-------------

See this zendesk ticket: https://govuk.zendesk.com/agent/tickets/3823235

Who can review
--------------

Not @tlwr

### DIFF
--- a/terraform/prod-lon.vpc_peering.json
+++ b/terraform/prod-lon.vpc_peering.json
@@ -6,7 +6,7 @@
         "subnet_cidr": "172.16.0.0/22"
     },
     {
-        "peer_name": "dit-services_datasci",
+        "peer_name": "traderemedies-services_datasci",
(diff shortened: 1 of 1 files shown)


tlwr:
For a paas-reviewer:

- check that `terraform/scripts/generate_vpc_peering_tfvars.rb` which generates peer_name is not actually used by terraform.

- check that `terraform/scripts/generate_vpc_peering_opsfile.rb` generates CAPI config which sets up security groups

PR #2092, opened 2019-10-08
[Firebreak] [#169004501] Concourse deploy markers in Prometheus

46bit:
What
----

The changes should be fairly self-explanatory when looking at the commits, but the nuts and bolts are:

* The tasks are created in meta and are referred to as anchors throughout (because of reuse);
* The annotations are created in grafana's internal data store - this removes the need for it to be 'pulled' by prometheus, meaning it should all be pretty real-time;
* Annotations are created as a 'point', with `*add-grafana-job-annotation` and `*add-grafana-overview-annotation`, with the start and end time being the same. Once the respective `end-*` tasks are run, the end time for the annotation is updated to the current time, and the annotation changes to a range. This happens in real-time, so you can watch the deploy on grafana;
* Job annotations are for each individual build job;
* Overview annotations show the start and end of the whole pipeline.

How to review
-------------

We seem to think this is good to go. Maybe deploy to a dev env if you really want to.

Who can review
--------------

Not @tnwhitwell.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -378,6 +378,17 @@ resource_types:
     tag: f2a80c95481056aa57d837e3b14f0012b542fdb3
 
 resources:
+  - name: metadata
+    type: metadata
+
+  - name: overview-marker-id
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: overview_marker_id
+      initial_version: "-"
+
   - name: pipeline-trigger
     type: semver-iam
     source:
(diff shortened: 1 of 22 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -44,9 +44,236 @@ meta:
     curl-ssl: &curl-ssl-image-resource
       type: docker-image
       source:
-        repository: governmentpaas/curl-ssl
-        tag: 54cc3c9f49f9032e46dd4538c626f2533bf90a94
+        repository: tnwhitwell/curl-ssl
+        tag: 36db896bf1e6dd8f4034b539f814e1622a583966
```
46bit: I think this was to pull in `jq` and we should be able to use our own image updated with https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/173.

tlwr:
`jq` is doing the right thing WRT `echo`

I'm going to push up some changes that I hope will be non-controversial that I made during review

PR #2097, opened 2019-10-14
[#168322288] Postgres 11 & MySQL 8 for tenants

46bit:
What
----
Two things.

**Thing the first**
We've refactored the config billing file generation, because with the addition of the new Postgres and MySQL plans, it was set to get out of hand. It's now been split out in to its components, and the generation script recombines it.

**Thing the second**
We've added new Postgres 11 and MySQL 8.0 plans, and all of the associated tests.

How to review
-------------
1. Code review
2. Run it down a pipeline and do some manual testing until you're happy that creating, (un)binding, and deleting the new services work, and that they operate as expected when bound.

Who can review
--------------
Anyone


AP-Hunt:
Rebased on `master` and regenerated billing in a new commit

config/billing/generate_config.rb
```diff
@@ -18,14 +19,39 @@ def price(key)
...
-  File.open("#{__dir__}/output/#{region}.json", "w") do |f|
-    f.write g.generate(template)
+  begin
+    g = Generator.new(region, pricing_data)
+    json_content = g.generate(template)
+
+    JSON.parse json_content
+
+    File.open("#{__dir__}/output/#{region}.json", "w") do |f|
+      f.write json_content
```
richardTowers: You could get a properly indented output (and a way nicer git diff in the output) if you did this:

```suggestion
      f.write JSON.pretty_generate(json, indent: "\t")
```

richardTowers: Sorry - I missed saying that you'd also have to assign `json = JSON.parse json_content` on line 39.

46bit: Done

46bit:
I've addressed all of @richardTowers feedback. Ready for re-review.

richardTowers:
I've seen this working in Miki's dev env now - I can see that they've been created and deleted. Let's ship this. Still to do - documentation (including how to upgrade from 9.5.x and 10.x plans).

PR #2098, opened 2019-10-15
Add a dashboard showing billing vs. spending

richardTowers:
What
----

For apps and services.

A couple of things to note:

* These graphs just compare how much we're billing with how much we're
  spending. They should be a sanity check to protect us against any
  regressions in paas-billing that accidentally change how much we're
  charging.
* The `paas_aws_cost_explorer_by_service_pounds` metric got renamed to
  `paas_aws_cost_explorer_by_service_dollars`, so we need to have old
  and new graphs to compensate for the change

We can probably update some of our alerting use the calculations in this
dashboard, as they're a little more smooth than the ones we're using at
the moment.

How to review
-------------

* Have a quick look at the JSON
* Check out the dashboard in my dev environment [Grafana](https://grafana-1.towers.dev.cloudpipeline.digital/d/billing-compared-to-spending/billing-compared-to-spending) (which was
  deployed by the pipeline) (note: not up yet)

Who can review
--------------

Anyone except @richardtowers

### DIFF
--- a/manifests/prometheus/dashboards.d/billing.json
+++ b/manifests/prometheus/dashboards.d/billing.json
@@ -1,0 +1,726 @@
+{
+  "dashboard": {
+    "annotations": {
+      "list": [
+        {
+          "builtIn": 1,
+          "datasource": "-- Grafana --",
+          "enable": true,
+          "hide": true,
+          "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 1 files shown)


richardTowers:
I'm going to do a dashboard that actually shows the SLIs instead / as well.

PR #2109, opened 2019-10-18
Update create-org.sh to annotate orgs with their owner

richardTowers:
What
----

So when we create new organisations they're tagged with the right owner.

Also adds a script to get the current list of orgs and their owners.

How to review
-------------

* Code review
* Run `scripts/list-org-owners.sh` against the prod environments
* Run `scripts/create-org.sh` against your dev environment
* Check that the org created has the right annotation

Who can review
--------------

Not @richardtowers

### DIFF
--- a/scripts/create-org.sh
+++ b/scripts/create-org.sh
@@ -13,7 +13,7 @@ usage() {
   cat <<EOF
 Usage:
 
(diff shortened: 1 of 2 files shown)


tlwr:
[x] Code review
[x] Manual testing

PR #2111, opened 2019-10-18
Actually generate the new exchange rate

tlwr:
What
----

The commit "ac1f5401 Update exchange rate to prove we can" updated the
exchange rate, but did not actually generate the required output files

How to review
-------------

Code review

Who can review
--------------

@richardTowers (who generated the new exchange rate), or anyone else really


tlwr:
Good points, though I haven't thought of a neat solution for testing, or even looked into if they cannot be generated at use 🙈

46bit:
It's OK, I was just trying to nerdsnipe you 🙉

PR #2116, opened 2019-10-23
[#168704085] Upgrade CF to v12.1.0

mogds:
What
----

Upgrade cf-deployment to v12.1.0

How to review
-------------

* [x] Deploy to your dev env
* [x] Review https://github.com/alphagov/paas-uaa/pull/11 and https://github.com/alphagov/paas-uaa-release/pull/13
* [x] Update last commit with the final build

Who can review
--------------

Not me

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -501,7 +501,7 @@ resources:
     type: git
     source:
       uri: https://github.com/cloudfoundry/cf-acceptance-tests
-      branch: cf11.2
+      branch: cf12.1
 
   - name: cf-smoke-tests-release
     type: git
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -507,7 +507,7 @@ resources:
     type: git
     source:
       uri: https://github.com/cloudfoundry/cf-smoke-tests-release
-      tag_filter: "40.0.116"
+      tag_filter: "40.0.119"
       submodules:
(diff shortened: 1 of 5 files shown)


richardTowers:
Force pushed to rebase on master and update some version pinning to make the tests happy.

Code review looks totally fine. Changes went down my development environment fine, but there was some app and API downtime reported:

* [API](https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/api-availability-tests/builds/6) - 99.6% reliability reported, target 99.95%. Looks like 25 seconds of unavailable stats - probably not something to worry about.
* [App](https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/app-availability-tests/builds/6) - 99.77% reliability reported, target 100%. Looks like there was a 10 second period where all 100 of the healthcheck requests failed with 502 or 404, otherwise it was fine. Looks like it maybe correlates with the routers being updated. A 10 second outage would reduce our daily uptime slightly below 99.99% ([source](https://uptime.is/99.99)), but this might be fine as a one off. It's also possible that the outage would be shorter in production, as there are 6 routers instead of 2 for a dev env. Staging has 3 routers.

richardTowers:
I re-ran the upgrade with my development environment set to 6 routers, instead of the default 2. This resulted in 99.994% application uptime during the actual deployment, but I did see a longer blip of downtime when the healthcheck redeployed in post-deploy. I'm pretty sure that's a red herring, so I think it should be fine for us to push this out.

API reliability was 99.8% this time - I think good enough.

PR #2117, opened 2019-10-23
[SUPPORT] Update CFSyslogDrains alert to track adapter numbers

AP-Hunt:
What
----

The alert was hardcoded to go off when any of the schedulers reached more than
250 drains. However, the documentation that we link to in the alert indicates
that the number of adapters should be:

    (cf-syslog-drain.scheduler.drains / 250) + 2

Previously, we were running only 2 instances across the 2 availability zones,
and it was running fine, so experience suggests that the below is probably fine
as a gauge of the number adapters needed:

    (sum(drains) / 250)

In this commit, I've included the
`manifests/cf-manifest/env-specific/${ENV_SPECIFIC_BOSH_VARS_FILE}` in the vars
used to generate the Prometheus manifest, so that the alerts have access to
useful variables about the scale of different components. I've then used the
`adapter_instances` variable in the alert ops file so that in future the alert
threshold will automatically scale with the number of instances of the adapter.

How to review
-------------

1. Code review
2. This ought to be run down a dev env, just in case the variables from the newly included vars file override something in an unintended way.

Who can review
--------------
Anyone

### DIFF
--- a/manifests/prometheus/alerts.d/cf-syslog-drains.yml
+++ b/manifests/prometheus/alerts.d/cf-syslog-drains.yml
@@ -6,11 +6,15 @@
   value:
     name: CFSyslogDrains
     rules:
+    - record: "firehose_value_metric_cf_syslog_drain_scheduler_drains:required_adapters"
+      expr: ceil(sum(firehose_value_metric_cf_syslog_drain_scheduler_drains) / 250)
+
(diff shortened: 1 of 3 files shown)


AP-Hunt:
Rebased this on `master` because my local `master` was behind  when I branched off.

PR #2120, opened 2019-10-24
Ship audit to cyber

tlwr:
🐝 WIP

What
----

How to review
-------------

Who can review
--------------

Paired with @olliejc


OllieJC:
We have now created a destination in `eu-west-1`.

Could you do the following?
Change the secret to `arn:aws:logs:REGION:xxx:destination:xxx` and use a replace function here:
https://github.com/alphagov/paas-cf/pull/2120/files#diff-6eca3adac458bc6a5e23d3a72ea8d5ecR27
`destination_arn = "${replace(var.csls_kinesis_destination_arn, "REGION", var.region)}"`

PR #2122, opened 2019-10-24
Move the DopplerDroppedEnvelopes alert to "Other"

richardTowers:
What
----

It's constantly firing which is causing alert blindness. Ideally we'd
fix it, rather than ignoring it. We are going to raise a story to do
that. In the meantime though it's more important that we notice the other
alerts when they fire.

Also makes the "Other" alert have a dark grey background instead of the
almost-red orange we were using before.

How to review
-------------

* Code review is probs enough

Who can review
--------------

Not @richardtowers


tlwr:
shiperino

PR #2128, opened 2019-10-28
Update S3 Bucket limit

mogds:
What
----

The s3 bucket limit in the prod account has been increased to 250. This updates
the alert to reflect the new value.

How to review
-------------

Code review

Who can review
--------------

Not me

### DIFF
--- a/manifests/prometheus/alerts.d/s3_bucket_count.yml
+++ b/manifests/prometheus/alerts.d/s3_bucket_count.yml
@@ -7,7 +7,7 @@
     name: S3BucketCountCloseToLimit
     rules:
       - alert: S3BucketCountCloseToLimit
-        expr: paas_aws_s3_buckets_count > ((aws_limits_s3_buckets)) / 100 * 80
+        expr: paas_aws_s3_buckets_count > ((aws_limits_s3_buckets)) * .8
         for: 1h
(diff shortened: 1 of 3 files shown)


manifests/prometheus/alerts.d/s3_bucket_count.yml
```diff
@@ -7,7 +7,7 @@
     name: S3BucketCountCloseToLimit
     rules:
       - alert: S3BucketCountCloseToLimit
-        expr: paas_aws_s3_buckets_count > ((aws_limits_s3_buckets)) / 100 * 80
```
richardTowers: I think the intent was for this line not to change - `80 / 100 == 200 / 250`, so they're equivalent. It's just supposed to say `paas_aws_s3_buckets_count` is more than 80% of the configured limit.

richardTowers: Probably `* 0.8` or `* 80/100` would be a more elegant way of expressing that mind you.

mogds: Agreed, I did think it was an unnecessary way of getting to > 80% but went with the existing format. I'll switch it to  `* 0.8`

PR #2129, opened 2019-10-29
Explicitly disable keep alives with warning

tlwr:
What
----

In future versions of cf-deployment, keep alives are enabled by default

We cannot enable keep alives until routing-release 0.194 which contains changes to keep alive health checks.

How to review
-------------

Code review

Run it down your pipeline and check gorouter's configuration on a VM, which should have `disable_keep_alives: true`

Who can review
--------------

Not @tlwr


tlwr:
Superseded by #2162

PR #2137, opened 2019-10-30
Set LB deregistration timeout

tlwr:
What
----

We should sync the routing table more often to avoid the situation where gorouter misses a message from NATS about a route deregistration (during instance evacuation)

We should specify a value for `deregistration_delay` for our `aws_lb`s which is equivalent for the `aws_elb`s that they replace (I added one for loggregator as well, which is less crucial) (read the commit for this one).

How to review
-------------

Code review

Run down your pipeline

Who can review
--------------

Not @tlwr


tlwr:
Force pushed some tests and a missing dereg for doppler tg

PR #2151, opened 2019-11-06
[#167590572] Reenable cf exporter 

AP-Hunt:
What
----

Re-enabled `cf_exporter` in our Prometheus deployment in order give us better metrics for tenant apps. Also slims down the number of collectors used.

How to review
-------------
1. Code review
2. Run it down your pipeline and check that `cf_*` metrics are being published, and are available in Prometheus

Who can review
--------------

Anyone but me


tlwr:
Running this down my pipeline

Nice commit history, and messages, although I have question about: https://github.com/alphagov/paas-cf/pull/2151/commits/30597b046098775ebde1e396b460b0d592fd7194 is it intentional?

AP-Hunt:
> Nice commit history, and messages, although I have question about: [30597b0](https://github.com/alphagov/paas-cf/commit/30597b046098775ebde1e396b460b0d592fd7194) is it intentional?

Yep, this was intentional. I reverted the content of the PR which disabled `cf_exporter`, figuring if the tests were worthwhile then, they're probably worthwhile now

tlwr:
@AP-Hunt when I read the diff, we are deleting the entirety of `manifests/prometheus/spec/operations/monitor_cf_spec.rb`

What I think we want instead is this:

```
RSpec.describe "Monitor CF" do	
  it "adds the firehose instance_group" do	
    firehose = manifest_with_defaults.get("instance_groups.prometheus2")	
    expect(firehose).not_to be_nil	
    expect(firehose["vm_type"]).to eq("medium")	
    expect(firehose["networks"]).to eq([{ "name" => "cf" }])	
  end	

  it "adds the cf_exporter job" do	
    expect(manifest_with_defaults.get("instance_groups.prometheus2.jobs.cf_exporter")).to_not be_nil	
  end	

  it "adds the cloudfoundry_dashboards job" do	
    expect(manifest_with_defaults.get("instance_groups.grafana.jobs.cloudfoundry_dashboards")).to_not be_nil	
  end	
end
```

i.e. `it adds the cf_exporter job` and all the other tests

```
  it "adds the cf_exporter job" do	
    expect(manifest_with_defaults.get("instance_groups.prometheus2.jobs.cf_exporter")).to_not be_nil	
  end
```

AP-Hunt:
Yep, agree with you on the spec @tlwr. 

However, I worry about needing to keep a 4m scrape interval. We originally disabled `cf_exporter` because it was effectively DoSing CAPI with its 4m scrape interval, and afaik we didn't get enough value out of it at the time to justify keeping it turned on. Is it not possible to adjust what Prometheus thinks is stale per scrape config?

tlwr:
I'll reinstate the spec.

@AP-Hunt if we're worried about rate limits in production, which we may or may not mitigated using v3, how do you think about the following:

- Proceed with 10m scrape interval
- Observe the usage and then calculate what it would look like ~2.5x higher

The only configuration parameter relating to staleness seems to be Consul specific.

AP-Hunt:
> how do you think about the following:
> * Proceed with 10m scrape interval
> * Observe the usage and then calculate what it would look like ~2.5x higher

Sounds like a plan

tlwr:
Alright I've added the spec, lets ship it. I think if we can't make the series continuous then we should either submit a PR to the `cf_exporter` to make it async, or not use it at all.

Once I'm done massaging pipelines then I'll merge this

PR #2155, opened 2019-11-12
[#167507410] Enable AWS Shield

AP-Hunt:
🚨 Do not merge this PR until PR 193 in the account wide terraform has been merged and applied in all environments 🚨 

What
----
We've subscribed to AWS Shield Advanced in our AWS parent account, and that gives our other accounts access. This applies DDoS protection to the application and system load balancers using an AWS WAF rate based rule.

How to review
-------------

1. Code review
2. Run it down your dev env, then check in the AWS console that your app and system load balancers are protected by the WAF (see the integrated services tab)

Who can review
--------------
Anyone

### DIFF
--- a/terraform/cloudfoundry/shield_and_waf.tf
+++ b/terraform/cloudfoundry/shield_and_waf.tf
@@ -1,9 +1,46 @@
 resource "aws_shield_protection" "shield_for_app_gorouter_alb" {
-  name = "${var.env}-app-gorouter-shield"
+  name         = "${var.env}-app-gorouter-shield"
   resource_arn = "${aws_lb.cf_router_app_domain.arn}"
 }
 
 resource "aws_shield_protection" "shield_for_system_gorouter_alb" {
-  name = "${var.env}-system-gorouter-shield"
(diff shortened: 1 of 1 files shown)


tlwr:
Would it be possible to document reproduce how this 600k number can be calculated by a reviewer?

AP-Hunt:
> Would it be possible to document reproduce how this 600k number can be calculated by a reviewer?

Yep. In a Kibana data table visualisation, we searched for gorouter log lines with `@source.component:gorouter`, then we ...
1. split the table using the terms in the `@source.deployment` field to get a table of London and Ireland
2. bucketed lines using a date histogram with an interval of 5m
3. split columns using the `gorouter.header.x_forwarded_for.keyword` field (intended to give us an originating IP, since we don't have a field on the gorouter logs which is solely that)
4. Set the time range to the last 14 days

Finally, we downloaded the data as CSV, ordered it by request count descending, and looked for the biggest count in any 5 minute period.

Looking back, I'm not entirely happy with that method, so I'll redo it this morning and update as necessary.

PR #2156, opened 2019-11-12
[#168215129] Declarative platform user management with support for Global Auditors

paroxp:
What
----

This replaces our existing CF admin user management with a new tool which does a bit more and requires people to sign in with Google. For more details see and review the `README.md` file included 🙂 

How to review
-------------

* Code review;
* I've manually run this successfully against staging 🎉;
* I've run this through the pipelines against my `miki` dev environment;
* I've run a modified read-only version of this against both prods, and the output is all happy. There's some Global Auditors being removed or replaced but they're OK (see the Pivotal Tracker story for details on that.)

## How to release

Since this performs admin user management I suggest we're a little cautious. Pause the prod pipelines and roll this out to staging. Check the output in Concourse and that you still have CF admin permissions. Then unpause the prods one at a time. This way if anything does surprise us (which I doubt) we'll have fewer things to fix.

## Who can review

Definitely not @46bit. @tlwr and @paroxp both have context and have worked on this, but that might be a good reason for them to review.


tools/user_management/main.rb
```diff
@@ -0,0 +1,36 @@
...
+USERS_CONFIG_PATH = ENV.fetch('USERS_CONFIG_PATH')
+
+uaa_client = RestClient::Resource.new(
+  UAA_API_URL,
+  headers: {
+    'Authorization' => CF_TOKEN,
+    'Content-Type' => 'application/json'
+  }
+)
+users_config = YAML.safe_load(File.read(USERS_CONFIG_PATH)).map { |uo| User.new(uo) }
```
46bit: I've discovered a bug. 😮 We pass `deploy_envs` from the YAML file onto the `User` class… but we never use it. We need to filter users down to those present in the current `DEPLOY_ENV`. 🤦‍♀

46bit:
I think we might be better off removing the `deploy_envs` feature than implementing it. Before this work we needed it to restrict the access people had, but now that we can add people as Global Auditors it seems much less risky to give everyone access to all envs.

tlwr:
Miki raises a good point WRT `deploy_envs`

That feature was implemented because we used to email people when we did user account creation: `deploy_envs` was added so that people didn't get spammed when development environments were created.

Now we aren't emailing people I'm cool with removing the `deploy_envs` feature

The only case in which it would be useful is when we have a developer who is a `cf_admin` in dev but not staging/prod (e.g. Seb presently)...

paroxp:
Not a bug, it's a feature. New commit removing the parameter.

46bit:
> The only case in which it would be useful is when we have a developer who is a `cf_admin` in dev but not staging/prod (e.g. Seb presently)...

Argh, that's a good point. Looks like we'll cross that bridge when we come to it. I think a `prod_access` boolean might be nicer nowadays than a list of deploy_envs.

tools/user_management/lib/group.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def initialize(name, users)
+    @name = name
+    @users = users
+  end
+
+  def remove_unexpected_members(uaa_client)
+    get_members(uaa_client).each do |member|
+      member_guid = member['value']
+      unless @users.any? { |user| user.guid == member_guid || member['origin'] == 'uaa' }
```
46bit: What is this line of code trying to do? Preserve the `admin` user?

paroxp: Yeah... The member object only contains three fields...

```json
{
  "value": "00000000-0000-0000-0000-000000000000",
  "type": "USER",
  "origin": "uaa" // or `google`
}
```

I thought that If we want to use google as our source of truth for that, we can essentially ignore users with origin of `uaa`, which coincidentally include the `admin` user. And we don't want to loose that...

46bit: I ended up rewriting things to use the list members endpoint which does include username. Deploying to `miki` this morning and hopefully ship it after :)

46bit:
ℹ️ In my dev env, `51358f4` did what it was meant to but with one very important quirk. Smoke test and billing smoke test users were de-admined while those tests were running. 😱

My idea for working around this is to only remove users from groups where the user was created >1 hour ago. 🤔

46bit:
I tried running this in a read-only way against `stg-lon`, and hit an annoying problem. UAA's `/Groups/$GUID/members?includeEntities=true` endpoint is returning a 404 because some of the group members no longer exist as users. Which is awkward. There's something like 9 of them per group. 😱

There's a bunch of ways out of this mess. I'll check if production is the same tomorrow. If it is I'll have to adapt this PR to handle the problem.

46bit:
Yesterday afternoon I successfully applied this to `stg-lon`. 🎉 No unexpected users with global permissions, only the smoke tests. It added some global auditors as expected.

I tried a read-only version of this against prods, and those have some global auditors not configured in our `config/users.yml`. I'll investigate if they still need access and once that's done we can ship this. 😄

46bit:
**This is ready for review.** 🚀

PR #2164, opened 2019-11-19
Deploy app shipping audit events to Splunk

paroxp:
What
----

We'd like to be constantly pushing this to our environment allowing
splunk to receive the audit events from CF API.

We're adding set of credentials into the system, that will be then used.

Splunk URL is not really a secret, therefore doesn't deserve a place in
our secret store.

How to review
-------------

- Sanity check
- See `rafalp` environment

**Note:** I have uploaded the certs to all envs, but doesn't hurt to do again.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -292,6 +292,7 @@ groups:
       - post-deploy
       - deploy-paas-accounts
       - deploy-paas-admin
+      - deploy-paas-audit-to-splunk
       - deploy-paas-billing
       - deploy-paas-metrics
       - smoke-tests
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3106,6 +3117,71 @@ jobs:
...
                 fi
 
+  - name: deploy-paas-audit-to-splunk
+    serial: true
+    plan:
+      - *add-grafana-job-annotation
+      - get: paas-audit-to-splunk
+        trigger: true
+
+      - in_parallel:
```
46bit: Is there a reason that this `get: paas-audit-to-splunk` task isn't in the `in_parallel:`?

paroxp: Oversight :D Probably doesn't matter. But moved in for making it consistent.

PR #2169, opened 2019-11-21
Remove Tom Whitwell from admin users

richardTowers:
What
----

He shouldn't need this access anymore. We should check that the
pipeline's removed him properly once this is released.

This may conflict slightly with #2156, but it shouldn't be difficult to
fix so I think they can go out in either order.

How to review
-------------

* Check 'em

Who can review
--------------

Not @richardtowers


richardTowers:
Hold off on merging this - I just found some more things to remove.

PR #2173, opened 2019-11-27
get chris a working account

paroxp:
What
----

Farms is currently locked out of his account, due to the management of admin vs tenants accounts.

I'm taking the approach of 🙈 and granting farms the powers of an admin back. One team one dream.

How to review
-------------

- Sanity check


richardTowers:
As discussed on slack, we don't trust @chrisfarms 🙅

PR #2182, opened 2019-12-03
[#170056541] Add instructions to the platform users config

46bit:
What
----

This commit adds information to `config/users.yml` explaining what it
is and what permissions are granted. This commit then gives
instructions on how to add brand new users, including how to avoid a
frustrating edge-case that we investigated in [Pivotal #170056541](https://www.pivotaltracker.com/n/projects/1275640/stories/170056541).

How to review
-------------

* Read the instructions and see if they make sense to you;
* I checked that this comment didn't cause problems for the things using this file. My `miki` dev environment deployed fine with this comment.

Who can review
--------------

Not @46bit.

### DIFF
--- a/config/users.yml
+++ b/config/users.yml
@@ -1,4 +1,24 @@
 ---
+
+# This file configures platform-wide users.
+# - All users in this file get access to Grafana.
+# - Users with `cf_admin: true` get full admin privileges in Cloud Foundry.
+# - Users without the `cf_admin` option or where it is false get
+#   Global Auditor, which allows them to see everything except secrets in
+#   app environment variables.
+#
+# HOW TO ADD NEW USERS
+# - This file requires users to sign in using Google SSO. To do that we need
+#   their Google ID, which is a large number.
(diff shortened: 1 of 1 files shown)


config/users.yml
```diff
@@ -1,4 +1,24 @@
...
+# - Users without the `cf_admin` option or where it is false get
+#   Global Auditor, which allows them to see everything except secrets in
+#   app environment variables.
+#
+# HOW TO ADD NEW USERS
+# - This file requires users to sign in using Google SSO. To do that we need
+#   their Google ID, which is a large number.
+# - If a new user is configured in this file who didn't already exist, they
+#   lose Global Auditor when they first log in. Subsequent pipeline runs will
+#   permanently re-add it. This seems to be UAA behaviour.
```
46bit: We discovered the source of this weirdness and fixed it in #2185. These instructions are still good as they tell us the Google ID, but we can remove this warning. I've set a reminder to edit them later :)

PR #2185, opened 2019-12-04
[#170056541] UAA group member's origin should equal "uaa"

46bit:
What
----

The `origin` field of a group member is not the same as the `origin` field of a user. From https://docs.cloudfoundry.org/api/uaa/version/74.4.0/index.html#add-member:

> `origin`: The originating IDP of the entity, or "uaa" for groups and internal users

So users signing in with Google should have an origin of `google`, but their group memberships should have an origin of `uaa`.

How to review
-------------

* See that the tests go green;
* Code review.

I've run this manually against staging and it worked as expected 🎉:
* The existing group memberships with `origin=google` were all removed.
* New group memberships with `origin=uaa` were created.
* Subsequent runs don't make any unexpected changes.

Who can review
--------------

Not @46bit.

### DIFF
--- a/tools/user_management/lib/group.rb
+++ b/tools/user_management/lib/group.rb
@@ -14,31 +14,31 @@ class Group < UAAResource
 
   def remove_unexpected_members(uaa_client)
     desired_user_guids = @users.map(&:guid)
-    unexpected_members = get_members(uaa_client).select do |member|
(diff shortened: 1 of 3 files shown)


46bit:
Confession: I think this could do with a refactoring to better encapsulate lots of things. But I don't think that should be a priority–the tests and README hopefully document things well.

PR #2188, opened 2019-12-05
[#170129714] Specify broker when enabling Elasticsearch service

mogds:
[Story](https://www.pivotaltracker.com/story/show/170129714)

What
----

The deploy-aiven-service-broker task currently fails if another broker is added
with an elasticsearch service. This updates the command to specify the aiven
broker.

How to review
-------------

Code review

Who can review
--------------

Not me


tlwr:
Added the story that Venus raised.

LGTM, but while we are here, would it be possible to do this for all instances of `enable-service-access` so this doesn't happen in future?

PR #2195, opened 2019-12-09
Remove docker from credhub_shell and simplify

richardTowers:
What
----

... and add ASCII art, of course.

People on Linux machines can't always use docker and the VPN at the same
time, because anyconnect insists on owning the entire routing table.

There's really no need for us to be using docker in this script though -
everyone's going to have SSH already installed locally, and it's not
difficult to `brew install credhub-cli`.

How to review
-------------

* Code review
* Check you can still access credhub with this version of the script
* Confirm that the shellcheck changes are safe (either by running it down a dev pipeline, or by sanity checking by hand)

Who can review
--------------

Not @richardtowers


scripts/credhub_shell.sh
```diff
@@ -2,26 +2,32 @@
 
 set -eu
 
-SSH_PATH=${SSH_PATH:-"/Users/${USER}/.ssh/id_rsa"}
+SSH_PATH=${SSH_PATH:-"${HOME}/.ssh/id_rsa"}
```
tlwr:
```suggestion
SSH_PATH=${SSH_PATH:-"${HOME}/.ssh/id_rsa"}
```

We can remove this and use the key in the SSH agent, this will allow Yubikeys

PR #2198, opened 2019-12-11
Revert the change to RDS database kick off in 0242505330c7

AP-Hunt:
What
----
In commit 0242505330c72550b85d817c81c5328b5ee66bf2, the shebang for
`concourse/scripts/rds_instances.sh` was changed to `/bin/bash`. However, the
container iin which this script runs does not contain `/bin/bash`, only
`/bin/sh`. This resulted in environment kickoffs being broken.

How to review
-------------

1. Code review

Who can review
--------------
Anyone but me

### DIFF
--- a/concourse/scripts/rds_instances.sh
+++ b/concourse/scripts/rds_instances.sh
@@ -1,6 +1,5 @@
-#!/bin/bash
-
-set -euo pipefail
+#!/bin/sh
+set -euo
 
 ACTION=$1
 CMD=""


richardTowers:
The reason for doing it in 0242505 was that `set -o pipefail` doesn't work properly in non-bash shells.

Is `/bin/sh` set to `dash` in this container, or is it actually `bash` anyway? Maybe a better shebang would be `/usr/bin/env bash`

AP-Hunt:
> The reason for doing it in [0242505](https://github.com/alphagov/paas-cf/commit/0242505330c72550b85d817c81c5328b5ee66bf2) was that `set -o pipefail` doesn't work properly in non-bash shells.
> 
> Is `/bin/sh` set to `dash` in this container, or is it actually `bash` anyway? Maybe a better shebang would be `/usr/bin/env bash`

The [container in which this runs](https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/master/awscli/Dockerfile) is ultimately [based on Alpine 3.9](https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/master/curl-ssl/Dockerfile#L1), so `bash` isn't available at all.

```
$ docker run -it governmentpaas/awscli:91fe1e826f39798986d95a02fb1ccab6f0e7c746 /usr/bin/env bash
env: can't execute 'bash': No such file or directory

$ docker run -it governmentpaas/awscli:91fe1e826f39798986d95a02fb1ccab6f0e7c746 /bin/bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown.
```

In this case, I guess we should just drop pipefail?

richardTowers:
Yeah, agree. Drop pipefail, it doesn't work in dash anyway. Thanks shellcheck!

AP-Hunt:
> Yeah, agree. Drop pipefail, it doesn't work in dash anyway. Thanks shellcheck!

Done :+1:

PR #2201, opened 2019-12-13
Use paas-trusted-people user configuration file

richardTowers:
What
----

We had to handle a slightly different configuration file format, so:

```
- email: ''
  roles: []
```

instead of:

```
- email: ''
  cf_admin: true
```

so we're fixing the tools/user_management scripts to work with the new
format.

paas-trusted-people is a private repo, so we have to upload it to S3 and
then download it in the pipeline (because the pipeline can't clone
private repos).

How to review
-------------

* Code review
* Upload paas-trusted-people to S3 in your dev env by running `make upload-paas-trusted-people DEPLOY_ENV=blah`
* Run the create-cloudfoundry pipeline and check that the `post-deploy/sync-admin-users` task does something sensible

Who can review
--------------

Not @richardtowers or @schmie

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ showenv: ## Display environment information
 	@scripts/showenv.sh
 
 .PHONY: upload-all-secrets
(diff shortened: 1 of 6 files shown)


tlwr:
Works in my dev env

PR #2203, opened 2019-12-13
[WIP] Move PaaS environment config out of the Makefile

46bit:
What
----

This PR moves our `make dev`/`make stg-lon`/etc environment variables from the Makefile into a separate YAML file. It isn't a long-term solution, but it makes it easy to go on and experiment with not using a Makefile at all (or at least revising how we use it.)

The motivation for doing this now is that we plan to merge paas-cf, paas-bootstrap and paas-release-ci. They have Make targets with the same name and some rethinking is probably needed. Moving our config out of the Makefiles allows us to try and find a neater solution.

More info is in the commit message.

How to review
-------------

* Code review;
* See if the `make dev`/etc Make targets still work fine for you;
* Check that all the env variables removed from the Makefile are in the YAML.

Who can review
--------------

Not @46bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,87 +138,22 @@ update_merge_keys:
 
 .PHONY: dev
 dev: ## Set Environment to DEV
-	$(eval export AWS_DEFAULT_REGION ?= eu-west-1)
-	$(eval export AWS_ACCOUNT=dev)
-	$(eval export MAKEFILE_ENV_TARGET=dev)
-	$(eval export PERSISTENT_ENVIRONMENT=false)
-	$(eval export ENABLE_DESTROY=true)
-	$(eval export ENABLE_AUTODELETE=true)
-	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
-	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
-	$(eval export ALERT_EMAIL_ADDRESS?=govpaas-alerting-dev@digital.cabinet-office.gov.uk)
-	$(eval export NEW_ACCOUNT_EMAIL_ADDRESS?=the-multi-cloud-paas-team+dev@digital.cabinet-office.gov.uk)
-	$(eval export ENABLE_ALERT_NOTIFICATIONS ?= false)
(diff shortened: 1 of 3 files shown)


richardTowers:
I really like this idea, and it's cool that we can make this more maintainable without changing the user experience we're all used to (e.g. things like the insane `make dev deployer-concourse bootstrap` command).

I'd be interested in discussing where we think the end game for this is though. I think there's probably two parts to it:

* how do the environments differ from each other? (e.g. prod shouldn't run acceptance tests, staging should run load tests)
* how do we tell our scripts what their targets should be interactively?

For the first one I agree we should definitely have a configuration file that's much more readable than the mess of environment variables set in make targets that we've got now. 

For interactive use I'd like to get away from using environment variables as the primary way of modifying settings, preferring flags. So something like:

```
gds paas concourse set-pipelines --aws-account dev --deploy-env towers
gds paas credhub --aws-account prod --region london
```

perhaps these could fall back to env vars.

👍  from me on this though - a good step in the right direction.

46bit:
@richardTowers I’m thrilled you like this step. Given free reign for long enough I’d do away with the Makefile, remove lots of the environment variables, and write a small CLI in an expressive language (Ruby.) 

I think using environment variables is the right thing so long as we’re triggering Shell scripts run by Concourse (very happy to be proven wrong!) But we don’t need as many degrees of freedom as are in the YAML file right now.

On Pay I really liked having environment names which described the AWS account *and* the particular deployment. Right now we have `MAKEFILE_ENV_TARGET` doing some of the disambiguation that env names such as `dev-miki` would.

My dream is one YAML per environment configuring like this YAML does, rather than passing lots of flags by hand. But I’m only owning getting the core PaaS repos merged—where we go from there is up to all of us :)

46bit:
I've marked this as Work In Progress again. Merging `paas-cf`, `paas-release-ci` and `paas-bootstrap` meant so much examining and revising the Makefiles that I've given up and am writing a PaaS CLI within https://github.com/alphagov/gds-cli.

The YAML file in this PR is a simpler form of what will power the CLI. I'm massively happy to gotten the YAML file compatible with the existing `paas-cf` `Makefile`, but there's not much point shipping this. We've postponed merging the repos until the start of January, by which time the CLI will be ready to replace the `Makefile`s.

PR #2197, opened 2019-12-10
[#169775886] Upgrade to cf v12.15.0

AP-Hunt:
What
----
v12.18.0 was released 7 days ago (2019-12-04) relative to now (2019-12-10). There are more recent releases 2 days ago, 18 hours ago and 16 hours ago, but I’m choosing v12.15.0 for the sake of not releasing something which other people haven’t had chance to test.

Upgrade notes are in [this hackmd document](https://hackmd.cloudapps.digital/AOadTZQPTOGidskEPFSAnQ?view)

How to review
-------------
1. Review my notes
1. Review the PR
1. Check my pipeline (N.B. acceptance tests failed because I'd pushed a custom buildpack whilst it was running. There's a second run which passes once I deleted the buildpack)
1. Run it down your pipeline

Who can review
--------------
Anyone

PR #2204, opened 2019-12-16
Upgrade BOSH DNS, set log level, update stemcells

tlwr:
What
----

See https://github.com/alphagov/paas-bootstrap/pull/331 for info WRT stemcells

Since we did #2197 the log-api instance has had high CPU, and has been generating many logs, this is because of changes to trafficcontroller (a partially deprecated loggregator component).

The reason for this is that many successful DNS requests are being logged, which is a bit pointless, and very noisy. In 1 hour it generated 8 million DNS requests in my development environment.

We're updating BOSH DNS because we were a bit behind, and we are setting the log level to WARN instead of DEBUG/INFO

Problems with DNS resolution should be caught by the blackbox DNS canary on each VM. We added some spec tests for BOSH DNS as well

Because BOSH DNS is an addon, we are doing it in the same PR as a stemcell upgrade, so that we don't have to roll all instances multiple times.

How to review
-------------

Code review

Who can review
--------------

Done as pair with @schmie

### DIFF
--- a/manifests/cf-manifest/operations.d/020-bosh-set-stemcells.yml
+++ b/manifests/cf-manifest/operations.d/020-bosh-set-stemcells.yml
@@ -4,4 +4,4 @@
   value:
     - alias: default
       os: ubuntu-xenial
-      version: "621.26"
+      version: "621.29"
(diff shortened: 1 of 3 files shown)


tlwr:
This passed in `dev/tlwr`

PR #2218, opened 2020-01-09
Add usage analytics to scripts

AP-Hunt:
What
----
We don't know which scripts in the `scripts/` directory are being used, and which are dead and broken. To get a handle on this, we're introducing super super basic analytics, in the form of logging the name of the script to a file (`~/.paas-script-usage`) whenever it's executed. In a month or so, we'll combine everyone's logs, and count the entries. Any scripts not used in a month we can assume are fair game to remove.

How to review
-------------
1. Code review
2. Run one of each type of script and check it logs the relative-to-paas-cf name to `~/.paas-script-usage`

Who can review
--------------
Anyone

### DIFF
--- a/scripts/audit-services.rb
+++ b/scripts/audit-services.rb
@@ -1,6 +1,9 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+script_path = File.absolute_path(__FILE__).sub!(Dir.pwd + "/", "")
+File.open(File.expand_path('~/.paas-script-usage'), 'a') { |f| f.puts script_path }
+
 require 'httparty'
 require 'optparse'
 require 'json'
--- a/scripts/cf_subshell_scoped_login.sh
+++ b/scripts/cf_subshell_scoped_login.sh
@@ -1,6 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
+
+echo "${0#$PWD}" >> ~/.paas-script-usage
 
 if [ $# -lt 1 ]; then
   echo "Usage $0 <target>" 1>&2
--- a/scripts/find_diverged_forks.py
+++ b/scripts/find_diverged_forks.py
@@ -2,6 +2,12 @@
 import argparse
 from github import Github
 
+import os
+
+script_path = os.path.relpath(__file__)
+with open(os.path.expanduser('~/.paas-script-usage'), 'a') as f:
+  f.write(script_path + "\n")
+
 BOLD = '\033[1m'
 ENDC = '\033[0m'
 
--- a/scripts/lib/mail_credentials_helper.rb
+++ b/scripts/lib/mail_credentials_helper.rb
@@ -5,6 +5,9 @@ require 'tempfile'
 require 'aws-sdk'
 require 'mail'
 
+script_path = File.absolute_path(__FILE__).sub!(Dir.pwd + "/", "")
+File.open(File.expand_path('~/.paas-script-usage'), 'a') { |f| f.puts script_path }
+
 module EmailCredentialsHelper
   DEFAULT_REGION = "eu-west-1".freeze
 
--- a/scripts/cleanup-deleted-cf-users.sh
+++ b/scripts/cleanup-deleted-cf-users.sh
@@ -2,6 +2,8 @@
 
 set -e -u -o pipefail
 
+echo "${0#$PWD}" >> ~/.paas-script-usage
+
 delete=no
 
 for i in "$@"
--- a/scripts/cleanup_unused_cdn_broker_iam_certs.rb
+++ b/scripts/cleanup_unused_cdn_broker_iam_certs.rb
@@ -1,5 +1,8 @@
 #!/usr/bin/env ruby
 
+script_path = File.absolute_path(__FILE__).sub!(Dir.pwd + "/", "")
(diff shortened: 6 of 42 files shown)


scripts/audit-services.rb
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+`echo "#{File.absolute_path(__FILE__).sub!(Dir.pwd + "/", "")}" >> ~/.paas-script-usage`
```
richardTowers: Feels a bit meh to shell out to echo just to append a line to a file 🤔

scripts/cf_subshell_scoped_login.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
 
+[[ "$0" = /* ]] && echo "${0#$PWD}" >> ~/.paas-script-usage || echo "$0" >> ~/.paas-script-usage
```
richardTowers: This is pretty cryptic - I'm not even sure what the `[[ "$0" = /* ]]` condition does.

Would `"${0#$PWD}"` not be better as `"$(basename "$0")"`?

richardTowers: Ah, I see - you specifically want the path from the current directory to the executable, not just the executable's name.

scripts/find_diverged_forks.py
```diff
@@ -2,6 +2,10 @@
 import argparse
 from github import Github
 
+import os
+
+os.system("echo \""+os.path.abspath(__file__).replace(os.getcwd(), "")+"\" >> ~/.paas-script-usage")
```
richardTowers: Again, seems a bit weird to shell out to echo to append to a file.

scripts/cf_subshell_scoped_login.sh
AP-Hunt: Yeah, I wasn't super happy with this either. The aim was to print out the path of the file relative to the paas-cf directory, when $0 could be anything from just the name of the file (excluding path), to the absolute path to the file.

It reads as
* `[[ "$0" = /* ]] `: does `$0` begin with `/`
* `${0#$PWD}`: strip the value of `$PWD` from the front of `$0`
* `echo $0`: otherwise just `$0`

Is there better way to do this in Bash?

scripts/audit-services.rb
AP-Hunt: I was going for the most obvious one-liner that was similar to the one we'd use in the Bash scripts. I'm up for having it replaced with a conventionally Ruby line.

scripts/find_diverged_forks.py
AP-Hunt: Same explanation as for Ruby, combined with me not knowing Python at all.

46bit:
@AP-Hunt @richardTowers I don't think this is worthwhile refactoring if it works as-is. Can it be merged?

richardTowers:
I'll have a think and see if I can make some suggestions for actual improvements, rather than just criticising :)

scripts/audit-services.rb
richardTowers:
```ruby
script_path = File.absolute_path(__FILE__).sub!(Dir.pwd + "/", "")
File.open(File.expand_path('~/.paas-script-usage'), 'a') {|f| f.puts 'script_path' }
```

The `'a'` argument means "append".

scripts/find_diverged_forks.py
richardTowers:
```python
script_path = os.path.relpath(__file__)
with open(os.path.expanduser('~/.paas-script-usage'), 'a') as f:
  f.write(script_path + "\n")
```

scripts/cf_subshell_scoped_login.sh
richardTowers: I think this could just be:

```bash
echo "${0#$PWD}" >> ~/.paas-script-usage
```

I think there's no need for the conditionals because it will do the right thing anyway:

```bash
BANANA=BANANA
echo "${BANANA#BA}"
# => NANA
echo "${BANANA#BADGER}"
# => BANANA
```

No?

AP-Hunt:
Happy to accept the changes on all of these. Convention > consistency here, I think with hindsight. I'll update the commits soon.

scripts/lib/mail_credentials_helper.rb
```diff
@@ -1,3 +1,5 @@
+`echo #{File.expand_path(__FILE__, __dir__)} >> ~/.paas-script-usage`
```
richardTowers: looks like this is the last one that shells out

richardTowers:
Looks good to me, other than the one remaining file that shells out to `echo`.

I've tested one each of `shell` / `ruby` / `python` and they all wrote a line to the file, so happy to merge once the last echo has gone.

scripts/lib/mail_credentials_helper.rb
AP-Hunt: Bleg! Find and replace has betrayed me! I'll fix it, once and for all ⚔️

AP-Hunt:
> Looks good to me, other than the one remaining file that shells out to `echo`.
> 
> I've tested one each of `shell` / `ruby` / `python` and they all wrote a line to the file, so happy to merge once the last echo has gone.

The final line that shells out has been taken out 🗡

PR #2220, opened 2020-01-10
Make CF target available as bash functions in CF subshell

AP-Hunt:
What
----

Exposes three `_cf_target__*` bash functions in the subshell. They are intended
for users to incorporate in to their command prompts when inside the shell.

- `__cf_target_env`
Outputs the name of the CF deployment, eg "prod-lon"

- `__cf_target_org`
Outputs the org being targeted by CF CLI

- `__cf_target_space`
Outputs the space being targeted by CF CLI

Sample `.bashrc` commands for embedding the information
```
if [ ! -z $CF_HOME ]; then
  # e.g. "prod-lon/admin/billing"
  export PS1+="\$(__cf_target_env)/\$(__cf_target_org)/\$(__cf_target_space)"
fi
```

How to review
-------------
Code review

Who can review
--------------
Anyone

### DIFF
--- a/scripts/cf_subshell_scoped_login.sh
+++ b/scripts/cf_subshell_scoped_login.sh
@@ -31,6 +31,26 @@ esac
 
 TMPDIR=${TMPDIR:-/tmp}
 CF_HOME=$(mktemp -d "${TMPDIR}/cf_home.XXXXXX")
+
+cat <<EOF > "$CF_HOME/.bashrc"
+__cf_target_env ()
+{
+  echo "$TARGET"
+}
+
+__cf_target_org ()
+{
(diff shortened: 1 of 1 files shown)


richardTowers:
I would just export these as variables, instead of using functions for them, no?

I use a go program for my prompt, so I can't (easily) run bash functions.

richardTowers:
Ah, I see - the org / space is a bit more tricky because it will change within the subshell.

richardTowers:
Presumably this is no longer needed now we've got alphagov/paas-cf-prompt ?

AP-Hunt:
> Presumably this is no longer needed now we've got alphagov/paas-cf-prompt ?

Correct. Closing this.

PR #2230, opened 2020-01-20
Fix update_buildpacks to use latest version deps

schmie:
What
-------------

Fix update_buildpacks.sh to use latest rather than current release to read dependencies.

Also renamed the variable "release" to "latestRelease" to make the
variable name more descriptive (is this controversial?!)

How to review
-------------

Code review of `tools/buildpack/main.go`.
Run `scripts/update_buildpacks.sh` and check that `config/buildpacks.yml` is updated correctly, such that dependency versions match with (latest) buildpack release on GitHub.

Who can review
--------------

not @schmie

### DIFF
--- a/tools/buildpacks/main.go
+++ b/tools/buildpacks/main.go
@@ -87,14 +87,14 @@ func getAssetSha(assetDetails AssetDetails) (sha string) {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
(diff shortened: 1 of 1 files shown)


tlwr:
Good fix, nice work!

If you're asking for personal style choices (which don't reflect this PR review), I would say:

- `latestRelease` is a good name in `func main`
- `release` is a good name in `func readManifest`

PR #2237, opened 2020-01-30
[#170704347] Expose Redis cache cluster node metrics from paas-metrics

46bit:
What
----

This makes `paas-metrics` expose a `aws.elasticache.cluster.nodes.count` metric, giving the number of nodes in each Elasticache (Redis) cluster as well as some related Elasticache and CF metadata.

How to review
-------------

* See if the tests pass;
* Code review;
* Deploy in a dev env and try with varying elasticache cluster sizes.

Who can review
--------------

Not @46bit @AP-Hunt.


AP-Hunt:
The contents of this PR are covered in #2224 too, because this is branched from it. I'm closing the PR so that the reviewer can review one PR with many small commits.

PR #2240, opened 2020-02-03
Increase size of  CF prometheus persistent disk

tlwr:
What
----

We are now storing up to 370 days of metrics, which will require > 100GB of storage

This will also give us more IOPS, which should reduce resource usage during compaction

The cost implications in prod are:
- before: 4 * 100 GB = 42.92USD
- after:  4 * 500 GB = 288.52USD

There are no cost implications in development environments

We also have recently had problems with memory usage, when [prometheus is compacting](https://prometheus.io/docs/prometheus/latest/storage/#compaction). The increased disk size will give us more IOPS, by a factor of 5 (because we are increasing the disk size by 5). This may make compaction a bit more efficient and require less memory usage.

How to review
-------------

Code review

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/operations.d/500-prometheus.yml
+++ b/manifests/cf-manifest/operations.d/500-prometheus.yml
@@ -40,7 +40,7 @@
     azs: [z1, z2, z3]
     stemcell: default
 
-    persistent_disk_type: 100GB
+    persistent_disk_type: 500GB
     vm_type: xlarge
 
     networks:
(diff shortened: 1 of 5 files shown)


AP-Hunt:
This is a big jump. What calculations got you to that number?

46bit:
> The cost implications in prod are:
>
> * before: 4 * 100 GB = 42.92USD
> * after: 4 * 500 GB = 288.52USD

Is that per month?

tlwr:
The price quoted is per month from the AWS calculator, I also made a typo:

image
_After_

image

_Before_

@schmie good catch

I'll update this in a sec

PR #2244, opened 2020-02-06
[#168322617] Add high IOPS database plans

LeePorte:
What
----

We have had a number of tenants ask for database plans with greater IOPS. Here they are added with 5 times the amount of disk space and therefor 5 times the amount of IOPS.

How to review
-------------

Code review and run `cf marketplace -s postgres` and `cf marketplace -s mysql` in Lee's dev env

Who can review
--------------

Anyone


tlwr:
There is some custom code in pazmin to hide some plans, but we don't need to explicitly add them.

It might be useful to use CF labels to control what gets displayed in Pazmin/Billing instead of the regular expression munging that happens right now.

I'm running this down my dev env, as I got a 502 trying to check out your calculator

LeePorte:
I'm still deploying the first run of the day, so that'll cover the 502.

PR #2257, opened 2020-02-17
Use user identity for credhub shell

tlwr:
[Story](https://www.pivotaltracker.com/story/show/170894390)

What
----

We are now using user identities for BOSH and Credhub shells

Let's use them instead of the shared client

How to review
-------------

Code review

Try it out

Who can review
--------------

@AP-Hunt

### DIFF
--- a/scripts/credhub_shell.sh
+++ b/scripts/credhub_shell.sh
@@ -12,10 +12,11 @@ function cleanup () {
 }
 trap cleanup EXIT
 
-BOSH_CA_CERT="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-CA.crt" -)"
-BOSH_IP=$(aws ec2 describe-instances \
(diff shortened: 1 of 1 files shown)


scripts/credhub_shell.sh
```diff
@@ -12,10 +12,11 @@ function cleanup () {
 }
 trap cleanup EXIT
 
+BOSH_CA_CERT="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-CA.crt" -)"
 BOSH_IP=$(aws ec2 describe-instances \
```
richardTowers: Do you actually still need BOSH_IP? Or can you just use `bosh-external.${SYSTEM_DNS_ZONE_NAME}` instead?

tlwr: 9f4b313

tlwr: Good shout

tlwr:
Flakey travis test, going to merge this tomorrow

PR #2261, opened 2020-02-18
[#169623326] Retain SSH logs from VMs

mogds:
What
----

Adds an auth logs Cloudwatch log group and ships the auth.log from all instances to this group. It also adds a Cloudwatch subscription so these logs are shipped to Splunk.

How to review
-------------

Code review and deploy to your dev env 

Who can review
--------------

Not me

### DIFF
--- a/manifests/cf-manifest/operations.d/320-cc-ship-logs-for-auditing.yml
+++ b/manifests/cf-manifest/operations.d/320-cc-ship-logs-for-auditing.yml
@@ -14,3 +14,9 @@
             log_stream_name: "{{instance_id}}"
             initial_position: start_of_file
             datetime_format: "%Y-%m-%dT%H:%M:%S"
+          - name: /var/log/auth.log
+            file: /var/log/auth.log
+            log_group_name: auth_logs_((deployment_name))
+            log_stream_name: "{{instance_id}}"
+            initial_position: start_of_file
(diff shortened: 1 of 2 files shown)


tlwr:
Can we not use a [BOSH runtime config addon](https://bosh.io/docs/runtime-config/#placement-rules) instead of having the same operation on each instance group?

```
addons:
  - name: node_exporter
    jobs:
      - name: node_exporter
        release: node-exporter
        properties:
          awslogs-xenial:
            region: ((terraform_outputs_region))
            awslogs_files_config:
              - name: /var/log/auth.log
                file: /var/log/auth.log
                log_group_name: auth_logs_((deployment_name))
                log_stream_name: "{{instance_id}}"
                initial_position: start_of_file
                datetime_format: "%Y-%m-%dT%H:%M:%S"
    include:
      stemcell:
        - os: ubuntu-trusty
        - os: ubuntu-xenial
    exclude:
      jobs:
        - awslogs
      deployments:
        - concourse
```

And then we should only need to to add the modify UAA and CC instance groups

tlwr:
I really think we should use an addon instead, if not we _definitely_ need tests to ensure that every instance group has the `awslogs` job

tlwr:
Mo reworked this PR and the equivalent in paas-bootstrap, so that everything is done as an addon.

PR #2271, opened 2020-03-03
Remove user creation script

paroxp:
What
----

This functionality is being moved to PaaS Admin.

See alphagov/paas-admin#686

How to review
-------------

- Sanity check


tlwr:
I think you can also delete:

- `scripts/reset-org.py`
- `scripts/list-org-owners.sh`
- `scripts/org-usage-report.rb`
- `scripts/reset-org.py`

tlwr:
Also although it is logically separate you can also delete `scripts/find_diverged_forks.py`

paroxp:
Done

PR #2274, opened 2020-03-04
Remove extraneous slash in CF_HOME set by subshell script

AP-Hunt:
What
----
The extra slash was creating a value for CF_HOME which was not the actual path.

How to review
-------------
1. Look at the one character I removed

Who can review
--------------
Me, probably.

### DIFF
--- a/scripts/cf_subshell_scoped_login.sh
+++ b/scripts/cf_subshell_scoped_login.sh
@@ -32,7 +32,7 @@ case $TARGET in
 esac
 
 TMPDIR=${TMPDIR:-/tmp}
-CF_HOME=$(mktemp -d "${TMPDIR}/cf_home.XXXXXX")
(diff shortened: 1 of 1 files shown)


scripts/cf_subshell_scoped_login.sh
```diff
@@ -32,7 +32,7 @@ case $TARGET in
 esac
 
 TMPDIR=${TMPDIR:-/tmp}
```
46bit: Actually, should this now be:
```suggestion
TMPDIR=${TMPDIR:-/tmp/}
```

AP-Hunt: Do you mean in addition to my change, or instead of?

46bit: Without this extra change, aren't you running `mktemp -d "/tmpcf_home.XXXXXX"`?

PR #2281, opened 2020-03-19
[#164728382] Use upstream ipsec release

tlwr:
[Story](https://www.pivotaltracker.com/story/show/164728382)

What
----

Do not use our IPSec release, use SAP's (upstream) instead

We forked the release to change some file permissions, which has been merged upstream.

This doesn't actually change any packages at all, so [it didn't modify any instances during deployment](https://deployer.tlwr.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/151).

How to review
-------------

[Review the no-op changes upstream](https://github.com/alphagov/paas-ipsec-release/compare/gds_master...SAP:master)

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/operations.d/530-ipsec.yml
+++ b/manifests/cf-manifest/operations.d/530-ipsec.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: ipsec
-    version: 0.1.4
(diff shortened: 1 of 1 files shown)


AP-Hunt:
The upstream changes look fine to me, and I can see that the change was a no-op in your env.

PR #2285, opened 2020-03-23
Scale up cells in Ireland

LeePorte:
What
----

As there are some high demand services being deployed to the PaaS let's
make sure we have plenty of capacity.

How to review
-------------

Code review

Who can review
--------------

Anyone


LeePorte:
Yeah we want to be safe, the cost is minimal compared to any reputational damage

PR #2291, opened 2020-03-31
[#171978958] Communicate with tenants of GOV.UK PaaS

LeePorte:
What
----

As part of [#171978958] we need to communicate with the org managers rather than the developers. We didn't have a way of pulling this info out previously. Now we have a process that pulls out a CSV that contains, email address, org, role, admin link to org and the region the user is in.

As it wasn't clear if this code was going to live on past this the tests were added later.

How to review
-------------

Code review, check the tests pass and feel free to try it ` make prod show-tenant-comms-addresses > addresses_prod.csv MANAGEMENT=true`

Who can review
--------------

Anyone


tlwr:
`*.csv` in gitignore may cause some head scratching down the line, but we can address it if it happens

LeePorte:
I though let's go with headscratching over possible ICO data issues

PR #2292, opened 2020-04-01
Enable `service_instance_sharing` flag

LeePorte:
What
----

The ability to share service instances between spaces has been requested by some tenants as it enables some of our tenants to speed up the creation of PR test instances to aid in their CI/CD workflow.

How to review
-------------

Code review
Have deployed to my dev env without issue. feel free to check there

Who can review
--------------

Anyone

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2430,6 +2430,29 @@ jobs:
 
                 cf enable-feature-flag diego_docker
 
+      - task: enable-service_instance_sharing-feature-flag
+        tags: [colocated-with-web]
+        config:
+          platform: linux
+          inputs:
+            - name: paas-cf
+          params:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -2430,6 +2430,29 @@ jobs:
 
                 cf enable-feature-flag diego_docker
 
+      - task: enable-service_instance_sharing-feature-flag
```
paroxp, line 2433: All we're missing here is some PascalCase :trollface: 

Do you think it'd be worth keeping consistency?

LeePorte: I was going with consitency with https://github.com/alphagov/paas-cf/blob/f92d40bfbde04a0a71d1a07c29718800f14f38f1/concourse/pipelines/create-cloudfoundry.yml#L2411

PR #2299, opened 2020-04-14
[Support] Bump cells in London and reduce cells in Ireland

mogds:
What
----

Increases the number of cells in London and reduces the number in Ireland

How to review
-------------

Code review

Who can review
--------------

Not me


tlwr:
We will merge this after we have fixed a cell in London

PR #2303, opened 2020-04-21
Fix BOSH DNS CA certificate rotation

tlwr:
What
----

1. Add 2 pipeline jobs to manage the healthcheck deployment

During debugging of this problem, the logging pipeline was broken in staging. I wanted to re-run the pipeline, but could not because the healthcheck was set to "deployed".

The healthcheck could not be ascertained to be available by the pipeline because the logging pipeline was broken.

Add a task to check the state, and set the healthcheck deployed status to false, to allow us to run the pipeline in such situations.

2. Fix BOSH DNS CA certificate rotation

If you use `((ca.certificate))` when refering to a certificate's CA cert, then you will only get the latest CA certificate interpolated in the manifest.

You must instead use `((certficate.ca))`

When you use `((certificate.ca))` and the latest CA has the "transitional" flag, then BOSH will put the latest ca cert (transitional) and the latest non-transitional CA certificate.

As a practical example, imagine `my_ca` has been updated with `transitional`.

If you refer to it in a manifest:

```
ca: ((my_ca.certificate))
```

You will get:

```
ca: ----- BEGIN CERTIFICATE -----...----- END CERTIFICATE -----
```

If you refer instead to the `.ca` from the leaf:

```
ca: ((my_certificate.ca))
```

You will get:

```
ca: ----- BEGIN CERTIFICATE -----...----- END CERTIFICATE -----\n----- BEGIN CERTIFICATE -----...----- END CERTIFICATE -----
```

3. Add a test to stop this from happening in future

The test is kind of janky, but it finds all CA certificate usages (`ca:`) which are not variables, and checks that they use `var_name.ca` instead of `var_name.certificate`

How to review
-------------

Code review

CI

I'm currently running this in my pipeline

Check that my understanding is consistent with other certificates

Who can review
--------------

Not @tlwr

Appendix
----

The rough timeline which brought this to our attention:

- Smoke tests fail in staging because logs cannot be retrieved
- CA certificate rotation job had market BOSH DNS cert as transitional in previous run
- Investigation on VM saw many invalid TLS certificate error messages
- `/var/vcap/jobs/bosh-dns/config/certs/{server,client}_ca.crt` was not the CA certificate which signed the healthcheck server certificate
- `/var/vcap/jobs/bosh-dns/config/certs/{server,client}_ca.crt` only contained one certificate, but `/stg-lon/stg-lon/dns_healthcheck_tls_ca` had a transitional version
- Checked the manifest, and saw it referenced the certificate directly instead of `certificate.ca`

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -343,6 +343,8 @@ groups:
       - bump-patch-version
       - pipeline-check-lock
       - pipeline-release-lock
+      - healthcheck-check-state
+      - healthcheck-disable
       - cleanup-deleted-cf-users
   - name: tests
     jobs:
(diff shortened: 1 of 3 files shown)


tlwr:
This was deployed to my development environment in this [`cf-deploy job`](https://deployer.tlwr.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/17) which updated the manifest but didn't roll any VMs, which meant the certificates were the same (in my env), which gives me confidence in merging this and depoying to staging

PR #2307, opened 2020-04-24
Set Cloud Controller Worker job timeouts

AP-Hunt:
What
----

The default timeout for a job in the cloud controller worker queue is 4 hours.
This is a very long time for a job to be hanging about when it isn't
completing. This commit sets the timeout, and that of deleting from the
blobstore, to 1 hour. This should provide time for a long running job to
complete, without being so long as to clutter the queue with jobs that will
never complete.

How to review
-------------
1. Code review
2. See that [it went out OK in my dev env](https://deployer.andyhunt.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/5)
3. Run it down your own dev env

Who can review
--------------
Anyone


AP-Hunt:
Closing this PR because we resolved the incident that this was a reaction to in another way (iirc). Talked to @tlwr and agreed this isn't necessary any more.

PR #2312, opened 2020-04-30
Increase broker client timeout property in CC

tlwr:
What
----

1. Also reconcile properties for CC clock, as well as CC worker and API
2. Increase the broker client timeout property

(2) Creating cloudfront distributions sometimes take longer than 1 minute, which is the default timeout for CCs broker HTTP client

Increasing this will increase the likelihood that requests to the CDN broker do not timeout

More details on this zendesk ticket: https://govuk.zendesk.com/agent/tickets/4049154

The actually good solution would be to do the distribution creation in `cdn-cron` rather than within the context of a HTTP request/response.

How to review
-------------

Code review

Who can review
--------------

Not @tlwr


manifests/cf-manifest/operations.d/320-cc-increase-broker-client-timeout.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/broker_client_timeout_seconds?
+  value: 300
+
+- type: replace
+  path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/broker_client_timeout_seconds?
+  value: 300
+
+- type: replace
```
AP-Hunt, line 13: I don't see this property in the job config: https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_clock/spec

tlwr:
You are correct this doesn't apply to the clock, only worker and API

PR #2324, opened 2020-05-20
[#172697662] Add syslog agent in preparation for CF 13 upgrade

mogds:
What
----

Adds the `add-disabled-syslog-agent-for-upgrade.yml` ops file so we can avoid logs on app syslog drains being duplicated or dropped for much of the duration of the deploy when upgrading CF to >= 13.0.0

How to review
-------------

Deploy to your dev env

Who can review
--------------

Not me


tlwr:
This deployed cleanly to my development environment

PR #2326, opened 2020-05-22
Update silk and cf-networking to 2.30

tlwr:
What
----

The releases are built with Go 1.14 and also include a fix for https://github.com/cloudfoundry/cf-networking-release/issues/76

Which will fix the long-standing issue that @uktrade have with unreliability during cell rolling

How to review
-------------

Run this down your development environment

Read the release notes
- [silk](https://github.com/cloudfoundry/silk-release/releases/tag/2.30.0)
- [cf-networking](https://github.com/cloudfoundry/cf-networking-release/releases/tag/2.30.0)

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/operations.d/350-networking.yml
+++ b/manifests/cf-manifest/operations.d/350-networking.yml
@@ -1,0 +1,17 @@
+---
+
+- type: replace
+  path: /releases/name=silk
+  value:
+    name: "silk"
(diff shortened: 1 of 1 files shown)


tlwr:
This has gone green in my dev environment but is not currently mentioned in cf-deployment

PR #2327, opened 2020-05-27
[cf-d 13 part 2] [#172697662] Upgrade CF to v13.2

mogds:
What
----

Upgrades CF to [v13.2.0](https://github.com/cloudfoundry/cf-deployment/releases/tag/v13.2.0)

UAA PRs 
https://github.com/alphagov/paas-uaa/pull/19
https://github.com/alphagov/paas-uaa-release/pull/22

How to review
-------------

- Deploy to your dev env
- Update UAA WIP commit 

Who can review
--------------

Not me


tlwr:
Running Lee's updates down my pipeline but looks fine

There haven't been any template changes to UAA AFAICT

tlwr:
One thought that just came to me

We are removing the adapter instances in the same release as we are upgrading, but the adapter instances are quite early on in the manifest.

This means the adapter instances will get removed early in the deployment process.

Do we want to separate this into two releases, one to do the upgrade, and one to remove adapter instances?

LeePorte:
Based on the data It seems sensible to seperate them.

tlwr:
Actually, we need that commit in this PR to get the manifest to render.

The adapter instances are removed in the `cf-deployment` manifest, which doesn't seem to provide a path to a deployment without interruption to log delivery. I guess our only choice is to do it in one release, but with a maintenance window?

tlwr:
Before we merge this, we need to remove the separator commit and those preceding it.

Also https://github.com/alphagov/paas-cf/pull/2334 should be deployed before this is merged

PR #2328, opened 2020-05-27
[#170700038] Allow pazmin password resets

paroxp:
[Story](https://www.pivotaltracker.com/story/show/170700038)

What
----

We'd like to moce the forgotten password functionality from UAA to
Pazmin, to gain more control of what's happening and how it's being
presented to our beloved users.

The Password Reset API call requires either: `uaa.admin`, `oauth.login`
or `zones.uaa.admin` scope to perform it's action.

Now that we've moved that functionality into Pazmin, we need to make
sure our login page points to it.

We'll be sending new email from pazmin to our beloved tenants. We need
to point to the template in notify, which should be used for that action
to be performed.

How to review
-------------

- Deploy to your dev environment
- Review alphagov/paas-admin#893

⚠️ This needs to be merged after the alphagov/paas-admin#893 ⚠️ 
--------------------------------------------------------------------


tlwr:
I'm gonna split this into two PRs

Once to update the scope

The other to change CF to use Pazmin for password resets

PR #2334, opened 2020-06-05
[#172697662] Enable syslog agent as part of the upgrade to cf 13

LeePorte:
What
----

We have decided that we prefer a period of duplicate log delivery over a
period of log loss. So this PR enables the syslog agents

How to review
-------------

Code review and deploy to dev

Who can review
--------------

anyone


tlwr:
Pushed up an extra revert commit otherwise the disable operation stays around

LeePorte:
Cool that makes sense

PR #2336, opened 2020-06-11
[#172697662] CF Deployment 13 cleanup

tlwr:
[Story](https://www.pivotaltracker.com/story/show/172697662)

What
----

As part of the CF deployment 13 upgrade we added a temporary ops file to enable the syslog binding cache so that we would get double logs during the upgrade

This is inlined into the manifest as of cf-deployment 13

How to review
-------------

Code review

[Check the submodule has the correct operations](https://github.com/cloudfoundry/cf-deployment/blob/6a57656eafce1d9954fe6a81af13b0f23716808a/cf-deployment.yml#L1254)

Run it down your pipeline and check for a no-op

Who can review
--------------

Not @tlwr


barsutka:
ran this branch on my dev environment, everything looks good.

PR #2340, opened 2020-06-12
[#173220478] Making Elasticache service shareable

barsutka:
[Story](https://www.pivotaltracker.com/story/show/173220478)

What
----

The services that we provide as part of our platform are not shareable by default.
As service broker authors we have to implicitly opt-in for the broker to create 'shareable' service instances. To do that, we have to change the metadata of the service broker catalog definitions to make it a shareable service.
This PR will make the redis (based on Amazon ElastiCache) service to become shareable.

How to review
-------------

1. Code review
2. Deploy this branch to your dev environment
3. Make sure that `custom-broker-acceptance-test` job in `create-cloudfoundry` pipeline is passing.
4. [Optional Step] After you deployed this branch to your dev env - you can create an instance of redis service. Then run cf service <service-instance-name>, in the output you should see this line `This service is not currently shared.` just before the info about the last operation.

Who can review
--------------

@barsutka


tlwr:
LGTM

PR #2345, opened 2020-06-16
Update UAA-customised to include a11y improvements

LeePorte:
What
----

We have a new releae of UAA customised that used the up to date design
system. This releases that version.

How to review
-------------
Code review

Who can review
--------------

Anyone


LeePorte:
Ah PR clash with https://github.com/alphagov/paas-cf/pull/2344

PR #2351, opened 2020-06-22
Expose metric for the number of CDN route TLS certs issued by different certificate authorities

AP-Hunt:
What
----
We've asked our tenants to perform a `cf update-service` call to migrate their
CDN route services between certificate authorities. By exposing a metric for
the number of CloudFront instances presenting certificates signed by different
authorities, we can track how many tenants have performed the migration.

How to review
-------------
1. Code review

Unless you have an old CDN route service in your dev env, you won't be able to test that it counts >1 certificate authority. The unit tests should test that.

Who can review
--------------
Anyone

### DIFF
--- a/tools/metrics/gauges_tls.go
+++ b/tools/metrics/gauges_tls.go
@@ -106,3 +106,55 @@ func CDNTLSValidityGauge(
 		return w.WriteMetrics(metrics)
 	})
 }
+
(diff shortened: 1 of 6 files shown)


tlwr:
Was gonna say `FDescribe` should go but you got it already 🚤

PR #2349, opened 2020-06-19
[#173382894] Ruby 2.7

tlwr:
[Story](https://www.pivotaltracker.com/story/show/173382894)

What
----

- Use ruby 2.7.1
- Use Ubuntu Focal for running the tests
- Update Rubocop
- Rename some of the rubocops since they changed
- Do not use govuk-lint, as it is deprecated, there is a rubocop replacement
- Do not run `lint_ruby` as part of `lint` and therefore `test` make task

In a follow up PR I will fix all the linting errors, so this one is easier to review

How to review
-------------

CI

Check Travis is still running the tests, and linting

Maybe run on your development environment pipeline

PR #2353, opened 2020-06-22
[#173382894] Ruby 2.7 lint

tlwr:
[Story](https://www.pivotaltracker.com/story/show/173382894)

What
----

Review [part one](https://github.com/alphagov/paas-cf/pull/2349)

This PR re-adds linting, and makes the codebase conform to the rubocop-govuk gem, as such corresponds with the GDS Ruby style guide.

How to review
-------------

Travis should pass

Code review commit by commit

Linting should be re-enabled in the Makefile

Before merging, the separator and the commits before the separator should be removed via a `git rebase`

Who can review
--------------

Not @tlwr


AP-Hunt:
Are we going to squash the other commits after the separator too?

tlwr:
I'll keep them separate commits after dropping the separator, noisy history but good for `git bisect`ing if we discover a regression

tlwr:
Rebased now that #2349 is merged

PR #2356, opened 2020-06-23
Add 6 more cells in Ireland for Notify load test

tlwr:
What
----

This should be reverted when they are finished

How to review
-------------

Code review


tlwr:
There are 12 routers and they didn't seem troubled by the soak testing, just exercised a bit

PR #2358, opened 2020-06-25
cf: horizontally scale scheduler VM

tlwr:
What
----

We want to support more DNS requests per second served by service-discovery-controller

We want to have more instances, so that when we deploy we do not have a single instance of scheduler

Context
----

Yesterday one of our users (GOV.UK Notify) experienced an outage pertaining to internal DNS

The DNS request path is like this
image

During the incident we experienced a high rate of DNS failures
image

During the incident the number of DNS requests went up by roughly 4x (this is important)
image

We can observe that these are DNS requests to service-discovery-controller (that's what backs the DNS requests made by bosh-dns-adapter). We can observe that there was a spike in requests
image

We can correlate this to the number of concurrent socket allocations on the scheduler instances
image

We can correlate this to TCP listen overflows on the scheduler instances
image

We can observe in our logs lots of messages about `TLS handshake failure`
image

We can correlate these failures to `http: Accept error: accept tcp 0.0.0.0:8054: accept4: too many open files; retrying in 10ms`
image

`service-discovery-controller` runs via `bpm` with the default `ulimit` (number of open files/sockets) which for us is a soft limit of `1024`. The controller does not attempt to increase the limit, the controller has a limit of 1024 connections open.

If `bosh-dns-adapter` has a DNS request failure, it has a [naive retry mechanism](https://github.com/cloudfoundry/cf-networking-release/blob/develop/src/bosh-dns-adapter/sdcclient/service_discovery_client.go#L80-L94) that can denial of service (DoS) the DNS server. It retries 4 times in quick succession

Summary:
- GOV.UK Notify making many DNS requests
- BOSH deploy happens, one service-discovery-controller is available, one is being redeployed
- number of concurrent DNS requests is greater than the ulimit
- service-discovery-controller cannot handle any more DNS requests
- bosh-dns-adapter naively retries, immediately quadrupling the number of DNS requests being made
- ‼️ 

The above summary would also explain the reason why GOV.UK Notify experiences 502s when we re-deploy the scheduler instances

In the short term, we should increase the number of scheduler instances deployed, so that we can serve more internal DNS requests

How to review
-------------

Code review

CI

Who can review
--------------

@LeePorte

### DIFF
--- a/manifests/cf-manifest/env-specific/default.yml
+++ b/manifests/cf-manifest/env-specific/default.yml
@@ -4,6 +4,7 @@ router_instances: 2
 api_instances: 2
 doppler_instances: 6
 log_api_instances: 2
+scheduler_instances: 2
 cc_worker_instances: 2
 cc_hourly_rate_limit: 15000
 paas_region_name: dev
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -4,6 +4,7 @@ router_instances: 6
 api_instances: 12
 doppler_instances: 33
 log_api_instances: 6
+scheduler_instances: 4
 cc_worker_instances: 4
 cc_hourly_rate_limit: 20000
 paas_region_name: london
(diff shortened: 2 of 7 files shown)


chrisfarms:
:trophy: pull request context of the year nominee

rjbaker:
Can you increase the default `ulimit` for the `service-discovery-controller`? 1024 is pretty low.

tlwr:
@rjbaker see https://github.com/cloudfoundry/cf-networking-release/pull/80 mentioned in the table above

PR #2360, opened 2020-06-26
Deploy CDN broker 0.1.36

AP-Hunt:
What
----
Deploys a version of the CDN broker with the fix from https://github.com/alphagov/paas-cdn-broker/pull/40

How to review
-------------

1. Deploy and test that updating CDN route still works

Who can review
--------------
Anyone


AP-Hunt:
I deployed this to @barsutka's dev env and it all works OK

PR #2367, opened 2020-07-03
cf: prometheus in london has big disk

tlwr:
What
----

We have 370d retention on the metrics for our users

We are reaching 85% prometheus disk capacity, we should increase max capacity

Ireland does not need more disk, just London, so make it specific to London

How to review
-------------

[Check grafana](https://grafana-1.london.cloud.service.gov.uk/explore?left=%5B%22now-6h%22,%22now%22,%22prometheus%22,%7B%22expr%22:%22bosh_job_persistent_disk_percent%7Bbosh_job_name%3D%5C%22prometheus%5C%22%7D%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D)

CI

Who can review
--------------

Not @tlwr

### DIFF
--- a/manifests/cf-manifest/env-specific/prod.yml
+++ b/manifests/cf-manifest/env-specific/prod.yml
@@ -8,3 +8,6 @@ scheduler_instances: 4
 cc_worker_instances: 4
 cc_hourly_rate_limit: 20000
 paas_region_name: ireland
+
+prometheus_disk_size: 500GB
+prometheus_retention_size: 475GB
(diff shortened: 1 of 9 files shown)


manifests/cf-manifest/env-specific/prod.yml
```diff
@@ -8,3 +8,6 @@ scheduler_instances: 4
 cc_worker_instances: 4
 cc_hourly_rate_limit: 20000
 paas_region_name: ireland
+
+prometheus_disk_size: 500GB
+prometheus_retention_size: 475GB
```
AP-Hunt, line 13: Should we keep the `prometheus_disk_size - prometheus_retention_size = 50GB` in both regions?

tlwr: This was 475GB before

I made 750/700 arbitrarily (roughly the same proportion)

AP-Hunt: Ok. same proportion makes sense.

PR #2371, opened 2020-07-07
Use CF CLI v7

tlwr:
What
----

In https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/193 we started to use the generally available version of the CF CLI in our `cf-cli` docker image

This enables us to use `cf push --strategy=rolling` instead of using the blue-green plugin or autopilot plugin, and enables us to avoid the legacy `buildpack` parameter (instead of `buildpacks`) in our manifests

How to review
-------------

CI

Deploy to your development environment

Run all the deploy jobs:

- [ ] pazmin
- [ ] paas-accounts
- [ ] paas-billing
- [ ] aiven broker
- [ ] post-deploy

Who can review
--------------

Not @tlwr

PR #2372, opened 2020-07-08
[#173692258] Add pinger and ponger health check apps

tlwr:
[Story](https://www.pivotaltracker.com/story/show/173692258)

What
----

We are going to availability test internal networking during our pipeline

To do this we need some apps which use internal networking and internal DNS

This PR adds two new example apps:
- healthcheck-pinger
- healthcheck-ponger

The ponger app always returns 200 OK

The pinger app returns 200 OK if the ponger app returns 200 OK, otherwise 503

This PR deploys the two apps as part of the deploy-healthcheck step

In a later PR we will actually availability test them

How to review
-------------

Deploy to your development environment

(⚠️  this requires https://github.com/alphagov/paas-cf/pull/2371 because of how `cf add-network-policy` works ⚠️ )

`curl https://healthcheck-pinger.$env.london.cloudpipelineapps.digital` should return `OK` with status `200`

Who can review
--------------

Not @tlwr

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2722,6 +2722,20 @@ jobs:
                 cf push --strategy=rolling healthcheck
 
                 cd "${BUILD_ROOT}"
+                cd paas-cf/platform-tests/example-apps/healthcheck-pinger
+                cf cancel-deployment healthcheck-pinger || true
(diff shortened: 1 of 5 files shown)


tlwr:
rebased after cf cli v7

barsutka:
pipeline is green, and curl of healthcheck-pinger returns OK/200.
Looks good.

PR #2375, opened 2020-07-13
prometheus: actually drop the metrics properly

tlwr:
What
----

Miki was looking into this for a tenant broker story

relabel_configs only applies to dropping targets, we want to drop metrics instead

fix the test and the dropping to actually drop the metrics

How to review
-------------

Code review

Check [tlwr pipeline is green](https://deployer.tlwr.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/prometheus-deploy/builds/139)

Review timeseries showing number of metrics, before and after relabelling changes:
an elasticsearch instance being scraped

Who can review
--------------

@46bit

### DIFF
--- a/manifests/cf-manifest/operations.d/500-prometheus.yml
+++ b/manifests/cf-manifest/operations.d/500-prometheus.yml
@@ -92,6 +92,7 @@
                     regex: 'aws-((terraform_outputs_region))'
                     action: keep
 
+                metric_relabel_configs:
                   - source_labels: [__name__]
                     separator: ;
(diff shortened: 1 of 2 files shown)


philandstuff:
>relabel_configs only applies to dropping tags, we want to drop metrics instead

No, relabel_configs applies to dropping *targets*.

tlwr:
thanks for the correction :) that is me typing words wrong - afaik tags aren't even things WRT prometheus 🏷️

PR #2391, opened 2020-07-31
Update VPC peering for govuk_development

seb-szy:
What:
Update to the VPC subnet configured for GOV.UK Development peering.

Why:
"subnet_cidr" represents client subnet as per: step 1&2 of PaaS guide at https://team-manual.cloud.service.gov.uk/guides/vpc_peering/#content

### DIFF
--- a/terraform/prod.vpc_peering.json
+++ b/terraform/prod.vpc_peering.json
@@ -63,6 +63,6 @@
         "peer_name": "govuk-infrastructure-test",
         "account_id": "430354129336",
         "vpc_id": "vpc-9e62bcf8",
-        "subnet_cidr": "172.16.36.0/22"
+        "subnet_cidr": "10.200.0.0/16"
     }
 ]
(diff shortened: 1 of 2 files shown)


sengi:
Thanks, @seb-szy!

PaaS folks - I can confirm we definitely need to use our existing 10.200/16 network on the GOV.UK side for this. Sorry if this wasn't sufficiently clear from the original request. Even though we _could_ renumber our test environment, we definitely won't be able to renumber our other pre-existing environments (which are 10.1, 10.12 and 10.13).

sengi:
Assuming you're ok with this, would need to tweak https://github.com/alphagov/paas-cf/blob/master/terraform/scripts/generate_vpc_peering_test.go#L65 to just look for clashes with the two /16s which are actually in use rather than the whole of 10/8.

sengi:
Depends on #2394 I think

tlwr:
```
_, reservedRange, _ := net.ParseCIDR("10.0.0.0/8")
```

Could be instead `10.0.0.0/16` and `10.255.0.0/16` without the exception

sengi:
> Could be instead `10.0.0.0/16` and `10.255.0.0/16` without the exception

Lee mentioned to me on Slack that `10.1/16` is also in use somewhere. OTOH Richard didn't find that one when he went looking in the AWS console so I'm not entirely sure what the deal is with that one. (The exception approach sounds safer to me FWIW.)

PR #2392, opened 2020-07-31
A new tool for monitoring org quota upgrades

46bit:
What
----

This commit adds a small new tool named `org_quota_history`. When invoked using the instructions in its source code, it will output a CSV of the orgs that exist, when/if they were upgraded to be billable, etc.

Data is sourced from `paas-billing`'s `cfstore`. That stores a record  of the quotas PaaS orgs have had since we started recording data in mid-2018.

Ideally this should be put in a better place than `paas-cf` but that's left for later. I have some ideas for a private `paas-growth` repo to store visualisations, scripts, etc without having to build them into `paas-admin`.

How to review
-------------

The code is intentionally not very good--just about maintainable. This PR is largely so that Mark has somewhere to link people to learn how to regenerate the CSVs next time.

Who can review
--------------

Anyone.

### DIFF
--- a/tools/org_quota_history/main.rb
+++ b/tools/org_quota_history/main.rb
@@ -71,7 +71,7 @@ orgs_seen.each do |org_guid, _|
   orgs_seen[org_guid] = org_current_metadata["errors"] ? false : org_current_metadata
 end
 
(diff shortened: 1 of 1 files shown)


46bit:
Ready for review.

PR #2394, opened 2020-08-04
Ensure VPC peering CIDR blocks don't overlap with our infrastructure

AP-Hunt:
What
----

We are going to allow GOV.UK to set up a VPC peering which is outside of the
range we normally grant to tenants. This will introduce a wider set of CIDR
blocks being alloated to tenants, in a way that isn't wholly consistent. To
prevent future us from accidentally allocating a block that overlaps with our
own infrastructure, this commit adds a test to check that very thing.

How to review
-------------
1. Code review
2. Check the tests pass

Who can review
--------------
Anyone

### DIFF
--- a/terraform/scripts/generate_vpc_peering_test.go
+++ b/terraform/scripts/generate_vpc_peering_test.go
@@ -1,6 +1,7 @@
 package scripts_test
 
 import (
+	"fmt"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
@@ -16,6 +17,8 @@ import (
 var _ = Describe("VPC peering", func() {
 
 	var KEY_NAMES = []string{"peer_name", "account_id", "vpc_id", "subnet_cidr"}
+	var ENV_NAMES = []string{"dev", "prod", "staging"}
+	var REGION_NAMES = []string{"eu-west-1", "eu-west-2"}
 
 	Describe("config files", func() {
 		var files []string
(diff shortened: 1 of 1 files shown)


terraform/scripts/generate_vpc_peering_test.go
```diff
@@ -84,6 +87,44 @@ var _ = Describe("VPC peering", func() {
...
+
+							infraNets = append(infraNets, net)
+						}
+
+						for _, filename := range files {
+							file, _ := ioutil.ReadFile(filename)
+
+							var peers []map[string]string
+							err := json.Unmarshal([]byte(file), &peers)
+							Expect(err).To(BeNil(), "Config file: %s", filename)
```
tlwr, line 110: nit

```suggestion
							Expect(err).NotTo(HaveOccurred(), "Config file: %s", filename)
```

terraform/scripts/generate_vpc_peering_test.go
```diff
@@ -84,6 +87,44 @@ var _ = Describe("VPC peering", func() {
...
+
+							for _, peer := range peers {
+								_, peerNet, err := net.ParseCIDR(peer["subnet_cidr"])
+								peerName := peer["peer_name"]
+								Expect(err).To(BeNil(), "Config file: %s", filename)
```
tlwr, line 115: nit

```suggestion
								Expect(err).NotTo(HaveOccurred(), "Config file: %s", filename)
```

terraform/scripts/generate_vpc_peering_test.go
```diff
@@ -84,6 +87,44 @@ var _ = Describe("VPC peering", func() {
...
+
+								for _, infraNet := range infraNets {
+									intersection := (peerNet.Contains(infraNet.IP) || infraNet.Contains(peerNet.IP))
+									Expect(intersection).To(Equal(false), "Peer CIDR %s (%s) intersects with infra CIDR %s", peerNet, peerName, infraNet)
```
tlwr, line 119: nice

PR #2420, opened 2020-08-21
Uploading secrets fails safely

tlwr:
What
----

We use ruby scripts to upload secrets (and all sorts of other things)

Ruby has the concept of "execution strings" (eg "`" or "%x") where:
- the contents of the execution string is executed as a child process
- the exitstatus of the child process is populated in `$?` and `$CHILD_STATUS` global variables

When the exitstatus is non-zero, this does not stop the execution of the program, like a bash script without `set -e`

This is quite dangerous, especially within the context of uploading secrets. Yesterday I uploaded secrets but without correctly configured `.paas-pass` and `.paas-pass-high` symlinks. I expected the secrets uploading scripts to fail, but they did not.

This pull request:
- Ensures that an error is thrown if `pass` fails
- Adds rubocop custom linting rules to ensure that programmers check exitstatus and output
- Fixes other areas in the codebase where we were not being vigilant

The linting rules look like:

```
Offenses:

concourse/scripts/generate-public-key-vars.rb:11:3: C: CustomCops/MustCaptureXStr: Must capture execution string output
  `gpg -k #{id}`
  ^^^^^^^^^^^^^^
concourse/scripts/generate-public-key-vars.rb:11:3: C: CustomCops/MustCheckXStrExitstatus: After using execution string must check exitstatus
  `gpg -k #{id}`
  ^^^^^^^^^^^^^^
```

How to review
-------------

Clone the repository

```
$ gds aws paas-dev-admin -- make dev credhub
$$ make dev upload-all-secrets
```

Deploy to your development environment

Who can review
--------------

Not @tlwr


46bit (approved):
Impressive 👍

PR #2421, opened 2020-08-21
Use upload-runtime-config in cf-deploy

schmie:
What
----

Use upload-runtime-config instead of upload-config in cf-deploy.
The latter does not upload releases configured in the runtime config
(See [discussion on this GitHub issue](https://github.com/cloudfoundry/bosh/issues/2209)).

How to review
-------------

- Code review
- To test in dev (requires destroyed cloudfoundry) - or [see it working in my dev env](https://deployer.schmie.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/6)
  - `BOSH (your_dev) $ bosh delete-release syslog`
  - Run create-cloudfoundry and see the release uploaded in `update-paas-cf-runtime-config` / `cf-deploy` not failing with `Error: Release 'syslog' doesn't exist`

Who can review
--------------

Not @schmie

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1724,8 +1724,7 @@ jobs:
                   BOSH_CLIENT='admin'
                   export BOSH_CLIENT
 
-                  bosh -n update-config \
-                    --type=runtime \
+                  bosh -n update-runtime-config \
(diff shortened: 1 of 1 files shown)


46bit (approved):
Schmie's env has `cf-deploy`ed several times with this change so it looks good to me :)

PR #2424, opened 2020-08-25
block-routes: disabling create account route

chrisfarms:
What
----

changes how the "create account" routes are being blocked, by reserving
routes directly on the system domain instead of by creating a new
domain.

a [recent change][1] has prevented the creation of a domain that overlaps
with the system domain, which has broken how the create account routes
were being masked and is preventing the creation of a new env

[1]: https://github.com/cloudfoundry/cloud_controller_ng/commit/666bcd57ee8abf200b22863bc5f91ef3dcd202d4

How to review
-------------

After applying it should not be possible to visit:

* https://login.$DEPLOY_ENV.dev.cloudpipeline.digital/create_account
* https://login.$DEPLOY_ENV.dev.cloudpipeline.digital/create_account.do
* https://uaa.$DEPLOY_ENV.dev.cloudpipeline.digital/create_account
* https://uaa.$DEPLOY_ENV.dev.cloudpipeline.digital/create_account.do

Consider:

* Is anything lost by not reserving the route as a domain?
* What will the impact of this change be to an existing long-lived environment that has already managed to create the overlapping domain? We may need to delete the existing domains/routes for this to apply cleanly?

Who can review
--------------

Not @chrisfarms

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -2910,32 +2910,25 @@ jobs:
                 cd paas-cf/platform-tests/example-apps/create-account-holding-page/
                 cf push
 
-                if ! cf domains | grep -q "login.((system_dns_zone_name))"; then
-                  cf create-domain admin "login.((system_dns_zone_name))"
-                fi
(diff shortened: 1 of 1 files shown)


chrisfarms:
nope, didn't work

PR #2426, opened 2020-08-26
disable self service (account creation / password reset) endpoints

chrisfarms:
What
----

* disables the "self service" (account creation and password reset) endpoints
* updates the tests that check that account creation endpoint is disabled
* removes the tests for password reset

a [recent change in cloudcontroller](https://github.com/cloudfoundry/cloud_controller_ng/commit/666bcd57ee8abf200b22863bc5f91ef3dcd202d4) has prevented the creation of a domain that overlaps with the system domain, which has broken how the create account routes were being masked and is preventing the creation of a new env. (existing environments are unaffected as the routes already exist and so skip validation).

the only reason we had to block the endpoints this way previously was because there was no way to disable account creation without also disabling password reset, however password reset is now handled by pazmin and so we can ignore the built-in password reset feature.

How to review
-------------

* Confirm it is not possible to use the anonymous account creation endpoint: https://login.$DEPLOY_ENV.dev.cloudpipeline.digital/create_account ([example from autom8 env](https://login.autom8.dev.cloudpipeline.digital/create_account))
* Are the tests sufficient 

Who can review
--------------

Not @chrisfarms


46bit (changes requested):
Hey Chris. This is great. Given UAA can be... surprising... I'd prefer to keep the `/create_account.do` tests just-in-case 🙈

chrisfarms:
> Hey Chris. This is great. Given UAA can be... surprising... I'd prefer to keep the `/create_account.do` tests just-in-case

The problem is that requests to this endpoint now do (harder to test for) things ... like redirect to /login ... which redirects to /login which (sometimes) gets you to pazmin ... or maybe an infinite redirect loop:

Try this to see what I mean:

```
curl -L -vv -X POST https://login.autom8.dev.cloudpipeline.digital/create_account.do
```
We _could_ check for 302 status on that request, but that feels fragile.

So I fear those tests are now of limited value at best and at worst a flake factory.

I believe relying on checking that the `/create_account` page is 404ing is probably the best balance we will get between "create_account is disabled" and "the tests aren't too fragile".... unless you can think of another check?

46bit (approved):
> The problem is that requests to this endpoint now do (harder to test for) things ... like redirect to /login ... which redirects to /login which (sometimes) gets you to pazmin ... or maybe an infinite redirect loop:

Well that's horrifying. I agree let's not try to embrace the horror. 🦑

chrisfarms:
One thing I have just remembered ... This change removes the bit in the pipeline that pushes the holding page app and maps some routes ... however nothing will _remove_ that app/routes from the persistent environments ... I don't think this matters too much, but we may want to go manually tidy them up to keep the long-lived envs more in sync with the fresh ones

PR #2430, opened 2020-09-01
[#174595178] Fix postgres plan UUID

tlwr:
[Story](https://www.pivotaltracker.com/story/show/174595178)

What
----

[paas-billing](https://github.com/alphagov/paas-billing) relies on the unique name of service plans being a UUID

Billing breaks if the following happens:
- we have a plan with a unique name that is not a valid UUID
- someone uses that plan

This happened recently for the app-autoscaler, but is also a historical issue

This PR
- fixes a typo in the unique name of `medium-10-high-iops`
- add acceptance test which tests that all of the plans have UUID unique IDs

How to review
-------------

Code review

Deploy to your dev env

Observe that `post-deploy` fails during `register-rds-broker` task, with an unknown error

Run the following:

```
cd paas-bootstrap
make dev bosh-cli
bosh ssh api/0
cd /var/vcap/jobs/cloud_controller_ng
./bin/console
ServicePlan.find(name: 'medium-10-high-iops').delete
```

Restart `post-deploy`, it should go green

Who can review
--------------

Not @tlwr


tlwr:
I should point out, no one is using the `medium-10-high-iops` which is why it is okay to delete it, and also why we haven't observed billing throwing errors

PR #2433, opened 2020-09-02
Better alerts for billing that don't misrepresent how well we recharge

46bit:
What
----

This improves our alerts for problems with billing and recharging:

1. Adds a `PaaSBillingIsNotIncreasing` alert that will warn us if `paas-billing`'s totals have stopped increasing. This would have alerted us to #2428 much earlier
2. Fixes the alert message of the existing AWS recharge alerts. They have been converting a percentage to currency and misrepresenting a 50% recharge rate as us recharging all but 50p. 🤦🏻‍♀️
3. Reduces the expected RDS recharge rate to 50%, as this is in line with what I see in Grafana.

It may be that (3) needs further investigating and there are more costs to recoup, but right now all that is happening is the RDS recharge alert is going off constantly and isn't actionable.

How to review
-------------

See what you think 🤷 

Who can review
--------------

Not @46bit

### DIFF
--- a/manifests/prometheus/alerts.d/paas-billing-costs.yml
+++ b/manifests/prometheus/alerts.d/paas-billing-costs.yml
@@ -27,7 +27,7 @@
           layer: billing
         annotations:
           summary: "PaaS Billing DB costs exceed AWS RDS costs"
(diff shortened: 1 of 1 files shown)


tlwr:
It appears we are charging a much higher proportion in London compared to Ireland (59% vs 78%)

tlwr (approved):

46bit:
I filed Pivotal #174645077 to investigate that low recharge rate, since it suggests a lot of money GDS could save.

PR #2217, opened 2020-01-07
[BLOCKED] Introduce Elasticsearch 7.x plans

AP-Hunt:
Blocker
----
This is blocked by [the Aiven broker](https://github.com/alphagov/paas-aiven-broker) not allowing users to apply updates to their ES6 services. This will be necessary for tenants to upgrade from ES6 to ES7.

Story #170575362 covers the work

What
----
On 14th October 2019, our Elasticsearch provider (Aiven) announced support for
Elasticsearch 7.4.0. Since then, they have announced support for 7.4.2 and
7.5.1.

Our tenants love elasticsearch and we love our tenants, so this commit gives
them access to Elasticsearch 7.x plans!

How to review
-------------
1. Run it down your pipeline (or check mine, `DEPLOY_ENV=andyhunt`)
1. Check the acceptance tests pass
1. Provision yourself an ES 7 cluster using `cf cs elasticsearch tiny-7.x my-es-7`
1. Provision yourself an ES 6 cluster and upgrade it to ES7 using `cf cs elasticsearch tiny-6.x my-es-6; cf update-service my-es-6 -p tiny-7.x`

Who can review
--------------
Anyone

PR #2434, opened 2020-09-03
tests: skip elasticsearch provisioning test

tlwr:
What
----

We need to unblock the pipeline so we can release new changes

Aiven have EOLd 6.x ES, so this test is failing

Skip it, so that when we add ES 7.x we can unskip it using 7.x

How to review
-------------

Code review

Who can review
--------------

@LeePorte

### DIFF
--- a/platform-tests/aiven-broker-acceptance/elasticsearch_service_test.go
+++ b/platform-tests/aiven-broker-acceptance/elasticsearch_service_test.go
@@ -89,7 +89,7 @@ var _ = Describe("Elasticsearch backing service", func() {
 			pollForServiceDeletionCompletion(dbInstanceName)
 		})
 
-		It("is accessible from the healthcheck app", func() {
+		XIt("is accessible from the healthcheck app", func() {
 			By("allowing connections with TLS")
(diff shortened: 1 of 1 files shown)


LeePorte:
I'd have a preference that we picked in from https://github.com/alphagov/paas-cf/pull/2217 so that tenants can provision an elasticsearch whilst we work on the upgrade path

tlwr:
I can add the plans from the linked PR, but I'll leave the modifications to the test suite as we will still only be testing one version

LeePorte:
As 6 is going EOL, we may as well test with 7

tlwr:
Yes, but not 6.x and 7.x which is what the PR currently does, also the PR currently doesn't have the AdditionalMetadata or billing information so I'm adding that too

paroxp (approved):
At the moment, we need to be in the position where the pipeline is unblocked. We shouldn't rush any features out in the same time as unblocking the pipeline in the same PR.

LeePorte:
Ah as you're adding those in a separate PR then all good to ship

LeePorte (approved):

PR #2437, opened 2020-09-03
bosh: blackbox exporter is a runtime config addon

tlwr:
What
----

We use bosh dns in multiple deployments

We should canary dns in all those deployments

So we should deploy the blackbox exporter in all deployments using a runtime config addon

How to review
-------------

Code review

Check `count(ALERTS{alertstate="firing", alertname="PrometheusScrapeError"})` in your environment before and after the deployment

Who can review
--------------

Not @tlwr


46bit:
I'm happy with this change but can't test if it works. Happy to approve if you've already deployed it?

tlwr:
I have already deployed it and:
image

46bit (approved):

46bit:
Having fewer alerts will spark joy

PR #2440, opened 2020-09-07
Add initial version of SQS broker

chrisfarms:
### What

Add the [SQS Service Broker Boshrelease](https://github.com/alphagov/paas-sqs-broker-boshrelease/pull/2) to the deployment.

* Adds terraform, manifests and ops files for deploying the SQS service broker
* Sets the relevent IAM role and permission boundaries as defined in https://github.com/alphagov/paas-aws-account-wide-terraform/pull/226
* Advertises "standard" and "fifo" queue types as plans
* Registers the broker as part of the pipeline
* Enable service access only for the `admin` and `gds-tech-ops` orgs (for now)
* Add platform tests to ensure `cf marketplace`, `cf create-service`, `cf bind-service` all work as expected and the bindings produce valid useful credentials.
* Shrink broker deployment to 1x instance in dev

### Why

The people have spoken, and they said "esscueace" 

### Notes

This is NOT going out the public right away, we intend on getting this deployed but only enabled for a select few orgs, then iterating a little while working out the final parts. However this is a fully working implementation.

### Before merge

[x] update the manifest to point at the final boshrelease once https://github.com/alphagov/paas-sqs-broker-boshrelease/pull/2 is merged


AP-Hunt:
I enjoy this commit history. It makes me feel good to see other people doing exactly the same process of trial and error I do with these things.

barsutka:
Ran this PR through my dev environment, it's looking good (all green). 
created a sqs service instance and bound it to the healthcheck app, sqs_test endpoint works fine - so far so good.
Shall we start thinking about more in depth tests - application(s) that are using SQS for more complex use-cases?

chrisfarms:
> Shall we start thinking about more in depth tests - application(s) that are using SQS for more complex use-cases?

I think if we want tests for specific tenant use cases that we are documenting, then those test should probably live in the `paas-cf/platform-tests/...` I don't think we should be attempting to test much more than basic usage within the broker codebase where we are mostly concerned with ensuring we are translating provision/bind params -> cloudformation params.

The initial scope of this work is only to provide "a standard queue that can be configured to send failed messages to a dead-letter queue", so there's not really much in the way of "advanced" usage yet, but we are also covering the common use cases for setting up "fifo" queues and some basic tuning parameters for message size/visibility that we know will be required.

barsutka:
> Did you have anything in mind for further tests you would like added at this stage @barsutka ?

nothing concrete, tbh, just a common cases for messaging - things like a simple pub/sub and fan-out use-cases from the top of my head. If there is no user need for it right now, we should probably postpone the implementation of this coverage in our tests.

chrisfarms:
boshrelease bumped to the non-dev version ... ready :rocket:

paroxp (approved):
Tested on `autom8` pipeline and with code review. So did @barsutka.

Ready for lift off.

PR #2450, opened 2020-09-16
Pay Test VPC Peering

rorymalcolm:
What
----

Added VPC peering details for Pay's test environment

How to review
-------------

Sense check with @danworth or @rjbaker 

Who can review
--------------

PaaS team members

### DIFF
--- a/terraform/prod.vpc_peering.json
+++ b/terraform/prod.vpc_peering.json
@@ -70,5 +70,11 @@
         "account_id": "234617505259",
         "vpc_id": "vpc-08b7d799f1691fd11",
         "subnet_cidr": "172.16.36.0/22"
+    },
+    {
+        "peer_name": "govuk-pay-test",
+        "account_id": "734876687119",
+        "vpc_id": "vpc-0170bee868cc6fb34",
+        "subnet_cidr": "172.16.40.0/22"
     }
 ]


rjbaker:
I've checked the details in the destination account/VPC and they're correct.

terraform/prod.vpc_peering.json
```diff
@@ -70,5 +70,11 @@
         "account_id": "234617505259",
         "vpc_id": "vpc-08b7d799f1691fd11",
         "subnet_cidr": "172.16.36.0/22"
+    },
+    {
+        "peer_name": "govuk-pay-test",
+        "account_id": "734876687119",
+        "vpc_id": "vpc-031070d650a39111e",
+        "subnet_cidr": "172.17.36.0/22"
```
tlwr, line 78:
```suggestion
        "subnet_cidr": "172.16.40.0/22"
```
Can this be continuous within the 172.16 range (ignore govuk-infrastructure-test to see the pattern)

rorymalcolm: Updated - maybe sense check with @rjbaker but I think this is to spec now

terraform/prod.vpc_peering.json
```diff
@@ -70,5 +70,11 @@
         "account_id": "234617505259",
         "vpc_id": "vpc-08b7d799f1691fd11",
         "subnet_cidr": "172.16.36.0/22"
+    },
+    {
+        "peer_name": "govuk-pay-test",
+        "account_id": "734876687119",
+        "vpc_id": "vpc-0170bee868cc6fb34",
+        "subnet_cidr": "172.17.40.0/22"
```
rjbaker, line 78: Should this not be 172.16.40.0/22

rorymalcolm: it should, thanks

paroxp (approved):

PR #2452, opened 2020-09-16
Make travis faster [citation needed]

tlwr:
What
----

We have one large monolithic travis job which does all the things.

This means all of our tests run in serial, and we wait a long time for things to install. If we experience a flake during a test, we have to re-run everything.

This PR does some questionable things to the Makefile, as well as `travis.yml` with the goal of

- caching dependencies
- avoid misusing `GOPATH`
- split out the `make test` single job into multiple jobs which can (in theory) be run in parallel
- steps can be re-run individually

How to review
-------------

Code review

Travis should pass

Travis should do what it is supposed to

Who can review
--------------

Not @tlwr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ env:
     - SHELLCHECK_VERSION="0.7.1"
     - TMPDIR="/tmp"
 
+cache:
+  directories:
+    - /home/travis/.rvm/
+    - /home/travis/gopath/pkg
+    - /home/travis/.cache/pip
+
 addons:
   apt:
     sources:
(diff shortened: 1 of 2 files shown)


46bit:
I'm a little scared we might end up having to fix the cache in travis one day (I've been there.) Is it possible to not cache and still be faster than what we have now?

tlwr:
Travis cache can be busted via the UI but I'll check my dev branch to see what the cache/non-cache time differential is

tlwr:
I have followed your advice Miki, and added a new commit which doesn't cache the dependencies, with the reasoning in the commit message.

I feel that leaving the two commits as they are (~"cache deps", ~"do not cache deps") is useful for a future ~archaeologist~ developer

46bit (approved):
> I feel that leaving the two commits as they are (~"cache deps", ~"do not cache deps") is useful for a future archaeologist developer

As a former digital archaeologist, I approve ⚒️

PR #2453, opened 2020-09-16
[#174773005] Make IPSec optional

tlwr:
[Story](https://www.pivotaltracker.com/n/projects/1275640/stories/174773005)

Please review this in conjunction with part two: https://github.com/alphagov/paas-cf/pull/2454

What
----

We are removing ipsec because the release is deprecated and we do not need it anymore because we have mtls between gorouter and apps (which satisfies the e2e encryption requirement)

IPSec has two levels `require` and `use`. Where `require` rejects all packets which are not encrypted from the CIDR ranges configured for IPSec and `use` makes IPSec optional

How to review
-------------

Run this down your pipeline, it should do nothing whatsoever

Who can review
--------------

Not @alphagov/tech-ops-contributors


tlwr:
Sorry to all the folks who were tagged, I typed "Not @tlwr" and hit enter before GitHub autocomplete could catch up

PR #2456, opened 2020-09-17
sqs: ensure fifo queues are operational

chrisfarms:
What
----

Updates platform tests for sqs to ensure both "standard" and "fifo" plans are tested.

FIFO queues have some unusual restrictions and we did not notice that "fifo" plans were broken.

How to review
-------------

:warning: Do not merge for now

Currently this branch should fail at platform-tests.

Wait until broker is fixed and this PR is ammended with the sqs-broker version bump.

Who can review
--------------

?


chrisfarms:
We are testing this in the broker's own integration tests, I'm not sure it is valuable to do it in both locations (especially since duplicating the platform-tests slows down the deployment pipeline)

PR #2461, opened 2020-09-22
Introduce custom error page

paroxp:
What
----

We'd like to be able to display a custom error message whenever gorouter
has an issue, like a 404 page for instance.

This is certainly possible, however it only accepts a Go template.

Our custom page, should be using the govuk-frontend node module, mainly
for styles.

We'd like to make sure that static page is easily upgradable and
maintainable, which is the reason of throwing in some NodeJS and React
into the mix. TypeScript is a must, Rafal insists on having.

The application should be used to upgrade the YAML patch upon upgrades
or changes.

How to review
-------------

- deploy to your dev env (or see tlwr whilst it lasts)
- visit `https://404.${SYSTEM_DOMAIN}/`
- experience GOV.UK branded `Not Found` page
- optionally, attempt to run the application (see README.md)
- validate the tests are enough check


kr8n3r:
can't speak for the approach itself, but from the FE point of view the page looks as it should

tools/error_page/src/layout.tsx
```diff
@@ -0,0 +1,149 @@
...
+                <a className="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
+              </div>
+            </div>
+          </div>
+        </footer>
+      </body>,
+    )}
+  </html>`;
+
+  return html.replace(/{{ .Header.Get &quot;X-Cf-RouterError&quot; }}/g, '{{ .Header.Get "X-Cf-RouterError" }}');
```
chrisfarms, line 148: this looks funky. Is there any possibility of this causing injection if I manually set that header somehow? or somehow influence what gets in it?

paroxp: No. This is only a template for Go to parse. So:

Go template reading: `{{ .Header.Get &quot;X-Cf-RouterError&quot; }}` will not be able to fill in the variable later. Form this point onward, this would be responsibility of Go templating language to escape HTML properly.

Does that make sense?

paroxp: As in, this node stuff is only used to create a Go template statically. It doesn't run node on the fly.

paroxp:
> can't speak for the generating s static page approach itself in relation to CF, but from the FE point of view the page looks as it should. does this add more weight to cf?

More weight in therms of CloudFoundry, not really. The Gorouter module has one extra field of configuration added containing HTML page.

In terms of paas-cf repo, it does have extra weight in the form of 16 added files. Which shouldn't be significant I wouldn't think...

tools/error_page/src/layout.tsx
chrisfarms: I think my concern is that you have html escaping, but in what looks like a `js` context. It looked like if I set "X-Cf-RouterError" to something like `'') (function(){ alert(1) })('` I could do something trixy.

It's a bit hard to grok, just thought it worth making sure weird input from that header had been considered :+1:

chrisfarms: Ah I think I just got what you meant ... heh

kr8n3r:
> > can't speak for the generating s static page approach itself in relation to CF, but from the FE point of view the page looks as it should. does this add more weight to cf?
> 
> More weight in therms of CloudFoundry, not really. The Gorouter module has one extra field of configuration added containing HTML page.
> 
> In terms of paas-cf repo, it does have extra weight in the form of 16 added files. Which shouldn't be significant I wouldn't think...

ok, I see. 
we're not shipping this to anyone.
Instead of manually have to create a static page you've created an app that generates in and that gets injected into the gorouter config

You mention that "currently the application has to be manually ran" 
I guess that's only when the page" templates" gets updates and the resulting HTML updated in https://github.com/alphagov/paas-cf/blob/24c8c98c2e41419a75cc0fbcac184e1848aa5c27/manifests/cf-manifest/operations.d/310-router-custom-error-page.yml#L3 and file checked in?

paroxp:
That would be correct @kr8n3r 

Say, you'd like to change some syntax, content or update the GOV.UK Frontend dependency. One may find it difficult to edit these things in a YAML file, thus an app that separates all these concerns and attempts to make it simpler to do so.

After such updates are made, one runs the application (process described in README.md file) to generate new YAML and commit.

This could/should be automated, however probably was out of scope for that story. Let's see if this proves to be a pain, unless the interest is expressed.

kr8n3r (approved):
cool. personally I'm ok with that.
not sure how much of risk it is to rely on github - perhaps we don't even need those assets (but we do need the favicon, to reliably have x-browsers, as not all support svgs)

tlwr:
https://raw.githubusercontent.com always serves files with the header `Content-Type: text/plain; charset=utf-8` which is not going to work for some browsers.

Would it be possible to build a page consisting just of what we need using some webpack awfulness, and inline that instead?

kr8n3r:
i _think_ utf-8 is supported in all the browser we support
I understand the transport not on cloudapps.digital but all our apps are on that domain as well :(

paroxp:
We will need `fonts`, `favicons` and one image I believe from somewhere.

I wen't with GitHub, as it has version control, and I didn't want to include assets from different version.

An alternative, could be using one of our pazmin urls, as the assets are exposed publically.

paroxp:
We've decided to simplify the HTML page by removing branding. This includes assets and some styling.

kr8n3r:
produces a smaller css
```
$govuk-font-family: "HelveticaNeue", "Helvetica Neue", "Arial", "Helvetica", sans-serif;

@import "govuk-frontend/govuk/settings/all";
@import "govuk-frontend/govuk/core/template";
@import "govuk-frontend/govuk/core/global-styles";
@import "govuk-frontend/govuk/components/skip-link/index";
@import "govuk-frontend/govuk/objects/grid";
@import "govuk-frontend/govuk/objects/main-wrapper";
@import "govuk-frontend/govuk/objects/width-container";
```
i guess we can also remove some conditional comments
` `

paroxp:
Applied.

Managed to save 5kb! :trollface: 

The shiv, was not even pointing to the right place. Removed.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3419,6 +3419,30 @@ jobs:
...
+
+                  TEST_PAGE="https://404.${SYSTEM_DNS_ZONE_NAME}"
+
+                  echo "Hitting ${TEST_PAGE}"
+                  HTML_CONTENT="$(curl "${TEST_PAGE}")"
+
+                  echo "Testing for HTML content..."
+                  grep 'class="govuk-header"' <(echo "${HTML_CONTENT}") > /dev/null
+                  grep 'Not Found' <(echo "${HTML_CONTENT}") > /dev/null
+                  grep 'Requested route (&#39;404.tlwr.dev.cloudpipeline.digital&#39;) does not exist.' <(echo "${HTML_CONTENT}") > /dev/null
```
tlwr, line 3444: This needs to be `((apps_domain))` or something

paroxp: addressed

kr8n3r (approved):
we're now using a non-govuk branded page, as per [service manual](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk)
which is modelled on the one from [gov.uk prototype kit](https://govuk-prototype-kit.herokuapp.com/docs/templates/blank-unbranded) with minimal styles from govuk-frontend.

as we've removed the dependency on github as cdn, I'm going to approve this.
Happy to have this dismissed of something else needs to change

tlwr:
⌛ tlwr pipeline

PR #2462, opened 2020-09-22
Give a bit more wiggle room on availability tests

LeePorte:
What
----

Our tests being at 100% were too stringent causing pipelines to become
blocked on one failure out of 14,000.

How to review
-------------

Code review

Who can review
--------------

Anyone

### DIFF
--- a/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go
+++ b/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go
@@ -18,11 +18,9 @@ import (
 const (
 	availabilityTestRate             = 10
 	availabilityTestMaxDuration      = 12 * time.Hour
-	availabilityTestMaxLatency       = 300 * time.Millisecond
-	availabilityTestMaxErrors        = 0
-	availabilitySuccessRateThreshold = 100.0
 	minimumTestDuration              = 15 * time.Second
 	maximumErrorRate                 = 0.5
+	availabilitySuccessRateThreshold = 100 - maximumErrorRate
 
 	vegetaRunForever = 0 * time.Second
 	vegetaKeepAlive  = true


paroxp (approved):
This is a weird check...

99.5 is a lot in a term of a day. But it actually doesn't count 99.5 % of time, but 99.5% of requests made since the test started...

I'm happy with this to be set to 99.5, but it's not the SLO, we'd ever be sharing with anyone...

a better test would be 9 out of 10 request have to pass and have the test "reset/reassess" on tick.

Let's merge this, but I'd like to propose a story to think about it more...

AP-Hunt:
Proposing this patch:

```
diff --git a/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go b/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go
index b3ba1992c..58bd47676 100644
--- a/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go
+++ b/platform-tests/availability/app/healthcheck_availability_during_deployment_test.go
@@ -18,11 +18,9 @@ import (
 const (
 	availabilityTestRate             = 10
 	availabilityTestMaxDuration      = 12 * time.Hour
-	availabilityTestMaxLatency       = 300 * time.Millisecond
-	availabilityTestMaxErrors        = 0
-	availabilitySuccessRateThreshold = 99.5
 	minimumTestDuration              = 15 * time.Second
 	maximumErrorRate                 = 0.5
+	availabilitySuccessRateThreshold = 1 - maximumErrorRate
 
 	vegetaRunForever = 0 * time.Second
 	vegetaKeepAlive  = true
```

`availabilitySuccessRateThreshold` and `maximumErrorRate` effectively described the same thing from what I could tell, and the former is just `1 - maximumErrorRate`.

AP-Hunt:
Above patch applied to the commit and force pushed.

PR #2465, opened 2020-09-23
[#171309836] Empty isolation segments

tlwr:
[Story](https://www.pivotaltracker.com/story/show/171309836)

What
----

[Isolation segments](https://docs.cloudfoundry.org/adminguide/isolation-segments.html) are a way of telling Cloud Foundry where to schedule apps, like affinity rules in Kubernetes, or tags in Concourse.

An isolation segment relates _orgs_ or _spaces_ to cells, so that:
- an org's apps can be scheduled on a group of cells
- a space's apps can be scheduled on a group of cells
- multiple orgs' apps can be scheduled on a group of cells
- multiple spaces' apps from multiple orgs can be scheduled on a group of cells

This PR is a cleaned up version of some code we wrote during a spike, with some tests.

It adds:
- a environment specific workflow for creating isolation segments
- empty (number of instances = 0) egress restricted isolation segments for staging/prod/prod-lon
- environment specific tests (for scale)
- general tests for isolation segments

An isolation segment (YAML file) can have the following properties:
- `number_of_cells`
- `isolation_segment_size` (either blank or `small_cell`)
- `restricted_egress` (boolean)
- `isolation_segment_name`

If an isolation segment is not egress restricted, it behaves like a normal diego cell. This is useful if we want to enable specific tenants to have specific instances, eg support for ARM CPUs, high CPU instances, etc

If an isolation segment is egress restricted then we:
- use [`silk-cni` `deny_networks`](https://bosh.io/jobs/silk-cni?source=github.com/cloudfoundry/silk-release&version=2.33.0#p%3ddeny_networks) to ensure that only the VPC is accessible. These deny rules are applied before security groups
- override the DNS server to use [a local coredns DNS server](https://coredns.io/) which masks BOSH DNS and ensures that only `apps.internal` DNS responses are returned.

Egress restricted isolation segments therefore allow our tenants to push apps which cannot talk to the internet, but can communicate with services within the VPC (eg RDS, Redis, other apps, etc). If they need to connect to a subset of the internet, then the tenant can run an egress proxy (eg Squid) as an internal app (outside of the isolation segment) to control that themselves.

The way this is implemented is _big-brain_ or _kind-of-hacky_ depending on your perspective. Cloud Foundry defines each isolation segment as a BOSH instance group, with a diego job property set. In order to reduce the operator burden of keeping all of our diego-cells in sync other than the isolation segment specific properties, we want to essentially copy the rendered isolation segment, with some overrides. To do this, we:

- render the manifest without any isolation segments
- extract the diego cell definition from the manifest
- rewrite the diego cell definition for the isolation segment
- append the new definition to the manifest's instance groups section
- repeat for the remaining isolation segments

The only complexity here is the management of BOSH links, for which we introduce some operations files.

Todo
------

- [x] Update coredns in `paas-observability-release`

How to review
-------------

Cherry pick these commits to your development

Scale up `diego-cell-egress-restricted-1` to 1 instance using BOSH (either via a pipeline or `bosh deploy`)

Create a space `cf t -o admin` `cf create-space isolated`

Create an isolation segment: `cf create-isolation-segment egress-restricted-1`

Enable the isolation segment for `admin`: `cf enable-org-isolation admin egress-restricted-1`

Isolate the `isolated` space: `cf set-space-isolation-segment isolated egress-restricted-1`

Target the space: `cf t -s isolated`

Push an app `cd paas-cf/platform-tests/example-apps/static-app && cf push isolated-app -b staticfile_buildpack`

Check the app has no egress: `cf ssh isolated-app` `dig example.com` `curl https://www.gov.uk`

Who can review
--------------

I paired on this with @schmie, neither of us should review


manifests/cf-manifest/test-config/acceptance_tests.yml
```diff
@@ -8,22 +8,25 @@ test_password: ((secrets_test_user_password))
...
-include_route_services: true
+include_tasks: true
+include_v3: true
 include_zipkin: true
 artifacts_directory: "/tmp/artifacts"
 use_existing_user: false
 include_container_networking: true
 name_prefix: ((name_prefix))
+include_isolation_segments: false
+isolation_segment_name: dev-1
```
46bit, line 29: Unsure if this really wants to be hardcoded to `dev-1`

tlwr: resolved in ffa078f

manifests/cf-manifest/scripts/generate-manifest.sh
```diff
@@ -3,33 +3,58 @@
...
-  --vars-file="${WORKDIR}/terraform-outputs/cf.yml" \
-  --vars-file="${PAAS_CF_DIR}/manifests/variables.yml" \
-  --vars-file="${ENV_SPECIFIC_BOSH_VARS_FILE}" \
-  --vars-file="${WORKDIR}/environment-variables.yml" \
-  --ops-file="${CF_DEPLOYMENT_DIR}/operations/use-internal-lookup-for-route-services.yml" \
-  ${opsfile_args} \
-  --ops-file="${WORKDIR}/vpc-peering-opsfile/vpc-peers.yml" \
-  --ops-file="${WORKDIR}/tenant-uaa-clients-opsfile/tenant-uaa-opsfile.yml" \
-  "${CF_DEPLOYMENT_DIR}/cf-deployment.yml"
+manifest_without_isolation_segments="$(
```
AP-Hunt, line 24: Is it possible to make this script any simpler/more readable? It's a pretty important part of our pipeline.

manifests/cf-manifest/scripts/generate-isolation-segment-ops-file.rb
```diff
@@ -0,0 +1,157 @@
...
+    .find { |job| job["name"] == "silk-cni" }[
+    "properties"][
+    "dns_servers"] = ["169.254.0.3"]
+
+  isolation_segment
+    .fetch("jobs")
+    .find { |job| job["name"] == "silk-cni" }[
+    "properties"][
+    "deny_networks"] = {
+      "running" => %w[
```
AP-Hunt, line 100: We should note why and what these networks are covering. CIDR blocks ending in `/3` and `/2` aren't that common I don't think?

AP-Hunt:
I don't know if this is the plan already, but I think we could do to go back over some of the commit messages and provide a bit more explanation and context on them. E.g. `3802eb2d745447533918d3ea182256fdf233ee32` could explain why CoreDNS needs that IP, and why we need to add it to the local interface.

tlwr:
@ap-hunt I have added two commits which add some (hopefully) descriptive comments for future readers. Would you mind having another read?

AP-Hunt:
> @AP-Hunt I have added two commits which add some (hopefully) descriptive comments for future readers. Would you mind having another read?

All looks good to me, thanks

AP-Hunt (approved):

46bit (approved):
:shipit:

PR #2466, opened 2020-09-24
Configure SQS broker to only permit access from local VPC

philandstuff:
What
----

Depends on https://github.com/alphagov/paas-sqs-broker/pull/15 and a
corresponding boshrelease to push that out.

This adds a new managed policy, local_vpc_access_only, which can be
attached to an IAM User or Role to restrict use of that User/Role to
only be permitted from the local VPC.  This mitigates against
credentials that are leaked.

The S3 broker has a similar feature in
s3_broker_ip_restriction_policy_arn, but I decided to go with an
approach based on `aws:SourceVpc` rather than `aws:SourceIp`.  This is
so that: a) the policy doesn't have to care about specific IP
addresses, and b) the policy will work whether the AWS API is accessed
over public internet or via a private endpoint (at some point in the
future).

I deliberately made the local_vpc_access_only policy's name generic
because it could easily be used as-is for other brokers in future.

How to review
-------------

I guess you could deploy it and verify that SQS-managed users get the appropriate policy restriction.

Who can review
--------------

Do I really have to say "not @philandstuff" for each PR? 😕

### DIFF
--- a/manifests/cf-manifest/operations.d/760-sqs-broker.yml
+++ b/manifests/cf-manifest/operations.d/760-sqs-broker.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: sqs-broker
-    version: 0.1.6
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/sqs-broker-0.1.6.tgz
-    sha1: 00f49f6597f6858a88c63825706a94a639a31dc0
+    version: 0.1.8
+    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/sqs-broker-0.1.8.tgz
+    sha1: 55c741e1c968238cf4c2f03df80e7bfde8f46a75
 
 - type: replace
   path: /addons/name=loggregator_agent/exclude/jobs/-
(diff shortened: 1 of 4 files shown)


tlwr:
> Do I really have to say "not @philandstuff" for each PR? 😕

I sometimes just say "Not me" 👼 

---

Does SourceVpc work if we don't have an SQS VPC Endpoint? [Amazon docs suggest that this is the case for S3](https://aws.amazon.com/premiumsupport/knowledge-center/block-s3-traffic-vpc-ip/) but I'm not sure if it also applies for SQS

philandstuff:
> I sometimes just say "Not me" 👼

Would you accept a PR to change the PR template to remove this dance entirely? 😄 

@tlwr the blog post you linked to talks about `aws:SourceVpce`, not `aws:SourceVpc` 😕 

but oh bums, [the docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-sourcevpc) seem to say:

> Availability – This key is included in the request context only if the requester uses a VPC endpoint to make the request.

which means this plan won't work 😭

philandstuff:
I've updated to use `aws:SourceIp` instead of `aws:SourceVpc`.

Does anyone know why the s3 ip restriction policy is only scoped to `s3:*` actions? I see it was introduced in 60d915d75580c38ec7afeb95171ad1f2960ceefb but not much was said in that commit.  In any case, I think we can have a much more general policy which can be reusable.

paroxp:
@philandstuff 

Personally I always remove that section of who can review. I'll be glad to approve you removing the section!

cbaines (approved):

PR #2470, opened 2020-10-06
alerts: lock held once must breach for duration

tlwr:
What
----

These alerts are noisy

The firehose exporter has a bit of latency built into it, and so if a deployment happens, then the locks always appear held twice, even though a swap has happened

adding a 10m caveat to these alerts will reduce noise and hopefully make us trust these alerts more

For example, the following should not alert us, as it is a deployment happening:
auctioneer locks held in London

How to review
-------------

Code review

CI

---

⚠️ Please do not merge this pull request via the GitHub UI ⚠️


46bit (approved):
Thank you

PR #2492, opened 2020-10-16
cf: scale london gorouters from 6 to 15

tlwr:
What
----

CPU pegged during tenant load test (GDS VPS) which caused 5xxs across
the whole platform: we need more gorouters during high load

gorouter process cpu usage

My interpretation of this chart is that something (it may not necessarily be CPU) is constrained during high periods of load, I believe that horizontally scaling will alleviate this problem

BOSH processes additionally have CPU limits

How to review
-------------

Code review

---


46bit (approved):

tlwr:
Cannot approve my own PR but 👍

PR #2494, opened 2020-10-19
Allow to pass a default into get_secrets

schmie:
What
----
Add an optional argument to the `get_secrets` function in order to pass a default (dummy) value to be used if the lookup in `paas-credentials` fails. Also set a dummy Mailchimp API key via this new route.

Why 
----
Some secrets, such as the Mailchimp API key, are only required in prod, but we still need to set a dummy value in staging and dev environments as it is expected to be non-nil during deployment/execution.

How to review
-------------
- Code review
- Check out the branch, run `CREDHUB (DEPLOY_ENV) $ make dev upload-all-secrets` and make sure it works as expected

---

### DIFF
--- a/scripts/upload-secrets/upload_secrets.rb
+++ b/scripts/upload-secrets/upload_secrets.rb
@@ -8,10 +8,13 @@ require "English"
 require "yaml"
 require "tempfile"
 
-def get_secret(path)
+def get_secret(path, default = nil)
   out = `pass "#{path}"`
-  abort out unless $CHILD_STATUS.success?
-  out
+
+  return out if $CHILD_STATUS.success?
+  return default.to_s unless default.nil?
+
+  abort out
 end
 
 def upload_secrets(credhub_namespaces, secrets)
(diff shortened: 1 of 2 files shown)


scripts/upload-secrets/upload_secrets.rb
```diff
@@ -8,10 +8,18 @@
 require "yaml"
 require "tempfile"
 
-def get_secret(path)
+def get_secret(path, *default_out)
```
tlwr, line 11: This could be a named argument instead, eg

```
def get_secret(path, default=nil)
```

then your else becomes

```
abort out if default.nil?

out.to_s
```

scripts/upload-secrets/upload_secrets.rb
```diff
@@ -8,10 +8,18 @@
...
   out = `pass "#{path}"`
-  abort out unless $CHILD_STATUS.success?
-  out
+  if $CHILD_STATUS.success?
```
tlwr, line 13: I would use `return` early to avoid nested conditionals

```
out = `pass "#{path}"`

return out if $CHILD_STATUS.success?
return default.to_s unless default.nil?

abort out

schmie: done and done : )

tlwr (approved):
The code looks good, thanks for those amendments

PR #2498, opened 2020-10-29
Ignore account notification events in AWS health metrics

AP-Hunt:
What
----
Account notification events appear to be for long running things like
deprecations or other AWS service changes. We don't really want these to be
appearing in our alerts, because they're going to be long running and noisy.

We care about short term, service impacting events.

How to review
-------------

1. See if you agree with the change in principal
2. Code review
3. [See if I haven't included an event category that I should have](https://docs.aws.amazon.com/health/latest/APIReference/API_EventFilter.html#AWSHealth-Type-EventFilter-eventTypeCategories)

---


46bit (approved):
I am surprised that deprecations don't come in as `scheduledChange`. This PR looks good though. 👍

PR #2500, opened 2020-10-29
[#175232600] Clarify just how "free" service plans are

AP-Hunt:
What
----

Service plans have a boolean property `free` which tells CloudFoundry if the
plan is available to trial orgs. Unfortunately, `cf marketplace -e SERVICE`
says "free" whenever that property is true. That's a bit misleading for
billable orgs, who DO have to pay for it.

We don't have influence over the format of the `cf marketplace` command, so
instead of correcting it be more nuanced, we introduce a clarifying statement
in the description of the "free" plans.

How to review
-------------

1. Check you agree with the message.

---

### DIFF
--- a/config/service-brokers/aiven/config.json
+++ b/config/service-brokers/aiven/config.json
@@ -30,7 +30,7 @@
             "name": "tiny-6.x",
             "aiven_plan": "startup-4",
             "elasticsearch_version": "6",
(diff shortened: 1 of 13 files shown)


kr8n3r (approved):
words are good. we'll see if there is confusion later
:shipit:

PR #2504, opened 2020-11-03
Increase number of cells in Ireland

paroxp:


AP-Hunt (approved):
image
Based on this chart, 45 is a good number.

PR #2512, opened 2020-11-12
Pipecleaner supports set pipeline

tlwr:
What
----

Upgrade the version of Concourse dependency to 6.7.1

Support `set_pipeline` when validating Concourse pipelines

How to review
-------------

Code review

CI

---


richardTowers (approved):
Coincidentally, I ended up looking at this same issue. I ended up with pretty much the same solution (which is more or less the obvious way of dealing with the interface changes between concourse 4 and concourse 6). I think that counts as a decent code review. Nice to see the extra integration test to cover the new behaviour too.

I've also built pipecleaner from this branch and tried it out on some of GOV.UK's pipelines. Seems to work nicely, and fixes the issue where the current version whinges about set-pipeline steps because it doesn't understand them.

I'll let someone from the PaaS team do the merge :)

PR #2527, opened 2020-11-24
Scale Notify staging isolation segment to 24 cells

46bit:
What
----

We recently created an isolation segment for GOV.UK Notify. They are using it to conduct load testing against their staging environment without affecting the rest of the PaaS.

We chose 12 cells to start with. This was based on the maximum memory usage if their staging environment autoscaled all the way up. We already suspected they might run low on CPU resources. They have now run more load tests and we can see that the isolation segment's cells all went up to 100% CPU.

This commit doubles the isolation segment's size from 12 cells up to 24. This will provide twice as much CPU.

An alternative approach is to use an EC2 instance type that has more vCPU cores. We may still do that, and it is cheaper (at least if we ignore existing Reserved Instances.) For now though we want load testing results that broadly will apply to Notify's production environment, using the normal cells.

How to review
-------------

Code and justification review.

---

### DIFF
--- a/manifests/cf-manifest/isolation-segments/prod/govuk-notify-staging.yml
+++ b/manifests/cf-manifest/isolation-segments/prod/govuk-notify-staging.yml
@@ -1,3 +1,3 @@
 ---
-number_of_cells: 12
+number_of_cells: 24
 isolation_segment_name: govuk-notify-staging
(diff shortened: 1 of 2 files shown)


schmie (approved):
:+1:

PR #2530, opened 2020-11-26
[#175819180][Play after PR#2528] Remove update mode from certificate

AP-Hunt:
What
----
Removes `update_mode: converge` from certificate type variables so that we don't have a foot cannon in that area. We added it to begin with so that we could force the update of certificates as a result of a change to their properties; in this case adding subject alternative names.

We don't want that behaviour to persist in future. We should opt in to it by adding the flag back.

How to review
-------------
1. Check it fully reverts the `update_mode: converge` additions from #2528
2. Deploy to your dev env after #2528 (or later) has been run down your env

---

### DIFF
--- a/manifests/cf-manifest/operations.d/030-bosh-dns.yml
+++ b/manifests/cf-manifest/operations.d/030-bosh-dns.yml
@@ -56,7 +56,6 @@
   value:
     name: dns_healthcheck_server_tls
     type: certificate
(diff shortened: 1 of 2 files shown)


mogds (approved):

AP-Hunt:
We didn't ever bother with this and nothing catastrophic has happened 🤷

PR #2545, opened 2020-12-14
change notify isolation-segment instance type

LeePorte:
What
----

Notify would like to test their app on instances with more CPU whilst keeping the memory the same

How to review
-------------

Code review

---

### DIFF
--- a/manifests/cloud-config/paas-cf-cloud-config.yml
+++ b/manifests/cloud-config/paas-cf-cloud-config.yml
@@ -212,6 +212,17 @@ vm_types:
     - ((terraform_outputs_elasticache_broker_clients_security_group))
     - ((terraform_outputs_default_security_group))
 
+- name: high_cpu_cell
+  cloud_properties:
+    instance_type: ((high_cpu_cell_instance_type))
+    ephemeral_disk:
+      size: 204800
+      type: gp2
+    security_groups:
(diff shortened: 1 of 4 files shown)


manifests/cloud-config/paas-cf-cloud-config.yml
```diff
@@ -212,6 +212,17 @@ vm_types:
     - ((terraform_outputs_elasticache_broker_clients_security_group))
     - ((terraform_outputs_default_security_group))
 
+- name: notify_cell
```
paroxp, line 215: Can we not call them notify cells? it's a type and more generic one could be useful elsewhere?

AP-Hunt: +1

Naming it for what it is (a cell type with more CPU power) would be better

paroxp (commented):
In general, I'd rather see something like: `c5.2xlarge` type and call it something along the lines of: `2xlarge_compute_cell`

manifests/cloud-config/paas-cf-cloud-config.yml
LeePorte: I was going for the Ronseal approach whilst testing was happening

paroxp (approved):

PR #2548, opened 2020-12-18
Add 6 extra routers in Ireland to support Notify load tests

AP-Hunt:
What
----
Notify are load testing at almost 2000 req/s, and causing all of the routers to sit at 70%+ CPU usage. Adding 6 more should take the pressure off a bit.

How to review
-------------
Check I did my maths right, and that it's a multiple of 3.

Tell me if you disagree with the approach.

---


kr8n3r (approved):
it is in fact a multiple of 3

PR #2549, opened 2020-12-18
Add notify production isolation segment

idavidmcdonald:
Mimics the segment given to their staging environment after load tests
on the staging segment have seen a significant impact in terms of
performance.

The plan is hopefully to only keep these around for a few months and pay
the cost of that whilst we bring things under control and can look at
reducing our CPU usage in a more long term manner.

### DIFF
--- a/manifests/cf-manifest/isolation-segments/prod/govuk-notify-production.yml
+++ b/manifests/cf-manifest/isolation-segments/prod/govuk-notify-production.yml
@@ -1,0 +1,4 @@
+---
+number_of_cells: 24
+isolation_segment_name: govuk-notify-production
+vm_type: high_cpu_cell


manifests/cf-manifest/isolation-segments/prod/govuk-notify-production.yml
```diff
@@ -0,0 +1,4 @@
+---
+number_of_cells: 24
```
idavidmcdonald, line 2: This is the same number of cells in our staging segment. Load tests have been good on that. One thing I've considered, but think we might be OK without is to add a few extra cells in production, just in case. Then after a few days of seeing that things are fine we could bring it down to 24 if appropriate.

Any metrics on your side to help support whether 24 is a good enough number would be useful but at the moment I feel reasonably confident.

AP-Hunt (approved):

PR #2554, opened 2020-12-24
billing: Update database

paroxp:
What
----

We'd like to deprecate the older database versions, meaning, along with our tenants, we need to move on with our lives.

The idea here is to run the update manually once on all of the environments, and have this merged as oppose to over-engineering and future proofing it.

How to review
-------------

- Sanity check

---


46bit (approved):
Agreed that it's probably not worth pipelining the actual upgrade, and this change LGTM.

PR #2556, opened 2021-01-04
Pin cell counts chart to default cell segment only

AP-Hunt:
What
----

Recently introduced and reverted a change that differentiated cell memory
metrics by isolation segment. Now that it's reverted, new metric values are no
longer differentiated, but old ones are. In order to have the cell counts chart
display properly, we need to explicitly look at the `diego-cell` segment.

How to review
-------------
1. Check I got the expression right
2. Compare it [a modified version in prod](https://grafana-1.cloud.service.gov.uk/d/HVQu5D-Mk/andy-user-impact-prod?orgId=1&refresh=5s)

---

### DIFF
--- a/manifests/prometheus/dashboards.d/user-impact.json
+++ b/manifests/prometheus/dashboards.d/user-impact.json
@@ -502,7 +502,7 @@
                 "stack": false,
                 "steppedLine": false,
                 "targets": [{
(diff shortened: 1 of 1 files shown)


46bit (approved):
LGTM

PR #2559, opened 2021-01-06
Configure GitHub actions to run tests on pull requests

AP-Hunt:
What
----
Travis is getting awfully slow, and it's slowing down our deployments by quite a lot. Let's get the tests running on GitHub Actions

How to review
-------------
1. Check all the tests pass
2. Code review
3. Make sure I didn't miss anything from the original Travis config

---


.github/workflows/test_on_pr.yml
```diff
@@ -0,0 +1,87 @@
+on: pull_request
+
+env:
+  TF_VERSION: "0.13.2"
```
LeePorte, line 4: we're now using 0.14.4 in the containers and in account wide

LeePorte (changes requested):
Looking good, only change is the terraform version which has been upgraded since this work was started

LeePorte (approved):

PR #2562, opened 2021-01-08
Prevent Redis plans from being updated

AP-Hunt:
What
----

The broker stops plans from being updated [1], but the broker says they're allowed. This results
in a `cf update-service -p` call leading to a Redis service instance in a `failed` state. Rather
than that, we should make it clear that the plans can't actually be updated.

[1] https://github.com/alphagov/paas-elasticache-broker/blob/master/broker/broker.go#L212

How to review
-------------

1. Run this down your env
2. Check it prevents you from updating the plan of a Redis service instance, without getting in to a failed state

---

### DIFF
--- a/manifests/cf-manifest/operations.d/730-elasticache-broker.yml
+++ b/manifests/cf-manifest/operations.d/730-elasticache-broker.yml
@@ -120,7 +120,7 @@
                     - elasticache
                     - redis
                   bindable: true
-                  plan_updateable: true
+                  plan_updateable: false
                   plans:
                     # Deprecated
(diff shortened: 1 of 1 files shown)


46bit:
Deploying this in my `miki` dev env now.

46bit:
Still slightly ugly output, but at least it's very clear:

```
$ cf update-service test-redis -p large-ha-5.x
Updating service instance test-redis as miki.mokrysz@digital.cabinet-office.gov.uk...
FAILED
Server error, status code: 400, error code: 110004, message: The service does not support changing plans.
```

And the service is still in `status: create succeeded` :)

46bit (approved):

PR #2563, opened 2021-01-11
Update terraform 14

LeePorte:
What
----

Update docker containers to the latest hash, this upgrades us to terraform 0.14.4

How to review
-------------

Code review and deploy to dev env or check out https://deployer.leeporte.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-terraform/builds/36

---


paroxp (changes requested):
There is an extra space in front of these tags. This _should_ not cause a problem, but I am not certain if we'd like to maintain this state, just in case.

LeePorte:
> There is an extra space in front of these tags. This _should_ not cause a problem, but I am not certain if we'd like to maintain this state, just in case.

Fixed it

paroxp (approved):

PR #2565, opened 2021-01-12
Add 3 cells to London

AP-Hunt:
What
----

We've been seeing a lot of HTTP 502s coming out of GoRouter recently. We think that's caused by apps not responding in time.

We're seeing high CPU usage in around 1 in 8 of the cells (memory consumption remains fine). We think this could be the cause of apps not responding in time. Adding more cells provides more CPU cores to carry out the work of the apps.

In 7 days we'll check back on the rate of HTTP 502s to see if adding cells had a positive impact.

## CPU usage across all cells in the last 7 days
image

## Number of cells whose 1m average CPU usage is above 50% over the last 7 days
image

How to review
-------------
1. Do you agree with the rationale?
1. Do you agree with the number of cells being added?
---

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -2,7 +2,7 @@
 uaa_instances: 3
 nats_instances: 3
 diego_api_instances: 3
-cell_instances: 81
+cell_instances: 84
 router_instances: 15
 api_instances: 12
 doppler_instances: 39


46bit (approved):
This can help a bit, but app instances running on the busy cell will need to be redeployed before there's a substantial improvement.

46bit:
This might be a good example of where CPU Entitlement would help–minimising the effect of high CPU apps without crippling them. https://www.cloudfoundry.org/blog/better-way-split-cake-cpu-entitlements/

Not saying do that here though! Would need research and experimentation.

PR #2567, opened 2021-01-13
Fix some incorrect behaviour when run on Linux

whi-tw:
What
----

Fixed some incorrect behaviour when run on linux

* TMPDIR was being set without a trailing slash, and mktemp could not
  create `/tmpcf_home.XXXXXX`
* `open` does not exist on linux. test if it's available, and if not,
  use `xdg-open` instead.

How to review
-------------

Use `scripts/cf_subshell_scoped_login.sh` on a mac to ensure it still functions correctly.

---

### DIFF
--- a/scripts/cf_subshell_scoped_login.sh
+++ b/scripts/cf_subshell_scoped_login.sh
@@ -31,7 +31,7 @@ case $TARGET in
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
Looks solid to me, and doesn't break when I use it on Mac.

PR #2583, opened 2021-01-26
[#176574936] Remove `postgis` from default Postgres extensions

AP-Hunt:
What
----

We believe that the `postgis` extension is causing problems when upgrading
Postgres major versions. We're making a change the RDS broker to force it to be
disabled when upgrading, but in order to do that the extension has to be
removed from the list of defaults.

How to review
-------------

1. Make sure I didn't miss anywhere
2. See if it [deployed OK in my environment](https://deployer.andyhunt.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry)
3. See if you can enable and disable the `postgis` extension on a Postgres database

---


46bit (approved):
I'm pretty sure I've already deployed and tested this, when I did the force-push above. Approved :)

PR #2627, opened 2021-03-15
[#177267449] Remove Postgres 9.5 plans from RDS broker configuration

AP-Hunt:
What
----
Removes Postgres 9.5 plans from RDS broker configuration.

Extracts the Postgres service configuration from the Postgres 9.5 plans file, and does the same thing to the MySQL service configuration so that they match.

By convention, the RDS broker configuration has service offerings and plans in separate files, with plans separated by version and high iops-ness.

Billing information is retained, so that billing isn't broken.

How to review
-------------
1. Run it down your pipeline and see how it goes
1. Check that billing isn't emitting any errors related to Postgres 9.5 plans

🚨 Caution🚨  
-------------
This should not be deployed until have zero Postgres 9.5 databases left, because removing the plans from the catalog means the broker can't find the plan to determine if an upgrade path is valid.

---


paroxp (approved):
> This should not be deployed until have zero Postgres 9.5 databases left, because removing the plans from the catalog means the broker can't find the plan to determine if an upgrade path is valid.

We're past the point of no return.

PR #2629, opened 2021-03-17
Enlarge Firehose VMs to avoid exhausted CPU credits

46bit:
What
----

We have "Firehose" VMs that run the [Firehose Exporter](https://github.com/bosh-prometheus/firehose_exporter).

Until now our Firehose VMs have used the t3.medium instance type. But in London they have been experiencing exhausted CPU credits for months now. CPU usage is not very high but they use just a bit more CPU than they have CPU credits. Moving them to m5.large will resolve this by giving them consistent amounts of CPU.

Here's a Firehose in London:

Screenshot 2021-03-17 at 11 18 31

Here's a Firehose in Ireland:

Screenshot 2021-03-17 at 11 18 23

You can see that the CPU credit usage isn't all that much higher in London, but it's enough to exhaust them. I think it's better to just move all Firehose VMs rather than just doing some.

How to review
-------------

Code review should be good enough. I'm pretty confident the `path` is correct based on https://github.com/alphagov/paas-cf/blob/b8763f4bd04cb5151d1a8ce822378f6536ede39a/manifests/prometheus/operations.d/400-set-instance-count.yml#L32

---

### DIFF
--- a/manifests/prometheus/operations.d/221-monitor-cf.yml
+++ b/manifests/prometheus/operations.d/221-monitor-cf.yml
@@ -5,7 +5,7 @@
 
 - type: replace
   path: /instance_groups/name=firehose/vm_type
-  value: medium
+  value: large
 
 - type: replace
   path: /instance_groups/name=firehose/jobs/name=firehose_exporter/properties/firehose_exporter/doppler/idle_timeout?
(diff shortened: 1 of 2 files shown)


46bit:
Actually this patch needs to be in the prometheus manifest, not the CF one.

46bit:
Finally figured this out! Not sure how I missed the correct YAML location when searching.

paroxp (approved):

PR #2641, opened 2021-04-14
Reduce number of targets to one

paroxp:
What
----

We couldn't do this before, as some of the variables in `dev` target are
generated based on DEPLOY_ENV value.

We can override those to however, making it all seemless and making Make
happy with a single target at all times...

How to review
-------------

- Sanity check

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,15 +220,19 @@ dev: ## Set Environment to DEV
 	@true
 
 .PHONY: dev01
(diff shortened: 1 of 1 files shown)


0atman (approved):
Thank you @paroxp !

AP-Hunt (approved):
This adds a small maintenance burden, but honestly it's kinda hypothetical anyway since I don't think those domains have changed in 5 years.

Let's do it.

PR #2644, opened 2021-04-19
Increase London's diego cell count to 87

AP-Hunt:
What
----
Utilisation has been steadily increasing over the last 30 days

image

Raises from 84 to 87.

How to review
-------------
Check it's a multiple of 3.

---

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -2,7 +2,7 @@
 uaa_instances: 3
 nats_instances: 3
 diego_api_instances: 3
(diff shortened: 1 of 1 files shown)


paroxp (approved):

paroxp:
It's a shy headroom being proposed. Will work, but means we're revisit this file in about a month? :D

```js
function increased(current) {
  return current + Math.round(current * 0.1/3.0) * 3;
}

const new = increased(84); // 93
```

AP-Hunt:
I think it's a sufficiently small change and low effort PR that I'm fine with that amount of toil versus the cost of running more cells than we need.

PR #2663, opened 2021-05-13
Allow the continuous deployment of PaaS admin to be disabled

AP-Hunt:
What
----

PaaS admin is usually re-deployed every time a new release is detected. In some environments, we don't want this to happen, because it wrecks developer productivity, or overwrites a deliberate change that is being tested.

Running the below will now disable the continuous deployment in an environment

  make ENV pipelines ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY=false

How to review
-------------

See if you agree with the changes

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,7 @@ dev: ## Set Environment to DEV
 	$(eval export ENABLE_MORNING_DEPLOYMENT=true)
 	$(eval export SLIM_DEV_DEPLOYMENT ?= true)
 	$(eval export CA_ROTATION_EXPIRY_DAYS ?= 360)
+	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
 	@true
 
 .PHONY: dev01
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -217,6 +217,7 @@ dev: ## Set Environment to DEV
 	$(eval export ENABLE_MORNING_DEPLOYMENT=true)
 	$(eval export SLIM_DEV_DEPLOYMENT ?= true)
 	$(eval export CA_ROTATION_EXPIRY_DAYS ?= 360)
+	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
```
0atman, line 220: does this work!? I thought you couldn't `export` envs to the calling shell from within a makefile, because it's running its own subshel?

paroxp (approved):
Fancy! I wonder if we should perhaps disable it for dev by default though ;)

0atman (approved):
apart from my astonishment about the `eval export` stuff, this makes sense :ok_hand:

Makefile
AP-Hunt: Doesn't export it to the calling shell, I don't think. It exports it for all of the scripts run by Make.

AP-Hunt:
> Fancy! I wonder if we should perhaps disable it for dev by default though ;)

Possibly. Let's let it run as is for now, and we can add that on if we feel the need.

Makefile
0atman: aaah, all subshells. That totally makes sense now.

PR #2673, opened 2021-06-03
Remove the `paas_currency_real_ratio` metric

AP-Hunt:
What
----

The API we used to get the value is now behind a pay wall. We haven't had data for the metric since 4th April, and nobody has noticed, so we can safely say we don't need it.

image

How to review
-------------

Make sure I didn't miss anything

---


paroxp (approved):
Don't we have some alerts based on this?

AP-Hunt:
> Don't we have some alerts based on this?

I couldn't find any:
```
$ echo $PWD; git grep -r 'paas_currency_real_ratio'; echo $?
...github.com/alphagov/paas-cf
1
```

46bit:
Not since https://github.com/alphagov/paas-cf/pull/2147

paroxp:
Thanks Miki!

PR #2683, opened 2021-06-15
[#178302212] create-cloudfoundry pipeline: add `DISABLED_AZS` variable 

risicle:
What
----

This variable takes a space separated list of names of AZs that should be omitted from cf's `instance_groups`. AZ names are expected to be as understood by bosh, e.g. `z2`.

Shutting down an AZ at this level should cause bosh to reallocate the desired number of resources across the AZs that are still up, rather than just tolerating operation at a lower capacity which is what would happen using the VPC-level disabling mechanism. Normal state is expected to be unset/empty string.

Related things _also_ done as part of this PR:

 - Renamed the existing `disable-az-*` jobs to `disable-az-*-vpc` to make clear what the difference is between this and the new mechanism. Renaming done with an `old_name` so history etc should be retained.
 - Prevented the `disable-az-*-vpc` jobs from disabling the VPC's `infra` subnet, as it turns out resources without redundancy could be on this subnet. In our case it was concourse's RDS db, which of course made concourse non-operational.
 - Disable the bosh resurrector as part of the `disable-az-*-vpc` jobs to prevent it fruitlessly attempting to fix the instances in the disabled AZ and continually occupying locks in doing so. Open question here about what we should do about re-enabling the resurrector. Should the `enable-az-*-vpc` jobs turn it back on? Or would it be better to just have some gentle alerting going all the time that the resurrector's off to remind people to re-enable it once the incident's over?

How to review
-------------

 - Firstly bring up an environment with `SLIM_DEV_DEPLOYMENT=false` so that you have all AZs up.
 - Trigger the `disable-az-c-vpc` job from the pipeline's `operator` page.
 - Set the pipeline's `DISABLED_AZS` to `z3` & re-trigger the deployment.
 - Watch the deployment go out, moving resources to the remaining AZs. Eventually the desired number of resources should be reached.
 - Trigger the `enable-az-c-vpc` job from the pipeline's `operator` page.
 - Set the pipeline's `DISABLED_AZS` back to an empty string & re-deploy.

If you're feeling particularly adventurous you could try disabling multiple AZs.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,6 +218,7 @@ dev: ## Set Environment to DEV
 	$(eval export SLIM_DEV_DEPLOYMENT ?= true)
 	$(eval export CA_ROTATION_EXPIRY_DAYS ?= 360)
 	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
+	$(eval export DISABLED_AZS)
 	@true
 
 .PHONY: dev01
@@ -258,6 +259,7 @@ stg-lon: ## Set Environment to stg-lon
 	$(eval export AWS_REGION=eu-west-2)
 	$(eval export CA_ROTATION_EXPIRY_DAYS=335)
 	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
+	$(eval export DISABLED_AZS)
 	@true
 
 .PHONY: prod
@@ -282,6 +284,7 @@ prod: ## Set Environment to Production
 	$(eval export AWS_REGION=eu-west-1)
 	$(eval export CA_ROTATION_EXPIRY_DAYS=30)
 	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
+	$(eval export DISABLED_AZS)
 	@true
 
 .PHONY: prod-lon
@@ -306,6 +309,7 @@ prod-lon: ## Set Environment to prod-lon
 	$(eval export AWS_REGION=eu-west-2)
 	$(eval export CA_ROTATION_EXPIRY_DAYS=30)
 	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
+	$(eval export DISABLED_AZS)
 	@true
 
 .PHONY: bosh-cli
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -206,12 +206,12 @@ groups:
       - healthcheck-check-state
       - healthcheck-disable
       - cleanup-deleted-cf-users
-      - disable-az-a
-      - enable-az-a
-      - disable-az-b
-      - enable-az-b
-      - disable-az-c
-      - enable-az-c
+      - disable-az-a-vpc
+      - enable-az-a-vpc
+      - disable-az-b-vpc
+      - enable-az-b-vpc
+      - disable-az-c-vpc
+      - enable-az-c-vpc
   - name: tests
     jobs:
       - app-availability-tests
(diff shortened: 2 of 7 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1569,6 +1569,7 @@ jobs:
               - |
                 echo "RUNNING THIS ACTION FOR AZ-A, WILL MEAN THIS CONCOURSE IS BEING USELESS!"
                 echo "You will need to run the 'revert' action by hand, to get it back to working state."
+                # ^ is this still true?
```
AP-Hunt, line 1572: Shouldn't be true any more, I don't think.

Makefile
```diff
@@ -218,6 +218,7 @@ dev: ## Set Environment to DEV
 	$(eval export SLIM_DEV_DEPLOYMENT ?= true)
 	$(eval export CA_ROTATION_EXPIRY_DAYS ?= 360)
 	$(eval export ENABLE_PAAS_ADMIN_CONTINUOUS_DEPLOY ?= true)
+	$(eval export DISABLED_AZS)
```
AP-Hunt, line 221: How should this be used at the command line, given the input is space separated? As below?

```
make ENV pipelines DISABLED_AZS="z1 z2"
```

AP-Hunt:
We should definitely look at how to alert on 

1. the resurrector being offline 
2. AZ count being reduced

as a way to catch us forgetting to re-enable things

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -1786,6 +1822,24 @@ jobs:
                 ruby paas-cf/concourse/scripts/extract_tf_vars_from_terraform_state.rb \
                 < vpc-tfstate/vpc.tfstate > terraform-variables/vpc.tfvars.sh
 
+      - task: disable-bosh-resurrector
```
AP-Hunt, line 1832: We should consider how and when to re-enable the resurrector

concourse/pipelines/create-cloudfoundry.yml
risicle: Cool I'll remove

Makefile
risicle: Bingo.

concourse/pipelines/create-cloudfoundry.yml
risicle: Yeah you can see up above I've had some thoughts on it. Thought it might be a bit of a footgun having the `enable-az` jobs blindly turn it back on. I think the answer might be alerting - should I create a story for that do you think?

Makefile
AP-Hunt: Confirmed it worked for my own sanity

concourse/pipelines/create-cloudfoundry.yml
AP-Hunt: I'm not convinced that `bosh deploy` doesn't turn it back on to be honest, so we might be having a moot conversation. We should check it.

AP-Hunt: But also yes, let's create a story for the alerting. IMO we're better off getting this out there and then expanding it, over trying to get everything in place. An availability zone could go away tomorrow.

risicle:
re alerts... perhaps an unconventional idea, but has anyone ever used a periodic concourse task that simply pokes something and passes/fails depending on its status just for the purpose of providing a red/green light on the pipeline?

AP-Hunt:
We've not done that before, no. I'm not sure it would work for us as a set of engineers, because of how distributed we are. 

We shouldn't get noise for the N environments we run, only the 1 environment we bother to disable an AZ in: `prod` or `prod-lon`, so I think alerting through the normal channels would be just fine.

risicle:
Wait - it's not that unconventional - that's just a smoke test that's not hooked up to an alert.

AP-Hunt:
> Wait - it's not that unconventional - that's just a smoke test that's not hooked up to an alert.

Oh yeah we do that with our smoke tests already. I thought you meant a point-your-eyes-at-the-pipeline type of alert.

risicle:
Well, you get the point-your-eyes-at-a-pipeline alert for free if you just don't give it slack creds...

PR #2692, opened 2021-07-01
Increase number of doppler instances

paroxp:
What
----

We're starting to see dropped messages from the system. The rule of
thumb is that we keep the ratio of cells to dopplers around 2:1.

This increase should suffice for a short while.

Ireland does not grow, and we didn't see issues thus far. If anything,
it is generously provisioned.

How to review
-------------

- Sanity check

---


LeePorte (approved):

schmie (approved):
:+1: LGTM - we might also decrease the number of diego-cells at the moment, but the usage goes up steadily, so I am not sure how valuable that is.

PR #2698, opened 2021-07-06
Make instance count changes

paroxp:
What
----

> Traffic Controller resources are usually scaled in line with Doppler
> resources. The recommended formula for determining the number of Traffic
> Controller instances is:

> Number of Traffic Controller instances = Number of Doppler instances / 2

As we do not work in single instances, we increase things in pairs of threes.

We're also refining the tests for number of dopplers... As we seem to be
just on the edge and to incorporate the following:

> Doppler resources can require scaling to accommodate your overall log
> and metric volume. The recommended formula for estimating the number of
> Doppler instances you need to achieve a loss rate of < 1% is:

> Number of Doppler instances = doppler.ingress / 16 000
> where doppler.ingress represents the per-second rate of change of the
> ingress metric from the doppler source, summed over all VMs in your
> deployment.

Now, this isn't as simple as our rule of thumb ratio 2:1 with cells. But
is fairly close...

After examining the logs, we're just on the edge after recent changes,
missing that one extra doppler...

Along with the refined tests, I'm increasing the counts.

In addition, We'd like to eliminate the flakyness of the tests, by throwing
more resource at the system, hopefully handling more actions, more
promptly.

This is a finger in the air change, that should be small enough but may
be informative to our investigation into flaky smoke tests.

How to review
-------------

- Sanity check

---

### DIFF
--- a/manifests/cf-manifest/spec/manifest/env_spec.rb
+++ b/manifests/cf-manifest/spec/manifest/env_spec.rb
@@ -1,3 +1,7 @@
+def round_up(value, increment)
+  increment * ((value + increment - 1) / increment)
+end
+
 RSpec.describe "Environment specific configuration" do
   let(:default_manifest) { manifest_without_vars_store }
   let(:prod_manifest) { manifest_for_env("prod") }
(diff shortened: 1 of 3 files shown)


mogds (approved):

manifests/cf-manifest/spec/manifest/env_spec.rb
```diff
@@ -41,14 +41,15 @@ def get_instance_group_instances(manifest, instance_group_name)
...
 
       describe "doppler" do
-        it "instance count should be at least 1:2 with cell count" do
+        it "instance count should be at least half of the cell count" do
           doppler_instance_count = env_manifest.fetch("instance_groups.doppler").dig("instances").to_f
           cell_instance_count = env_manifest.fetch("instance_groups.diego-cell").dig("instances").to_f
 
-          ratio = cell_instance_count / doppler_instance_count
+          half = cell_instance_count / 2
+          half_with_headroom = half + 3
```
AP-Hunt, line 49: I think this should be half + AZ count

AP-Hunt: Or maybe (half + AZ count) rounded to the nearest multiple of AZ count

AP-Hunt (commented):
Yeah I like this. Just the question about giving headroom in terms of AZ count, rather than hardcoded to 3.

manifests/cf-manifest/spec/manifest/env_spec.rb
paroxp: Yeah, you're right!

paroxp: Added round_up :)

PR #2718, opened 2021-08-05
[#177724957] elasticache broker: add plans for redis 6.x

risicle:
What
----

Config mostly cloned from 5.x plans except we no longer have any control over minor redis versions.

Here I have also switched the common parameter `reserved-memory` to `reserved-memory-percent`, which seems to be the newer more broadly supported form of the parameter, because it seems like `redis6.x` has quietly dropped support for `reserved-memory`. Still, since this is universally set to 0, surely 0% of X = 0, right?

Thoughts welcome.

How to review
-------------

Deploy. Try to provision a redis 6.x instance.

---

### DIFF
--- a/config/billing/config-parts/redis_pricing_plans.json.erb
+++ b/config/billing/config-parts/redis_pricing_plans.json.erb
@@ -136,6 +136,64 @@
 	  "guid" => "f60443d9-3e9f-4c9b-8653-c83b6418292e",
 	  "ha" => true,
 	  "aws_price" => price("elasticache_m5_2xlarge")
+	},
+
+	# Redis 6.x
+	{
+	  "name" => "micro-6.x",
+	  "valid_from" => "2021-08-01",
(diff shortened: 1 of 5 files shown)


paroxp (approved):
Looks good. Pazmin marketplace steady, used dev02 to setup an instance and tear it down.

risicle:
Cool - you don't think there'll be any issues with existing redis instances do you? I was assuming that the parameter groups were only touched on service creation.

paroxp:
I think you are correct. The existing services should remain intact. Though if we're _really_ worried... We could raise a support ticket with AWS to double check?

risicle:
Oh I've also forgotten to re-run the billing `generate_config.rb` script

paroxp:
Yeah, there are entries missing from `config/billing/output/${AWS_REGION}.json`. Good spot!

paroxp (approved):

PR #2721, opened 2021-08-12
[#178640787] rds broker: add plans for postgres 13, shake out extensions

risicle:
What
----

Added postgres 13 plans for the RDS broker based on the postgres 12 plans, omitting the deprecated `large` plans.

Updated billing config to match.

Updated the custom broker acceptance tests to expect the new plans and to use a postgres 13 plan when creating a test instance.

Updated the set of available extensions to match the current set listed by AWS for 13. This includes some that I'm uncertain whether they can actually be used by our tenants with the permissions we give them.

Switched *all* plans `postgresql-hll` extension to be named simple `hll` as it appears `postgresql-hll` was _never_ the name to access it by, even in postgres 10, so creating a database with this extension would always fail. Fortunately this means we shouldn't have any existing databases with this extension to worry about migrating, right?

How to review
-------------

Deploy to dev paas. Check the custom broker acceptance tests pass. Try creating some databases, perhaps with some unusual extensions.

---


risicle:
Update: going to significantly cut down the selection of allowed extensions for pg13, but first need to make the RDS broker aware of pg13 for shared preload handling: https://github.com/alphagov/paas-rds-broker/pull/139

manifests/cf-manifest/operations.d/710-rds-broker.yml
```diff
@@ -3,9 +3,9 @@
   path: /releases/-
   value:
     name: rds-broker
-    version: 1.8.0
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/rds-broker-1.8.0.tgz
-    sha1: 9264469b832061f15479df96440ee6ad1280692b
+    version: 0.0.1629903410
```
tomrosier, line 6: Just a reminder to update this before release, as it may be hidden in the commit messages.

tomrosier (approved):

LeePorte (approved):

risicle:
Could you squash the "update RDS broker" commits down into the `TMP ... !UPDATE WITH FINAL!` commit? If I start a commit message with `TMP` it means "this is not supposed to make it into the merge".

risicle:
No!!

PR #2733, opened 2021-09-09
[#179252883] Bump cf deployment to 16.23

schmie:
What
----

- Bump cf-deployment to 16.23, excluding buildpacks
- Bump cflinuxfs3 to 0.254.0 in line with cf-deployment 16.23
- Bump CAPI to 0.115 in line with cf-deployment 16.23
- Bump bionic-stemcell to 1.25 in line with cf-deployment 16.23
- Address dependabot:
  - Go module miekg/dns in tools/metrics
  - node.js dependencies for error_page
  - Ruby deps addressable and rexml for bundler (omitting rack as required by cf-uaac) 

How to review
-------------

- Code review commit by commit
- Deploy to dev env (See also below)
- Check against acceptance criteria in (https://www.pivotaltracker.com/n/projects/1275640/stories/179252883).

How to deploy
-------------

- Paas-bootstrap stemcells should also be upgraded with this ([PR here](https://github.com/alphagov/paas-bootstrap/pull/472)).

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #2743, opened 2021-09-21
[#179252161] mass bump of buildpacks 2021-09

risicle:
What
----

Bumped the buildpacks using `scripts/update_buildpacks.sh`.

I wonder how much I should try and understand the result. As for the `dependencies` listing multiple versions of another package, I guess I should assume that means they're allowed alternatives?

~I should probably also try pushing this out to a dev env on top of #2733 (which should be merged before this).~

Do we have any acceptance tests for our buildpacks? I see our `custom-acceptance-tests` _listing_ them, but don't see anything ensuring they _work_.

How to review
-------------

See the relevant section of https://team-manual.cloud.service.gov.uk/guides/upgrading_CF,_bosh_and_stemcells/#buildpacks and play your part in the dance.

---

### DIFF
--- a/platform-tests/example-apps/logging-pipeline/manifest.yml
+++ b/platform-tests/example-apps/logging-pipeline/manifest.yml
@@ -8,5 +8,5 @@ applications:
    buildpacks: [go_buildpack]
    stack: cflinuxfs3
    env:
-     GOVERSION: go1.15
+     GOVERSION: go1.16
      GOPACKAGENAME: github.com/alphagov/paas-cf/platform-tests/example-apps/logging-pipeline


risicle:
This includes a bump of the golang buildpack which removes 1.15, breaking some of our own apps.

risicle:
This now depends on https://github.com/alphagov/paas-aiven-broker/pull/32 (and probably more other PRs to come)

cadmiumcat (commented):
The PR looks good to me and it seems to work on my dev environment. But I have the same questions as you! So perhaps we should get someone else involved in the review

whpearson:
The email announcement has been sent out for deployment on the  5th

risicle:
Neat.

Going to need a review of #2747 ahead of the rollout.

PR #2749, opened 2021-10-01
Replaces `git` resource with docker-image ubuntu version

cadmiumcat:
What
----

Workaround for issues with Let's Encrypt concourse/concourse#7630
The server certificate verification fails when using the concourse version of the resource_type git. As a temporary fix we have docker image of the resource. This resource_type can be deleted when the issue is resolved

How to review
-------------

---


whpearson:
I think we need the hash to be 40 characters long 

https://github.com/alphagov/paas-cf/blob/main/concourse/spec/image_resource_spec.rb#L40

risicle (approved):

PR #2769, opened 2021-11-01
csls-splunk broker: enable access from gds-security in prod

risicle:
What
----

Given `gds-security` org access to the csls splunk broker. Looks like they must have had it in the past, but no longer do. This should make it permanent.

How to review
-------------

Look at it. I don't think deploying to a dev instance will have any effect.

---


paroxp (approved):
Ireland.

PR #2750, opened 2021-10-07
Update docker container hash

LeePorte:
What
----

Update docker containers to get the latest terraform

How to review
-------------

Check the hash matches the latest from https://concourse.build.ci.cloudpipeline.digital/teams/main/pipelines/build-docker-containers/jobs/build-terraform/builds/7

Who can review
--------------

Anyone

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #2777, opened 2021-11-08
Bugfix: Replaces deprecated terraform command

cadmiumcat:
What
----
The [upgraded terraform](https://github.com/alphagov/paas-cf/pull/2750) has changed a couple of things and now this script is broken. To fix it:
- `-force` has been replaced by `-auto-approve` as the former is no longer recognised
- ~~use `-chdir` to change the directory from which the command is executed https://www.terraform.io/docs/cli/commands/#switching-working-directory-with-chdir~~

How to review
-------------
Run the `destroy-cloudfroundry` pipeline in a dev environment and check that terraform doesn't complain about running this command

---

### DIFF
--- a/concourse/pipelines/destroy-cloudfoundry.yml
+++ b/concourse/pipelines/destroy-cloudfoundry.yml
@@ -296,11 +296,12 @@ jobs:
                 cp cf-tfstate/cf.tfstate updated-cf-tfstate/cf.tfstate
                 cd paas-cf/terraform/cloudfoundry || exit
                 terraform init
-                terraform destroy -force -var env="((deploy_env))" \
+                terraform destroy \
+                  -auto-approve -var env="((deploy_env))" \
                   -var-file="../../../paas-cf/terraform/((aws_account)).tfvars" \
                   -var-file="../../../paas-cf/terraform/cloudfoundry/((aws_account)).tfvars" \
                   -var-file="../../../paas-cf/terraform/((aws_region)).tfvars" \
-                  -state=../../../updated-cf-tfstate/cf.tfstate paas-cf/terraform/cloudfoundry
+                  -state=../../../updated-cf-tfstate/cf.tfstate
         ensure:
           put: cf-tfstate
           params:


concourse/pipelines/destroy-cloudfoundry.yml
```diff
@@ -296,7 +296,8 @@ jobs:
                 cp cf-tfstate/cf.tfstate updated-cf-tfstate/cf.tfstate
                 cd paas-cf/terraform/cloudfoundry || exit
```
risicle, line 297: I think the `-chdir` is unnecessary because we're already in that directory due to this line.

concourse/pipelines/destroy-cloudfoundry.yml
```diff
@@ -296,7 +296,8 @@ jobs:
...
                 terraform init
-                terraform destroy -force -var env="((deploy_env))" \
+                terraform -chdir=../../../paas-cf/terraform/cloudfoundry destroy \
+                  -auto-approve -var env="((deploy_env))" \
                   -var-file="../../../paas-cf/terraform/((aws_account)).tfvars" \
                   -var-file="../../../paas-cf/terraform/cloudfoundry/((aws_account)).tfvars" \
                   -var-file="../../../paas-cf/terraform/((aws_region)).tfvars" \
```
risicle, line 303: And I think you may need to remove the trailing `paas-cf/terraform/cloudfoundry` on the line following this one :point_down: . Due to the `-chdir` this means it's now looking in `paas-cf/terraform/cloudfoundry/paas-cf/terraform/cloudfoundry` for the plan/state files..?

risicle (commented):
Fails for me on `dev01` (lookie there), have left reasons why i think that is.

concourse/pipelines/destroy-cloudfoundry.yml
cadmiumcat: 🤦 
yes and yes. Thank you

PR #2780, opened 2021-11-11
Update dev* make targets and add branch target

whi-tw:
What
----

* Rather than hardcoding "devxx" as targets, use a wildcard rule.
* Add `current-branch` target, that populates the `BRANCH`
  environment variable as the current checked out branch

How to review
-------------

* Code review, really.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,8 +200,8 @@ dev: ## Set Environment to DEV
 	$(eval export ENABLE_AUTODELETE=true)
 	$(eval export ENABLE_TEST_PIPELINES=true)
 	$(eval export ENABLE_AZ_HEALTHCHECK ?= false)
-	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -222,18 +222,9 @@ dev: ## Set Environment to DEV
...
-	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
-	$(eval export APPS_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipelineapps.digital)
-	@true
-
-.PHONY: dev02
-dev02: dev
-	$(eval export DEPLOY_ENV=dev02)
+.PHONY: $(filter-out dev%,$(MAKECMDGOALS))
+dev%: dev
+	$(eval export DEPLOY_ENV=$@)
```
AP-Hunt, line 227: We need to include those extra lines because they otherwise don't include the new deploy env name. `eval export` does it at the time it's invoked, with the context it's invoked in.

AP-Hunt (commented):
Big yes to this change, I was thinking the exact same thing yesterday.

Makefile
AP-Hunt: I misread the diff

PR #2783, opened 2021-11-16
Update makefile help generation

whi-tw:
What
----

Rather than the old way of just pulling out the targets, categorise and sort the targets into categories and then alphabetically sort the targets within these groups. This allows targets to be identified and found easier.

This script isn't intended to be run manually - it is the default target in the Makefile, so is only run if running `make` with no  arguments, or `make help`

How to review
-------------

Run `make help` and ensure the output is sensible

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 .DEFAULT_GOAL := help
 
 .PHONY: help
-help:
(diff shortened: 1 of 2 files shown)


whi-tw:
(tagged some random people for review, no real demand that you do it - don't worry!)

paroxp:
Would you be able to document usage of the script and describe when one should be using it?

whi-tw:
> Would you be able to document usage of the script and describe when one should be using it?

Hopefully that's more better! 💎

whi-tw:
This is really just a nice-to-have, and not really something people should spend time reviewing!

PR #2776, opened 2021-11-05
[#180081291] update aiven broker

whi-tw:
What
----

* Add updated config for Aiven broker
* Add plans for OpenSearch
* Add billing data for OpenSearch

How to review
-------------

* Code review
* Check it's gone down a pipeline successfully

## BEFORE DEPLOY
Before deploying, this needs `REMOVEME: force deploy to use branch` commit to be dropped. This is just to ensure that the right branch is deployed during development.

## AFTER DEPLOY
#2784 needs to be merged.

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #2784, opened 2021-11-17
[#180081291] stop removing envvar

whi-tw:
What
----

- Revert "REMOVEME: remove vestigial envar"

This was just to ensure an old envvar was removed on the first deployment. It is now not needed, as it has been run once.

How to review
-------------

* Code review

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -6092,10 +6092,6 @@ jobs:
                     logit-syslog-drain \
                     -l "syslog-tls://${LOGIT_ADDRESS}:${LOGIT_PORT}"
                 fi
-                
-                ## TODO: remove me after deployed everywhere
-                # remove vestigial environment variable
-                cf unset-env aiven-broker "SERVICE_NAME_PREFIX" || true
 
                 ruby -ryaml -e "
                   env = {


cadmiumcat:
Can you explain like I'm 5? 
I got confused about whether or not I need to remove a commit, and to me it looks as if these are the same changes as in #2776 
Do I just need more ☕

whi-tw:
> Can you explain like I'm 5? I got confused about whether or not I need to remove a commit, and to me it looks as if these are the same changes as in #2776 Do I just need more coffee

😂 no, I think I'm bad at describing!

So, the first one needs to be merged first. once it's merged, this one will look more sensible, as the `Update Aiven Broker deployment for OpenSearch` commit will drop off, and it'll just be the revert commit in the changeset. does that make more sense? ☕

cadmiumcat:
> > Can you explain like I'm 5? I got confused about whether or not I need to remove a commit, and to me it looks as if these are the same changes as in #2776 Do I just need more coffee
> 
> 😂 no, I think I'm bad at describing!
> 
> So, the first one needs to be merged first. once it's merged, this one will look more sensible, as the `Update Aiven Broker deployment for OpenSearch` commit will drop off, and it'll just be the revert commit in the changeset. does that make more sense? ☕

What you're saying makes sense, that's what I thought would happen. But I'm seeing the same commits in both PRs, and suddenly things don't have sense anymore 😭

whi-tw:
> What you're saying makes sense, that's what I thought would happen. But I'm seeing the same commits in both PRs, and suddenly things don't have sense anymore sob

So, I've just branched from my original branch, so all the commits are there. as it's a revert it's easier to see what'll happen if the commit that's being reverted is also in this branch. I *think* it'll all make sense after the other PR is merged. maybe.

cadmiumcat (approved):
All good 👍

PR #2786, opened 2021-11-19
Changes condition for checking if admin user exists in paas-accounts

cadmiumcat:
What
----

This step was failing to find a user in paas-accounts (we were getting 404), and also not creating one. 
This change that an user is created when we get 404

How to review
-------------

Code review + try out on dev environment 

---


AP-Hunt (approved):

risicle (approved):
That'll do it

PR #2790, opened 2021-11-26
Adds a useful script for cleaning up unused test orgs

AP-Hunt:
What
----

Our continuous smoke tests, acceptance tests and the like
create organizations in CF, and provision backing services.
Sometimes the tests fail and the resources don't get cleaned
up. This script tidies up.

How to review
-------------
1. Code review
2. Try it out in `dev02`

---

### DIFF
--- a/scripts/cleanup_leftover_test_orgs.sh
+++ b/scripts/cleanup_leftover_test_orgs.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -e -u -o pipefail
+
+# shellcheck disable=SC2089
+ORGS='{"orgs":[]}'
+NEXT_PAGE_NUM=1
+while true; do
+  ORG_RESPONSE=$(cf curl "/v3/organizations?per_page=5000&page=${NEXT_PAGE_NUM}")
+  ORGS=$(
+    echo "$ORGS" | \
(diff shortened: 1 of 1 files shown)


scripts/cleanup_leftover_test_orgs.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+set -e -u -o pipefail
+
+if [[ "$(cf api | head -n1 | awk '{print $3}')" =~ ^.*cloud.service.gov.uk$ ]]; then
+  echo "Don't try to use this script in prod. It's a bit too risky."
```
LeePorte, line 6: Actually given that there is later confirmation, I disagree

AP-Hunt: Fair point!

paroxp (approved):
Few considerations.

 If it has to be used by operators, why wouldn't we want that functionality in pazmin, so Mark could also have a go?
 
 And if it's intends to be automated, should we also be adding manually triggerable pipeline?
 
 In general, good and much needed idea.

schmie (approved):
:+1: works like a charm (tested pagination in my dev by reducing to 10).

PR #2791, opened 2021-11-29
Create dummy password for prometheus to prevent cf-deploy failing

cadmiumcat:
What
----

This bug means that when the job first runs there is no
'PLATFORM_PROMETHEUS_PASSWORD' to get. We seem to have been doing it manually
on the first run of new dev environments. 

How to review
-------------

Code review. 
You can delete the prometheus password from credhub, and confirm that this (running cf-deploy) doesn't fail and creates a dummy one.
Also check that the it doesn't break the current flow, i.e. if the password already exists, then it is used

---


schmie (approved):
:+1: works like a charm : )

PR #2794, opened 2021-12-01
[WIP] Refactor Github Actions workflow to re-use a workflow

AP-Hunt:
What
----

We have a handful of groups of tests that we want to run against the
repository, and they all share the setup steps for the environment. In the
past, Github Actions didn't support any kind of task re-use, so we made do with
a matrix to run the groups of tests as different configurations.

GHA now supports task re-use, so we should refactor the workflow to properly
represent each set of tests as a different job with a shared task.

How to review
-------------

---

### DIFF
--- a/.github/workflows/prepare-unit-test-environment.yml
+++ b/.github/workflows/prepare-unit-test-environment.yml
@@ -1,0 +1,95 @@
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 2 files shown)


AP-Hunt:
I don't think there's enough value in this to chase it up, given how hard it is to test.

PR #2803, opened 2021-12-14
Notify isolation segments management

risicle:
What
----

Notify may have some extra traffic ahead and want to be able to do some load testing.

How to review
-------------

Look at it, check for typos. Don't think it's really testable.

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #2804, opened 2021-12-14
Comment out high infra instance count check

schmie:
What
----
This test does not take into account diego-cells in isolation segments.
We need to scale Notify isolation segments for load-testing in anticipation of higher utilisation and not bottleneck on the infra
level.

This is required to pass https://github.com/alphagov/paas-cf/pull/2803

How to review
-------------

- Code review

---


paroxp (approved):
Do we have a plan for the future with these? I assume we'd like to revert at some point soon(tm)?

schmie:
Story to revert/correct is here: https://www.pivotaltracker.com/story/show/180635817

PR #2806, opened 2021-12-15
Updating uaa release to fix log4j by updating to log4j 2.16

whpearson:
What
----
Updating uaa release to fix log4j by updating to log4j 2.16

See  https://lists.apache.org/thread/83y7dx5xvn3h5290q1twn16tltolv88f
for details

How to review
-------------

Code review

---


paroxp (approved):
Ideally, we should run this through dev env to test. But I'm also happy that staging will catch potential issues, as we're only really bumping a dependency.

What do you think Will?

whpearson:
It don't think it would be a change likely to cause damage to an environment (unlike database changes or whatever). So in the interests of getting it out, it makes sense to go ahead without a full dev test, if people are happy.

PR #2810, opened 2021-12-16
waf: exclude notify tester ip from waf per-ip rate limiting

risicle:
What
----

Notify need to do some load testing, so exclude their load tester's ip from our WAF rate limiter.

How to review
-------------

Roll out to dev env, check it doesn't break anything.

---


risicle:
This rolls out fine, but we haven't confirmed that it works.

However notify have decided that the existing limit is good enough for them so I'm going to close this.

PR #2811, opened 2021-12-17
Increase notify staging isolation segments

whpearson:
What
----

Notify asked us to increase the isolation segments to 36 for staging

Increasing other cells to roughly match

There were problems with doppler in the last load test, which is why
they are increased a bit more.

How to review
-------------

Make sure that the increases make sense.

---


paroxp (approved):

schmie (approved):
Have to have another look at the distribution of api_instance - not sure if it was 2x or 3x - if the tests are happy, I am, though. :+1:

PR #2815, opened 2021-12-21
Scale down diego-cells in Ireland to 27

schmie:
What
----

Scale down diego-cells in Ireland to 27, as we have not seen the required cell amount to exceed 27 in the past year.

How to review
-------------

- Code review

---


risicle:
What's happening with this? Should we have merged it a month ago?

risicle:
Actually, seeing as notify staging is going back in to the `shared` pool, should we be keeping it like this instead?

PR #2819, opened 2021-12-23
prometheus: upgrade firehose instance group to high_cpu_xlarge

risicle:
What
----

It looks like the first envelopes dopplers start dropping are destined for prometheus' firehose instances. Perhaps because these appear to be built without the anti-backpressure measures later implementations of loggregator components were. These firehose instances do appear to spike up to ~70% cpu utilization under heavy load, so we should probably upgrade them similarly to how we've upgraded other doppler clients.

How to review
-------------

Look at `dev03` and be happy.

---


whpearson (approved):
Looks good, dev03 was happy

PR #2823, opened 2021-12-30
diego cells: increase ephemeral disk size

risicle:
https://www.pivotaltracker.com/story/show/180312173

What
----

Increased the size of the ephemeral disks attached to the diego cells, to give them more headroom and prevent them alerting us so frequently.

I've chosen to multiply the current sizes by 1.5x. Perhaps we could go further, but to me this seems like a good conservative place to start.

---


whpearson (approved):
Seems like a sensible change

PR #2827, opened 2022-01-18
prod isolation-segments: remove govuk-notify-staging

risicle:
What
----

Deleted the `govuk-notify-staging` isolation segment from `prod` as per their request. Their load testing is complete for the time being.

Need to make sure they have migrated their space(s) away from this segment and restaged their apps before deploying this.

Not sure if we should also reduce the loggregator-related instance groups to match.

How to review
-------------

Not sure you can.

---


schmie (approved):
LGTM :+1:

PR #2832, opened 2022-01-20
set_security_groups_from_manifest: don't scrape cf output, use cf curl

risicle:
What
----

It looks like cf's output has changed enough to make the regex not work correctly anymore, so use json from `cf curl` instead.

We could probably take some further steps to ensure this has gone through ok, e.g. try to read back the configuration after an attempted update, but we want to get this out.

How to review
-------------

Look at `dev01`.

---

### DIFF
--- a/concourse/scripts/set_security_groups_from_manifest.rb
+++ b/concourse/scripts/set_security_groups_from_manifest.rb
@@ -47,17 +47,10 @@ private
   end
 
   def fetch_security_groups
-    groups = []
+    response = `cf curl /v3/security_groups?per_page=1000`
+    abort "security_groups request failed" unless $CHILD_STATUS.success?
 
-    security_groups = `cf security-groups`
-    abort security_groups unless $CHILD_STATUS.success?
-
-    security_groups.each_line do |line|
-      if line =~ /\A#\d+\s+(\S+)\s*/
-        groups << Regexp.last_match(1)
-      end
-    end
-    groups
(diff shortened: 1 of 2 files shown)


cadmiumcat (approved):
I can see it works 👍 
But another pair of 👀 would be great

concourse/scripts/set_security_groups_from_manifest.rb
```diff
@@ -47,17 +47,10 @@ def existing_security_groups
...
-    security_groups = `cf security-groups`
-    abort security_groups unless $CHILD_STATUS.success?
-
-    security_groups.each_line do |line|
-      if line =~ /\A#\d+\s+(\S+)\s*/
-        groups << Regexp.last_match(1)
-      end
-    end
-    groups
+    JSON.parse(response)["resources"].map {|sg| sg["name"]}
```
AP-Hunt, line 53: I like this. It's simple, concise, and self-explanatory

AP-Hunt (approved):

risicle:
Incidentally this is why I don't like low-level heavily mocked unit tests. They don't catch things that are broken IRL and they make it painful to fix.

risicle:
omg robocop don't even get me started.

concourse/scripts/set_security_groups_from_manifest.rb
risicle: Too concise for rubocop it turns out

PR #2836, opened 2022-01-28
Fix creation of temporary table locking issues

whpearson:
What
----
The creation of temporary tables was causing triggers to be run that
attempted to lock tables to add permissions. This is unnecessary for
temporary tables, as they are scoped to a user.

This updates the broker to make the triggers no longer run.

How to review
-------------

Have a look in dev03 (where it is deployed) and see that it creating a temporary table is no longer verbose.

---


risicle (approved):
Confirmed.

PR #2844, opened 2022-02-11
18107319- Add CDN Broker acceptance test

nimalank7:
Description:
---
- Adds a basic acceptance test for the CDN Broker which tests if a CDN service with a custom domain can be setup

How to review
-------------

- Code review.
- Deploy to dev01.

---

### DIFF
--- a/platform-tests/broker-acceptance/cdn_broker_test.go
+++ b/platform-tests/broker-acceptance/cdn_broker_test.go
@@ -1,0 +1,56 @@
+package broker_acceptance_test
+
+import (
+	"fmt"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/cf"
+	"github.com/cloudfoundry-incubator/cf-test-helpers/generator"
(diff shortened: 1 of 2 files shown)


cadmiumcat (changes requested):
Just a tiny change and a question.

From the requirements, I think we'd also like to check that when the attempt to creation fails then the instances are deleted. I don't think that's covered here - is it easy/possible to test?

cadmiumcat:
> Not sure I understand. If the CDN instance fails to create then all that will be left is the org + space + user and so the test will fail.

hmmm I'm not sure either. From the ticket it says
```
- attempt creating a CDN instance for domain <randomstring>-1.example.com, check failure.
- delete failed CDN instance
```
Which makes me thing there might be other reasons for failure. I feel like I've read about it somewhere... but I can't find it anymore 🤦

cadmiumcat (approved):
I'm happy that this is testing the happy path. It sounds like we can't test more than that 😢

PR #2846, opened 2022-02-14
Pins aws version for psn

cadmiumcat:
What
----
Here's my naive attempt to fix psn trying to use a newer version of the aws provider. I'm assuming it should be like others in https://github.com/alphagov/paas-cf/commit/a6195f00fa571c2cdec95ea9a187a7493cdab108

How to review
-------------

I'm not sure how to test as it only seems to run in prod

---

### DIFF
--- a/terraform/psn/versions.tf
+++ b/terraform/psn/versions.tf
@@ -1,0 +1,1 @@
+../versions.tf


paroxp (approved):
Looks good. Nice solution.

We won't be able to test it sooner than prod /shrug

PR #2847, opened 2022-02-15
181073179 - Add enable private domain creation step to create Cloud Foundry pipeline

nimalank7:
Description:
- Acceptance tests for the CDN broker require the private_domain_creation feature flag to be enabled
- Prod and the other environments already have this enabled so this will ensure it is turned on when
Cloud Foundry is deployed


paroxp (approved):
Looks good.

> An Org Manager can create private domains for that org. Minimum CC API version: 2.12.
https://docs.cloudfoundry.org/adminguide/listing-feature-flags.html

PR #2851, opened 2022-02-17
instance types: switch dopplers back to "large"

risicle:
What
----

Dopplers turn out to be quite memory-bound, and aren't really using the extra cpu power granted to them by the high-cpu-xlarge upgrade (https://github.com/alphagov/paas-cf/pull/2814).

What's more we have reserved `m5.large` instances which we're now not using.

---

### DIFF
--- a/manifests/variables.yml
+++ b/manifests/variables.yml
@@ -1,4 +1,9 @@
 ---
+#
+# Before you go changing instance types, remember we may have reserved
+# instances of some types.
+#
+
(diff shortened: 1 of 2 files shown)


manifests/variables.yml
```diff
@@ -1,4 +1,9 @@
 ---
+#
+# Before you go changing instance types, remember we may have reserved
+# instances of some types.
```
paroxp, line 4: ☝️ Good shout with the coment!

PR #2853, opened 2022-02-23
[#180965549] Upgrades the TLS cipher suite we use on load balancers

AP-Hunt:
What
----
Upgrades from to 'ELBSecurityPolicy-TLS-1-2-2017-01' to 'ELBSecurityPolicy-TLS-1-2-Ext-2018-06' for ALB/NLB and 'ELBSecurityPolicy-TLS-1-2-2017-01' for classic load balancers. As it stands today (2022-02-23), AWS does not have an ELB security policy which supports TLS 1.3 on Application Load Balancers. One is avaibale for Network Load Balancers, but in the interests of consistency we will only use it when ALBs have the same support.
    
The policy names for ALB/NLBs and classic load balancers have diverged in AWS, so we've had to diverge the two properties here too.

The presence of the 'SslPolicy' field on the 'ModifyListener' action of AWS' Elastic Load Balancing API [1] suggests that we will be able to modify the policy without downtime. Equally, the AWS Terraform provider documentation makes no mention of needing to destroy the listener before making the change.

[1] https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_ModifyListener.html

How to review
-------------

Did I miss anywhere?

---

### DIFF
--- a/terraform/globals.tf
+++ b/terraform/globals.tf
@@ -36,9 +36,15 @@ variable "aws_account" {
   description = "the AWS account being deployed to"
 }
 
+# See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#tls-security-policies
+variable "default_load_balancer_security_policy" {
+  description = "Which Security policy to use for ALBs and NLBs. This controls things like available SSL protocols/ciphers."
+  default     = "ELBSecurityPolicy-TLS-1-2-Ext-2018-06"
+}
+
 # See https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html
-variable "default_elb_security_policy" {
-  description = "Which Security policy to use for ELBs. This controls things like available SSL protocols/ciphers."
+variable "default_classic_load_balancer_security_policy" {
+  description = "Which Security policy to use for classic load balancers. This controls things like available SSL protocols/ciphers."
   default     = "ELBSecurityPolicy-TLS-1-2-2017-01"
 }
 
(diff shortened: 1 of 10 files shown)


terraform/globals.tf
```diff
@@ -38,7 +38,12 @@ variable "aws_account" {
 
 # See https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html
```
cadmiumcat, line 39: I think there's a table for the ALBs
```suggestion
# See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#tls-security-policies
```

terraform/globals.tf
```diff
@@ -38,7 +38,12 @@ variable "aws_account" {
...
 variable "default_elb_security_policy" {
-  description = "Which Security policy to use for ELBs. This controls things like available SSL protocols/ciphers."
+  description = "Which Security policy to use for ALBs and NLBs. This controls things like available SSL protocols/ciphers."
+  default     = "ELBSecurityPolicy-TLS-1-2-Ext-2018-06"
+}
+
```
cadmiumcat, line 44:
```suggestion
# See https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html
```

AP-Hunt:
Good catch, Cat. Commit updated.

cadmiumcat (approved):
👍

AP-Hunt:
[The equivalent `paas-bootstrap` PR is also here](https://github.com/alphagov/paas-bootstrap/pull/489)

PR #2860, opened 2022-03-10
check-certificates: improve error message for soon-expiring certs

risicle:
What
----
Change the error message displayed by the `check-certificates`(`-after-rotation`) job when certificates are expiring very soon.

The previous error message was vague and might lead the operator to retry just the rotate-certs job, which it turns out would be a bad thing to do.

How to review
-------------

:eyes: 

---

### DIFF
--- a/concourse/scripts/check-certificates.rb
+++ b/concourse/scripts/check-certificates.rb
@@ -68,11 +68,17 @@ end
 unless expiring_certificate_names.empty?
   separator
 
-  puts "The following certificates are expiring and require operator intervention:"
+  puts "The following certificates are expiring very soon:"
 
   expiring_certificate_names.each do |cert|
     puts cert.red
   end
+
+  puts <<~HELP
+    The deployment pipeline must be re-run from the beginning to complete the certificate rotation process before these have a chance to expire.
+
+    You may need to do this multiple times before this message goes away.
+  HELP
 end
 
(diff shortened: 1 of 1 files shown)


concourse/scripts/check-certificates.rb
```diff
@@ -68,11 +68,13 @@
...
   separator
 
-  puts "The following certificates are expiring and require operator intervention:"
+  puts "The following certificates are expiring very soon:"
 
   expiring_certificate_names.each do |cert|
     puts cert.red
   end
+
+  puts "\nThe deployment pipeline must be re-run from the beginning before these certificates have a chance to expire."
```
AP-Hunt, line 77: Is it possible to explain why? The more information we can provide here the better, honestly. It's a scary thing to have go awry.

risicle: The problem is, there are multiple possible cases in which this error could be shown and I don't want to say anything that would be incorrect in any cases.

risicle: In fact I've noticed that it could happen that this message is triggered on the _first_ stage of the process, in which case re-running the pipeline would show you the same message again. So I'm going to have to provision for that in the message too.

cadmiumcat (approved):

risicle:
Damn I just used the merge button

PR #2866, opened 2022-03-16
create-cloudfoundry: ensure logit-syslog-drain exists before creating cloudwatch exporter

risicle:
What
----

The `deploy-cloudwatch-exporter` task of the `prometheus-deploy` job expects the `logit-syslog-drain` service to exist in the `admin` / `monitoring` space, making no effort to create it if it doesn't.

Looks like the place this is normally created is in `deploy-paas-metrics`, which comes _after_ it in the pipeline, which is not ideal. So copy the existing pattern to ensure it exists before attempting to deploy.

How to review
-------------

Technically you'd have to destroy a dev env and recreate it from scratch to see this in action. But you could just deploy to an existing dev env and check it doesn't break everything.

Who can review
------------
Not @risicle or @nimalank7 as we are both complicit.

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -3198,6 +3198,9 @@ jobs:
               API_ENDPOINT: ((api_endpoint))
               CF_ADMIN: ((cf_admin))
               CF_PASS: ((cf_pass))
+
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
This looks like a very sensible change

PR #2871, opened 2022-03-31
cf-deployment: 17.1.0 -> 19.0.0

risicle:
What
----
https://www.pivotaltracker.com/story/show/181415115

This sits on top of #2870, which will need to be merged and released to production first.

This does the actual upgrade to cf-deployment 19.0.0.

The test added for the reordering mechanism is of the "torture test" variety rather than the sort that breaks down each individual feature. It may not seem as pretty, but I find this gives the biggest "bang for the buck" and doesn't suffer the issue of failing to test whether the various features work _together, at the same time_.

Commit comments have more detail, but this deployment should prevent the new log cache instances from being used until the last minute when everything should be happily sending traffic to them already.

How to review
-------------

Following a release of #2870 (to a **non-slim** dev env), deploy this while observing `cf logs` availability (possibly using the https://github.com/cloudfoundry/uptimer tool)

(if anyone wants to play with a working cf19 deployment, `dev01` is currently running it)

---


AP-Hunt (approved):
Approved. Can we also prepare a PR for undoing the temporary things after, please?

risicle:
Sure, was just delaying that till after I thought there weren't going to be more changes, as I'm already having to do double-rebases with the other PR.

schmie (approved):

PR #2877, opened 2022-04-12
logit lostash filters: interpret various more accurate timestamp fields

risicle:
What
----

From a trawl of a month's worth of logs, these appear to be the fields already present that contain earlier, more accurate, host-generated timestamps. Some are ISO timestamps, some are unix epoch floats.

The ISO timestamps seem to be interpreted at some point after the logstash filters (?) (though not set to be the `@timestamp` value). The unix epoch values are ingested as strings, so not sortable.

How to review
-------------

Think we need to run `make logit-filters` and use the result to populate one of our logit stacks logstash configurations and testing it out.

---

### DIFF
--- a/config/logit/30_various_timestamps.conf
+++ b/config/logit/30_various_timestamps.conf
@@ -1,0 +1,118 @@
+# Use more accurate timestamp fields, instead of the timestamp of when logit received the log
+
+date {
+  match => [ "[aiven_service_discovery][timestamp]", "dd/MMMM/yyyy:HH:mm:ss Z", "dd/MMM/yyyy:HH:mm:ss Z", "ISO8601", "UNIX" ]
+  target => "@timestamp"
+}
+date {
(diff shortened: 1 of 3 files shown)


risicle:
Hmmmmmmm this still doesn't appear to be interpreting unix epoch timestamps, will need to tinker further.

config/logit/30_various_timestamps.conf
```diff
@@ -0,0 +1,116 @@
+date {
```
cadmiumcat, line 1: Add comment for context

```suggestion
# Use more accurate timestamp fields, instead of the timestamp of when logit received the log

date {
```

config/logit/output/generated_logit_filters.conf
```diff
@@ -1195,4 +1195,120 @@ filter {
       add_field => { "[@source][deployment]" => "%{[app][data][deployment]}" }
     }
   }
+  date {
```
cadmiumcat, line 1198:
```suggestion
# Use more accurate timestamp fields, instead of the timestamp of when logit received the log

  date {
```

cadmiumcat (approved):
Makes sense to me. I'd add a comment to explain why they're there

risicle:
Will squash

PR #2885, opened 2022-04-29
alerts: improve expression for certificate check failure alerts

risicle:
What
----
I started writing up a ticket for this (https://www.pivotaltracker.com/story/show/182035914) but decided to just do it as I'd already done most of the work anyway.

Previous expression would substitute all missing values with 0, which would get picked up as resets by `increase()`, meaning a spotty value would get counted many times. Instead only use 0 when we haven't seen any values in the last 12h otherwise use `max_over_time` from the last 12h.

Ideally we'd use `last_over_time`, but that's only available from prometheus 2.26 onwards, which we don't quite have yet.

How to review
-------------
Deploy to a dev env, see perhaps see if alertmanager alerts appear following repeated failures? Could take multiple days. Could otherwise just suck it and see.

---

### DIFF
--- a/.github/workflows/test_on_pr.yml
+++ b/.github/workflows/test_on_pr.yml
@@ -3,7 +3,7 @@ on: pull_request
 env:
   TF_VERSION: "1.0.8"
   BOSH_CLI_VERSION: "2.0.48"
-  PROMETHEUS_VERSION: "2.6.1"
+  PROMETHEUS_VERSION: "2.22.2"
   DEPLOY_ENV: "github"
   SHELLCHECK_VERSION: "0.7.1"
   GO_VERSION: "1.17"
@@ -55,6 +55,7 @@ jobs:
         run: |
           wget -q -O prometheus.tgz "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz"
           sudo tar xzf "prometheus.tgz" -C /usr/local/bin --wildcards --wildcards-match-slash --strip-components=1 '*promtool'
+          promtool --version
 
       - name: "Install Go ${{env.GO_VERSION}}"
         uses: actions/setup-go@v2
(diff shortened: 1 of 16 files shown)


risicle:
Urg the `promtool` our tests uses is too old to know about `absent_over_time`

manifests/prometheus/spec/alerts/uaa-5xx-errors.test.yml
```diff
@@ -1,7 +1,7 @@
 ---
 rule_files:
   # See alerts_validation_spec.rb for details of how stdin gets set:
-  - spec/alerts/fixtures/rules.yml
+  - fixtures/rules.yml
```
AP-Hunt, line 4: What made these need to change?

cadmiumcat: it seems to be from upgrading the promtool (see individual commits) 

> new version appears to now expect rule_files paths to be relative to the test file

risicle: the promtool bump

risicle: (see the commit comment)

if you think about it it makes more sense for these to be relative to the file rather than the cwd of the caller.

.github/workflows/test_on_pr.yml
```diff
@@ -55,6 +55,7 @@ jobs:
         run: |
           wget -q -O prometheus.tgz "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz"
           sudo tar xzf "prometheus.tgz" -C /usr/local/bin --wildcards --wildcards-match-slash --strip-components=1 '*promtool'
+          promtool --version
```
cadmiumcat, line 58: Was this leftover from checking, or is it meant to stay?

risicle: Thought it was a useful thing to leave. Confirms that we've got it installed and working properly.

manifests/prometheus/spec/alerts/uaa-5xx-errors.test.yml
AP-Hunt: :+1:

AP-Hunt (approved):

PR #2898, opened 2022-05-12
Split ips in x-forwarded-for

cadmiumcat:
What
----

X-forwarded-for field in kibana is a string of comma separated ips. We want to be able to easily filter by ip. This PR converts the field into an array of ips

How to review
-------------

- Code review
- [Check that you can filter this field by individual ips ](https://kibana.logit.io/s/c7358874-2b30-44a6-8271-554ad3996e50/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-15m,mode:quick,to:now))&_a=(columns:!(gorouter.header.x_forwarded_for),filters:!(('$state':(store:appState),exists:(field:gorouter.header.x_forwarded_for),meta:(alias:!n,disabled:!f,index:'logstash-*',key:gorouter.header.x_forwarded_for,negate:!f,type:exists,value:exists))),index:'logstash-*',interval:auto,query:(match_all:()),sort:!('@timestamp',desc)))

After merging
-------------
Copy `config/logit/output/generated_logit_filters.conf` and paste in Logit --> configure logstash pipelines, replacing the one in the editor. Test configuration --> apply
Do this for prod and staging

This will restart Kibana, so let people know!

---

### DIFF
--- a/config/logit/20_custom_cf_filters.conf
+++ b/config/logit/20_custom_cf_filters.conf
@@ -110,3 +110,19 @@ mutate {
     "[gorouter][bytesreceived]" => "integer"
   }
 }
+
+# Process x_forwarded_for in gorouter logs in same way as done for RTR logs above
+if [gorouter][header][x_forwarded_for] {
+  # Set [gorouter][header][x_forwarded_for]
+  mutate {
+    gsub => ["[gorouter][header][x_forwarded_for]","[\s\"]",""] # remove quotes and whitespace
+    split => ["[gorouter][header][x_forwarded_for]", ","] # format is client, proxy1, proxy2 ...
+  }
+
+  # Set [gorouter][remote_addr]
+  # NOTE this value could potentially be spoofed - correct x-forwarded-for
+  # handling in logstash is very tricky
+  mutate {
+    add_field => ["[gorouter][remote_addr]", "%{[gorouter][header][x_forwarded_for][0]}"]
+  }
+}
--- a/config/logit/output/generated_logit_filters.conf
+++ b/config/logit/output/generated_logit_filters.conf
@@ -1187,6 +1187,24 @@ filter {
       "[gorouter][bytesreceived]" => "integer"
     }
   }
+
+
+  # Process x_forwarded_for in gorouter logs in same way as done for RTR logs above
+  if [gorouter][header][x_forwarded_for] {
+    # Set [gorouter][header][x_forwarded_for]
+    mutate {
+      gsub => ["[gorouter][header][x_forwarded_for]","[\s\"]",""] # remove quotes and whitespace
+      split => ["[gorouter][header][x_forwarded_for]", ","] # format is client, proxy1, proxy2 ...
+    }
+
+    # Set [gorouter][remote_addr]
+    # NOTE this value could potentially be spoofed - correct x-forwarded-for
+    # handling in logstash is very tricky
+    mutate {
+      add_field => ["[gorouter][remote_addr]", "%{[gorouter][header][x_forwarded_for][0]}"]
+    }
+  }
+
   # `paas-billing` provides the value of DEPLOY_ENV through an
   # `app.data.deployment` field. We want to set `@source.deployment` so
   # that it can be queried like the rest of the platform.
@@ -1195,7 +1213,7 @@ filter {
       add_field => { "[@source][deployment]" => "%{[app][data][deployment]}" }
     }
   }
-# Use more accurate timestamp fields, instead of the timestamp of when logit received the log
+  # Use more accurate timestamp fields, instead of the timestamp of when logit received the log
 
   date {
     match => [ "[aiven_service_discovery][timestamp]", "dd/MMMM/yyyy:HH:mm:ss Z", "dd/MMM/yyyy:HH:mm:ss Z", "ISO8601", "UNIX" ]


cadmiumcat:
> I might be missing something, but what difference does this make? It seems to look quite a bit like before.
> 
> I think what we probably need is to generate a new field based on `x_forwarded_for` which _only_ has the "real ip". Perhaps called `real_ip` or something. But there's a gotcha in that we should make sure we only do it when we trust the trailing IPs in the field.
> 
> I should possibly have put this in the story but I didn't want to be accused of "solutionizing".

(We had a call and I showed risicle that it did make a difference XD)

cadmiumcat:
> I think my remaining concern is that we'll end up with thousands of e.g. `10.*.*.*` addresses (and our LB's IPs) in the dataset, drowning out the "real IPs". And I don't know how effectively we can "ignore" the intermediate IPs with the tools kibana gives us if we want to e.g. build a ~histogram~ rank-frequency plot of IPs - does "excluding" `10.0.1.12` actually result in all log messages that include that IP being thrown out? Because that's not we'd want - we'd want to just not include those IPs in the ~histogram~ rank-frequency plot.
> 
> Basically, this leaves a lot of the work up to kibana, and I don't know if kibana's powerful enough to manage that.

I think I see your point. However, in terms of the problem this is intended to solve, it is still a good solution, or at least good progress. We'll be able to identify any IPs with a "high" number of requests much quicker than we do right now

AP-Hunt:
> I think my remaining concern is that we'll end up with thousands of e.g. 10.*.*.* addresses (and our LB's IPs) in the dataset, drowning out the "real IPs"

We could use the [Logstash CIDR filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-cidr.html) to filter out `10.0.0.0/8`, and things like CloudFront IP addresses (which we could pull in when we compile the filters)

cadmiumcat:
Instead of filtering out what we don't want, I've added a field `client_ip` that contains the first ip in the `x_forwarded_for` field.
I've just added the change to dev in case you want to have a look (`gorouter.header.client_ip`)
How does that sound, @AP-Hunt @risicle?

risicle:
I think that's the approach I originally imagined, and it's what I was talking about when I mentioned "safe" ways of interpreting x-forwarded-for. If we simply use the leftmost value we open ourselves to spoofing: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#parsing. The "full fidelity" approach described in that article might be a bit much to implement in logstash, but a slightly simpler approach of always picking the third-from-right (?) IP might work in lieu of that.

risicle:
Actually come to think I'm not sure that would be flexible enough to deal with traffic that is/isn't coming via cloudfront. Which makes me think we might also have to exclude all cloudfront IPs from this field too.

risicle:
Note there is actually a proper plugin for this https://github.com/IngaFeick/logstash-filter-forwarded but I don't think we're going to go about getting logit to install/maintain it for us.

risicle:
Having looked deeper at how practical it is for us to do this "properly" in logit (it isn't, especially if we consider that tenants may be pointing their own CDN at us and expecting us to pass on _their_ x-forwarded-for headers) and noticing that handling of the RTR logs also does basically exactly this (the most recent solution, see line 565 of the combined file onwards), I think we're going to have to go with this. I think I'd be happier if we called the resultant field `likely_remote_addr`, but :shrug: .

As long as we still have the split, multi-indexed `x_forwarded_for` field, it should still be possible to dig out a spoofing intermediary with a little effort.

risicle:
For clarification, when I mentioned tenants' external CDNs it's because I was entertaining the idea that we should scrub all x-forwarded-for headers that came into the system and solve the problem once and for all, for our tenants too, but this would annoy tenants who are using foobarcdn and are expecting to get x-forwarded-for through it.

risicle:
Pushed a version which is as much a copy of the RTR field's handling as possible, but fundamentally the same as Cat's most recent version.

config/logit/20_custom_cf_filters.conf
```diff
@@ -110,3 +110,19 @@ mutate {
...
+  mutate {
+    gsub => ["[gorouter][header][x_forwarded_for]","[\s\"]",""] # remove quotes and whitespace
+    split => ["[gorouter][header][x_forwarded_for]", ","] # format is client, proxy1, proxy2 ...
+  }
+
+  # Set [gorouter][remote_addr]
+  # NOTE this value could potentially be spoofed - correct x-forwarded-for
+  # handling in logstash is very tricky
+  mutate {
+    add_field => ["[gorouter][remote_addr]", "%{[gorouter][header][x_forwarded_for][0]}"]
```
AP-Hunt, line 126: Could we make sure the field added has [the IP type](https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html)? That way we'll be able to do all the [types of query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-iprange-aggregation.html) that are supported on that type  

https://discuss.elastic.co/t/how-to-assign-a-new-field-a-certain-type-and-value/34148/2

risicle: Did not know about that at all

risicle: Looks like `ip` is not a valid value https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html#plugins-filters-mutate-convert, even if it exists as a type in ES.

AP-Hunt: Bugger. Oh well, let's get this shipped with the known limitations. Better than nothing.

AP-Hunt (approved):

PR #2901, opened 2022-05-17
Fix `make test`

cadmiumcat:
What
----

We no longer use travis for PR testing, so we can't rely on .travis being there. Instead look for the list of tasks on the github workflow and run `make` for each of them

How to review
-------------

Run `make test` and check it works as expected

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ check-env:
 	@./scripts/validate_aws_credentials.sh
 
 .PHONY: test
-test:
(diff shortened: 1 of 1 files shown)


AP-Hunt:
A more Make-idiomatic way to do this, I think, is to have it depend on the tasks by expanding a variable:

```
TEST_TARGETS := $(shell cat .github/workflows/test_on_pr.yml  | ruby -ryaml -e 'puts YAML.load(STDIN.read)["jobs"]["tests"]["strategy"]["matrix"]["make_task"].map { |j| puts j}')
.PHONY: test
test: $(TEST_TARGETS)# Run the same tests that run on a PR
```

Has the upside of not double invoking Make, which will make it easier to debug if we need to.

PR #2914, opened 2022-06-06
prod: scale down to 24 cell_instances

risicle:
What
----

We seem to be levelling around 20 cells' usage in `prod` - 36 cells is a waste. Haven't changed any of the rest of the infra to match due to the presence of Notify.

But I do note we have over twice as many router instances in `prod` as we do in `prod-lon` and only half the cells.

How to review
-------------

:eyes: 

---


schmie (approved):
👍

PR #2918, opened 2022-06-23
manifest spec: reintroduce uaa version checks outlining necessary testing

risicle:
What
----

Now that we've unforked our uaa dependency we're potentially vulnerable to breakage on uaa updates in areas that aren't covered by automated testing.

This is possible because of the tight coupling of our uaa customization, problems with which prevented us from unforking for so long in the first place.

This is a precursor to the cf-deployment 21.x upgrade (https://www.pivotaltracker.com/story/show/182344728), which will be the first time in ages we've done a "routine" uaa upgrade.

How to review
-------------

Try overriding the version in the manifest, see if the error message is sensible?

---

### DIFF
--- a/manifests/cf-manifest/spec/manifest/release_versions_spec.rb
+++ b/manifests/cf-manifest/spec/manifest/release_versions_spec.rb
@@ -14,6 +14,17 @@ RSpec.describe "release versions" do
     end
   end
 
+  def normalise_version(version)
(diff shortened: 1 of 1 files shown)


schmie (approved):
👍

risicle:
Can't believe I'm getting away with this.

PR #2931, opened 2022-07-20
external prometheus: increase prometheus_disk_size to 250G in prod & prod-lon

risicle:
What
----
https://www.pivotaltracker.com/story/show/182769592

This is as opposed to setting a size-based retention policy, which would cause us to stop getting alerts when we're no longer able to retain our requested 450d of metrics.

Looking back in history the "450d" comes from a desire to have more than a year's worth of metrics available to us for analysis.

How to review
-------------

:eyes: 

---


AP-Hunt (approved):
In principle, I agree with this. I can only hope that the Bosh AWS CPI will do something sensible and increase the size of the existing disk, rather than make a new one.

schmie (approved):

PR #2932, opened 2022-07-21
fix prometheus manifest generation, bump disk sizes

risicle:
What
----

https://www.pivotaltracker.com/story/show/182769592

Firstly this changes the "generic manifest tests" (called "integration" for the cf manifest) to run once for each env, because some important values e.g. disk types are affected by the env config and we want to make sure they're all valid.

Secondly this changes the way the `prometheus` deployment's manifest is generated. Previously we were prioritizing env-specific variables from the top-level/"cf" deployment over those specified in the prometheus deployment's env-specific vars. This meant that the `prometheus_disk_size` settings we had set there were never actually being used. Instead the "external" prometheus' was using the setting intended for just the "internal" prometheus.

Set the external prometheus `prometheus_disk_size` values to match their previous _actual_ value, and bump each up a tier in storage size to solve the original problem purportedly addressed in #2931

How to review
-------------

:eyes: 

---

### DIFF
--- a/manifests/cf-manifest/spec/integration/integration_spec.rb
+++ b/manifests/cf-manifest/spec/integration/integration_spec.rb
@@ -1,176 +1,180 @@
 require "ipaddr"
 
 RSpec.describe "generic manifest validations" do
-  let(:manifest) { manifest_with_defaults }
+  %w[prod prod-lon stg-lon].each do |env|
+    context "for the #{env} environment" do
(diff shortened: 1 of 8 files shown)


AP-Hunt (approved):
I had a bit of a hard time comparing the tests to see if you'd not ported any over, but I _think_ you haven't missed any

PR #2948, opened 2022-08-01
loggregator instances: switch doppler and log cache to more appropriate instance types

risicle:
What
----

https://www.pivotaltracker.com/story/show/182790768

Doppler doesn't use the memory we give it and log-cache could use more if we gave it. Following this we might be able to horizontally downscale log-cache a little further, particularly if we increased the percentage of memory it was allowed to use for logs.

Following some stress-testing on dev02, it was possible to push `c5.large` dopplers to receiving 8k messages/sec each without their memory usage going above 15%. This is well above what our current doppler load is, so I'm happy this halving of the dopplers' memory will have no ill effects.

As for the log-cache instances, that is effectively an _upgrade_, doubling their memory.

So all in all I'm quite confident in this change not having immediate ill effects.

How to review
-------------

Unfortunately I destroyed the prometheus instance that had all the performance test data you would have been able to look at :sweat: 

---

### DIFF
--- a/manifests/variables.yml
+++ b/manifests/variables.yml
@@ -4,6 +4,10 @@
 # instances of some types.
 #
 
+#
+# Spot bid prices tend to be chosen as something around the 1y 100% upfront RI price
+#
+
 compilation_vm_instance_type: c5.large
 
 nano_vm_instance_type: t3.nano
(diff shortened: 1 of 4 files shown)


manifests/variables.yml
```diff
@@ -21,6 +21,12 @@ large_vm_spot_bid_price: 0.05
 xlarge_vm_instance_type: m5.xlarge
 xlarge_vm_spot_bid_price: 0.1
 
+high_cpu_large_vm_instance_type: c5.large
+high_cpu_large_vm_spot_bid_price: 0.05
```
AP-Hunt, line 25: Can you provide some evidence for this bid price?

risicle: Ah hold on do you think it should approximately match the on-demand price or the 1y 100% upfront reversed price? Here it's based on the reserved price but it probably makes more sense to use the on-demand price (which is approx double).

The above, `m5.xlarge`, appears to be approx based on the reserved price (which is actually ~$0.13 according to what I'm looking at)

(Either way this is only used for dev instances...)

risicle: Have added a comment, no idea if it's what we _should_ be doing but :shrug:

AP-Hunt (approved):

PR #2949, opened 2022-08-02
182427769-Fix-GOPATH-error

nimalank7:
Description:
- GOPATH is already defined in the docker image so no need to define it here
- Fixes the error as "$PWD" is not in the PATH


cadmiumcat (approved):
:magic:
Looks good. Running in dev before merging

PR #2950, opened 2022-08-02
Update tags for apps that use ginkgo

cadmiumcat:
What
----

We've updated ginkgo to v2 in [paas-auditor](https://github.com/alphagov/paas-auditor/pull/19) and [paas-accounts](https://github.com/alphagov/paas-accounts/pull/20). This PR updates the tags to make sure we're using those version

How to review
-------------

Check that the tags correspond to the correct release
---


schmie (approved):
:+1:

PR #2952, opened 2022-08-05
Raise maximum number of CF network policies per app source to 200

AP-Hunt:
What
----
We have a tenant making heavy use of private networking for their Prometheus
metric scraping, and they have hit the 100 policies per app limit. What they're
doing is smart and we don't want to discourage it, so the sensible thing to do
is raise the limit. The number we're using is pretty arbitrary anyway.

We shouldn't expect any adverse affects from this. Most other tenants make very
limited use of private networking policies, and their implementation can very
easily handle thousands per app (based on IPTables IIRC).

How to review
-------------

Agree with the change? Different number?

---


cadmiumcat (approved):
Seems sensible 👍

PR #2963, opened 2022-08-12
az-management: exclude backing services when disabling subnets

risicle:
What
----

https://www.pivotaltracker.com/story/show/182772579

This covers RDS & Elasticache instances - these won't automatically handle endpoints being unreachable, so disconnecting their subnets would just result in any non-HA (or HA-but-not-failed-over) service instances being cut off from their apps.

This was @cadmiumcat 's work mostly, I just tested it & fixed a syntax change.

How to review
-------------

Well, I'll describe how I tested it, you can decide to what extent you want to do similar:

- Deploy this to a dev env.
- Create a non-ha database in cf
- Look at the AWS console to see which AZ it was placed in. If it gets put in the same AZ as your rds broker (or elasticache broker if you've chosen an elasticache db), try again. Slim dev envs only have one of each broker and if you disable its AZ you won't be able to bind to it while the AZ is disabled.
- Check you can `cf conduit` to your db.
- Disable the AZ that your db resides in using the `disable-az-?-vpc` concourse job.
- Check you can still `cf conduit` to your db.
- Might be useful to do a `bosh vms` at this point to check the disablement _is actually_ disabling ec2 instances. If the AZ you disabled is c/`z3` it's possible your slim bosh env doesn't have any instances in it, so would be good to do another check afterwards with an AZ that does so you can ensure this.
- Re-enable az with `enable-az-?-vpc` job.
- Check everything's back to normal.

---

### DIFF
--- a/terraform/az-management/main.tf
+++ b/terraform/az-management/main.tf
@@ -2,23 +2,25 @@ provider "aws" {
   region = var.region
 }
 
-data "aws_subnet_ids" "selected" {
-  vpc_id = var.vpc_id
-
+data "aws_subnets" "selected" {
+  filter {
+    name = "vpc-id"
+    values = [var.vpc_id]
+  }
   filter {
     name   = "availability-zone"
     values = [var.az]
   }
 }
 
-data "aws_subnet" "infra_subnet" {
-  vpc_id = var.vpc_id
+data "aws_subnets" "excluded" {
+  filter {
+    name = "vpc-id"
+    values = [var.vpc_id]
+  }
   filter {
     name = "tag:Name"
-    # The wildcard covers the deploy env,
-    # but the VPC id locks the search to
-    # the correct deploy env anyway
-    values = ["*-infra-${var.az}"]
+    values = ["*-infra-${var.az}", "*-backing-service-*"]
   }
 }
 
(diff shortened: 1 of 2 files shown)


terraform/az-management/main.tf
```diff
@@ -2,23 +2,25 @@ provider "aws" {
   region = var.region
 }
 
-data "aws_subnet_ids" "selected" {
-  vpc_id = var.vpc_id
-
+data "aws_subnets" "selected" {
```
AP-Hunt, line 5: It's a very minor thing, but ease of comprehension could we call this `all_subnets` and rename `subnet_ids` to `selected_subnet_ids`. Having this named `selected` and not choose all of the ones from it is a mite misleading.

AP-Hunt (changes requested):
See the other comment I left

terraform/az-management/main.tf
cadmiumcat: Could it be confusing to say "all"? I thought they were the ones for a particular availability zone

I'm not a fan or repeating the resource type in the name (see [terraform naming conventions](https://www.terraform-best-practices.com/naming)) - but I guess we're not sticking to the conventions anyways so that's not a bit deal 😅

AP-Hunt: `all_subets_in_az` would be OK too

risicle: https://en.wiktionary.org/wiki/subet I'll let you dig into that yourself

risicle: I'm a bit confused over what I'm supposed to do - there _isn't_ a `subnet_ids` identifier anymore, other than the `aws_network_acl` attribute, which we don't get to choose the name of.

AP-Hunt: Oh bloody 'ell, that'll teach me to not expand the diff. I read that attribute as being a variable. Sorry folks, false alarm! Leave it as is if we can't change it.

AP-Hunt (approved):

PR #2971, opened 2022-08-26
dev environments: set CA_ROTATION_EXPIRY_DAYS to 345

risicle:
What
----

Having these set to 360 (i.e. having a cert rotation happen every 5 days) is too disruptive to the development process and may even make dev testing _more_ unreliable, as we will frequently be testing with setups that roll more nodes on a deploy than production will.

How to review
-------------

:eyes: :brain: 

---


AP-Hunt (approved):
A certificate rotation every 20 days in dev is a figure I'm happy with.

PR #2973, opened 2022-08-30
`app-autoscaler-cpu-usage` example app: revamp for better reliability, simplicity

risicle:
What
----

This is an app that is deployed with our autoscaler to continuously exercise it by spinning for a period of time and then sleeping for a period of time. But the way it's written it's a little flaky and quirky. If we're going to set up an alert to ensure this is always behaving correctly (as appears to have been the original intention), it would be nice to be able to eliminate any of these strangenesses. The goal here is to, as much as possible, have all instances of this app perform their spinning at the same time (two instances "out of phase" will result in an approximately constant average cpu usage, which won't result in the desired scale-up-scale-down behaviour). Using this technique, that is much more reliable.

Before:
Before

After (ignore gap while prometheus was being redeployed):
After

How to review
-------------

Deploy to a dev env, look at the pretty plot in grafana?

---

### DIFF
--- a/platform-tests/example-apps/app-autoscaler-cpu-usage/main.go
+++ b/platform-tests/example-apps/app-autoscaler-cpu-usage/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"log"
-	"math"
 	"os"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
This is a really clever, and sensible, change and I love it

risicle:
Careful you'll make me well up

PR #2978, opened 2022-09-09
Raise maximum disk space an app can request to 8gb

cadmiumcat:
What
----

One of our users requires extra disk space to be able to temporarily store and scan file uploads. This increases it to 8gb

How to review
-------------

Deploy to dev environment, check that you can scale an app to use more than 6gb of disk space, e.g. `cf scale paas-admin -k 8GB`


risicle:
Not gonna turn this into a variable?

cadmiumcat:
> Not gonna turn this into a variable?

We seem to define the variables per environment, so we'd end up changing it in 4 files instead of one

risicle (approved):

PR #2985, opened 2022-09-29
app-autoscaler-cpu-usage test app: add `app-autoscaler-cpu-usage-multiproc`

risicle:
What
----
https://www.pivotaltracker.com/story/show/183165018

Here we add a multi-process continuously autoscaling test app to catch issue affecting only such apps. We leave in place (but rename) the single-process one and ensure there is an alert for each.

Also reduce the memory & disk allocation for each app, as each actually uses <16M, this is still generous.

How to review
-------------
How I did it:
 - deploy to a dev env
 - see both new apps running in grafana, scaling up and down
 - look at env's alertmanager, see no relevant alert
 - detach one of the apps' scaling policies, e.g. `cf dasp app-autoscaler-cpu-usage-multiproc`
 - do something else for a couple of hours
 - check alertmanager for a relevant alert

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -3537,6 +3537,10 @@ jobs:
                 cf push --strategy=rolling app-autoscaler-cpu-usage
                 cf aasp app-autoscaler-cpu-usage policy.json
 
+                cf cancel-deployment app-autoscaler-cpu-usage-multiproc || true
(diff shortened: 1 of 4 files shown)


malcgds (approved):
In addition to the steps listed in the "How to review" section of the PR, I re-attached the autoscaling policy to the multiproc app, observed the app scaling up within Grafana and observed the AppAutoscalingIsNotScalingMultiProc alert being removed from Alert Manager.

PR #2997, opened 2022-10-13
Pin Ginkgo CLI version to the one used in packages

schmie:
What
----
Pin Ginkgo CLI version to the one used in packages. Using latest has caused acceptance tests to fail with a version mismatch.

How to review
-------------

- Double-check in a dev env that deploy-paas-metrics/acceptance-tests pass

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -5615,7 +5615,7 @@ jobs:
               - -e
               - -c
               - |
-                go install github.com/onsi/ginkgo/v2/ginkgo@latest
+                go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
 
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -5615,7 +5615,7 @@ jobs:
               - -e
               - -c
               - |
-                go install github.com/onsi/ginkgo/v2/ginkgo@latest
+                go install github.com/onsi/ginkgo/v2/ginkgo@2.1.4
```
cadmiumcat, line 5618:
```suggestion
                go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.4
```
Works with the v

cadmiumcat (approved):
I've added the v. If you're happy, I'm happy

PR #2998, opened 2022-10-14
Run ginkgo better for paas-billing and paas-metrics

whi-tw:
What
----

`paas-billing` and `paas-metrics` were throwing errors when running acceptance tests, because we were trying to run them with a different version of ginkgo.

This was caused by this release: https://github.com/onsi/ginkgo/releases/tag/v2.3.1

This is a solved problem with go modules.

Rather than using the binaries via `go install`, instead, vendor via [tools.go](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module)
and run the tools with `go run`.

`ginkgo` and `godog` are now run this way.

How to review
-------------

1. run down a pipeline
1. check the tests:
    1. run
    1. pass

## Before merging

🚨⚠️ reference to `183452476_upgrade_paas_cf` must be reverted` after https://github.com/alphagov/paas-billing/pull/159 is merged, and before this is merged⚠️🚨

---


tools/metrics/Makefile
```diff
@@ -10,15 +10,17 @@ endif
 
 .PHONY: acceptance-tests
 acceptance-tests:
-	ginkgo acceptance
+	go run github.com/onsi/ginkgo/v2/ginkgo acceptance
 
 .PHONY: pkg-tests
+# This looks broken. i can't see how this ever ran. should we remove it?
+# (`. "github.com/onsi/ginkgo/v2"` is missing???)
```
risicle, line 17: Is this not added by your paas-billing PR?

risicle:
(got rid of the paas-billing branch reference now I've merged that)

tools/metrics/Makefile
whi-tw: the comment is, the test is not. i can't see how `make pkg-tests` ever ran correctly, so I dropped a comment in to draw attention to it in future / make it so the next person doesn't end up as confused as I was when I couldn't get these tests to run

risicle: Yeah ok noticed this is on the paas-metrics project I see.

risicle (approved):
I'll merge this

PR #3006, opened 2022-10-21
Improve diego cell memory alerts

risicle:
What
----

https://www.pivotaltracker.com/story/show/183164959

This slightly improves the wording for the former `DiegoCellRepMemoryCapacity` alert, making it `DiegoCellRepsReachingTotalMemoryCapacity`.

This then also adds a gentle alert for there being fewer than 3 cells with enough memory for a 16GiB app instance (and therefore not enough for full HA), and a pagerduty alert for there being fewer than 3 cells with enough memory for a 4GiB app. I'd only expect these to be triggered in fairly extreme circumstances, probably when we're already running a depleted number of cells.

How to review
-------------

Currently running in `dev02`, Where you can see it being triggered from the low number of cells there generally are in slim deployments: https://alertmanager-1.dev02.dev.cloudpipeline.digital

---

### DIFF
--- a/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
+++ b/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
@@ -1,0 +1,26 @@
+# Source: bosh-exporter
+---
+- type: replace
+  path: /instance_groups/name=prometheus2/jobs/name=prometheus2/properties/prometheus/custom_rules?/-
+  value:
+    name: DiegoCellMemoryFewCandidates
+    rules:
+      - alert: DiegoCellMemoryFewCandidates4G
+        expr: (count(firehose_value_metric_rep_capacity_remaining_memory{bosh_job_name="diego-cell"} > 4096) or vector(0)) < 3
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          summary: Few candidate cells for a 4GiB app
+          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 4GiB app instance. Some tenant app(s) may be scaling out of control or we may have depleted cell capacity.
+          url: https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR021-cell-capacity-assignment-2/#decision
+
+      - alert: DiegoCellMemoryFewCandidates16G
+        expr: (count(firehose_value_metric_rep_capacity_remaining_memory{bosh_job_name="diego-cell"} > 16384) or vector(0)) < 3
+        for: 15m
+        labels:
+          severity: warning
+        annotations:
+          summary: Few candidate cells for a 16GiB app
+          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 16GiB app instance. Some tenant app(s) may be scaling out of control or we may have depleted cell capacity.
+          url: https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR021-cell-capacity-assignment-2/#decision
--- a/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
+++ b/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
@@ -33,7 +33,8 @@
           severity: warning
         annotations:
           summary: Few candidate cells for a 16GiB app
-          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 16GiB app instance. The threshold is 3 (1 per AZ).
+          description: |
+            Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 16GiB app instance. The threshold is 3 (1 per AZ).
 
             The likely impact is some tenants are unable to schedule some apps, but failures may be sporadic.
 
--- a/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
+++ b/manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
@@ -12,7 +12,18 @@
           severity: critical
         annotations:
           summary: Few candidate cells for a 4GiB app
-          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 4GiB app instance. Some tenant app(s) may be scaling out of control or we may have depleted cell capacity.
+          description: |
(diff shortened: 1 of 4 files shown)


manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
```diff
@@ -0,0 +1,26 @@
...
+    name: DiegoCellMemoryFewCandidates
+    rules:
+      - alert: DiegoCellMemoryFewCandidates4G
+        expr: (count(firehose_value_metric_rep_capacity_remaining_memory{bosh_job_name="diego-cell"} > 4096) or vector(0)) < 3
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          summary: Few candidate cells for a 4GiB app
+          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 4GiB app instance. Some tenant app(s) may be scaling out of control or we may have depleted cell capacity.
```
AP-Hunt, line 15: Could we tweak this to make it more explicit 

1. what the user impact could be
2. what an operator needs to do

manifests/prometheus/alerts.d/diego_cell_rep_memory_capacity.yml
```diff
@@ -38,16 +38,14 @@
...
           severity: warning
         annotations:
-          summary: Rep low free memory capacity
+          summary: Reps low free memory capacity
           bosh_job_name: "{{ $labels.bosh_job_name }}"
           description: >
-            Low free memory {{ $value | printf "%.0f" }}% for the advertised rep memory capacity
-            in the last 2 hours on average.
-            Review if we need to scale...
+            Low free memory {{ $value | printf "%.0f" }}% for the total advertised rep memory capacity in the last 2 hours on average - if we lost an AZ we may not be able to find space for all the app instances we're currently running. Review if we need to scale.
```
AP-Hunt, line 50: This doesn't read quite right. Suggest, perhaps, something like

> Reps are reporting low free memory on average (x%) over the last 2 hours. If we lost an AZ now, we may not have space to redistribute all the apps running in the lost AZ. Review if we need to scale the cells.

AP-Hunt (changes requested):

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
cadmiumcat: +1 to these

risicle: If you want to add that, you can add that. I'm not adding that because I don't have definite or succinct answers for either of those. And if we did have definite answers for either of those, why wouldn't we just script the action?

manifests/prometheus/alerts.d/diego_cell_rep_memory_capacity.yml
risicle: > Reps are reporting low free memory on average (x%) over the last 2 hours

The _whole point_ of rewriting this was to avoid the ambiguity that is inherent in this. Is it individual reps that are reporting low free memory? What is "low free memory"? Is that low amounts of memory free on the whole bosh instance? _Actual_ memory? _Committed_ memory? Is this just because an app instance has started using more memory than was expected? Is the "on average over the last 2 hours" taken per-cell?

AP-Hunt:
I've pushed a commit with my proposed changes. I'd be interested to see if they make sense to people who aren't as close to this at Rob.

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
```diff
@@ -12,7 +12,17 @@
...
           summary: Few candidate cells for a 4GiB app
-          description: Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 4GiB app instance. Some tenant app(s) may be scaling out of control or we may have depleted cell capacity.
+          description: |
+            Only {{ $value | printf "%.0f" }} cell(s) are advertising enough free memory to run a 4GiB app instance. The threshold is 3 (1 per AZ).
+
+            The likely impact is some tenants are unable to schedule some apps, but failures may be sporadic.
+
+            It means we don't have enough capacity to run all tenant workloads. This is probably caused by natural growth, but it is also possible
+            that one or more tenant applications are scaling out beyond their intended limtis (e.g. because of a bug in the autoscaler, or a fault
+            in the tenant's autoscaling policy).
```
risicle, line 22: I wouldn't focus so closely on the autoscaler, someone could equally have triggered a hundred tasks.

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
```diff
@@ -12,7 +12,17 @@
...
+
+            Consider adding more cells, or finding and scaling down the offending application.
```
risicle, line 24: I would emphasize the latter over the former. If something's managed to get to this point I would assume it would swallow more cells worth of memory within a matter of hours

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
AP-Hunt: Very good point. Will reword.

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
AP-Hunt: Mm. Will swap the order of operations there. Something like "check for something consuming all the resources it can gets its hands on, then scale if it's not that"

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
AP-Hunt: Addressed

manifests/prometheus/alerts.d/diego_cell_mem_few_candidates.yml
AP-Hunt: Addressed

AP-Hunt:
I will fix the test failures once we agree on the messages, because it's those causing the failures.

cadmiumcat:
> I've pushed a commit with my proposed changes. I'd be interested to see if they make sense to people who aren't as close to this at Rob.

I think the description makes it clearer for me and gives me an idea of where to start

cadmiumcat:
Although I might understand it better because we had a knowledge share 🤷

PR #3009, opened 2022-10-24
add "cell memory" dashboard

risicle:
What
----

https://www.pivotaltracker.com/story/show/183164959

Added a dashboard showing information about how much available memory our cells have across the instance group.

Screenshot 2022-10-24 at 11 27 26

How to review
-------------

Deploy to a dev env and look for the dashboard in "General"?

https://grafana-1.dev02.dev.cloudpipeline.digital/d/cell-memory/cell-memory?orgId=1&var-environment=dev02&var-bosh_director=dev02&var-bosh_deployment=dev02&var-bosh_job_name=diego-cell&var-bosh_job_id=All&var-bosh_job_ip=All&var-zone=All

---

### DIFF
--- a/manifests/prometheus/dashboards.d/cell-memory.json
+++ b/manifests/prometheus/dashboards.d/cell-memory.json
@@ -1,0 +1,701 @@
+{
+  "dashboard": {
+    "annotations": {
+      "list": [
+        {
+          "builtIn": 1,
+          "datasource": "-- Grafana --",
+          "enable": true,
+          "hide": true,
+          "iconColor": "rgba(0, 211, 255, 1)",
+          "name": "Annotations & Alerts",
+          "type": "dashboard"
+        }
+      ]
+    },
+    "editable": true,
+    "gnetId": null,
+    "graphTooltip": 1,
+    "id": null,
+    "iteration": 1666363845398,
+    "links": [],
+    "panels": [
+      {
+        "cards": {
+          "cardPadding": null,
+          "cardRound": null
+        },
+        "color": {
+          "cardColor": "#b4ff00",
+          "colorScale": "sqrt",
+          "colorScheme": "interpolatePlasma",
+          "exponent": 0.5,
+          "mode": "spectrum"
+        },
+        "dataFormat": "timeseries",
+        "datasource": null,
+        "description": "Available memory advertised by a Rep divided by the actual available memory on its cell.\n\nA value much greater than 100% suggests a cell that might be in danger of running out of actual memory, resulting in app memory going into swap or apps unexpectedly crashing due to machine OOM.\n\nA value much lower than 100% suggests a cell that will probably stop accepting new app instances when it still has plenty of actual memory left, which is suboptimal but safe.",
+        "fieldConfig": {
+          "defaults": {},
+          "overrides": []
+        },
+        "gridPos": {
+          "h": 8,
+          "w": 12,
+          "x": 0,
+          "y": 0
+        },
+        "heatmap": {},
+        "hideZeroBuckets": false,
+        "highlightCards": true,
+        "id": 4,
+        "legend": {
+          "show": false
+        },
+        "pluginVersion": "7.5.15",
+        "reverseYBuckets": false,
+        "targets": [
+          {
+            "exemplar": true,
+            "expr": "1e6 * firehose_value_metric_rep_capacity_remaining_memory{bosh_job_name=~\"$bosh_job_name\", bosh_job_ip=~\"$bosh_job_ip\", bosh_job_id=~\"$bosh_job_id\", zone=~\"$zone\"} / on(bosh_job_ip) avg(label_replace(node_memory_MemAvailable_bytes, \"bosh_job_ip\", \"$1\", \"instance\", \"(.*):.*\")) by (bosh_job_ip)",
+            "interval": "",
+            "legendFormat": "",
+            "refId": "A"
+          }
+        ],
+        "timeFrom": null,
+        "timeShift": null,
+        "title": "Cell available memory ratio (advertised/actual)",
+        "tooltip": {
+          "show": true,
+          "showHistogram": true
+        },
+        "type": "heatmap",
+        "xAxis": {
+          "show": true
+        },
+        "xBucketNumber": null,
+        "xBucketSize": null,
+        "yAxis": {
+          "decimals": null,
+          "format": "percentunit",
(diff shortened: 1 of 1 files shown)


AP-Hunt:
I'm finding the concept of cell memory debt intuitively difficult, even after reading the tooltip-y thing in Grafana. It might be useful to add a little deeper explainer on the page. Perhaps include what a cell's memory being in debt means for the OS, and what it means in terms of app scheduling?

risicle:
That could easily turn into a multi-page manual.

The "ratio" debt is essentially saying "This cell is advertising n% of the actual free memory it has". "200%" means it's advertising twice as much memory as is actually available.

The "absolute" debt is just saying that the cell is advertising n MiB more memory than it actually has free.

I would love grafana to have a usable scatter plot to better display the relationship between these, but there ain't one and I'm not going to start playing with plugins.

manifests/prometheus/dashboards.d/cell-memory.json
```diff
@@ -0,0 +1,550 @@
...
+        "color": {
+          "cardColor": "#b4ff00",
+          "colorScale": "sqrt",
+          "colorScheme": "interpolatePlasma",
+          "exponent": 0.5,
+          "mode": "spectrum"
+        },
+        "dataFormat": "timeseries",
+        "datasource": null,
+        "description": "Amount of available memory advertised by rep over the actual amount of available memory on cell. >100% is in-debt as a result of overcommitting.",
```
AP-Hunt, line 37: Perhaps

> The amount of uncommitted memory (reported by rep) as a percentage of the system's actual free memory
> (e.g. 18gb uncommitted/12gb free = 150%)
> A figure between x% and y% is healthy. Below that, we're underutilising. Beyond that, we risk exhausting memory in the event of an AZ failure

risicle: That's not what it is - below, we're underutilizing. Above, we're risking apps crashing with OOM if they try to use all their memory.

cadmiumcat: I think I'm missing the difference between the two statements. Is it that @AP-Hunt's suggestion makes sense as long as the last sentence is changed to mention crashing with OOM instead of AZ failures?

risicle: Yes.

Though having said that, the latest version of these dashboards has been changed a bit and I'll have to adapt this sentence a bit more. Haven't added the latest changes to this PR because it requires some manual copypasta which i haven't done yet.

risicle:
Updated, take another look.

Usable as "Cell memory - more" in london.

manifests/prometheus/dashboards.d/cell-memory.json
```diff
@@ -0,0 +1,701 @@
...
+        "color": {
+          "cardColor": "#b4ff00",
+          "colorScale": "sqrt",
+          "colorScheme": "interpolatePlasma",
+          "exponent": 0.5,
+          "mode": "spectrum"
+        },
+        "dataFormat": "timeseries",
+        "datasource": null,
+        "description": "Amount of available memory advertised by rep divided by the actual amount of available memory on cell.\n\nA value much greater than 100% suggests a cell that might be in danger of running out of actual memory and having app memory go into swap or even have apps unexpectedly crash due to machine OOM.\n\nA value much less than 100% suggests a cell that is probably stopping accepting new app instances when it still has plenty of actual memory left. Suboptimal but safe.",
```
cadmiumcat, line 37:
```suggestion
        "description": "Available memory advertised by the Reps divided by the actual available memory on the cell.\n\nA value much greater than 100% suggests a cell that might be in danger of running out of actual memory, resulting in app memory going into swap or apps unexpectedly crashing due to machine OOM.\n\nA value much lower than 100% suggests a cell that is not accepting new app instances when it still has plenty of actual memory left. It is suboptimal but safe.",
```

manifests/prometheus/dashboards.d/cell-memory.json
```diff
@@ -0,0 +1,701 @@
...
+            "exemplar": true,
+            "expr": "1e6 * firehose_value_metric_rep_capacity_remaining_memory{bosh_job_name=~\"$bosh_job_name\", bosh_job_ip=~\"$bosh_job_ip\", bosh_job_id=~\"$bosh_job_id\", zone=~\"$zone\"} / on(bosh_job_ip) avg(label_replace(node_memory_MemAvailable_bytes, \"bosh_job_ip\", \"$1\", \"instance\", \"(.*):.*\")) by (bosh_job_ip)",
+            "interval": "",
+            "legendFormat": "",
+            "refId": "A"
+          }
+        ],
+        "timeFrom": null,
+        "timeShift": null,
+        "title": "Cell available memory ratio",
```
cadmiumcat, line 68:
```suggestion
        "title": "Available memory advertised/Available memory on the cell",
```

cadmiumcat (changes requested):
I've suggested minor wording changes

Otherwise I'm happy for this to be merged

manifests/prometheus/dashboards.d/cell-memory.json
risicle: > the Reps

I would probably make it `the Rep` if anything because I wanted to make it clear that this is a value calculated per-rep.

> is not accepting new app instances

This was deliberately in the future tense because it's more that it's projecting that it _will_ stop accepting new apps early - it may still be accepting apps.

risicle:
Will include, squash & adjust..

manifests/prometheus/dashboards.d/cell-memory.json
risicle: I'm not sure this is clearer to people who don't expect `/` to mean "divide", instead thinking it'll mean "alternatively".

manifests/prometheus/dashboards.d/cell-memory.json
cadmiumcat: ah good, I was wondering about that plural.

For the second one, we could do:

> ...suggests a cell that might stop accepting new app instances

manifests/prometheus/dashboards.d/cell-memory.json
cadmiumcat: Good point. How about:
> Ratio of available memory: advertised/actual
That way we mention that it is a ratio

cadmiumcat (approved):
Nicely done

PR #3014, opened 2022-10-27
Fix timing issue between acceptance tests and custom acceptance tests

corlettb:
What
----

We currently have a timing issue between acceptance tests and custom acceptance tests. These tests run at the same time. The acceptance test will add buildpacks and the custom acceptance tests checks for no extra build packs.

An example of this failure can be seen [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/2#L635940ea:46).

This fix moves the check for the build packs from the custom acceptance tests to the post-deploy job. 

How to review
-------------

Modified pipeline builds can be seen [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/post-deploy/builds/7#L635a7f5e:1)

There are many different ways to fix this issue. Ensure you are happy with this fix.

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -4194,6 +4194,55 @@ jobs:
                 ./paas-cf/concourse/scripts/uaa_set_email_domains.sh '["*.*", "*.*.*", "*.*.*.*", "*.*.*.*.*", "*.*.*.*.*.*"]'
 
     - in_parallel:
+      - task: ensure-buildpacks-exist
+        tags: [colocated-with-web]
+        config:
+          platform: linux
+          image_resource: *cf-cli-image-resource
+          inputs:
+            - name: paas-cf
+            - name: cf-manifest
+          params:
+            CF_ADMIN: admin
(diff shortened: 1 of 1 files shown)


whi-tw (commented):
This change does make a lot of sense - we've seen this a few times over the past few weeks, and at a first look does seem like it would resolve the problem

risicle:
Conceptually I think putting one of the custom acceptance tests in the job of the upstream acceptance tests adds an extra wtf-trap for people unfamiliar with this part of the pipeline. I realize in some ways it's a little uglier, but making the test able to ignore buildpacks of a certain naming pattern would make more sense to me.

corlettb:
If we get it to ignore the CATS buildpacks we could end up with a situation where they linger about because of failed builds or bad pipelines.

I kind of like the check. So maybe we go down the route of moving it to it's own task that runs before or after?

corlettb:
Moved into post-deploy job.

AP-Hunt (approved):
Happy with this. It seems like a sensible place to check that only the right buildpacks exist, given the other things happening in that job.

PR #3016, opened 2022-10-28
added check-billing-logit-ssl-drain-service-app-bindings.sh to replac…

jackjoy-gds:
##What

I identified the issue detailed in the build log below, during the building of a new environment...

https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/deploy-paas-billing/builds/1#L63593f2b:218

Since the upgrade of cli7 to cli8, the output has changed, which was breaking the existing check for "create succeeded" in certain circumstances.

The fix was to replace the existing check which counted the number of instances of "create succeeded", with a series of cf curls with return JSON, and then query that JSON for the expected values.

This test was previously sometimes passing, probably because a single app binding (before the 2nd was started) would have produced the expected count of 2.

##How to review

Deploy an environment using this code change, and check that the output matches the below...

https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/deploy-paas-billing/builds/8#L63599b9c:204

### DIFF
--- a/concourse/scripts/check-billing-logit-ssl-drain-service-app-bindings.sh
+++ b/concourse/scripts/check-billing-logit-ssl-drain-service-app-bindings.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+# This script checks that bindings exist between the service billing-logit-ssl-drain and the billing apps. This is to ensure that the billing apps have been deployed successfully.
+
+set -e -u -o pipefail
+
+APP1_NAME='paas-billing-api'
+APP2_NAME='paas-billing-collector'
+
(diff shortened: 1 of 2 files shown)


concourse/scripts/check-billing-logit-ssl-drain-service-app-bindings.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
```
corlettb, line 2: set -e -u -o pipefail

cadmiumcat (commented):
Just a couple of style change suggestions

I'll try it on a dev env

PR #3018, opened 2022-11-01
Fix cloudfoundry destroy pipeline.

corlettb:
What
----

Fixes the cloudfoundry destroy pipeline so that manual removals are not necessary.

- Set the force_destroy option in dev environments to the cloud foundry s3 buckets.
- Remove the auditor postgres db before the bosh destroy happens.

How to review
-------------

A successful log file of a full destroy can be seen [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/destroy-cloudfoundry/jobs/delete-deployment/builds/6#L635a75f6:1) and [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/destroy-cloudfoundry/jobs/terraform-destroy/builds/7#L635a76f2:1).
Ensure you are happy with the protection of non-dev buckets. 

---

### DIFF
--- a/concourse/pipelines/destroy-cloudfoundry.yml
+++ b/concourse/pipelines/destroy-cloudfoundry.yml
@@ -163,6 +163,15 @@ jobs:
           DB_NAME: accounts-db
           BOUND_APPS: paas-accounts
 
+      - task: remove-auditor-db
+        tags: [colocated-with-web]
+        file: paas-cf/concourse/tasks/remove-db.yml
+        params:
+          ORG: admin
+          SPACE: billing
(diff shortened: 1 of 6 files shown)


AP-Hunt (changes requested):
Can we add an explicit `bucket_force_destroy = false` for prod envs, please? Just in case we default to true in the future.

corlettb:
> Can we add an explicit `bucket_force_destroy = false` for prod envs, please? Just in case we default to true in the future.

Done.

AP-Hunt (approved):

PR #3019, opened 2022-11-02
Allow new org so that they can use splunk

cadmiumcat:
What
----

We've had a request to enable splunk for this org. We need to add it to the allow list too

How to review
-------------

Name of the org matches `cs-conclave-evidence-locker`?

---


cadmiumcat:
Amazing. Thank you

PR #3010, opened 2022-10-24
[#183403347] Upgrade to Prometheus bosh release v27.1.0

AP-Hunt:
What
----

Upgrades Prometheus and Grafana. Grafana moves to 7.5.16 and receives a fix to CVE-2022-29170.

How to review
-------------

1. Check it goes [out down a pipeline OK](https://deployer.dev02.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/prometheus-deploy/builds/233).

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3023, opened 2022-11-04
Upgrade prometheus bosh release

cadmiumcat:
What
----
We have two different releases (27.0.0 and 27.1.0) at the moment. We [pointed to the upstream](https://github.com/alphagov/paas-cf/pull/3010) recently, but we need to also update the ops file.

How to review
-------------

- Check the versions match the [release](https://bosh.io/releases/github.com/cloudfoundry-community/prometheus-boshrelease?all=1)
- It's [happy in a dev environment](https://deployer.dev02.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/316)

---


risicle:
Yup this brings back memories.

Wonder why we don't have a `spec` checking the equality of all of these.

PR #3026, opened 2022-11-11
paas-billing-acceptance-tests: set `APP_ROOT`, move to before deployment

risicle:
What
----

In https://github.com/alphagov/paas-billing/pull/161 I'm trying to give paas-billing proper, robust migration handling. Part of this involves changing how the acceptance tests do their database schema setup, and setting `APP_ROOT` correctly makes this a lot easier.

While I'm at it, moved the acceptance tests (which aren't strictly acceptance tests in the sense we use it - more like integration tests) to *before* the deployment - because this way they can prevent a broken release going out. This is possible because they don't actually depend on any of the deployed resources.

How to review
-------------

Send down a dev env? Look at `dev03`?

---


risicle:
Wait - I'm not sure the reordering is going to work, because the tests *do* also contact the existing API. For some reason.

concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -5321,6 +5321,53 @@ jobs:
...
+                cp "paas-cf/config/billing/output/${AWS_REGION}.json" ./src/github.com/alphagov/paas-billing/config.json
+                cp "paas-cf/config/billing/tests/${AWS_REGION}_billing_rds_charges.feature" ./src/github.com/alphagov/paas-billing/gherkin/features/
+
+                export APP_ROOT="$PWD"
+
+                wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
+                echo "deb http://apt.postgresql.org/pub/repos/apt/ bionic-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list
+
+                apt-get -q update
+                apt-get -q -y install postgresql-12 postgresql-client-12
```
corlettb, line 5357: Do we really want to install postgres every pipeline run rather than including it in the docker image? I know we did that before, but now we are doing it twice.

risicle: Not doing it twice - I've removed it from the latter task. But yeah, I'm trying (lol) to keep my changes focused. My guess is that they didn't go down the route of adding it to the docker image because it was seen as a temporary stopgap, and didn't want to load up our acceptance tests docker image with a postgres.

corlettb (approved):

PR #3027, opened 2022-11-14
Switch billing-collector to use a http endpoint for probes

corlettb:
What
----

This change switches billing-collector from a process check to a http call. This change is linked to a change in [paas-billing](https://github.com/alphagov/paas-billing/pull/160).

How to review
-------------

You should be familiar with the change in paas-billing.

---


risicle (approved):
Let's merge once we're sure https://github.com/alphagov/paas-billing/pull/160 has made it to prod

PR #3039, opened 2022-12-07
router: set route_services_timeout to 900s

risicle:
What
----

Current default of 60s causes problems for some users: https://govuk.zendesk.com/agent/tickets/5118907

Now begins the debate of what value to set it to. I chose 900s (15m) as it matches the timeout we set on our AWS load balancers - with the idea that any query we would expect them to handle should be possible to handle with a route service.

How to review
-------------

---

### DIFF
--- a/manifests/cf-manifest/operations.d/310-router.yml
+++ b/manifests/cf-manifest/operations.d/310-router.yml
@@ -14,6 +14,11 @@
   value: 8443
 
 - type: replace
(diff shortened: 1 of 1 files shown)


risicle:
Un-drafting but I still haven't tested this.

risicle:
This has now been deployed on dev02 without noticeable issues.

PR #3040, opened 2022-12-08
Upgrade cflinuxfs3 version to 0.346

corlettb:
What
----

Update cflinuxfs3 version 0.343. Not the latest, but the latest the current release of cloud foundry is using.

How to review
-------------

This has been tested in dev04.

https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/pipeline-lock/builds/22

---


risicle:
> Not the latest, but the latest the current release of cloud foundry is using.

Normally we haven't been this conservative and just bumped to the latest-latest.. wonder if we should though?

risicle:
Conclusion from slack is to just go all the way, so I'll bump this and re-test.

risicle (approved):

PR #3043, opened 2022-12-13
Add Malcolm's gpg key

malcgds:
What
----

Added my gpg key..

How to review
-------------

Visual inspection of the files that have been amended or run the branch through a dev pipeline.

---


whi-tw (approved):
I like the addition of the check, here - feels like a sensible one.

Might it be an idea to just check one keyserver, or is the configuration throughout our estate to use both of these? If so, should this be mentioned in onboarding docs, for local GPG configuration?

Approved, except for the validity of the key - I can't confirm if this is definitely Malcolm's.

I wonder if we should also check against a canonical source of 'good' team key IDs - maybe in PaaS-trusted-people?

Finally, we don't need to do a full submodule pull in the checkout - that could probably be removed.

Oh, and one last thing: as this is a workflow-only script, perhaps it should be embedded into the workflow definition, to make it more portable?

```yaml
run:|
  key_servers=( "hkp://keyserver.ubuntu.com" "hkps://keys.openpgp.org")
  exit_code="0"

  while read -r key; do
    found="0"
    server_found_in=""
    for server in "${key_servers[@]}"; do
        if gpg --keyserver "${server}" --dry-run --quiet --recv "${key}"; then
            found="1"
            server_found_in="$server"
            break
        fi
    done

     if [[ "$found" == "0" ]]; then
          echo "${key}: not found"
          exit_code="1"
     else
          echo "${key}: found in ${server_found_in}"
     fi
  done < .gpg-id

  exit "$exit_code"

```

AP-Hunt:
> Are we specifically using the Ubuntu key servers everywhere, and if so, should this be mentioned in onboarding docs, for local GPG configuration?

No we're not, that's why I added a list of them; just two common ones for now. I don't think we have a strong opinion on where people put their public key


> Approved, except for the validity of the key - I can't confirm if this is definitely Malcolm's.
> 
> I wonder if we should also check against a canonical source of 'good' team key IDs - maybe in PaaS-trusted-people?

We don't keep the GPG fingerprints in `paas-trusted-people`. I think checking who it actually belongs to is still a human job. The goal of this check was to ensure it was definitely available on a well known key server, where Concourse and humans can find it.

> Finally, we don't need to do a full submodule pull in the checkout - that could probably be removed.

Good catch. Will amend.

AP-Hunt:
@whi-tw my thought on not including it inline was to make it possible to run locally, should someone need to.

PR #3044, opened 2022-12-13
Reduce diego-cells in prod, prod-lon to 144, 21 respectively

risicle:
What
----

App usage appears to have reduced to allow this quite comfortably. We'll get a notification if/when we need to scale back up.

How to review
-------------

No sensible way to test as this is env-specific. Just :eyes: I guess.

---

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -2,7 +2,7 @@
 uaa_instances: 3
 nats_instances: 3
 diego_api_instances: 3
-cell_instances: 150
+cell_instances: 144
 router_instances: 30
 api_instances: 15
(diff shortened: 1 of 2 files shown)


cadmiumcat (approved):

manifests/cf-manifest/env-specific/prod-lon.yml
```diff
@@ -2,7 +2,7 @@
 uaa_instances: 3
 nats_instances: 3
 diego_api_instances: 3
-cell_instances: 150
+cell_instances: 143
```
cadmiumcat, line 5:
```suggestion
cell_instances: 145
```

risicle: oh duh

PR #3052, opened 2023-01-03
Add vpc peering connections with Notify's accounts

sakisv:
What
----

Create VPC peering connections with Notify's AWS accounts. 

We will need these connections in place to migrate the database, traffic, etc.

How to review
-------------

Check names and json formating

---

### DIFF
--- a/terraform/prod.vpc_peering.json
+++ b/terraform/prod.vpc_peering.json
@@ -4,5 +4,23 @@
         "account_id": "430354129336",
         "vpc_id": "vpc-9e62bcf8",
         "subnet_cidr": "10.200.0.0/16"
+    },
+    {
+        "peer_name": "notify-preview",
+        "account_id": "302763885840",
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
These CIDR blocks are outside of our normal range, but we accept that for the special case of Notify migrating out.

sakisv:
Thanks for the approval Andy, what are the next steps to setup the connections? 

Also, out of curiosity, you mentioned

> outside of our normal range

is the normal range anything other than `10/8` ?

AP-Hunt:
Next steps are to get this merged. Our production pipeline will then create the VPC peering on our side, which can be accepted on your side. We're on code freeze until Jan 9th, so we'll merge it after that.

Our normal range is [sets of `/22` in `~172.0.0.0/16`](https://github.com/alphagov/paas-cf/blob/main/terraform/prod-lon.vpc_peering.json#L6)

PR #3060, opened 2023-01-11
[#183793130] Replace spots with on-demand in dev

malcgds:
What
----

I have changed the generate-cloud-config.sh script such that the operations file that replaced on-demand instances with spot instances in the dev env is no longer applied.

I have also removed the workflow test that ensures that there are spot instances in the dev environment.

The associated ticket provides details of the requirement.

How to review
-------------

You can look at the code or run it into one of the dev envs.

---


manifests/cloud-config/scripts/generate-cloud-config.sh
```diff
@@ -10,10 +10,6 @@ for i in "${PAAS_CF_DIR}"/manifests/cloud-config/operations.d/*.yml; do
   opsfile_args+="-o $i "
 done
 
-if [ "$AWS_ACCOUNT" = "dev" ]; then
-  opsfile_args+="-o ${PAAS_CF_DIR}/manifests/cloud-config/operations/use-spot-instances-in-dev.yml"
```
cadmiumcat, line 14: This means we can [delete that file](https://github.com/alphagov/paas-cf/blob/main/manifests/cloud-config/operations/use-spot-instances-in-dev.yml) and the [variables for spot prices ](https://github.com/alphagov/paas-cf/blob/main/manifests/variables.yml#L8) as well, unless it's used somewhere else?

cadmiumcat (commented):
This makes sense - have you already ran it in a dev environment?

manifests/cloud-config/scripts/generate-cloud-config.sh
malcgds: I ran it into dev03. 

I was wondering whether to leave everything else in place just in case the estimation of costs in the ticket was wildly out and spots had to be re-implemented quickly. On reflection, I'm going with your suggestion.

cadmiumcat (approved):
Looks good 👍

PR #3062, opened 2023-01-16
Introduce DMS configuration for Notify test database

AP-Hunt:
What
----

We're performing a test migration with Notify to understand .. stuff better. This commit adds a new DMS migration instance and task ready to go.

* Allocated storage is ~1.2tb, which is 50% of the source database size.
* Maintenance window matches the source database
* AZ matches the source database
* Replication instance class is a guess

How to review
-------------
Do you think I've missed anything?
Should we tweak anything?

---

### DIFF
--- a/terraform/prod.dms.json
+++ b/terraform/prod.dms.json
@@ -1,0 +1,26 @@
+    [
+        {
+            "name": "GOV.UK Notify test migration",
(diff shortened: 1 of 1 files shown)


corlettb (approved):
Looks okay

PR #3048, opened 2022-12-30
grafana dashboards: add usage-by-organisation and usage-by-space

risicle:
What
----

Intended for use in tracking down tenants whose activity is having a detrimental effect on the platform.

These can be seen currently in the "general" folder on London's grafana.

How to review
-------------

These can be seen currently in the "general" folder on London's grafana. Have a play.

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3068, opened 2023-01-20
grafana dashboards: add "container start rate" to usage-by-org/space

risicle:
What
----

Nobody asked for this, but the other day's exercise where the platform was seeing elevated app crashes gave me an itch about the visibility of this sort of thing, so I added a (slightly hacky) plot for container start rate to the dashboards from #3048.

How is this useful? If we're seeing elevated crashes this could help determine if the problem is restricted to one tenant or if it could be a wider platform problem.

This is currently in place in london's grafana-1 under the `(noodling)` dashboard copy.

From the panel's description, in case it's interesting:

> Lacking a real (and sufficiently-labeled) metric for containers being started, this actually measures the number of very young containers attributable to an organization.
> 
> A surge of apparent container starts *could* indicate a lot of crashes and attempted restarts.
> 
> Attempted container starts that don't even manage to live for a minute will probably get missed from this plot.

The query effectively asks "over the last 8 minutes, how many different containers did you see claiming to be less than two minutes old?". This gives us an average figure for the last 10 minutes because the oldest possible container counted at any point in time would have been 2m (`2 * 60 * 1e9` nanoseconds in the query) old 8 minutes ago.

Why two minutes? Give the container a chance to miss its first metrics report somehow.

Why not just query for all containers less than 10m old? Because then a container which lasts e.g. barely more than a minute would only be counted in figures for that one minute. Doing the `min_over_time` forces any container that gets caught by even 1 metrics report to be accounted for for at least 8 minutes.

How to review
-------------

:eyes: 

---

### DIFF
--- a/manifests/prometheus/dashboards.d/usage-by-organization.json
+++ b/manifests/prometheus/dashboards.d/usage-by-organization.json
@@ -34,7 +34,7 @@
         "fill": 10,
         "fillGradient": 0,
         "gridPos": {
-          "h": 20,
+          "h": 10,
           "w": 8,
           "x": 0,
           "y": 0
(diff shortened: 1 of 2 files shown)


risicle:
Pushed an extra fix for these dashboards - I noticed that the "log rate" metric I'm using is actually Bytes/sec.

nimalank7 (approved):
LGTM!

PR #3069, opened 2023-01-20
Site2site vpn

dark5un:
What
----

Introducing the ability for Paas CF to create site to site VPN connections.

How to review
-------------

> Concourse need to be able to create db parameter groups.
### To create the VPN(s)
- Make the appropriate terraform/<env>.vpn.json file in a branch
- Apply the branch to the create-cloudfoundry pipeline
- Run the pipeline
- See `vpn-terraform` job go green applying the VPN changes
- See the `<env>-postgres12-logical-replication` & `<env>-postgres13-logical-replication` appear in AWS console
### To remove the created VPN(s)
- Set the content of the appropriate terraform/<env>.vpn.json file to `[]` in a branch
- Apply the branch to the create-cloudfoundry pipeline
- Run the pipeline
- See `vpn-terraform` job go green removing the VPN configurations 
- See the `<env>-postgres12-logical-replication` & `<env>-postgres13-logical-replication` in AWS console

### The format of terraform/<env>.vpn.json
```
[{
    "name": "<a name that will be used in tags so no spaces please>",
    "aws_customer_gateway": {
        "bgp_asn": <a number in range of 64512...65535>,
        "ip_address": "<the ip address of the gateway>"
    },
    "aws_vpn_connection": {
        "destination_cidr_blocks": [
            "<a destination cidr block>"
        ]
    }
}]
```
---


terraform/vpn/variables.tf
```diff
@@ -0,0 +1,20 @@
+variable "tunnel1_preshared_key" {
+  description = "The preshared key of the first VPN tunnel. The preshared key must be between 8 and 64 characters in length and cannot start with zero(0). Allowed characters are alphanumeric characters, periods(.) and underscores(_)."
```
risicle, line 2: Ew symmetric keys. But looks like the alternative is needlessly complicated :(

risicle:
Should we account for this new `tunnel1_preshared_key` credhub var in one of the credentials-rotation jobs?

AP-Hunt:
> Should we account for this new `tunnel1_preshared_key` credhub var in one of the credentials-rotation jobs?

Probably not, because we likely won't rotate this in the weeks/months it lives. If we rotate it, we have to inform and share with BJSS.

AP-Hunt:
Could we set the VPN options explicitly, so that they match [one of the two IPSec profiles recommended by NCSC](https://www.ncsc.gov.uk/guidance/using-ipsec-protect-data), please? It's something we agreed to with our information assurance, and CAA/BJSS (see the doc linked in the story)

risicle:
This works, I'll just note that it uses a different pattern from the rest of our terraform-that-uses-credhub-creds.

The place this falls down is the initial run of the deployment, because the manifest declaring the credhub var hasn't yet been run and so it doesn't exist. Other credhub vars that get used in the terraform jobs are pre-created by the `generate-secrets` job into the `/concourse/main` namespace, allowing them to be referenced directly in the concourse task.

Don't know if we should be using that approach here for uniformity. Will defer - @AP-Hunt ?

risicle:
Hold on @AP-Hunt won't that mean the var needs to be declared in paas-bootstrap? Think I overlooked this.

AP-Hunt:
I don't recall anymore, it's been too long :( Can I leave this with you two please, @risicle @dark5un?

AP-Hunt (approved):

PR #3073, opened 2023-01-27
[chore] Add a script to simplify suspending cf orgs

whi-tw:
What
----

Allows suspending an org by name, and includes a 'hidden' unsupend method.

This will make suspension simpler, as we won't have to copy and paste from the docs or remember the syntax.

There is also a sanity check step to make sure you're suspending the right org in the right environment.

How to review
-------------

1. Log in to a cf instance
1. Choose an org to suspend
1. Check suspending works: `scripts/suspend-org.sh $ORG_NAME` (should return '"suspended": true' and show 'SUSPENDED' in paasmin)
1. Check unsuspending works: `UNSUSPEND=1 scripts/suspend-org.sh $ORG_NAME` (should return '"suspended": false' and no longer show 'SUSPENDED' in paasmin)

---

### DIFF
--- a/scripts/suspend-org.sh
+++ b/scripts/suspend-org.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+ORG_TO_SUSPEND=${1:-}
+
+SUSPENDED="true"
+
+function usage() {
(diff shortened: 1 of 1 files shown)


corlettb (approved):
Tested against dev01.

PR #3084, opened 2023-02-06
cloud controller: set `default_app_log_rate_limit_in_bytes_per_second`

risicle:
What
----
https://www.pivotaltracker.com/story/show/184235512

A value of 1M should be enough for 99.9...% of apps.

This itself won't stop people raising it beyond 1M but they would have to go out of their way to do so for new apps. I think if we set this and wait a month for it to gradually kick in (as people re-deploy their apps) we'd have a much easier time if we did decide to introduce an actual quota (via e.g. #3074).

The current effective setting for this is `-1`.

How to review
-------------

Deploy to dev env

---


malcgds (approved):
Loaded into dev env and the manifest reflected the new properties

AP-Hunt (approved):

PR #3086, opened 2023-02-06
[#184324630] Refactor dms and fix security group issue.

corlettb:
What
----

- Remove secrets manager endpoint from dms terraform. Now in vpc terraform.
- Add aws endpoint checks to acceptance tests.
- Fix dms egress security group issue.
- Rename cloud foundry security group to generic endpoints name.
- Obtain endpoints cidrs from terraform output (remove now redundant test).

How to review
-------------

Tested in dev-01.

---


platform-tests/acceptance/aws_endpoint_check_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/cloudfoundry/cf-test-helpers/cf"
+	"github.com/cloudfoundry/cf-test-helpers/generator"
+)
+
+type instanceIdentity struct {
+	Region string `json:"region"`
+}
+
+func GetAWSRegion() (string, error) {
+	resp, err := http.Get("http://169.254.169.254/latest/dynamic/instance-identity/document")
```
AP-Hunt, line 22: Have you seen IMDSv2? What's your opinion, should we be using it? Or are we confident enough in our existing security posture to continue using IMDSv1?

platform-tests/acceptance/aws_endpoint_check_test.go
```diff
@@ -0,0 +1,130 @@
...
+				"-c", "python hello.py",
+			).Wait(testConfig.CfPushTimeoutDuration())).To(Exit(0))
+			Expect(cf.Cf("start", appName).Wait(testConfig.CfPushTimeoutDuration())).To(Exit(0))
+		})
+
+		AfterAll(func() {
+			Expect(cf.Cf("delete", appName, "-f").Wait(testConfig.DefaultTimeoutDuration())).To(Exit(0))
+		})
+
+		It("should connect to secrets manager endpoint", func() {
```
AP-Hunt, line 78: These could be easier to read as [a table test](https://onsi.github.io/ginkgo/#table-specs)? Thoughts?

terraform/dms/dms.tf
```diff
@@ -32,7 +31,6 @@ resource "aws_dms_endpoint" "target" {
   ssl_mode                        = "require"
   secrets_manager_arn             = try(data.aws_secretsmanager_secret.target[each.key].arn, null)
   secrets_manager_access_role_arn = data.aws_iam_role.secrets_manager.arn
-  extra_connection_attributes     = try("secretsManagerEndpointOverride=${aws_vpc_endpoint.secrets_manager[0].dns_entry[0].dns_name}", null)
```
AP-Hunt, line 35: I think these extra connection attributes are still needed, it's just the names will come from elsewhere. Is that right @dark5un?

terraform/dms/security_groups.tf
```diff
@@ -5,30 +5,22 @@ data "aws_security_groups" "rds_broker_db_clients" {
   }
 }
 
-resource "aws_security_group" "secrets_manager_dms_access" {
-  name        = "${var.env}-secrets-manager-dms"
+resource "aws_security_group" "dms" {
+  name        = "${var.env}-dms"
   description = "Allow HTTPS inbound traffic for security manager vpc endpoint"
```
AP-Hunt, line 10: Description doesn't really match the name

platform-tests/acceptance/aws_endpoint_check_test.go
corlettb: I hadn't seen IMDSv2. I can understand peoples concerns about the security of IMDS. To be honest I don't like this getting the region from IMDS. I think it should come from the tests config. It currently isn't there. Let me see how hard it would be to add it.

platform-tests/acceptance/aws_endpoint_check_test.go
corlettb: Yeah not a bad idea. Let me give it a try.

platform-tests/acceptance/aws_endpoint_check_test.go
```diff
@@ -56,7 +30,8 @@ var _ = Describe("AWS Endpoint Check", Ordered, func() {
 		)
 
 		BeforeAll(func() {
-			region, err = GetAWSRegion()
+			region = os.Getenv("AWS_REGION")
```
AP-Hunt, line 33: I was wondering if we had that env var, but I assumed you'd thought of it already

AP-Hunt (approved):

platform-tests/acceptance/aws_endpoint_check_test.go
corlettb: Yeah, probably should have gone there first.

terraform/dms/dms.tf
corlettb: These are not needed after your vpc dns changes. This has been tested in dev.

platform-tests/acceptance/aws_endpoint_check_test.go
corlettb: Changed to use the environment variable.

terraform/dms/security_groups.tf
corlettb: Fixed

AP-Hunt (approved):

PR #3091, opened 2023-02-09
[#181970904] Add smtp-alert to prometheus

malcgds:
What
----

This change is to satisfy the requirements of ticket [alerts for missing alerts](https://www.pivotaltracker.com/n/projects/1275640/stories/181970904). A new boshrelease has been created and added to the prometheus manifest.

The smtp_alert script runs every 15 mins.

How to review
-------------

1. The new pagerduty secrets have to be put into credhub for each env: ```make dev04 upload-pagerduty-secrets```. This operation needs to run with the new branch checked out so that the updated upload-pagerduty-secrets.rb script is used.
1. Run the branch into a dev env.
1. ssh onto the alertmanager vm.
1. Check for an smtp_alert file in /etc/cron.d.
1. Change the "pass" var in smtp_alert.rb to raise an incident on pagerduty.
1. Set the "pass" var back to the correct password and check that the pagerduty incident has been resolved.

---

### DIFF
--- a/manifests/prometheus/operations.d/311-add-smtp-alert.yml
+++ b/manifests/prometheus/operations.d/311-add-smtp-alert.yml
@@ -1,0 +1,22 @@
+- type: replace
+  path: /instance_groups/name=alertmanager/jobs/name=smtp_alert?/properties/smtp_alert?
+  value:
+    auth_username: ((terraform_outputs_ses_smtp_aws_access_key_id))
+    auth_password: ((terraform_outputs_ses_smtp_password))
+    smarthost:  ((terraform_outputs_ses_smtp_host))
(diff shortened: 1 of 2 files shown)


whpearson (approved):
Tested the deployment in Dev01.

Changing the pass in the smtp_alert.rb file caused a pagerduty alert to fire and with "        535 Authentication Credentials Invalid" as the reason. Changing it back resolved it.

The upload-pagerduty-secrets needs to be run either before merging or quickly after merging in staging and prod.

whpearson:
Pagerduty alert is here https://governmentdigitalservice.pagerduty.com/incidents/Q31X330GKCJ31R

PR #3092, opened 2023-02-10
prod-lon: reduce number of log caches 18 -> 15

risicle:
What
----

This brings us back to half what we had before vertically scaling the instances, returning us to approximately the same spend as with 30 log-caches. The log caches are currently peaking at ~30% cpu so there should be plenty of headroom for this.

How to review
-------------

Can't really

---

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -6,7 +6,7 @@ cell_instances: 147
 router_instances: 30
 api_instances: 15
 doppler_instances: 39
-log_cache_instances: 18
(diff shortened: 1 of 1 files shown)


corlettb (approved):
What could possibly go wrong.

risicle:
Have looked at the metrics over the weekend and it appears that the log caches do start spiking up towards 100% cpu usage occasionally - _once they are filled_. This suggests that the taxing operation is what it calls "truncation".

Still, it's now not spiking enough to give us alerts, so I think we can continue with this reduction.

risicle:
Actually I take that back, we are still getting high cpu utilization alerts, but more occasionally :(

PR #3079, opened 2023-02-02
[#183231038] Disable all Postgres 10 plans

whi-tw:
## What

- Remove postgres10 from rds-broker manifest
- Set end date for postgres 10 in billing

## How to review

- Check tests pass
- On a deploy env this branch is deployed to (dev04 as-of 2023-02-03), ensure that postgres10 plans do not show up in `cf marketplace -e postgres` (ie. `cf marketplace -e postgres | grep '\-10'` returns nothing).

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3099, opened 2023-02-16
Revert "[#183231038] Disable all Postgres 10 plans"

risicle:
Reverts alphagov/paas-cf#3079


corlettb (approved):
Missing broker-acceptance tests from https://github.com/alphagov/paas-cf/pull/3088/files.

PR #3102, opened 2023-02-16
Disable access to the Postgres 10 plans

AP-Hunt:
What
----

We want them to exist in the RDS broker's catalogue, but not be accessible by tenants.

How to review
-------------

Did I miss anything? Correct way of doing it?

---


corlettb:
Do we not want to consider turning the enable-service-access to disables? with an ignore error and a comment to remove after a fixed time?

Did you test if after disabling the user can still upgrade/deprovision?

AP-Hunt:
I've turned these two `disable-service-access` already, yeah. I don't think `cf disable-service-access` errors when told to disable something that's already disabled.

I haven't tested if users can upgrade or deprovision once disabled, no. I ought to look at that today.

corlettb (approved):

risicle:
This did go out on a dev env though didn't it?

AP-Hunt:
I haven't had opportunity yet, no. I will get it down one shortly 👍

AP-Hunt:
To @corlettb's question

> Did you test if after disabling the user can still upgrade/deprovision?

I read the source code to see what permissions on the service plan were required for those two actions, then walked him through it for confirmation.

AP-Hunt:
@risicle Have now seen this run through a dev env successfully

PR #3103, opened 2023-02-17
Allow per-task configuration for DMS deployments

AP-Hunt:
What
----

During initial testing with the Notify team, we discovered a few things:
1. we set the wrong `TargetTablePrepMode` value for their use case
2. we'll want to exclude certain tables
3. log groups were not automatically created.
4. network speed dropped off on a "large" instance type after 1 hour.
5. terraform was applying task changes every time.

This PR contains changes to allow dms settings to overridden on a per replication basis as well as fixes to the above issues.

Commit messages contain detail for each change.

We are also adding notify-preview replication as part of this pr.

How to review
-------------
1. Code review. Does it look right? Sensible?
2. Is there a better way to do this?

---


terraform/dms/dms.tf
```diff
@@ -86,8 +86,14 @@ resource "aws_dms_replication_task" "default" {
...
-  replication_task_settings = local.task_settings
-  table_mappings            = local.table_mappings
+  replication_task_settings = jsonencode(
+      merge(
+        jsondecode(local.default_task_settings),
+        coalesce(each.value.task.settings_overrides, {})
+      )
+    )
+
+  table_mappings = jsonencode(coalesce(each.value.task.table_mappings, jsondecode(local.default_table_mappings)))
```
corlettb, line 96: Why do we have a difference in the way replication_task_settings and table mappings are calculated? 

Shouldn't table_mappings use a merge as well? Something I'm missing maybe.

AP-Hunt: It's because `merge` doesn't do a recursive merge, so I think for table mappings it should be a matter of define everything for a config, or use the default

https://github.com/alphagov/paas-cf/pull/3103/commits/8f17f4596307218bf45b09ac820f6752c2456bb3

AP-Hunt: I've used `merge` in the task settings because the task settings are much more expansive, and there's a greater likelihood of wanting to change only a small part

corlettb (approved):

terraform/dms/dms.tf
```diff
@@ -107,3 +107,9 @@ resource "aws_dms_replication_task" "default" {
     Name        = "${var.env}-${each.key}"
   }
 }
+
+resource "aws_cloudwatch_log_group" "default" {
+  for_each          = local.migrations
+  name              = "dms-tasks-{$var.env}-${each.key}"
```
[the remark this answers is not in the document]
AP-Hunt: No I mean the syntax is wrong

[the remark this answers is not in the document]
corlettb: My bad. Fixed.

jackjoy-gds (approved):
Changes look good

jackjoy-gds (approved):
Looks good

PR #3105, opened 2023-02-20
log-cache: increase `prunes_per_gc` from 3 -> 5

risicle:
What
----

https://www.pivotaltracker.com/story/show/184518513

This should reduce the frequency at which go's garbage collector has to scan the whole process memory for unreachable objects, which is quite taxing and probably a significant CPU burden.

How to review
-------------

Can check it _works_ on a dev env (example deployment https://deployer.dev05.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/cf-deploy/builds/3). Probably can't check whether it has a beneficial effect to production on anything except production.

---


cadmiumcat (approved):
This makes sense and I think we should give it a go

PR #3106, opened 2023-02-20
alerts: allow log-cache instances to reach 80% cpu util before warning

risicle:
What
----
https://www.pivotaltracker.com/story/show/184518513

log-caches can surge to high utilization occasionally and it appears to be "fine", at least to the point we would like to quieten down this warning until we can decide on a proper fix.

How to review
-------------

:eyes: at pre-merge tests and see it going out on a dev env without bringing everything down in flames. Actually trying to force a dev env's log cache to exceed the alert limits might not be a productive way to spend an afternoon.

---


risicle:
Looks like we don't need this now we've reduced the GC frequency.

PR #3115, opened 2023-02-23
dashboards: add "average envelope size" to "forwarder agents" dashboard

risicle:
What
----
...even though technically it's a metric of the "loggregator agent", downstream from the "forwarder agent", the figure should be similar. I don't think there's enough information about the "loggregator agent" to warrant giving it its own dashboard.

How to review
-------------

Look at dev04's grafana where this PR has been deployed.

---


corlettb (approved):
Checked in dev04. Looks good.

PR #3123, opened 2023-03-01
Upgrade app-autoscaler to v10.0.0

corlettb:
What
----

Upgrade autoscaler to v10.0.0.

A few issues were discovered during the upgrade:
- They've switched their .envrc file to be "not bash". So I've just put in the PATH changes into our pipeline.
- They failed to update the release url in our deployment template. [Here is the issue.](https://github.com/cloudfoundry/app-autoscaler-release/issues/1461)
- They failed to add the settings instances_retrievable and bindings_retrievable to the [app-autoscaler-deployment.yml](https://github.com/cloudfoundry/app-autoscaler-release/blob/main/templates/app-autoscaler-deployment.yml) template. We should look to migrate to their "tested" template [app-autoscaler.yml](https://github.com/cloudfoundry/app-autoscaler-release/blob/main/templates/app-autoscaler.yml).
- They have broken the memory tests when only using 128mb of memory. [Here is the issue](https://github.com/cloudfoundry/app-autoscaler-release/issues/1059). I've increase the node memory size to 256 to workaround this. 

How to review
-------------

[Tested in dev02](https://deployer.dev02.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/pipeline-lock/builds/389)

---

### DIFF
--- a/concourse/tasks/app-autoscaler-acceptance-tests-run.yml
+++ b/concourse/tasks/app-autoscaler-acceptance-tests-run.yml
@@ -26,6 +26,8 @@ run:
       echo "Generating configuration"
       CONFIG="$(pwd)/config.json"
       export CONFIG
+      # TODO: setting node_memory_limit to 256MB is a workaround for https://github.com/cloudfoundry/app-autoscaler-release/issues/1059
+      #       remove once this is fixed.
       cat <<EOF > "$CONFIG"
       {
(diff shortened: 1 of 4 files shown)


whpearson (approved):
Deployed to dev03.

Looks good, hope the issues get fixed in app-autoscaler 11

PR #3127, opened 2023-03-02
[#180362804] Logit flatline alerts

whi-tw:
What
----

Add script to create the content of alert yaml files for logit

The one alert I've created is for a flatline alert - if fewer than 1200 log entries are received in 10 minutes, the alert will fire.

This alert is sent to pagerduty. Currently the pagerduty alert is only for GOV.UK PaaS Email only - this is changed from within PagerDuty, though.

Also, I've fixed the `backtick` custom cop - it didn't allow me to do a thing (which it should have) so I have fixed that.

This is blocked by https://github.com/alphagov/paas-credentials/pull/242 - as currently the secrets are only present on my local machine.

How to review
-------------

- Code Review

---

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1406,6 +1406,7 @@ GEM
       hashdiff
 
 PLATFORMS
+  arm64-darwin-21
   x86_64-darwin-21
 
 DEPENDENCIES
--- a/Makefile
+++ b/Makefile
@@ -484,8 +484,16 @@ logit-filters:
 		-v $(CURDIR):/mnt:ro \
 		-v $(CURDIR)/config/logit/output:/output:rw \
 		-w /mnt \
+		--platform linux/amd64 \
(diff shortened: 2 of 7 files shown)


Gemfile.lock
```diff
@@ -1406,6 +1406,7 @@ GEM
       hashdiff
 
 PLATFORMS
+  arm64-darwin-21
```
risicle, line 1409: Yeah sorry I think this got removed as part of a reversion.

config/logit/alerts.d/flatline.yaml
```diff
@@ -0,0 +1,24 @@
+---
```
risicle, line 1: Suggest naming this file `.yaml.erb`

whi-tw: I did originally, but then I have to strip it in the ruby and eugh. done now!

risicle (approved):
This appears to WFM.

Guess we need to apply this to all envs now then.

PR #3135, opened 2023-03-08
improve buildpacks update process

whi-tw:
What
----

Improve process of creating buildpack release notes

- generate highlights automatically for pre-defined 'interesting' dependencies
- omits unchanged buildpacks
- temp files are now markdown - makes formatting nicer in an IDE
- initial attempt at outputting to HTML.

The HTML thing is under-developed, but the reasoning is:

As we can paste preformatted text into the message input box of Google Groups, we could actually produce a formatted message (ie. with headers / bullets etc) which might be easier to consume by end-users. Currently, we just dump some pseudo-markdown into the email in plaintext, which is fine, but definitely not screen-reader-able and frankly just looks a bit gross.

How to review
-------------

0. confirm that the output looks a bit nicer and takes less work
0. not really sure.

---


AP-Hunt:
I spoke to Tom about this elsewhere. The long and the short of it is that I don't think the highlights should be auto generated.

PR #3140, opened 2023-03-09
fly_sync_and_login: don't try to determine if we're on concourse via `$USER`

risicle:
Instead just use the basic auth creds if we have them, don't if we don't.

Sync'ed back from a change I'm proposing for paas-bootstrap in https://github.com/alphagov/paas-bootstrap/pull/565 

How to review
-------------

Try it locally and on a dev env, or just see it working for the other PR.

---


jackjoy-gds (approved):
Look good

PR #3141, opened 2023-03-13
[#180635817] Improve tests

whi-tw:
### Improve instance_type_counts tests

- when checking for evenly distributable instance groups, check all
  instance groups that start with the given name, not just the first one
  that matches. This means we take isolation segment diego-cell count
  into account

- remove commented out code - we have managed to last for a while
  without it, so should be fine to remove it

###  Increase number of cells in egress-restricted-1

This is required because we are now testing the number of cells in isolation segments, and 2 is not divisible by 3.

## How to review

- ensure the tests run correctly in GHA
- code review :eyes:

---


risicle (approved):
I think I understand this

PR #3144, opened 2023-03-16
grafana dashboards: update usage-by-(org/space) with request time panel

risicle:
What
----

This replaces the "Total service instances" panel with "Total request time", which exposes tenants with a lot of long-running connections (which can set off gorouter alerts for us, so it's useful to be able to narrow down apparently odd gorouter metrics to a specific tenant's activity).

Sad thing about the "Total request time" panel is it doesn't break down by status code like "Total request rate" does, so I've tried to make it as obvious as possible that the status code selector doesn't apply to it.

The service panel was never quite a fit here, because e.g. suddenly adding a large number of services isn't something we're likely to see and it doesn't put a particular strain on shared infrastructure. Could always bring it back via git history and hide it under a collapsing row or something.

This also switches existing panels to new "time series" plot type which is less glitchy in corner cases (particularly for container start rate panel).

How to review
-------------

Deploy to a dev env or look at dev05 where this has recently been deployed.

---

### DIFF
--- a/manifests/prometheus/dashboards.d/usage-by-organization.json
+++ b/manifests/prometheus/dashboards.d/usage-by-organization.json
@@ -22,49 +22,81 @@
     "links": [],
     "panels": [
       {
-        "aliasColors": {},
-        "bars": false,
-        "dashLength": 10,
-        "dashes": false,
-        "datasource": null,
+        "datasource": {
+          "type": "prometheus",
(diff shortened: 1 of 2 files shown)


malcgds:
I checked the dashboard on dev05 and ran it onto dev01. It all looks good to me.

malcgds (approved):
I checked the dashboard on dev05 and ran it into dev01. It looks good to me.

PR #3149, opened 2023-03-20
app-autoscaler base manifest switch

risicle:
What
----

https://www.pivotaltracker.com/story/show/184549339

This involves moving almost all jobs out into their own instance groups and changing a lot of naming to be based on `((deployment_name))`, meaning quite a bit of hostname adaptation etc.

It also currently points `manifests/app-autoscaler/upstream` at our own fork until https://github.com/cloudfoundry/app-autoscaler-release/pull/1653 or a similar fix makes it upstream. Left as a draft until that happens (or we abandon all hope).

(This also means this is likely to bump to the latest autoscaler version once that happens so we can include that change)

How to review
-------------

Deploy to a dev env, see the autoscaler acceptance tests passing.

Check the dev env's alertmanager (`https://alertmanager-1.<env-domain>`) to ensure `AppAutoscalingIsNotScalingSingleProc` and `AppAutoscalingIsNotScalingMultiProc` alerts aren't going off.

---

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
 	url = https://github.com/bosh-prometheus/prometheus-boshrelease
 [submodule "app-autoscaler-release"]
 	path = manifests/app-autoscaler/upstream
-	url = https://github.com/cloudfoundry/app-autoscaler-release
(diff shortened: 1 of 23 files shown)


kr8n3r:
is this still relevant?

PR #3159, opened 2023-03-31
[#184749358] Fix intermittent failure on test cases caused by timing issue.

corlettb:
What
----

Fixing a timing issue in the elb gauges test code. [Example failure](https://github.com/alphagov/paas-cf/actions/runs/4467412084/jobs/7846839945#step:12:58).

This is happening because of the nonblocking:

`go server.ListenAndServe()` 

If the listener doesn't start before we start the tests we will get a failure.

The fix starts the port listen non-blocking. 

For extra safety I'm also increasing the http request timeout to 2 seconds. 

How to review
-------------

The unit tests will run as part of the PR. So look at the code.

---

### DIFF
--- a/tools/metrics/fakes/http.go
+++ b/tools/metrics/fakes/http.go
@@ -1,6 +1,7 @@
 package fakes
 
 import (
+	"net"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


whpearson (approved):
Code looks okay.

PR #3162, opened 2023-04-01
[#184581852] Make terraform jobs run in parallel

malcgds:
What
----

The changes in this PR implement the requirement of the associated [Pivotal Tracker Story](https://www.pivotaltracker.com/n/projects/1275640/stories/184581852). All the terraform jobs have been removed from the serial group "cf-deploy" and now run in parallel.

How to review
-------------

Run the branch into a dev env and perform some cf commands.

Who can review
---

Any paas developer.

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1244,7 +1244,6 @@ jobs:
                   aws s3 rm "s3://((state_bucket))/${JOB_FILE}"
 
   - name: cf-terraform
-    serial_groups: [cf-deploy]
     serial: true
     plan:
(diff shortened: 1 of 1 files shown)


whpearson (changes requested):
Changes look good. Tested on dev04. Could you rebase the commits so there is only one, to keep the git commit history clean?

whpearson (approved):
Changes look good. The jobs do not interfere with each other in a dev environment.

If we didn't have locking, there could be an edge case where the psn-terraform is allowed to run and there is generate-cf-config backed up and then it runs with an out of date config vs  what is out there in the wild potentially causing breakages. But I don't think this is a problem.

PR #3157, opened 2023-03-28
move healthcheck docker startup to systemd service

jackjoy-gds:
What
----

move healthcheck docker startup to systemd service in cloudinit

How to review
-------------

Look at the code

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3164, opened 2023-04-13
Revert "move healthcheck docker startup to systemd service"

risicle:
Reverts alphagov/paas-cf#3157

Breaks in prod.


jackjoy-gds (approved):
Looks Good

PR #3169, opened 2023-04-14
generate_buildpack_bump_pr workflow: update body and fix variable substitution

risicle:
What
----

Firstly add a note to the buildpack PR body about regenerating the PR if it's more than a week old.

*Also* fix variable substitution in the markdown so we don't end up with literal `$GITHUB_SHA` in the body. This broke when the markdown got moved into a separate file.

How to review
-------------

See the test PR created with this @ #3168

---


risicle:
Depends on #3170

dark5un (approved):
👍

PR #3172, opened 2023-04-18
alerts: internalize "ephemeral disk full" alerts with higher threshold for concourse

risicle:
What
----

Concourse's disk usage pattern falsely triggers this all the time when it has a remotely full disk, even though concourse's garbage collector will generally kick in and prevent it actually running out of space. Increase the threshold for these alerts to 90% for concourse-worker nodes.

Here we have to do the same trick we did previously in the `BoshHighCPUUtilisation` upstream alerts which we wanted to alter - disable upstream's one with a very high threshold and then add our own with the same name.

Consciously no tests for this - it's a copypasta of upstream and upstream doesn't have any tests, What's more it seems the deeper you dig into the specifics of prometheus' `predict_linear` behaviour the less sane it shows itself to be.

How to review
-------------

:eyes: ? See it having been deployed to `dev05` @ https://deployer.dev05.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/prometheus-deploy/builds/36

---

### DIFF
--- a/manifests/prometheus/alerts.d/bosh-job-ephemeral-disk-full.yml
+++ b/manifests/prometheus/alerts.d/bosh-job-ephemeral-disk-full.yml
@@ -1,0 +1,33 @@
+# Source: bosh-exporter
+---
+
+- type: replace
+  path: /instance_groups/name=prometheus2/jobs/name=prometheus2/properties/prometheus/custom_rules?/-
+  value:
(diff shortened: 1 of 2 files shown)


dark5un (approved):
LGTM, I had to check the impact of `predict_linear()`. There are some edge cases that might trigger that but the calculations seem reasonable.

PR #3077, opened 2023-02-01
Add nginx-host-reload to nginx boxes as part of the prometheus deployment

corlettb:
What
----

Add nginx-host-reload to nginx boxes as part of the prometheus deployment

How to review
-------------

Tested in dev02

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3173, opened 2023-04-19
184821965 prometheus grafana 28.0.0

dark5un:
What
----

Upgraded Prometheus boshrelease to [v28.0.0](https://github.com/bosh-prometheus/prometheus-boshrelease/tree/v28.0.0).
Part of the change:
- removes [Add nginx-host-reload to nginx boxes as part of the prometheus deployment commit](https://github.com/alphagov/paas-cf/pull/3077/commits/4f90bc4b8481fba7b516e58001b9db9c6847a821).
- upgrades to [Prometheus boshrelease v28.0.0](https://github.com/bosh-prometheus/prometheus-boshrelease/tree/v28.0.0).
- reference to ops-file manifests/operators/enable-cf-api-v3.yml has been deleted, use of v3 API is implied in new version of exporter.

How to review
-------------

- Apply the branch to a dev foundation and run the pipeline
- Watch the pipeline go green
- Grafana / Prometheus still work and collect data
- ssh to nginx_z1 of prometheus deployment and `cat /var/vcap/jobs/nginx/monit`. It should contain the following block:
  ```
  check file hosts with path /etc/hosts
    if changed timestamp then exec "/usr/bin/bash -c '/usr/bin/kill -s SIGHUP $(cat /var/vcap/sys/run/nginx/nginx.pid)'"
  ```
---


risicle (approved):
LGTM WFM

PR #3179, opened 2023-04-24
184602179 healthcheck cloudinit refactor

jackjoy-gds:
What
----

Move the docker run startup command for the az-healthcheck app out of cloudinit and in to its own systemd startup script. Also add terraform config to replace the healthcheck nodes upon a change to the contents of userdata

How to review
-------------
👀  and test in dev/stage. Make sure that the healthcheck nodes are replaced and that the az-healthchecks are green after the deploy (in the health group of the create-cloudfoundry concourse view)

---

### DIFF
--- a/terraform/az-monitoring/module/main.tf
+++ b/terraform/az-monitoring/module/main.tf
@@ -62,13 +62,35 @@ resource "aws_instance" "healthcheck" {
   user_data = <<-EOF
     #!/bin/bash
     set -ex
-    sudo yum update -y
-    sudo amazon-linux-extras install docker -y
-    sudo service docker start
-    sudo usermod -a -G docker ec2-user
-    su - ec2-user
-    docker run -d -p 3000:3000 ghcr.io/alphagov/paas/simple-healthcheck
+
+    echo '
+[Unit]
+Description=simple-healthcheck-service
+Requires=docker.service
+After=docker.service
+[Service]
+Restart=always
+ExecStartPre=/bin/bash -c "/usr/bin/docker ps -q -f name=simple-healthcheck | grep -q . && /usr/bin/docker stop simple-healthcheck || true"
+ExecStartPre=/bin/bash -c "/usr/bin/docker ps -aq -f name=simple-healthcheck | grep -q . && /usr/bin/docker rm simple-healthcheck || true"
+ExecStart=/usr/bin/su ec2-user -c "/usr/bin/docker run --name simple-healthcheck -p 3000:3000 ghcr.io/alphagov/paas/simple-healthcheck"
+Restart=on-failure
+RestartSec=12s
+[Install]
+WantedBy=multi-user.target
+' >/etc/systemd/system/simple-healthcheck.service
+
+    systemctl daemon-reload
+ 
+    yum update -y --setopt=retries=0
+    amazon-linux-extras install docker -y --setopt=retries=0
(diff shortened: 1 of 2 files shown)


risicle:
Could you squash these into one or two commits please?

risicle:
Difference between this and the last version https://github.com/alphagov/paas-cf/compare/c30498852625248e363e8d0122ec3ba32eeecfba...ebf33cb5707c538d3b1b1bbcb26fea0f9e586b48#diff-5b88920650ff855a655963cb48798abb5377c90ca05f25583c7300173aea3911

terraform/az-monitoring/module/main.tf
```diff
@@ -62,13 +62,44 @@ resource "aws_instance" "healthcheck" {
...
+    sudo systemctl start simple-healthcheck
+
+    healthcheck_ok='false'
+    while [ "$healthcheck_ok" != 'true' ]; do
+      if healthcheck_response=$(curl -s -m 30 "http://127.0.0.1:3000" 2>&1) && $(echo "$healthcheck_response" | jq >/dev/null 2>&1); then
+        healthcheck_ok='true'
+      else
+        sleep 3
+      fi
+    done
```
risicle, line 99: What happens if this loop never terminates? Does the machine somehow figure out that all is not well  and reprovision or something?

jackjoy-gds: In short no - if the local health-check docker service never responds with valid JSON, then the cloud-init will just retry forever. I'm not sure what we can do about this without making it mush more complex because we are relying on the cloud-init mechanism to set all this up. It's certainly more robust than the code we have in place right now as it will insulate us from the ghcr/docker-hub outages.

risicle: What role does this loop play in insulating us from the ghcr/dockerhub outages? It seems to just wait until it comes up and then does nothing with that information.

jackjoy-gds: Yes I see your point - I was thinking that this would block the parent step and make sure that the health-checks were up and running before they got tested in the pipeline - but this isn't true for cloud-init as it's not blocking anything. I would have to move that while loop into the pipeline itself

risicle: This is getting into the territory of https://www.pivotaltracker.com/story/show/184950003, but such checks in the pipeline sound awfully close to the jobs we already have checking the healthchecks. We _could_ feed the `pipeline-trigger` into them from the output of this and ensure they pass before we continue. But again this strays into the territory of https://www.pivotaltracker.com/story/show/184950003 - do we actually want the pipeline blocked if an AZ is down?

I think any such check would fall into the same trap.

risicle: So... shall we get rid of this loop if it has no function?

jackjoy-gds: loop removed - blocking functionality moving to pipeline in other ticket as discussed in the team...

risicle:
Why have you changed the prometheus manifest in this PR?

jackjoy-gds:
> Why have you changed the prometheus manifest in this PR?
This is just the result of a submodule update I think - should I not have done that?

risicle (approved):
Appears to WFM, though we should release it through staging with caution

PR #3181, opened 2023-04-25
s3 bucket terraform: enable ACL mode via ownership_controls BucketOwnerPreferred

risicle:
What
----

Necessary since https://aws.amazon.com/about-aws/whats-new/2022/12/amazon-s3-automatically-enable-block-public-access-disable-access-control-lists-buckets-april-2023/ otherwise bringing up an env from scratch will fail with Access Denied when attempting to set up the buckets.

Uses `depends_on` in an attempt to ensure settings are put in place early enough. I *think* I'm doing the right thing with `aws_s3_bucket_public_access_block` for `cdn_broker_bucket` but would love if someone would confirm this for me.

How to review
-------------

Bring up a fresh dev env using this?

---

### DIFF
--- a/terraform/cloudfoundry/buckets.tf
+++ b/terraform/cloudfoundry/buckets.tf
@@ -4,9 +4,18 @@ resource "aws_s3_bucket" "droplets-s3" {
   force_destroy = var.bucket_force_destroy
 }
 
+resource "aws_s3_bucket_ownership_controls" "droplets-s3" {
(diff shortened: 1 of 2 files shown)


jackjoy-gds (approved):
LGTM

fearoffish:
Rebuilt CF on dev02 with this, it fixed the problem. I didn't rebuild bosh with it though, just cf. It works.

LGTM

PR #3184, opened 2023-04-27
custom-acceptance-tests aws endpoint check: remove SMS service, add cloudwatch

risicle:
What
----

The AWS endpoint-poking tests in the `custom-acceptance-tests` are intended to check there isn't any weird aws configuration we've added that would stop our tenants being able to use these services on their own terms from within apps.

AWS SMS ("Server Migration Service") was included in these but since the service was retired recently its endpoint has disappeared, preventing these tests from passing.

In its place I've added CloudWatch endpoints as I'm fairly certain at least one of our tenants feeds metrics/logs to CloudWatch from their apps.

How to review
-------------

Deploy to dev env, check tests pass?

---


dark5un (approved):
LGTM passed dev tests

PR #3191, opened 2023-05-09
Upgrade to prometheus 29.2.0

corlettb:
What
----

Upgrade to prometheus 29.2.0. 

We are using the 'attic' setting for now. There [is a separate ticket](https://www.pivotaltracker.com/story/show/185099769) to migrate to the latest firehose.

How to review
-------------

I deployed to dev03 in dev05.

Please deploy into a different dev environment and check it deploys okay.

Issue
----

I saw an issue when deploying that was hard to reproduce. Namely the error:

"Database locked, sleeping then retrying"

In the grafana log. My suspicion is that this problem is related to the monit change to use bpm in this release. There might be a timing issue where we end up with two grafana's running for a short amount of time. I resolved this when I saw it with a "bosh recreate".

---

### DIFF
--- a/manifests/cf-manifest/operations.d/009-add-prometheus-uaa-clients-attic.yml
+++ b/manifests/cf-manifest/operations.d/009-add-prometheus-uaa-clients-attic.yml
@@ -1,0 +1,1 @@
+../../prometheus/upstream/manifests/operators/deprecated/cf/add-prometheus-uaa-clients-attic.yml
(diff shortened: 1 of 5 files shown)


risicle:
So.. you think this is a glitch we'll only see the first time we migrate from a non-bpm-using grafana to a bpm-using grafana?

risicle:
Shouldn't we be pointing to https://github.com/bosh-prometheus/prometheus-boshrelease/commit/173e7df0a3a3730b4ef4ff2f317dffd6165344c4 ? This appears to point to its parent.

risicle (approved):
WFM on dev05

PR #3206, opened 2023-05-25
[#184950003] Block cf deploy if failing az-healthcheck node

fearoffish:
What
----

X can be a,b or c

- Block deployments if the `az-X-healthcheck` tasks are failing
- Allow an override to be used to ignore failing healthchecks (`DISABLE_AZ_HEALTHCHECK_BLOCK`)
  - This gives us the ability to override the checks and work around the az outage if we have an az outage

How to review
-------------

Deploy cloudfoundry with the following flags set:

- Enable availability zone healthchecks `ENABLE_AZ_HEALTHCHECK=true`
- Disable availability zone healthchecks block `DISABLE_AZ_HEALTHCHECK_BLOCK=false`

We need to simulate an az being down and we can do that by simply hopping into the AWS console and stopping an az healthcheck instance. The healthcheck node will fail the curl against that instance and we should see deploys blocked.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,6 +238,7 @@ dev%: dev
 	$(eval export DEPLOY_ENV=$@)
 	$(eval export ENABLE_AUTODELETE=false)
 	$(eval export ENABLE_AZ_HEALTHCHECK ?= false)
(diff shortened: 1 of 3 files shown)


fearoffish:
Closing to use a different implementation

PR #3216, opened 2023-06-02
#184950003 Block cf-deploy if az healthcheck node fails to start

fearoffish:
What
----

Added a blocker to the terraform responsible for building the az-healthcheck instances. The blocker waits until the healthcheck service is responding before allowing the pipeline to continue. New tasks after the terraform wait for the healthcheck service to start, and double check it's active just in case.

This can be disabled with the already existing `DISABLED_AZS=$ZONE`, where ZONE is a space separated AZ string e.g. `a b`. Disabling an az indicates degradation in AWS service, so this keeps the instances up but disables the checks.

How to review
-------------

Deploy the branch to a dev environment using `SLIM_DEV_DEPLOYMENT=false ENABLE_AZ_HEALTHCHECK=true SELF_UPDATE_PIPELINE=false`. You should see the `az-healthcheck-terraform` node build the instances and then a new task runs curl against them. Then disable a zone by adding `DISABLED_AZS=a` and you'll see those checks should skip.

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -1608,12 +1608,18 @@ jobs:
             TF_VAR_region: ((aws_region))
             AWS_DEFAULT_REGION: ((aws_region))
             ENABLE_AZ_HEALTHCHECK: ((enable_az_healthcheck))
+            DISABLED_AZS: ((disabled_azs))
           run:
             path: sh
             args:
               - -e
               - -c
               - |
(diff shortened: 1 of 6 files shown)


corlettb (approved):
Performed the following tests in dev05:

- Deployed without AZ_HEALHCHECK. Success
- Deployed with AZ_HEALTHCHECK. Success.
- Redeployed no changes. Success
- Deployed with DISABLED_AZS. Success.
- Deployed without AZ_HEALTHCHECK or DISABLED_AZS. Success.

PR #3227, opened 2023-06-16
Get logit pipeline building working

whi-tw:
What
----

Make it so we can build the logit pipelines again.

Yep, this is janky and gross

openjdk doesn't work properly when run through qemu (x86 on arm64):
https://github.com/docker/for-mac/issues/6060

Instead, build an image from the last 9.2 Dockerfile from the jruby git
history (before it was removed).

How to review
-------------

- Code review
- run the make target and ensure no changes are changed

---


fearoffish (approved):
Works fine for me and my m1

AP-Hunt (approved):
All good over here too. Would be good to see this still works on something else, though, if we can.

whi-tw:
> All good over here too. Would be good to see this still works on something else, though, if we can.

Just about to test it on a amd64 ubuntu box. stand by.

PR #3232, opened 2023-06-21
create-concourse: add nonadmin user creation, use for billing acceptance tests

risicle:
What
----

https://www.pivotaltracker.com/story/show/185167233

This expands the way we create/delete temporary cf users in the `create-cloudfoundry` pipeline to also be able to create non-admin users (which can optionally be made a billing manager of a named org).

It then uses both admin and non-admin accounts generated this way for the billing acceptance tests. 

How to review
-------------

Deploy to a dev env, see existing acceptance tests passing and billing acceptance tests passing following a deploy.

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -5829,43 +5829,74 @@ jobs:
             PREFIX: billing-acctests-admin
             DISABLE_ADMIN_USER_CREATION: false
             UAA_ADMIN_CLIENT_PASS: ((uaa_admin_client_secret))
-        - task: paas-billing-acceptance-tests
-          tags: [colocated-with-web]
-          config:
-            platform: linux
-            image_resource: *cf-acceptance-tests-image-resource
+        - do:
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -5818,38 +5821,62 @@ jobs:
...
+                - -c
+                - |
+                  cf api "${API_ENDPOINT}"
 
-                cf target -o admin -s billing
+                  cf auth "$(cat admin-creds/username)" "$(cat admin-creds/password)"
+                  # if we don't target a space with these creds we will run into
+                  # https://github.com/cloudfoundry/cli/issues/2086 and not be able to
+                  # delete the account
+                  cf target -o admin -s billing
```
AP-Hunt: Are we confident this org and space exists at this point in the pipeline on the first ever run?

risicle: Yes because that's what we've just deployed it to

AP-Hunt (approved):

PR #3309, opened 2023-07-14
Temporarily remove DLUHC VPC peering

AP-Hunt:
What
----

The team at DLUHC are re-creating their VPC, which means their VPC ID will change and our peering will be broken. We're removing our side of the peering while they re-create their shiny new VPC.

How to review
-------------

Is the file still valid? Did I miss anything?

---


dark5un (approved):
👍

risicle (approved):
Looks right to me

fearoffish:
Unfortunately we block vpc peer changes, so we'll need another ticket to allow this.

PR #3316, opened 2023-07-21
Split post-deploy into 2 separate jobs to fix race condition

corlettb:
https://www.pivotaltracker.com/n/projects/1275640/stories/185154682

What
----

Split the "post-deploy" into "post-cf-deploy" and "post-bosh-deploys" jobs to fix timing issue.

This change may confuse people as the change updates the shape of the pipeline.

Some of the dependencies have been simplified with the intention of making the pipeline more visually simple without affecting the actual functionality.

Why
----

We have an issue with the create-cloudfoundry pipeline during the aws keys rotation.

The way this works is we delete the keys from the terraform state each month as part of a scheduled job. This will mean terraform will regenerate new keys the next time a deployment happens. Then we remove the old access keys as part of post-deploy.

The problem is that post-deploy runs at the same time as prometheus-deploy. The credentials that alert manager uses will be invalid for a time. This will trigger the alert "DOWN alert: [prod] PaaS AWS SMTP is down".

How to review
-------------

Deploy to a test environment. Try not to be horrified by the new look of the pipeline.

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -146,7 +146,8 @@ groups:
       - cf-deploy
       - app-autoscaler-deploy
       - prometheus-deploy
-      - post-deploy
+      - post-cf-deploy
+      - post-bosh-deploys
       - deploy-paas-accounts
       - deploy-paas-admin
       - deploy-paas-auditor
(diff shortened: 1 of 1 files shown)


malcgds (approved):
Code looks good and ran into dev01 where the pipeline completed without errors.

jackjoy-gds (approved):
ok

PR #3330, opened 2023-07-28
Update rds root and intermediate certificates

corlettb:
What
----

Update rds root and intermediate certificates from https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem.

These were last updated in 2019.

Why
----

New AWS RDS root-ca's were created in 2021, however a user must select the root certificate for their database on create or modify.

We are currently using the "Certificate authority": "rds-ca-2019" across all of our rds instances. This expires on Aug 22 17:08:50 2024 GMT.

I suspect there will be a big push next year from AWS to get people to manually switch their database root ca, before forcing it at a set date.

Guess we have that to look forward to, but adding the new rootca's now should mean we don't run the risk of an outage.

I did a test switching a test db to the new root ca and it did it without needing a reboot.

The new root certificates have an expiry of either 40 or a 100 years. So we shouldn't need to worry again after this.

How to review
-------------

Look at the changes
---


risicle:
So this one actually _does_ affect tenants because it updates the certs our cflinuxfses trust.

PR #3333, opened 2023-07-28
[#185696848] Introduce script for decommissioning an organisation

AP-Hunt:
What
----

Finally decommissioning an org means removing all of the users from it. Doing so can be a tedious affair, so this script makes that easier.

```
./scripts/decommission_organisation --org SOME_ORG
```

How to review
-------------

Create an organisation in a dev env, invite a user, and then use the script

---


malcgds:
I created an org, a space and three users. I ran the script and the script flagged that the org was not suspended and that it had an existing space. I suspended the org and re-ran the script. The script flagged that the org had an existing space. I deleted the space, re-ran the script and then the script removed the users.

LGTM.

PR #3276, opened 2023-06-30
Switch to using the bosh-prometheus new 7.0 firehose

corlettb:
What
----

Switch to using the bosh-prometheus new 7.0 firehose.

The new jobs require bosh dns in order to function so adding it as part of this PR.

This PR also fixes our dashboards as the metrics have been changed as part of the firehose 7.0 release.

How to review
-------------

Review the changes
Deploy to an environment.

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3341, opened 2023-07-31
remove `uaa_clients_firehose_exporter_secret` handling in pipelines & scripts

risicle:
What
----

Looks like this is no longer needed since the "new" firehose_exporter (switched-to in #3276), and a pipeline expecting it to exist will choke when it hasn't been generated. e.g. when a newly-created env is brought up.

How to review
-------------

Deploy to a dev env? Look at dev03 which was brought up with this (commit applied to another branch...). dev04 had this change deployed to it directly (without a rebuild) and it seems happy.

---


fearoffish (approved):
🆗

PR #3342, opened 2023-07-31
terraform: bump ~all~ most providers

risicle:
What
----

https://www.pivotaltracker.com/story/show/185714476

Only required minor alterations to our terraform.

How to review
-------------

Deploy to dev env. This has been used to bring up dev03 from scratch with a non-slim deployment, including az-healthchecks.

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #3344, opened 2023-08-03
waf: migrate to WAF v2, adding ability to conditionally obey `x-forwarded-for`

risicle:
What
----

https://www.pivotaltracker.com/story/show/185574049

This depends on https://github.com/alphagov/paas-aws-account-wide-terraform/pull/372, #3342 ~and (tangentially) #3341~

WAF v2 gives us the ability to use more complex sets of rules for configuration of rate limiting.

Firstly we exclude an environment's own egress IPs and "system" (currently just concourse) IPs from rate limiting to prevent us blocking either our own tests or tenants that are using externally routed inter-app requests.

We then have a rate limit rule for requests we've determined should *not* have `x-forwarded-for` respected. We determine this by looking for a header with a "secret" value only set by our blessed CDNs. So that the header is not visible to tenant apps we abuse the header `x-cf-instanceid` for this. This is because neither load balancers or the gorouter have the ability to remove arbitrary headers from forwarded requests, so instead we have to choose one that we know will be blindly overwritten by the gorouter.

Finally there is a similar rate limit rule for requests that *should* have `x-forwarded-for` respected. these are expressed as two separate rules with the same header-checking condition inverted because abort-early/fall-through logic won't behave as we want it to here - a rate-limiting rule will only make processing abort for a request when the rate limit is tripped.

The secret for `x-forwarded-for` enablement is generated by credhub and woven through the creation pipeline. This will presumably be needed by the cdn broker to eventually be able to set the appropriate header. until that is done, the waf should continue to work as it always previously has, not obeying `x-forwarded-for`, so there's no race to set up both ends of this at the same time.

How to review
-------------

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -894,6 +894,7 @@ jobs:
                     external_uaa_database_password
                     external_silk_controller_database_password
                     external_policy_server_database_password
+                    waf_xff_auth_key
                   )
 
                   app_autoscaler_credentials=(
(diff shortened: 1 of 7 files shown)


fearoffish (approved):
I haven't tested it myself, but the code changes seem reasonable

risicle:
Have also added some configuration to get the cloudwatch exporter exporting metrics for this.

risicle:
And this now also adds a `WafThrottlingIPsMaxRate` alert.

AP-Hunt (approved):

PR #3369, opened 2023-08-24
[#185529919] Work associated with AWS retirement of db-instance-classes

malcgds:
What
----

We have received a notification that AWS will retire the t2, m4 and r4 db instance classes by April 2024. This PR replaces the deprecated classes with t3, m5 and r5 classes as appropriate in the affected marketplace offerings.

This change also contains a script - upgrade_instance_class.rb - which will upgrade the instance class in existing rds instances where appropriate.

How to review
-------------

1. Review the code. Any feedback gratefully received.
2. Run the branch into a dev env.
3. Check that no cf rds marketplace services offer any of the deprecated instance classes.
4. Using aws ro credentials run the script in "--dry-run" mode. You should see no "not authorized" errors meaning the script is not attempting to update any aws resources.
5. I have changed the terraform for the cdn broker instances so that they use a t3 class instead of a t2 class. Consequently, you will see that the cdn broker instance in the dev env that you are using will not need to be upgraded by the script.
6. When the script is run with aws admin creds and without the --dry-run flag it will set those rds instances that use a deprecated class to be upgraded during the next maintenance window. The upgrade can be viewed under the "Maintenance & backups" tab, within the "Maintenance" section under "Pending modifications" of the impacted rds instance.
7. The logger class within the script is currently set to INFO.

---

### DIFF
--- a/scripts/upgrade_instance_class.rb
+++ b/scripts/upgrade_instance_class.rb
@@ -1,0 +1,242 @@
+#!/usr/bin/env ruby
+
+require "aws-sdk-rds"
+require "logger"
+require "singleton"
+require "optparse"
+require "rubygems/version"
+
+class MyLogger
+  include Singleton
+
+  def initialize
+    @logger = Logger.new($stdout)
+    @logger.level = Logger::WARN
+  end
+
+  def log_info(message)
+    @logger.info(message)
+  end
+
+  def log_warn(message)
+    @logger.warn(message)
+  end
+
+  def log_fatal(message)
+    @logger.fatal(message)
+  end
+end
+
+def display_help
+  puts "Usage: upgrade_instance_types.rb [options]"
+  puts "Options:"
+  puts " -h, --help   Display this help message"
+  puts " --dry-run    Run the script in dry-run mode"
+end
+
+def closest_compatible_version(current_version, compatible_versions)
+  current_version_obj = Gem::Version.new(current_version)
+  compatible_versions = compatible_versions.map { |v| Gem::Version.new(v) }
+
+  closest_version = nil
+  min_difference = Float::INFINITY
+
+  compatible_versions.each do |version|
+    next if version <= current_version_obj
+
+    difference = version.segments.map(&:to_i).zip(current_version_obj.segments.map(&:to_i)).map { |a, b| (a - b).abs }.sum
+
+    if difference < min_difference
+      min_difference = difference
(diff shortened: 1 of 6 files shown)


scripts/upgrade_instance_class.rb
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env ruby
+
+require "aws-sdk-rds"
+require "versionomy"
```
risicle, line 4: Where does versionomy come from? Is it already included in the repo's Gemfile?

risicle: In `manifests/cf-manifest/spec/manifest/release_versions_spec.rb` we appear to use `Gem::Version`

scripts/upgrade_instance_class.rb
```diff
@@ -0,0 +1,202 @@
...
+  case instance.db_instance_status
+  when /modifying|rebooting/
+    MyLogger.instance.log_info("Instance #{instance.db_instance_identifier} is in a transitional state. Waiting for it to become available ...")
+    rds_client.wait_until(:db_instance_available, db_instance_identifier: instance.db_instance_identifier)
+    "available"
+  when "available"
+    MyLogger.instance.log_info("Instance #{instance.db_instance_identifier} is available.")
+    "available"
+  else
+    MyLogger.instance.log_info("Instance #{instance.db_instance_identifier} is in a state that can't be modified. Moving on ...")
```
risicle, line 60: Surely this should at least be a `log_warn`?

Gemfile
```diff
@@ -11,4 +11,5 @@ gem "rest-client", "~> 2.1.0"
 gem "rspec", "~> 3"
 gem "rubocop-govuk", "~> 4.8"
 gem "simplecov", require: false, group: :test
+gem "versionomy", "~> 0.5.0"
```
risicle, line 14: So.. there *is* a specific reason to use versionomy over `Gem::Version` is there?

scripts/upgrade_instance_class.rb
malcgds: Code now uses Gem::Version

Gemfile
malcgds: Removed versionomy gem

risicle: Is `blockenspiel` still needed?

malcgds: It has really gone this time.

PR #3370, opened 2023-08-24
load balancers: push logs to centralized bucket & update logit filters

risicle:
What
----

https://www.pivotaltracker.com/story/show/185739015

This is @whi-tw 's work, barely modified. Looks sensible and seems to work for me in a dev env.

Logs go into a per-account bucket because we have per-account logit stacks. As such this depends on ... having been run in your aws account.

How to review
-------------

:eyes: , look at dev03? Deploy to your own env?

---


malcgds (approved):
I checked it out in dev03 and it looked good. I've also run it into dev01 and I see logs being created.

risicle:
This needs modification to work with the multi-region shenanigans...

risicle:
Ok good now, though think it might need a re-review.

config/logit/filters.d/99_clean_alb_access.conf
```diff
@@ -0,0 +1,14 @@
+#
+# Fix up the datapoints for the alb_access log type
+#
+if [@input] == "alb_access" {
+    ## remove extraneous fields
+    mutate {
+        remove_field => [ '[@shipper][name]', '[@source][component]', '[@source][type]' ]
```
AP-Hunt, line 7: I think we quite regularly filter on `@source.component` in Kibana. It would be nice if we could have a meaningful value in that field for ALB access logs.

PR #3387, opened 2023-09-07
Reduce memory usage of yace by splitting the staging process

corlettb:
What
----

Reduce memory usage of yet-another-cloudwatch-exporter by splitting the staging process into two apps. The staging process will use 3GB and the running version will use 128MB.

Mask the AWS secret from the concourse logs.

Why
----

As part of the yet-another-cloudwatch-exporter upgrade we had to increase memory from 128M to 3GB. This was because the new aws client requires a lot of RAM to compile.

Sadly cloud foundry does not support different staging and running memory sizes. This process of downloading the droplet from one app and upload as another is a workaround to this issue.

How to review
-------------

Look at the changes.
[Tested in dev05](https://deployer.dev05.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/prometheus-deploy/builds/122#L64f74a61:4)

---

### DIFF
--- a/concourse/pipelines/create-cloudfoundry.yml
+++ b/concourse/pipelines/create-cloudfoundry.yml
@@ -3377,7 +3377,7 @@ jobs:
               - name: config
               - name: yet-another-cloudwatch-exporter
             run:
-              path: sh
+              path: bash
               args:
                 - -e
                 - -c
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-cloudfoundry.yml
```diff
@@ -3424,7 +3443,17 @@ jobs:
...
+                  cf stage cloudwatch-exporter-staging
+
+                  DROPLET=$(cf droplets cloudwatch-exporter-staging | tail -n +4 | awk '{print $1}' | head -n 1)
+
+                  cf download-droplet cloudwatch-exporter-staging --droplet "${DROPLET}" --path droplet.tgz
+
+                  set -o pipefail
+                  # mask the aws secret in the concourse output
+                  cf push cloudwatch-exporter -f manifest.yml --droplet droplet.tgz --strategy rolling 2>&1 | \
+                    sed "s/${YACE_AWS_SECRET_ACCESS_KEY}/********/g"
```
risicle, line 3456: I'm not sure an AWS key is safe to put in a sed pattern, all sorts of funny characters in there.

risicle: The stackoverflow rabbit-hole for how to do this robustly without installing more packages is depressing.

The simplest thing might be to just use `grep -Fv` and accept that matching lines will just be silently swallowed.

corlettb: Okay updated to a:
```
sed -r "s/AWS_SECRET_ACCESS_KEY: .+/AWS_SECRET_ACCESS_KEY: ********/g"
```

risicle (approved):
Yup, appears to work well for me

PR #3411, opened 2023-09-22
Scale down London and Ireland cell counts

AP-Hunt:
What
----

Both were a good number higher than was necessary. Numbers have been chosen to give a little wiggle room on top of what's already accounted for, so that we have time to address it if we're growing into that buffer.
London
image

Ireland
image

How to review
-------------

Do you agree with the numbers? Suggest others?

---

### DIFF
--- a/manifests/cf-manifest/env-specific/prod-lon.yml
+++ b/manifests/cf-manifest/env-specific/prod-lon.yml
@@ -2,7 +2,7 @@
 uaa_instances: 3
 nats_instances: 3
 diego_api_instances: 3
-cell_instances: 135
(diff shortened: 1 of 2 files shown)


risicle:
I think they're both a bit extreme - we'd be getting `DiegoCellRepMemoryCapacity` before the end of the week as it momentarily touched the limit for some reason.

AP-Hunt:
What would you suggest?

risicle:
Maybe 123 and 18?

AP-Hunt:
Have adjusted to those numbers

risicle (approved):

PR #3421, opened 2023-09-26
logit logstash filters: change `alb_access` log type to `aws_lb_access`

risicle:
What
----
https://www.pivotaltracker.com/story/show/185739015

This is what I unknowingly added the inputs in logit as in prod and staging, and it will be painful to change now. Also it may be more accurate for LBs that are not ALBs

How to review
-------------
:eyes: 

---


risicle:
Hold on, `alb_access` may be a special magic value we need to use.

risicle:
It is.

PR #3448, opened 2023-10-09
Update notify-staging dms process to TRUNCATE_BEFORE_LOAD

corlettb:
What
----

Update notify-staging dms process to TRUNCATE_BEFORE_LOAD

Why
----

This is a much more sensible choice for operation in the case of failure for notify. We want to be able to do a full retry of the db load. Truncate the target tables makes sense when managing the schema outside of dms.

How to review
-------------

Look at the changes.
---


fearoffish (approved):
Looks gooderer than before.

PR #3449, opened 2023-10-09
paas-s3-broker: bump to ..., provide with common user policy, expand custom-broker-acceptance-tests

risicle:
What
----
https://www.pivotaltracker.com/story/show/186090782

Roll this change out to cf, also making our custom-broker-acceptance-tests check they can successfully call `s3:GetObjectTagging`. In an ideal world we would have actually set a tag on the object to more completely test this works, but then we'd need to add `s3:PutObjectTagging` which we currently can avoid. So here we just test for the returned `TagSet`'s emptiness.

How to review
-------------

:eyes: https://deployer.dev03.dev.cloudpipeline.digital/teams/main/pipelines/create-cloudfoundry/jobs/custom-acceptance-tests/builds/21 ?

---

### DIFF
--- a/manifests/cf-manifest/operations.d/750-s3-broker.yml
+++ b/manifests/cf-manifest/operations.d/750-s3-broker.yml
@@ -4,9 +4,9 @@
   path: /releases/-
   value:
     name: s3-broker
-    version: 0.0.1697806230
-    url: https://s3-eu-west-1.amazonaws.com/gds-paas-build-releases/s3-broker-0.0.1697806230.tgz
-    sha1: f35d6a6c418c1bcd6e47d3c18224d72f7025b105
+    version: 0.1.25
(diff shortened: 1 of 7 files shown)


risicle:
I think we need some more alterations before this will work for `aws s3 sync`. Don't know if it's worth merging before that.

risicle:
Ok, this now includes the "common user policy" feature from https://github.com/alphagov/paas-s3-broker/pull/56 and provides that policy to the job.

It also expands the custom-broker-acceptance-tests to ensure different buckets' credentials can't access each other - something that could potentially happen if the common user policy were misconfigured.

dark5un (approved):
I reviewed the deployment of the broker on a dev environment with @malcgds and went really well. I also tested the sync functionality following the instructions that @malcgds  wrote and managed to sync the paas s3 bucket with one I made under an experimental account. The rest of the review (code etc has been carried out by @malcgds ). Seems good to go from my end.

PR #3513, opened 2023-11-23
[#186518458] Added script update-s3-broker-users-with-permissions-bou…

malcgds:
What
----

Added a script that will add the S3BrokerUserPermissionsBoundary policy to paas-s3-broker users that do not have it.

This change is associated with the work to allow s3 bucket synchronisation from PaaS to a tenant's AWS account.

How to review
-------------

1. Review the code.
2. Run the branch into a dev env and try various different variations of parameter. The policy_name that it will work with is S3BrokerUserPermissionsBoundary. If you are using an env where the latest s3 broker has been implemented you will find that the script reports that the policy already exists for certain s3 users.

---


fearoffish (approved):
worked on dev02

PR #3531, opened 2023-12-08
Remove PSN vpn endpoint in production london

corlettb:
What
----

Remove the PSN vpc endpoint in production London.

Why
----

DCS has now been decommissioned. They have requested we remove the PSN vpc_endpoint. See:

https://govuk.zendesk.com/agent/tickets/5578456

How to review
-------------

Look at the changes.

---


terraform/psn/psn.tf
```diff
@@ -17,61 +17,3 @@ variable "security_group_name" {
...
-  source_security_group_id = aws_security_group.psn_endpoint.id
-}
-
-data "aws_network_interface" "psn_interface" {
-  for_each = aws_vpc_endpoint.psn_service.network_interface_ids
-
-  id = each.value
-}
-
-output "psn_security_group_seed_json" {
```
AP-Hunt, line 70: We should validate that the outputs aren't required further down the pipeline

corlettb: Good spot. I've updated I will check how it was being used.

